PR #12, opened 2025-12-14
Migrate SDK to Agent-Centric REST API with Multi-Framework Support

vlad-thenvoi:
# Thenvoi SDK Architecture Reference

## Overview

The Thenvoi SDK follows a layered architecture with clean separation of concerns. Each layer can be used independently, enabling everything from ultra-light platform connectivity to full SDK-managed agent lifecycles.

---

## Layer Architecture

```
┌─────────────────────────────────────────────────────────────────┐
│                    INTEGRATION LAYER (Layer 4)                   │
│  ThenvoiAnthropicAgent, ThenvoiClaudeSDKAgent,                  │
│  ThenvoiLangGraphAgent, ThenvoiPydanticAgent                    │
├─────────────────────────────────────────────────────────────────┤
│                    AGENT LAYER (Layer 3)                         │
│                      BaseFrameworkAgent                          │
├─────────────────────────────────────────────────────────────────┤
│                    RUNTIME LAYER (Layer 2)                       │
│  AgentRuntime │ RoomPresence │ ExecutionContext │ AgentTools    │
├─────────────────────────────────────────────────────────────────┤
│                    PLATFORM LAYER (Layer 1)                      │
│                 ThenvoiLink │ PlatformEvent                      │
├─────────────────────────────────────────────────────────────────┤
│                    CLIENT LAYER (Layer 0)                        │
│              WebSocketClient │ AsyncRestClient                   │
└─────────────────────────────────────────────────────────────────┘
```

---

### Layer 0: Client (`thenvoi/client/`)

**Purpose:** Raw wire protocol - WebSocket and REST API clients.

| Class | File | Responsibility |
|-------|------|----------------|
| `WebSocketClient` | `client/streaming/client.py` | Phoenix channels WebSocket connection, topic subscriptions, payload validation |
| `AsyncRestClient` | `client/rest/__init__.py` | HTTP REST API client (re-exported from `thenvoi_rest`) |
| `MessageCreatedPayload` | `client/streaming/client.py` | Pydantic model for message events |
| `RoomAddedPayload` | `client/streaming/client.py` | Pydantic model for room_added events |
| `RoomRemovedPayload` | `client/streaming/client.py` | Pydantic model for room_removed events |

**When to use directly:** Never - use ThenvoiLink instead.

---

### Layer 1: Platform (`thenvoi/platform/`)

**Purpose:** Platform connectivity abstraction - unified event dispatch, subscription management.

| Class | File | Responsibility |
|-------|------|----------------|
| `ThenvoiLink` | `platform/link.py` | Live connection to Thenvoi (WebSocket + REST), subscription lifecycle, unified event callback |
| `PlatformEvent` | `platform/event.py` | Unified event wrapper with type-safe accessors and payload converters |

**Key Features:**
- Single `on_event` callback for all platform events
- Subscription tracking for reconnection
- Message delivery protocol: `mark_processing()`, `mark_processed()`, `mark_failed()`
- Crash recovery: `get_next_message()` for backlog sync

**When to use directly:** Ultra-light integrations, custom event routing, non-agent use cases.

---

### Layer 2: Runtime (`thenvoi/runtime/`)

**Purpose:** Agent execution infrastructure - room presence, execution contexts, tools.

| Class | File | Responsibility |
|-------|------|----------------|
| `RoomPresence` | `runtime/presence.py` | Cross-room lifecycle (join/leave), room filtering, event routing |
| `ExecutionContext` | `runtime/execution.py` | Per-room state: message queue, context hydration, crash recovery sync |
| `AgentRuntime` | `runtime/runtime.py` | Convenience wrapper combining RoomPresence + ExecutionContext factory |
| `AgentTools` | `runtime/tools.py` | LLM tool bridge: messaging, participants, schema generation |
| `Execution` | `runtime/execution.py` | Protocol for pluggable per-room execution strategies |

**Supporting Types:**
| Type | File | Purpose |
|------|------|---------|
| `SessionConfig` | `runtime/types.py` | Execution context configuration (hydration, retries, cache) |
| `AgentConfig` | `runtime/types.py` | Agent startup configuration |
| `PlatformMessage` | `runtime/types.py` | Normalized message format |
| `ConversationContext` | `runtime/types.py` | Hydrated room context with history |
| `MessageRetryTracker` | `runtime/retry_tracker.py` | Tracks processing attempts, prevents infinite retries |
| `ParticipantTracker` | `runtime/participant_tracker.py` | Tracks participant changes for LLM notification |

**Utility Functions:**
| Function | File | Purpose |
|----------|------|---------|
| `format_history_for_llm()` | `runtime/formatters.py` | Convert platform messages to LLM format |
| `build_participants_message()` | `runtime/formatters.py` | Format participant list for LLM |
| `render_system_prompt()` | `runtime/prompts.py` | Template-based system prompt generation |

#### Message Lifecycle Management (Automatic)

`ExecutionContext` automatically manages message processing state - **integrations don't handle this**.

```
ExecutionContext._process_event(event)
│
├── 1. PRE-CHECKS (skip if any fail)
│   ├── Is this my own message? → skip (prevent infinite loop)
│   ├── Already processed? → skip (dedupe cache)
│   ├── Permanently failed? → skip (exceeded retries)
│   └── Record attempt → skip if exceeded max_retries
│
├── 2. BEFORE HANDLER
│   └── await link.mark_processing(room_id, msg_id)
│       └── Server knows we're working on it
│
├── 3. EXECUTE HANDLER
│   └── await on_execute(ctx, event)
│       └── Your code runs here (BaseFrameworkAgent, custom handler, etc.)
│
└── 4. AFTER HANDLER
    ├── SUCCESS: await link.mark_processed(room_id, msg_id)
    │   └── Server removes from unprocessed queue
    └── EXCEPTION: await link.mark_failed(room_id, msg_id, error)
        └── Server records failure, may retry later
```

**What each status means:**

| Status | API Call | Server Behavior |
|--------|----------|-----------------|
| `processing` | `mark_processing()` | Message locked, won't be returned by `/next` |
| `processed` | `mark_processed()` | Message complete, removed from queue |
| `failed` | `mark_failed()` | Error recorded, may be retried based on server config |

**Retry tracking (client-side):**

`MessageRetryTracker` prevents infinite retry loops:
- Tracks attempts per message ID
- After `max_retries` (default 3), message is marked "permanently failed"
- Permanently failed messages are skipped on future encounters

**Why this matters:**

1. **Crash recovery** - If agent crashes mid-processing, server still has the message
2. **Visibility** - Platform knows which messages are being worked on
3. **Error tracking** - Failed messages with errors are logged server-side
4. **No duplicates** - `_processed_ids` cache + server state prevents reprocessing

#### Participant Management (Multi-Layer)

Participant tracking spans multiple layers - here's where each piece lives:

```
┌─────────────────────────────────────────────────────────────────────────┐
│ LAYER 1: ThenvoiLink                                                    │
│                                                                         │
│  WebSocket ──► participant_added event ──► PlatformEvent                │
│  WebSocket ──► participant_removed event ──► PlatformEvent              │
│                                                                         │
│  REST API: list_agent_chat_participants() for initial load              │
└─────────────────────────────────────────────────────────────────────────┘
                                    │
                                    ▼
┌─────────────────────────────────────────────────────────────────────────┐
│ LAYER 2: ExecutionContext                                               │
│                                                                         │
│  STATE:                                                                 │
│  - _participants: list[dict]           # Current participant list       │
│  - _participants_loaded: bool          # Has initial load happened?     │
│  - _last_participants_sent: list       # Snapshot for change detection  │
│                                                                         │
│  EVENTS (in _process_event):                                            │
│  - if event.is_participant_added:                                       │
│        self.add_participant(event.payload)                              │
│  - elif event.is_participant_removed:                                   │
│        self.remove_participant(event.payload.get("id"))                 │
│                                                                         │
│  INITIAL LOAD (during hydrate()):                                       │
│  - await self.load_participants()  # REST API call                      │
│                                                                         │
│  CHANGE DETECTION:                                                      │
│  - participants_changed() → bool    # Compare current vs last sent      │
│  - mark_participants_sent()         # Update snapshot after LLM notify  │
│                                                                         │
│  FORMATTING:                                                            │
│  - build_participants_message() → str   # "## Current Participants..."  │
└─────────────────────────────────────────────────────────────────────────┘
                                    │
                                    ▼
┌─────────────────────────────────────────────────────────────────────────┐
│ LAYER 3: BaseFrameworkAgent._dispatch_message()                         │
│                                                                         │
│  # Check if participants changed, format notification                   │
│  participants_msg = check_and_format_participants(ctx)                  │
│  if participants_msg:                                                   │
│      # Pass to integration for LLM injection                            │
│                                                                         │
│  await self._handle_message(                                            │
│      msg=msg,                                                           │
│      tools=tools,                                                       │
│      ctx=ctx,                                                           │
│      history=history,                                                   │
│      participants_msg=participants_msg,  # ◄── Passed to integration    │
│  )                                                                      │
└─────────────────────────────────────────────────────────────────────────┘
                                    │
                                    ▼
┌─────────────────────────────────────────────────────────────────────────┐
│ LAYER 4: Integration._handle_message()                                  │
│                                                                         │
│  if participants_msg:                                                   │
│      # Inject into LLM context (framework-specific)                     │
│      # Anthropic: add as user message                                   │
│      # LangGraph: add to messages state                                 │
│      # Pydantic AI: add to conversation                                 │
└─────────────────────────────────────────────────────────────────────────┘
```

**The participant change detection flow:**

```python
# ExecutionContext tracks:
_participants = [{"id": "1", "name": "Alice"}, {"id": "2", "name": "Bob"}]
_last_participants_sent = [{"id": "1", "name": "Alice"}]  # What LLM knows

# check_and_format_participants(ctx) does:
if ctx.participants_changed():  # True - Bob is new!
    msg = build_participants_message(ctx.participants)
    ctx.mark_participants_sent()  # Update snapshot
    return msg  # "## Current Participants\n- Alice (User)\n- Bob (Agent)"
```

**AgentTools also tracks participants** (for mention resolution):

```python
# AgentTools._participants - local cache
# Updated when:
# 1. Created from ExecutionContext: AgentTools.from_context(ctx)
# 2. After add_participant() tool call (immediate, before WS event)
# 3. After remove_participant() tool call (immediate, before WS event)

# This allows @mentions to work immediately after adding someone:
await tools.add_participant("Weather Agent")
await tools.send_message("Hello!", mentions=["Weather Agent"])  # Works!
```

**Formatted output for LLM:**

```
## Current Participants
- Alice (User)
- Bob (Agent)
- Weather Agent (Agent)

To mention a participant in send_message, use their EXACT name (e.g., 'Weather Agent', not an ID).
```

**When to use directly:** Custom agents, Letta-style persistent agents, custom execution strategies.

---

### Layer 3: Agent (`thenvoi/agents/`)

**Purpose:** Base agent implementation eliminating repeated boilerplate across all framework integrations.

| Class | File | Responsibility |
|-------|------|----------------|
| `BaseFrameworkAgent` | `agents/base.py` | Abstract base handling all common agent logic so frameworks only implement LLM-specific code |

#### What BaseFrameworkAgent Handles (so integrations don't have to)

**1. Lifecycle Management**
```python
# All integrations get this for free:
async def start()   # Creates ThenvoiLink, fetches metadata, starts AgentRuntime
async def stop()    # Stops runtime, disconnects link
async def run()     # start() + run_forever() + stop()
```

**2. Agent Metadata Fetch**
- Calls `/agent/me` API on startup
- Extracts `agent_name` and `agent_description` from platform
- Validates agent has required description

**3. Self-Message Filtering**
```python
# Prevents infinite loops - all integrations need this
if msg.sender_type == "Agent" and msg.sender_id == self._agent_id:
    return  # Skip own messages
```

**4. First-Message Detection + History Loading**
```python
# Tracks per-room LLM initialization state
is_first = not ctx.is_llm_initialized
if is_first:
    history = await self._load_history(ctx, msg)  # Load from platform
    ctx.mark_llm_initialized()
```

**5. Participant Change Detection**
```python
# Checks if participants changed, formats notification for LLM
participants_msg = check_and_format_participants(ctx)
if participants_msg:
    # "Participants updated: Alice, Bob, Charlie"
```

**6. PlatformEvent → PlatformMessage Conversion**
```python
# Converts WebSocket payload to normalized message format
msg = PlatformMessage(
    id=..., room_id=..., content=...,
    sender_id=..., sender_type=..., sender_name=...,
    message_type=..., metadata=..., created_at=...
)
```

**7. AgentTools Creation**
```python
# Creates tools bound to current room
tools = AgentTools.from_context(ctx)
```

**8. Error Reporting Helper**
```python
# Best-effort error notification to room
await self._report_error(tools, "Something went wrong")
```

#### What Integrations Implement

Subclasses only need to implement the framework-specific logic:

```python
class ThenvoiAnthropicAgent(BaseFrameworkAgent):

    @abstractmethod
    async def _handle_message(
        self,
        msg: PlatformMessage,      # Normalized message
        tools: AgentTools,          # Ready-to-use tools
        ctx: ExecutionContext,      # Room context
        history: list | None,       # Platform history (first message only)
        participants_msg: str | None,  # Participant update notification
    ) -> None:
        # Just implement your LLM call here!
        pass

    async def _cleanup_session(self, room_id: str) -> None:
        # Optional: cleanup framework state when leaving room
        pass

    async def _on_started(self) -> None:
        # Optional: post-start setup (e.g., initialize LLM client)
        pass
```

#### Without BaseFrameworkAgent

Every integration would need to duplicate:
- ~50 lines of lifecycle code
- ~20 lines of metadata fetch
- ~15 lines of self-message filtering
- ~25 lines of history loading
- ~10 lines of participant detection
- ~20 lines of message conversion

**Total: ~140 lines of boilerplate per integration.**

**When to use directly:** Creating new framework integrations.

---

### Layer 4: Integrations (`thenvoi/integrations/`)

**Purpose:** Framework-specific adapters implementing BaseFrameworkAgent.

| Class | File | Framework |
|-------|------|-----------|
| `ThenvoiAnthropicAgent` | `integrations/anthropic/agent.py` | Anthropic Python SDK |
| `ThenvoiClaudeSDKAgent` | `integrations/claude_sdk/agent.py` | Claude Agent SDK (MCP tools) |
| `ThenvoiLangGraphAgent` | `integrations/langgraph/agent.py` | LangGraph (graph factory) |
| `ThenvoiPydanticAgent` | `integrations/pydantic_ai/agent.py` | Pydantic AI |

**Supporting Classes:**

| Class/Function | File | Purpose |
|----------------|------|---------|
| `ClaudeSessionManager` | `integrations/claude_sdk/session_manager.py` | Per-room Claude SDK client cache |
| `create_thenvoi_mcp_server()` | `integrations/claude_sdk/tools.py` | MCP server factory for Claude SDK |
| `agent_tools_to_langchain()` | `integrations/langgraph/langchain_tools.py` | Convert AgentTools to LangChain tools |
| `graph_as_tool()` | `integrations/langgraph/graph_tools.py` | Wrap LangGraph as LangChain tool |
| `check_and_format_participants()` | `integrations/base.py` | Participant change notification helper |

**When to use:** Most users - pick the integration matching your LLM framework.

---

### Composition (owns lifecycle)
- `ThenvoiLink` **composes** `WebSocketClient`, `AsyncRestClient`
- `AgentRuntime` **composes** `RoomPresence`, `ExecutionContext` instances
- `ExecutionContext` **composes** `MessageRetryTracker`, `ConversationContext`
- `BaseFrameworkAgent` **composes** `ThenvoiLink`, `AgentRuntime`

### Aggregation (references, doesn't own)
- `AgentTools` **aggregates** `AsyncRestClient` (via ThenvoiLink.rest)
- `ClaudeSessionManager` **aggregates** `ClaudeSDKClient` instances

### Protocol Implementation
- `ExecutionContext` **implements** `Execution` protocol
- Custom execution strategies can implement `Execution` protocol

### Inheritance
- All integration agents **inherit from** `BaseFrameworkAgent`
- `ThenvoiLangGraphMCPAgent` **inherits from** `ThenvoiLangGraphAgent`

### Callbacks/References
- `ThenvoiLink` dispatches `PlatformEvent` to `on_event` callback
- `RoomPresence` invokes `on_room_joined`, `on_room_left`, `on_room_event` callbacks
- `ExecutionContext` invokes `on_execute` callback (ExecutionHandler)

---

## Data Flow

```
WebSocket Event
      │
      ▼
┌─────────────────┐
│  WebSocketClient │ ─── Validates payload against Pydantic models
└────────┬────────┘
         │
         ▼
┌─────────────────┐
│   ThenvoiLink   │ ─── Creates PlatformEvent, dispatches via on_event
└────────┬────────┘
         │
         ▼
┌─────────────────┐
│  RoomPresence   │ ─── Routes event to room's ExecutionContext
└────────┬────────┘
         │
         ▼
┌─────────────────┐
│ExecutionContext │ ─── Queues event, handles sync, invokes on_execute
└────────┬────────┘
         │
         ▼
┌─────────────────┐
│BaseFrameworkAgent│ ─── Builds AgentTools, calls _handle_message()
└────────┬────────┘
         │
         ▼
┌─────────────────┐
│ Integration     │ ─── Framework-specific LLM execution
│ (Anthropic,     │
│  LangGraph, etc)│
└─────────────────┘
```

---

## File Structure

```
src/thenvoi/
├── __init__.py              # Public API exports
├── client/
│   ├── rest/                # AsyncRestClient (from thenvoi_rest)
│   └── streaming/
│       └── client.py        # WebSocketClient, payload models
├── config/
│   └── loader.py            # YAML config loading utilities
├── platform/
│   ├── event.py             # PlatformEvent
│   └── link.py              # ThenvoiLink
├── runtime/
│   ├── types.py             # AgentConfig, SessionConfig, PlatformMessage, ConversationContext
│   ├── execution.py         # Execution protocol, ExecutionContext
│   ├── presence.py          # RoomPresence
│   ├── runtime.py           # AgentRuntime
│   ├── tools.py             # AgentTools, tool input models
│   ├── formatters.py        # LLM formatting functions
│   ├── prompts.py           # System prompt rendering
│   ├── retry_tracker.py     # MessageRetryTracker
│   └── participant_tracker.py # ParticipantTracker
├── agents/
│   └── base.py              # BaseFrameworkAgent
└── integrations/
    ├── base.py              # check_and_format_participants()
    ├── anthropic/
    │   └── agent.py         # ThenvoiAnthropicAgent
    ├── claude_sdk/
    │   ├── agent.py         # ThenvoiClaudeSDKAgent
    │   ├── session_manager.py # ClaudeSessionManager
    │   ├── prompts.py       # Claude SDK prompt generation
    │   └── tools.py         # MCP server factory
    ├── langgraph/
    │   ├── agent.py         # ThenvoiLangGraphAgent, ThenvoiLangGraphMCPAgent
    │   ├── langchain_tools.py # agent_tools_to_langchain()
    │   ├── graph_tools.py   # graph_as_tool()
    │   └── message_formatters.py # Message format converters
    └── pydantic_ai/
        └── agent.py         # ThenvoiPydanticAgent
```

Resolves https://github.com/thenvoi/thenvoi-sdk-python/issues/14
___
Resolves INT-5, INT-4, INT-6


src/thenvoi/core/session.py
```diff
@@ -0,0 +1,606 @@
...
+                    logger.debug(f"Session {self.room_id} cancelled")
+                    break
+
+        except Exception as e:
+            logger.error(f"Session {self.room_id} error: {e}", exc_info=True)
+            self._is_running = False
+
+        logger.debug(f"Session {self.room_id} loop exited")
+
+    async def _synchronize_with_next(self) -> None:
```
dan-naiger-thenvoi: 75 lines for what can be done in ~25 lines.

The problem:  
Agent was offline. Messages accumulated. Need to process backlog without duplicates.

Current approach:
- Second queue for WebSocket messages
- Peek at queue head using `queue._queue[0]` (CPython internals)
- Detect a “sync point” when `/next` matches the queue head
- Complex state machine

Simpler approach:

async def _on_room_added(self, room_id: str) -> None:
    self._backlog_ids[room_id] = set()
    self._backlog_done[room_id] = False

    # 1. Subscribe to WebSocket
    await self._subscribe_to_room(room_id)

    # 2. Process all unprocessed messages via /next
    while True:
        msg = await self._get_next_message(room_id)
        if msg is None:
            break
        self._backlog_ids[room_id].add(msg.id)
        await self._process_message(msg)

    self._backlog_done[room_id] = True

async def _on_message_created(self, msg_data) -> None:
    room_id = msg_data.room_id

    # During /next processing: ignore (will come via /next)
    if not self._backlog_done.get(room_id, True):
        return

    # After /next done: skip duplicates
    if msg_data.id in self._backlog_ids.get(room_id, set()):
        self._backlog_ids[room_id].discard(msg_data.id)
        return

    await self._process_message(msg_data)

Why this works:
- `/next` returns ANY unprocessed message (not just “old” backlog)
- Messages arriving during `/next` processing are also unprocessed
- So `/next` will return them too
- After `/next` returns None, all messages are processed
- WebSocket duplicates are skipped via set lookup

No second queue.  
No peeking.  
No sync point detection.

src/thenvoi/core/session.py
```diff
@@ -0,0 +1,606 @@
...
+            config: Optional session configuration
+        """
+        self.room_id = room_id
+        self._api_client = api_client
+        self._on_message = on_message
+        self._coordinator = coordinator
+        self.config = config or SessionConfig()
+
+        # Per-room state
+        self.queue: asyncio.Queue[PlatformMessage] = asyncio.Queue()
```
dan-naiger-thenvoi: This queue exists only because the /next sync algorithm needs to "peek" at the head. The problem: asyncio.Queue doesn't have a peek() method. So line 451 accesses CPython internals:

head_msg = self.queue._queue[0]  # type: ignore[attr-defined]
_queue is a private implementation detail (it's the underlying collections.deque). This could break on:
PyPy (different implementation)
Future Python versions
Any Python implementation that structures Queue differently
The # type: ignore comment is a red flag - we're telling the type checker to look away because we're doing something unsupported. Why do we need to peek? The sync algorithm wants to detect when /next returns the same message that's at the head of the WebSocket queue (the "sync point"). But this can be done with a set instead - see the comment on _synchronize_with_next. Also note: Phoenix channels client already has a queue per topic. This is a second queue on top of that.

dan-naiger-thenvoi:
## Import Structure Violates Agreed Design

From the meeting notes made by Gavri in Integrations group (which you reacted with 👍):

> from thenvoi.agent.core import ...
> from thenvoi.agent.langgraph import ...

The original `dev` branch followed this structure:
- `thenvoi.agent.core` → ThenvoiPlatformClient, RoomManager
- `thenvoi.agent.langgraph` → ThenvoiLangGraphAgent, connect_graph_to_platform

This PR:
1. **Deletes** all files from `src/thenvoi/agent/` (core, langgraph, crewai, letta, basic)
2. **Creates** `src/thenvoi/core/` instead (breaks `thenvoi.agent.core` import)
3. **Creates** `src/thenvoi/integrations/langgraph/` (breaks `thenvoi.agent.langgraph` import)
4. **Moves** the actual LangGraph agent to `examples/langgraph/thenvoi_langgraph_agent.py`
   - This is NOT installable via pip
   - Users must copy this file manually

The agreed import:

    from thenvoi.agent.langgraph import ThenvoiLangGraphAgent

Now requires:

    # Copy thenvoi_langgraph_agent.py into your project
    from thenvoi_langgraph_agent import ThenvoiLangGraphAgent

This is a significant deviation from what we agreed on.

examples/langgraph/thenvoi_langgraph_agent.py
dan-naiger-thenvoi: This file defines `ThenvoiLangGraphAgent` and `create_langgraph_agent` - the primary LangGraph integration.  
But it lives in `examples/` instead of `src/thenvoi/integrations/langgraph/`.

The examples import it as:

    from thenvoi_langgraph_agent import create_langgraph_agent

This means:
1. It's not part of the installable package (`pip install thenvoi`)
2. Users must copy this file into their project
3. Same pattern for Pydantic AI (361 lines) and Anthropic (512 lines) adapters

If these are meant to be reusable adapters, they should not be in examples.

examples/langgraph/thenvoi_langgraph_agent.py
```diff
@@ -0,0 +1,393 @@
...
+            raise RuntimeError("No graph available")
+
+        # Build messages list for this invocation
+        messages = []
+
+        if is_first_message:
+            # FIRST MESSAGE: Include system prompt + historical context
+            messages.append(("system", self._system_prompt))
+            logger.info(f"Room {room_id}: Sending system prompt (first message)")
+            logger.debug(f"System prompt:\n{self._system_prompt}")
```
dan-naiger-thenvoi: When user passes a pre-compiled `graph` (via the `graph` parameter), there's a problem:

If the graph was compiled with a system prompt:

    graph = create_react_agent(llm, tools, prompt="You are a weather assistant")

And the adapter injects its own at runtime (line 202):

    messages.append(("system", self._system_prompt))

The LLM receives TWO system messages:

    SystemMessage: You are a weather assistant    ← Baked into compiled graph
    SystemMessage: You are {agent_name}...        ← Injected by adapter
    HumanMessage: ...

This doesn't error, but it's likely unintentional. The LLM sees conflicting instructions.

If `graph` is provided (not `graph_factory`), the adapter should either:
1. Not inject any system prompt (respect user's compiled graph)
2. Document this behavior clearly

src/thenvoi/core/prompts.py
```diff
@@ -0,0 +1,100 @@
...
+Use `send_message(content, mentions)` to respond. Plain text output is not delivered.
+
+## CRITICAL: Always Relay Information Back to the Requester
+
+When someone asks you to get information from another agent:
+1. Ask the other agent for the information
+2. When you receive the response, IMMEDIATELY relay it back to the ORIGINAL REQUESTER
+3. Do NOT just thank the helper agent - the requester is waiting for their answer!
+
+## IMPORTANT: Always Share Your Thinking
```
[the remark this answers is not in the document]
vlad-thenvoi: You mix three different concepts:
  1. o1/o3 private reasoning - Internal, automatic, hidden chain-of-thought for problem-solving
  2. Claude extended thinking - Developer-controlled, returns thinking blocks for complex reasoning
  3. Claude "think" tool - Tactical pausing at decision points, not continuous thinking

  The distinction is:
  - LLM reasoning: Internal problem-solving (better answers)
  - SDK thought events: External observability (explain state, rationale for actions)

  These are orthogonal concerns - you can use extended thinking AND still need agents to declare their state.

examples/langgraph/thenvoi_langgraph_agent.py
vlad-thenvoi: fair point

src/thenvoi/core/session.py
dan-naiger-thenvoi: You're right about the race condition. I was wrong - if we discard messages during sync and one arrives after /next queries the DB but before it returns, we'd lose it.

The sync logic does need to handle this. I've been thinking about whether this could be solved server-side instead.

Instead of /next being a REST endpoint, what if the server pushed messages through the WebSocket one at a time? (it avoids the big message you were afraid of)

The flow would be:

- Client joins channel
- Server checks for unprocessed messages, pushes the first one
- Client processes it, sends mark_processed back through the channel
- Server receives mark_processed, pushes the next unprocessed message
- Repeat until caught up
- New real-time messages flow through the same channel normally

This is essentially the same logic as /next, but over WebSocket instead of REST. The server only sends the next message after receiving acknowledgment from the client. No two-channel coordination needed - everything flows through one channel with request/response semantics.

The key point: the server doesn't need to send all backlog messages in the join reply. It can be gradual - one message at a time, paced by client acknowledgment, or whatever strategy makes sense. The ordering problem goes away because the server controls the pace and knows exactly when the client is ready for the next message.

The advantage would be: one implementation on the server vs the same sync logic duplicated in every SDK (Python, JS, Go, etc.). Each client would just process messages as they arrive and send mark_processed.

Is there a reason this approach wouldn't work?

dan-naiger-thenvoi:
The explanation I promised regarding my thoughts on old abstractions:

## Why We Need PlatformClient and RoomManager

### PlatformClient — Infrastructure Layer

Every adapter needs to:
- Validate the agent exists on the platform
- Connect to WebSocket for real-time events
- Access the REST API for operations (send messages, fetch context, mark processed)

PlatformClient provides exactly this — connection and API access. It makes no assumptions about how you’ll use them.

class ThenvoiPlatformClient:
    def __init__(self, agent_id, api_key, ws_url, rest_url):
        self.api_client = AsyncRestClient(...)  # REST operations

    async def fetch_agent_metadata(self) -> Agent:
        # Validate agent, get name/description

    async def connect_websocket(self) -> WebSocketClient:
        # Connect to real-time events

That’s it.  
No message handling.  
No room management.  
No state.  
Just infrastructure.

---

### RoomManager — Lifecycle Events Layer

Built on top of PlatformClient. Handles room subscription mechanics and defines lifecycle hooks.

Our platform allows agents to be added/removed from rooms dynamically. When these events happen, different adapters need to do different things. The events are universal:

- room_added — agent joins a room  
- room_removed — agent leaves a room  
- message_received — message arrives  

But the behavior varies by adapter.

**LangGraph with persistent checkpointer:**
- on_room_added: Nothing — checkpointer loads state via thread_id
- on_message: Invoke graph directly

**LangGraph with InMemorySaver:**
- on_room_added: Maybe fetch history to prime checkpointer
- on_message: Invoke graph

**Letta (single-room per agent):**
- on_room_added: Maybe initialize agent context differently
- on_message: Maybe prefix with room info? Format differently?

**Future framework X:**
- Who knows what it needs

The point: we don’t know yet what every adapter will need. RoomManager with callbacks keeps this open:

class RoomManager:
    def __init__(
        self,
        platform: PlatformClient,  # Uses infrastructure layer
        on_room_added: Callable,   # Adapter decides behavior
        on_room_removed: Callable, # Adapter decides behavior
        on_message: Callable,      # Adapter decides behavior
    ):
        ...

RoomManager handles the mechanics.  
Adapters handle the behavior (what to actually do when these events happen).

---

### The Relationship

PlatformClient (infrastructure)
- WebSocket connection
- REST API client
- Agent validation
       ↓
RoomManager (lifecycle events)
- Uses PlatformClient for connection/API
- Subscribes to rooms
- Routes events to callbacks
- Adapters plug in their behavior
       ↓
Framework Adapter (behavior)
- Implements on_room_added, on_message, etc.
- Decides what its framework needs
- Full control over state management

Each layer has a single responsibility:
- PlatformClient: How do I talk to the platform?
- RoomManager: When do lifecycle events happen?
- Adapter: What do I do when they happen?

---

## What the New Architecture Does Wrong

It collapses RoomManager into ThenvoiAgent + AgentSession and bakes in specific behaviors:

# BaseFrameworkAgent._dispatch_message() — SDK decides for everyone
is_first = not session.is_llm_initialized
if is_first:
    history = await self._load_history(session, msg)
    session.mark_llm_initialized()

Now every adapter gets history passed to it, whether it wants it or not. The decision was made at the SDK level, not the adapter level.

For LangGraph with a persistent checkpointer, this is wasted work — the checkpointer already has history. For Letta, who knows if this is even the right approach.

---

## The Principle

Infrastructure belongs in PlatformClient.  
Lifecycle events belong in RoomManager.  
Behavior belongs in adapters.

This keeps flexibility for:
- Different LangGraph configurations (persistent vs in-memory checkpointer)
- Frameworks we haven’t integrated yet
- Future requirements we haven’t thought of

dan-naiger-thenvoi:
Not sure whether its new or not - using simple agent, I feel a lot of inconsistent behavior while using...
sending it messages, it does not reply - not sure whether its system prompt issue or something with message delivery

gavrie (commented):
# General
For naming consistency, I recommend following standard Python naming conventions. For example, Python's standard library has `threading.Lock` and `threading.Event` besides `asyncio.Lock` and `asyncio.Event` and this is considered fine.

## Platform layer
- Rename `ThenvoiLink` to `Connection` (from end user code it can then be accessed as `thenvoi.Connection`). This is more standard term than `Link` and the extra `Thenvoi` doesn't add anything here.
- `PlatformEvent` -> `thenvoi.Event`
- The use of callbacks is not very idiomatic in Python. I recommend using `async for` instead; see this draft PR: #16 . We may want to push this down into the Client layer instead or in addition to here; it also recurs in higher level layers.

src/thenvoi/integrations/langgraph/agent.py
```diff
@@ -0,0 +1,365 @@
...
+                for hist in history:
+                    role = hist["role"]
+                    content = hist["content"]
+                    sender_name = hist["sender_name"]
+
+                    # Format for LangGraph
+                    if role == "assistant":
+                        messages.append(("assistant", content))
+                    else:
+                        messages.append(("user", f"[{sender_name}]: {content}"))
```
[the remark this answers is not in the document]
dan-naiger-thenvoi: This fix is product decision. would be happy to elaborate if pr description is not clear
@roi-shikler-thenvoi @vlad-thenvoi 
see:
https://github.com/thenvoi/thenvoi-sdk-python/pull/20

dan-naiger-thenvoi:
> Not sure whether its new or not - using simple agent, I feel a lot of inconsistent behavior while using... sending it messages, it does not reply - not sure whether its system prompt issue or something with message delivery - trying to investigate...
> 
> here is an exmple:
> 
> Conversation:
> 
> [dan Naiger]: @simple agent hey what can you do for me? [dan Naiger]: @simple agent are you here? [dan Naiger]: @simple agent hello [dan Naiger]: @simple agent ??????? [dan Naiger]: @simple agent call your tools to answer me [dan Naiger]: @simple agent bro Agent Response: None Relevant Logs:
> 
> 16:41:48 [INFO] Room: Sending system prompt (first message) 16:41:48 [INFO] Room: Participants updated: ['dan Naiger', 'simple agent'] 16:41:48 [DEBUG] Room: Sending 3 messages to LangGraph 16:41:53 [DEBUG] Message processed successfully
> 
> 16:42:00 [DEBUG] Room: Sending 1 messages to LangGraph 16:42:02 [DEBUG] Message processed successfully
> 
> 16:42:10 [DEBUG] Room: Sending 1 messages to LangGraph 16:42:11 [DEBUG] Message processed successfully
> 
> 16:42:20 [DEBUG] Room: Sending 1 messages to LangGraph 16:42:22 [DEBUG] Message processed successfully
> 
> 16:42:36 [DEBUG] Room: Sending 1 messages to LangGraph 16:42:38 [DEBUG] Message processed successfully
> 
> 16:44:20 [DEBUG] Room: Sending 1 messages to LangGraph 16:44:22 [DEBUG] Message processed successfully Summary: All 6 messages were received and processed. No tool calls were made by the LLM (no on_tool_start/on_tool_end logs). The LLM appears to be responding with text instead of calling send_message, so no responses are delivered.
> 
> UPDATE: its probably not related to prompt. it happens every time agent joins a room it wasnt in before. if he was in the room when going up he answers mostly on first questions...

Root Cause Found: History Rehydration Corrupts LangGraph State + Prompt Issue

After extensive debugging, I identified two compounding issues causing the agent to not respond.

Issue 1: Prompt Teaching Text Output

The original system prompt showed examples with literal function call syntax:

[John Doe]: What's 2+2?
-> send_event("Simple arithmetic, answering directly.", message_type="thought")
-> send_message("4", mentions=["John Doe"])

This taught the model to output `send_event(...)` as text instead of using actual tool calls.

Fixed by changing to natural language workflow descriptions.

Issue 2: History Rehydration Corrupts Graph State

When an agent restarts and rejoins a room with existing history, the rehydration logic injects messages incorrectly.

Fresh room (no history) after prompt fix = Works perfectly - Tool calls work as expected:

[3] AIMessage: (empty)
     has tool_calls: True, value: [{'name': 'send_event', ...}]

After restart with history = Broken - The rehydrated history contains corrupted data:

[2] AIMessage: {"event": "on_tool_start", "data": {"input": {"content": "User Dan Naiger has greeted me...
     has tool_calls: False, value: []

The history stores raw event JSON (tool_call, tool_result events) and rehydrates them as plain AIMessage content. This teaches the model "respond with text, not tool calls" and breaks all subsequent interactions.

Why This Happens:

The current rehydration code injects all assistant role messages as plain text tuples:

if role == "assistant":
    messages.append(("assistant", content))

But the platform history includes:
- tool_call events (JSON blobs)
- tool_result events (JSON blobs)
- thought events
- Actual text messages

All get injected as AIMessage objects without `tool_calls`, corrupting the graph.

Fix Options Explored:

1. Filter by message_type (PR #20) - Only rehydrate text type messages.
   Did not work - model still learns to respond with text instead of tool calls from the reconstructed assistant messages.

2. Summarize conversation
   Instead of replaying individual messages, send one summary message with conversation context as a prefix.
   Avoids corrupting the graph with fake AIMessages.

3. Rebuild graph properly (requires platform changes)
   Store and expose full AIMessage structure including:
   - Agent's final response text (currently not visible separately)
   - tool_calls metadata with IDs
   - tool_call_id for ToolMessages

This would allow proper reconstruction of the LangGraph state with real tool call history.

Option 3 is the correct long-term fix but requires platform API changes. Option 2 could be a viable workaround.

PR #91, opened 2026-01-29
feat(tests): add session-scoped cleanup with optional --no-clean mode

yael-avioz-thenvoi:
## Summary
- Add `--no-clean` pytest option and `THENVOI_TEST_NO_CLEAN` env var to skip test cleanup
- Add `is_no_clean_mode()` helper function for fixtures to check cleanup mode
- Update `test_dynamic_agent` and `test_participant_permissions` fixtures with inline cleanup logic

## Use Cases
No-clean mode is useful for:
- Testing user limits (max agents, max chats)
- Finding edge case bugs with accumulated data
- Debugging test failures by inspecting created resources

## Usage
```bash
# Skip cleanup via flag
uv run pytest tests/integration/ -v --no-clean

# Skip cleanup via env var
THENVOI_TEST_NO_CLEAN=1 uv run pytest tests/integration/ -v
```

## Implementation
Simple per-fixture cleanup pattern (~10 lines per fixture):
```python
@pytest.fixture(scope="module")
async def my_fixture(client, request):
    resource = await create_resource()
    yield resource
    
    if not is_no_clean_mode(request):
        await delete_resource(resource)
    else:
        logger.info("[NO-CLEAN MODE] Skipping cleanup")
```

## Test plan
- [x] All unit tests pass
- [x] Ruff lint and format checks pass
- [x] Unit tests for `is_no_clean_mode()` helper (10 tests)
- [ ] Integration tests with cleanup enabled (default behavior)
- [ ] Integration tests with `--no-clean` flag

### DIFF
--- a/tests/integration/test_dynamic_agent.py
+++ b/tests/integration/test_dynamic_agent.py
@@ -45,10 +45,6 @@ class DynamicAgent:
     api_key: str
 
 
-# Module-level storage for the created agent
-_dynamic_agent: DynamicAgent | None = None
-
-
 @pytest.fixture(scope="module")
 def event_loop():
     """Create an event loop for the module scope.
@@ -82,8 +78,6 @@ async def dynamic_agent(module_user_api_client, request):
 
     Cleanup happens at the end of the module (unless --no-clean is specified).
     """
-    global _dynamic_agent
-
     if module_user_api_client is None:
         pytest.skip("THENVOI_API_KEY_USER not set")
 
@@ -114,7 +108,7 @@ async def dynamic_agent(module_user_api_client, request):
     agent = response.data.agent
     credentials = response.data.credentials
 
-    _dynamic_agent = DynamicAgent(
+    agent_info = DynamicAgent(
         agent_id=agent.id,
         agent_name=agent.name,
         api_key=credentials.api_key,
@@ -122,7 +116,7 @@ async def dynamic_agent(module_user_api_client, request):
 
     logger.info("\nCreated dynamic agent: %s (ID: %s)", agent.name, agent.id)
 
-    yield _dynamic_agent
+    yield agent_info
 
     # Cleanup: delete the agent (unless --no-clean mode)
     if not is_no_clean_mode(request):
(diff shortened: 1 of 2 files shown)


amit-gazal-thenvoi (commented):
## Code Review

This PR is well-documented and thoroughly tested. However, I have some feedback and a suggestion for a simpler approach.

---

## 🤔 Architectural Suggestion: Consider a Simpler Approach

If the main goal is `--no-clean` support for debugging, this could be achieved with ~10 lines per fixture instead of ~750 new lines:

```python
@pytest.fixture(scope="module")
async def dynamic_agent(module_user_api_client, request):
    # ... create agent ...
    yield _dynamic_agent
    
    # Simple flag check in existing cleanup
    if not is_no_clean_mode(request):
        await module_user_api_client.human_api.delete_my_agent(id=agent.id, force=True)
```

### Comparison

| Feature | Simple Flag Check | Centralized Tracker |
|---------|-------------------|---------------------|
| `--no-clean` support | ✅ | ✅ |
| Cleanup timing | Per-module | Session-end |
| Code complexity | ~5 lines/fixture | ~750 new lines |
| Thread-safety concerns | None | Yes (singleton) |

### Questions to Consider
1. Do tests across modules need to share dynamically created agents?
2. Is session-scoped cleanup (vs module-scoped) actually needed?
3. Is the resource summary at session end worth the added complexity?

If the answers are "no", the simpler approach would achieve the goal with significantly less code.

tests/conftest_integration.py
```diff
@@ -198,3 +421,126 @@ async def test_peer_id(api_client: AsyncRestClient | None) -> str | None:
...
+    """
+    yield  # Run all tests first
+
+    # Check if cleanup should be skipped
+    no_clean = is_no_clean_mode(request)
+    tracker = resource_tracker
+
+    # Print summary (using print() instead of logging for visibility without -s flag)
+    if tracker.agents:
+        print("=" * 60)
```
amit-gazal-thenvoi: **`print()` statements need `-s` flag anyway**

The comment above explains the rationale (visibility without `-s` flag), but pytest actually captures stdout by default, so users still need `-s` to see this output.

Consider:
- Using `pytest.config.hook.pytest_report_header` or similar pytest hooks for more reliable output
- Or document that `-s` is recommended when running integration tests

amit-gazal-thenvoi (approved):
Looks great

PR #94, opened 2026-01-29
fix(sdk): preserve tool history in Anthropic and Pydantic AI converters

yael-avioz-thenvoi:
## Summary

- Fix bug where Anthropic and Pydantic AI history converters silently dropped `tool_call` and `tool_result` messages during session history conversion
- Fix bug where tool results were not batched (each created a separate message, which "teaches Claude to avoid parallel calls")
- This caused LLMs to lose context about previous tool invocations when sessions were restored

### Tool Call Conversion
- **Anthropic converter**: `tool_call` → assistant message with `tool_use` content blocks
- **Pydantic AI converter**: `tool_call` → `ModelResponse` with `ToolCallPart`
- Multiple consecutive tool calls are batched into a single message

### Tool Result Conversion (Critical Fix)
- **Anthropic converter**: `tool_result` → user message with `tool_result` content blocks
- **Pydantic AI converter**: `tool_result` → `ModelRequest` with `ToolReturnPart`
- **Multiple consecutive tool results are now batched into a single message**

Per [Anthropic docs](https://docs.anthropic.com/en/docs/agents-and-tools/tool-use/implement-tool-use#parallel-tool-calls-not-working), sending separate user messages for each tool result "teaches Claude to avoid parallel calls."

**Before (incorrect):**
```json
[
  {"role": "assistant", "content": [tool_use_1, tool_use_2]},
  {"role": "user", "content": [tool_result_1]},
  {"role": "user", "content": [tool_result_2]}
]
```

**After (correct):**
```json
[
  {"role": "assistant", "content": [tool_use_1, tool_use_2]},
  {"role": "user", "content": [tool_result_1, tool_result_2]}
]
```

### Other Changes
- Added graceful handling for malformed JSON with warning logs
- Updated type aliases to reflect new output formats
- Use `ModelRequest | ModelResponse` instead of `Union[...]` syntax

## Test plan

- [x] Added `test_converts_tool_call_to_assistant_message` / `test_converts_tool_call_to_model_response`
- [x] Added `test_converts_tool_result_to_user_message` / `test_converts_tool_result_to_model_request`
- [x] Added `test_batches_multiple_tool_calls` for both converters
- [x] Added `test_batches_multiple_tool_results` for both converters
- [x] Added `test_handles_malformed_tool_call_json` / `test_handles_malformed_tool_result_json`
- [x] Updated `test_full_conversation_flow` integration tests
- [x] Updated `test_handles_interleaved_tool_calls_and_results` to verify correct batching
- [x] All 56 converter tests pass

Fixes INT-37

### DIFF
--- a/src/thenvoi/converters/pydantic_ai.py
+++ b/src/thenvoi/converters/pydantic_ai.py
@@ -8,6 +8,7 @@ try:
     from pydantic_ai.messages import (
         ModelRequest,
         ModelResponse,
+        RetryPromptPart,
         ToolCallPart,
         ToolReturnPart,
         UserPromptPart,
@@ -36,12 +37,15 @@ def _flush_pending_tool_calls(
 
 
 def _flush_pending_tool_results(
-    messages: PydanticAIMessages, pending_tool_results: list[ToolReturnPart]
+    messages: PydanticAIMessages,
+    pending_tool_results: list[ToolReturnPart | RetryPromptPart],
 ) -> None:
     """Flush pending tool results into a single ModelRequest.
 
     Similar to Anthropic's requirement, tool results should be batched
     together to enable parallel tool use patterns.
+
+    Uses ToolReturnPart for successful results and RetryPromptPart for errors.
     """
     if pending_tool_results:
         messages.append(ModelRequest(parts=list(pending_tool_results)))
@@ -56,12 +60,12 @@ class PydanticAIHistoryConverter(HistoryConverter[PydanticAIMessages]):
     - user messages → ModelRequest with UserPromptPart
     - other agents' messages → ModelRequest with UserPromptPart (with [name] prefix)
     - tool_call → ModelResponse with ToolCallPart
-    - tool_result → ModelRequest with ToolReturnPart
+    - tool_result → ModelRequest with ToolReturnPart (or RetryPromptPart if is_error=True)
     - this agent's text messages → skipped (redundant with tool results)
 
     Tool events are stored in platform as JSON:
     - tool_call: {"name": "...", "args": {...}, "tool_call_id": "..."}
-    - tool_result: {"name": "...", "output": "...", "tool_call_id": "..."}
+    - tool_result: {"name": "...", "output": "...", "tool_call_id": "...", "is_error": bool}
     """
 
     def __init__(self, agent_name: str = ""):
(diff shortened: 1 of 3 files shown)


amit-gazal-thenvoi (changes requested):
## Critical Bug: Tool results are not batched

The PR correctly batches `tool_call` messages into a single assistant message, but **`tool_result` messages are not batched** - each one creates a separate user message. This violates Anthropic's API requirements and will cause issues.

### The Problem

Looking at `test_handles_interleaved_tool_calls_and_results`, the expected output is:
```python
# Expected: [batched tool_use (2), tool_result, tool_result, tool_use, tool_result]
assert len(result) == 5
```

This means after an assistant message with 2 tool_use blocks, there are **two consecutive user messages** with tool_results. This is incorrect.

### What Anthropic Docs Say

From [Anthropic's tool use implementation guide](https://docs.anthropic.com/en/docs/agents-and-tools/tool-use/implement-tool-use#parallel-tool-calls-not-working):

> **Specifically for parallel tool use:**
> - ❌ **Wrong**: Sending separate user messages for each tool result
> - ✅ **Correct**: All tool results must be in a single user message

```json
// ❌ This reduces parallel tool use (current implementation)
[
  {"role": "assistant", "content": [tool_use_1, tool_use_2]},
  {"role": "user", "content": [tool_result_1]},
  {"role": "user", "content": [tool_result_2]}
]

// ✅ Correct format
[
  {"role": "assistant", "content": [tool_use_1, tool_use_2]},
  {"role": "user", "content": [tool_result_1, tool_result_2]}
]
```

The docs explicitly state this incorrect formatting "teaches Claude to avoid parallel calls."

### Suggested Fix

Add a `pending_tool_results` list similar to `pending_tool_calls`, and flush it when:
1. A new `tool_call` arrives (next turn started)
2. A `text` message arrives
3. End of history

### Same Issue in Pydantic AI Converter

The Pydantic AI converter has the same bug - each `ToolReturnPart` creates its own `ModelRequest` instead of batching them.

### Why Tests Didn't Catch This

The tests verify the *current behavior* works, but they don't validate against the *correct format* for the Anthropic API. The test `test_handles_interleaved_tool_calls_and_results` asserts `len(result) == 5`, which encodes the bug as expected behavior.

### Minor Issue

In `pydantic_ai.py`, use `ModelRequest | ModelResponse` instead of `Union[ModelRequest, ModelResponse]` per coding standards.

amit-gazal-thenvoi:
## Suggestion: Add Integration Tests

The current tests are all **unit tests with synthetic data**. Consider adding integration tests that verify the converters work with **real session history** from the platform.

### Why this matters

The unit tests assume a specific JSON format:
```json
{"name": "...", "args": {...}, "tool_call_id": "..."}
```

But if the actual platform stores tool events differently (different field names, nested structure, extra fields), the unit tests would pass while real usage fails.

### Suggested addition

```python
# tests/integration/test_history_converters.py

@pytest.mark.requires_api
async def test_anthropic_converter_with_real_session(api_client):
    """Verify Anthropic converter handles real platform tool history."""
    # 1. Create a chat and trigger a tool call
    # 2. Fetch session history from platform
    # 3. Convert with AnthropicHistoryConverter
    # 4. Verify tool_use and tool_result blocks are present
    # 5. Optionally: send to Anthropic API to verify it's valid
```

This would catch any mismatches between the expected format in the converter and the actual format stored by the platform.

amit-gazal-thenvoi (approved):

PR #139, opened 2026-02-23
fix(langgraph): combine context into user message to avoid Anthropic rejection

roi-shikler-thenvoi:
## Summary
- Anthropic rejects non-consecutive system messages in the message array
- Instead of appending participants and contacts as separate `("system", ...)` messages, they are now merged into the user message with `[Context]` and `[Message]` labels

## Test plan
- [ ] Verify LangGraph adapter works with Anthropic models (no system message rejection)
- [ ] Verify participants and contacts context is still visible to the LLM

### DIFF
--- a/src/thenvoi/adapters/langgraph.py
+++ b/src/thenvoi/adapters/langgraph.py
@@ -160,13 +160,20 @@ class LangGraphAdapter(SimpleAdapter[LangChainMessages]):
             if history:
                 messages.extend(history)  # Already converted by history_converter
 
+        # Combine context (participants, contacts) with the user message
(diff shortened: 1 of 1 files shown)


amit-gazal-thenvoi:
Closing — main already solved this. `src/thenvoi/adapters/langgraph.py:208-218` now sends `participants_msg` and `contacts_msg` as `("user", "[System]: ...")` instead of `("system", ...)`, which avoids the Anthropic non-consecutive-system-message rejection. Different approach than this PR's [Context]/[Message] merge, but same fix. Branch also conflicts.

PR #202, opened 2026-03-30
feat(integrations): deep Codex integration (INT-226)

yael-avioz-thenvoi:
## Summary

- **Structured error reporting**: Parse `codexErrorInfo` from error events and failed turns, emit structured metadata with error type, retryability, and remediation hints (7 known error types mapped)
- **Interactive approvals**: Session-level acceptance via `/approve-session`, approval audit trail as task events, enriched approval request metadata, new `/sandbox` and `/permissions` commands
- **Real-time streaming**: Forward reasoning deltas (`item/reasoning/*Delta`), plan deltas (`item/plan/delta`), and commentary-phase agent messages as incremental thought events with `streaming: true` metadata
- **Plan & task lifecycle**: Forward `turn/plan/updated` with per-step status tracking (pending/inProgress/completed), enriched turn lifecycle events with duration and token usage
- **Diffs & token usage**: Forward `turn/diff/updated` aggregated diffs, track `thread/tokenUsage/updated` per-thread, `/usage` and `/thread info` commands
- **All features opt-in** via new config flags (backwards compatible, all default `False` except `structured_errors=True`)

### New Config Flags

| Flag | Default | Purpose |
|------|---------|---------|
| `structured_errors` | `True` | Parse codexErrorInfo for typed errors + remediation |
| `stream_reasoning_events` | `False` | Stream reasoning deltas as thoughts |
| `stream_plan_events` | `False` | Stream plan deltas + structured plan steps |
| `stream_commentary_events` | `False` | Stream commentary-phase messages as thoughts |
| `emit_diff_events` | `False` | Forward aggregated diffs |
| `emit_token_usage_events` | `False` | Emit token usage task events |
| `emit_turn_lifecycle_events` | `False` | Enriched turn start/complete with duration |
| `client_close_timeout_s` | `10.0` | Bound transport close during cleanup (None = unbounded) |

### New Slash Commands

`/approve-session`, `/sandbox`, `/permissions`, `/threads`, `/thread info`, `/thread archive`, `/usage`

### Files Changed

| File | Changes |
|------|---------|
| `src/thenvoi/adapters/codex.py` | Event handling, config flags, slash commands, bounded cleanup |
| `src/thenvoi/integrations/codex/types.py` | New types: error metadata, plan steps, token usage, audit entries |
| `src/thenvoi/integrations/codex/sdk_client.py` | Optional SDK transport with tracked handler futures for clean shutdown |
| `src/thenvoi/integrations/codex/__init__.py` | Re-exports new public helpers |
| `tests/adapters/test_codex_adapter.py` | 176 tests total (113 on `dev`, +63 in this PR) |

## Review follow-ups addressed

- Added coverage for every remediation entry in `CODEX_ERROR_REMEDIATION` (all 7 types + fallback + non-dict `codexErrorInfo`)
- Added tests for `/thread info` (with and without a mapped thread) and `/permissions` reflecting a per-room sandbox override
- Added malformed-payload tolerance tests (non-dict `error`, missing `items`, non-list `steps`)
- Added cancellation-cleanup test (pending approvals cleared on `on_cleanup`)
- Added `emit_turn_lifecycle_events=False` parity test (no lifecycle task events)
- Replaced brittle `asyncio.sleep(0.01)` handshakes in approval tests with a `_wait_for_pending_approval` helper that polls adapter state
- Exported new helpers (`build_structured_error_metadata`, `parse_plan_steps`, `CodexTokenUsage`, `ApprovalAuditEntry`, `CodexPlanStep`, `CODEX_APPROVAL_METHODS`, `CODEX_ERROR_REMEDIATION`) from `thenvoi.integrations.codex`
- `CodexSdkClient.close()` now cancels any in-flight server-request handler coroutines scheduled via `asyncio.run_coroutine_threadsafe`
- `CodexAdapter.on_cleanup` bounds transport `close()` under `client_close_timeout_s` so `_rpc_lock` can't be held indefinitely by a hung subprocess

## Test plan

- [x] All 176 codex adapter tests pass (113 existing + 63 new)
- [x] All 2520 broader unit tests pass
- [x] Conformance tests pass for codex adapter
- [x] Ruff linting + formatting clean
- [ ] Verify structured error rendering in Thenvoi chat UI
- [ ] Verify streaming thought events aggregate correctly in UI
- [ ] Verify plan step checklist renders in UI
- [ ] E2E test with live Codex app-server


darvell-thenvoi:
Pausing review on this until INT-294 is handled since this will get affected.

src/thenvoi/integrations/codex/sdk_client.py
```diff
@@ -0,0 +1,433 @@
...
+        """Called by the SDK on the thread-pool thread for server requests.
+
+        If we are inside a ``request()`` call (``_in_request`` is True),
+        auto-handle to avoid deadlock — the adapter hasn't entered its
+        event loop yet so nobody would process the request.
+
+        Otherwise, schedule the adapter's async handler on the event loop,
+        block this thread until a result is ready, and return it to the SDK.
+        """
+        if self._in_request.is_set():
```
darvell-thenvoi, line 313: **MUST FIX — TOCTOU race between `close()` and `_server_request_bridge`**

This bridge runs on the SDK's callback thread while `close()` runs on the event loop. There's no `_connected` check at the top of this method, and even if there were, it would be TOCTOU vs. the insert at line 307.

Timeline:
1. SDK thread enters here, passes the `_in_request` / `_loop` / `_request_handler` checks
2. Event loop thread enters `close()`, flips `_connected = False`, acquires `_pending_lock`, iterates + clears `_pending_server_responses`, releases lock (sdk_client.py:258-262)
3. SDK thread acquires `_pending_lock` and inserts `(future, method)` into the just-cleared dict (line 307)
4. SDK thread schedules `_run_handler` via `run_coroutine_threadsafe`, blocks on `response_future.result(timeout=self._server_request_timeout_s)` — 300s default
5. `close()` proceeds to nil `self._sync_client`; the scheduled handler runs on a torn-down client
6. SDK callback thread sits blocked until the 300s timeout fires

Under "close while Codex is asking for approval" this fires every time, not occasionally.

**Fix:** check `_connected` at the top AND recheck under `_pending_lock` right before the insert:

```python
if not self._connected:
    return self._auto_handle_server_request(method, params)
if self._in_request.is_set():
    return self._auto_handle_server_request(method, params)
if self._loop is None or self._request_handler is None:
    return self._auto_handle_server_request(method, params)

req_id = self._allocate_synthetic_id()
response_future: concurrent.futures.Future[dict[str, Any]] = concurrent.futures.Future()
with self._pending_lock:
    if not self._connected:
        return self._auto_handle_server_request(method, params)
    self._pending_server_responses[req_id] = (response_future, method)
```

`_auto_handle_server_request` already returns `decline` for approval methods, so there's no silent-accept risk. Safe because `close()` already uses `_safe_default_response` for the same reason — no behavior drift.

src/thenvoi/adapters/codex.py
```diff
@@ -436,168 +568,417 @@ async def on_message(
...
-                turn_status=turn_status,
-                turn_error=turn_error,
-                final_text=final_text,
-                saw_send_message_tool=saw_send_message_tool,
+                if event.method == "transport/closed":
+                    result.turn_status = "failed"
+                    result.turn_error = "Codex transport closed unexpectedly"
+                    # Reset client state so _ensure_client_ready() rebuilds
+                    # on the next message instead of reusing a dead client.
+                    self._client = None
```
darvell-thenvoi, line 943: **SHOULD FIX — transport/closed leaves stale per-room state**

This handler nulls `_client` / `_initialized` but leaves `_room_threads`, `_token_usage`, `_prompt_injected_rooms`, `_needs_history_injection`, `_pending_approvals`, `_session_approved`, `_sandbox_overrides`, `_raw_history_by_room` intact.

The resume path at `_ensure_thread` (codex.py:1137) exists but isn't reached: line 1130 returns the cached `thread_id` unconditionally before the `is_session_bootstrap and history.has_thread()` branch. So after a transport-closed event, the next non-bootstrap turn sends `turn/start` for a thread_id the newly spawned subprocess doesn't know about. Token usage recorded under the old thread_id stays orphaned, and any futures in `_pending_approvals` from the dead session sit unresolved until per-room cleanup.

**Fix:** in this branch, clear `_room_threads` and `_raw_history_by_room` for every room and call `_clear_pending_approvals_for_room` for each — the pending-approval clear already resolves futures to `"decline"` (codex.py:3217-3221), so awaiters unblock cleanly. That way the next turn either resumes via `thread/resume` (if it's a bootstrap with `has_thread()`) or does a clean `thread/start`. Add a test that asserts recovery after `transport/closed` does a fresh thread/start rather than reusing the cached id.

src/thenvoi/adapters/codex.py
```diff
@@ -436,168 +568,417 @@ async def on_message(
...
+        manager guarantees cleanup even on exceptions, preventing stale
+        context from routing server requests to the wrong room.
+        """
+        self._sdk_request_context = (tools, msg, room_id)
+        try:
+            yield
+        finally:
+            self._sdk_request_context = None
+
+    async def _process_turn_events(
```
darvell-thenvoi, line 707: **NICE TO HAVE — `_process_turn_events` dispatch**

This method is a ~280-line `if event.method == "..."` chain (runs to ~line 940). Most branches already delegate to extracted handlers (`_forward_plan_steps`, `_forward_diff_event`, `_emit_token_usage_event`, `_handle_error_event`). The `Emit.TASK_EVENTS in self.features.emit` guard is re-checked inline at six sites and easy to drift.

A dispatch table (`method → handler`) with the feature-flag check moved into each handler (or a decorator) would make adding event types safer. Pure helpers like `_approval_summary` (3039+), `_approval_type`, `_session_approval_key`, `_canonical_sandbox_key`, `_normalize_sandbox_mode/_policy`, `_extract_turn_error`, `_extract_file_change_paths` are adapter-state-free and would fit naturally in `integrations/codex/`. The ~280-line `_handle_local_command` slash-command dispatcher is almost entirely UX string-building and could live alongside.

Not a blocker — `adapters/codex.py` is 3238 lines, and the next PR will pay twice if this doesn't happen soon.

src/thenvoi/integrations/codex/sdk_client.py
```diff
@@ -0,0 +1,433 @@
...
+        client_name: str = "thenvoi_codex_adapter",
+        client_title: str = "Thenvoi Codex Adapter",
+        client_version: str = "0.1.0",
+        experimental_api: bool = True,
+        server_request_timeout_s: float = _DEFAULT_SERVER_REQUEST_TIMEOUT_S,
+    ) -> None:
+        try:
+            from codex_app_server import AppServerClient, AppServerConfig  # type: ignore[missing-import]
+        except ImportError as exc:
+            if sys.version_info < (3, 12):
```
darvell-thenvoi, line 70: **NICE TO HAVE — Python 3.12 gating surfaces late**

`codex-app-server-sdk` is pinned with `python_version >= '3.12'` in pyproject. A user on 3.11 who sets `transport="sdk"` gets a late `ImportError` at `__getattr__` lazy-load time, not at config validation.

Raise `ValueError` at config resolution with a clear "CodexSdkClient requires Python 3.12+" message so the failure lands at startup rather than deep in the event loop.

tests/adapters/test_codex_adapter.py
```diff
@@ -3479,3 +3529,3416 @@ async def execute_tool_call(
...
+                contacts_msg=None,
+                is_session_bootstrap=True,
+                room_id="room-1",
+            )
+        )
+        await _wait_for_pending_approval(adapter, "room-1")
+
+        await adapter.on_cleanup("room-1")
+
+        assert adapter._pending_approvals.get("room-1", {}) == {}
```
darvell-thenvoi, line 6978: **SHOULD FIX — test doesn't distinguish cleanup from natural timeout**

This test claims to verify "pending approvals cleared on `on_cleanup`" but uses `approval_wait_timeout_s=0.05`, so the natural approval timeout almost always fires before `on_cleanup` acquires `_rpc_lock`. The assertion `_pending_approvals.get("room-1", {}) == {}` would pass whether cleanup cleared the dict or the timeout did — `_clear_pending_approvals_for_room` (codex.py:3217-3221) could be deleted and the test would still pass.

The next assertion `"room-1" not in adapter._room_threads` (line 6865) *is* cleanup-specific because only `on_cleanup` pops `_room_threads`, so the test isn't entirely toothless — but the headline property it claims to verify isn't actually verified.

**Fix:** bump the timeout (e.g. `approval_wait_timeout_s=30.0`), capture the pending future before calling cleanup, assert `on_cleanup` returns quickly AND `future.result() == "decline"` (set by `_clear_pending_approvals_for_room`, not by natural timeout).

src/thenvoi/adapters/codex.py
```diff
@@ -436,168 +568,417 @@ async def on_message(
...
+                        exc_info=True,
+                    )
+            result.turn_status = "interrupted"
+            result.turn_error = "Turn timed out"
+        return result
 
     async def on_cleanup(self, room_id: str) -> None:
         async with self._rpc_lock:
-            self._room_threads.pop(room_id, None)
+            thread_id = self._room_threads.pop(room_id, None)
```
darvell-thenvoi, line 1011: **SHOULD FIX — `_rpc_lock` scope blocks cleanup behind approvals**

`_rpc_lock` is adapter-wide and held by `on_message` for the entire turn, including the up-to-300s wait at `_resolve_manual_approval` (codex.py:2003). So when room B calls `on_cleanup` while room A is waiting on a human `/approve` click, cleanup blocks for up to `approval_wait_timeout_s` before it can reclaim room B's state.

The docstring at lines 379-384 mentions approvals blocking *turn processing* in other rooms but doesn't mention they also block *cleanup*.

Either:
- Document the behavior so operators understand cleanup latency, or
- Make per-room state mutations lock-free and only take `_rpc_lock` for the final `_client.close()` in the last-room path.

Related: `self._pending_approvals.clear()` at line 977 (inside the `finally`) globally wipes approvals for ALL rooms on the last-room cleanup path. That's defensive rather than a bug, but if any prior room's cleanup failed this masks the leak. Worth a one-line comment explaining intent.

darvell-thenvoi (commented):
Deep integration is solid — ticket criteria are implemented, tests are well-structured, and the `_wait_for_pending_approval` helper is a textbook fix for the flaky-polling pattern this repo has rejected before. Leaving a COMMENT review with findings grouped below; nothing here should block merge individually but I'd like the SDK bridge race and transport/closed recovery addressed before this lands.

**MUST FIX**
- sdk_client.py:295 — TOCTOU race between `close()` and `_server_request_bridge`

**SHOULD FIX**
- codex.py:898 — `transport/closed` leaves per-room state pointing at a dead thread_id
- codex.py:943 — `_rpc_lock` is adapter-wide so `on_cleanup` blocks behind another room's 300s approval wait
- test_codex_adapter.py:6864 — approval-cleanup test doesn't distinguish cleanup from natural timeout

**NICE TO HAVE**
- codex.py:662 — 280-line if/elif dispatch in `_process_turn_events` → dispatch table
- codex.py:138 (not in diff) — `_CodexClientProtocol` duplicates `BaseJsonRpcClient` interface. Consider moving into `integrations/codex/` next to the clients and adding a conformance test so both `CodexStdioClient` and `CodexSdkClient` are checked against one source of truth
- sdk_client.py:70 — codex-app-server-sdk Python-version error should surface at config validation, not lazy import
- Nothing shared with `claude_sdk.py` — approval audit, `/approve-session`, token-usage accumulation could all live in `runtime/` so other adapters inherit parity (claude_sdk.py:86-92 already has a parallel fork with a `# mirrors Codex adapter patterns` comment)

On PR body hygiene: test count ("176 total, +63") and files-changed table (5 rows) are stale — actual is 173 tests (+106 vs dev baseline of 67) and 9 files changed. Worth a quick update so reviewers can size the PR correctly.

src/thenvoi/integrations/codex/sdk_client.py
yael-avioz-thenvoi: Fixed in 9cc8427. Added `_connected` check before the early guards and a re-check under `_pending_lock` right before the insert, exactly as you suggested. The double-check closes the TOCTOU window — `close()` flips `_connected=False` and drains `_pending_server_responses` under the same lock, so a concurrent bridge call either exits early (first check) or sees the flip under the lock (second check). Also updated `test_server_request_bridge_auto_handles_during_request` which creates `CodexSdkClient` via `__new__` and needed `_connected = True` set.

src/thenvoi/adapters/codex.py
yael-avioz-thenvoi: Fixed in 9cc8427. On `transport/closed`, we now clear `_room_threads`, `_raw_history_by_room`, and call `_clear_pending_approvals_for_room` for each stale room. This ensures the next turn does a fresh `thread/start` instead of reusing a cached thread_id the new subprocess does not know about. Pending approval futures are resolved to `"decline"` so awaiters unblock cleanly. Added `test_transport_closed_clears_per_room_state` that pre-populates per-room state, triggers transport/closed, and verifies the state is cleared.

tests/adapters/test_codex_adapter.py
yael-avioz-thenvoi: Fixed in 9cc8427. Rewrote the test to directly populate adapter state (`_room_threads`, `_pending_approvals` with a real `_PendingApproval` + future) and call `on_cleanup` without going through `on_message`. This isolates cleanup behavior from `_rpc_lock` contention and verifies `_clear_pending_approvals_for_room` resolves the future to `"decline"` — the old 0.05s timeout meant the natural timeout always won the race.

src/thenvoi/adapters/codex.py
yael-avioz-thenvoi: Addressed in 9cc8427. Added a comment documenting the known cleanup latency limitation (cleanup for room B blocks if room A holds the lock during a pending approval). Also added a comment explaining the defensive `_pending_approvals.clear()` in the last-room finally block — it catches any leaked entries from rooms whose per-room cleanup failed, since `_clear_pending_approvals_for_room` should have already resolved them. Deferring the per-room lock-free approach to a follow-up since it requires more structural changes.

src/thenvoi/adapters/codex.py
yael-avioz-thenvoi: Agreed — the dispatch table would reduce drift risk and make the method easier to navigate. Deferring to a follow-up PR to keep this one focused on the review fixes.

src/thenvoi/integrations/codex/sdk_client.py
yael-avioz-thenvoi: Good call. The current code already raises a clear error with Python version info when the import fails (sdk_client.py:70-78), but surfacing it at config validation rather than lazy import time would be better UX. Will pick this up in a follow-up.

darvell-thenvoi:
LGTM after merge conflict fix.

amit-gazal-band (approved):
Approving after conflict fixes

PR #207, opened 2026-04-05
feat: add thenvoi-trigger CLI for process initialization (INT-227)

yael-avioz-thenvoi:
## Summary
- Adds `thenvoi-trigger` CLI command (`src/thenvoi/cli/trigger.py`) for process initialization
- Creates a chatroom, adds a target agent as participant, and sends a message — designed for cron jobs, CI/CD, and GitHub Actions
- Supports both agent and user authentication modes via `--auth-mode`
- Configuration via CLI arguments with environment variable fallbacks (`THENVOI_API_KEY`, `THENVOI_TARGET_HANDLE`, `THENVOI_MESSAGE`, etc.)
- Registered as `thenvoi-trigger` console script entry point in `pyproject.toml`

## Usage
```bash
thenvoi-trigger \
    --api-key "$THENVOI_API_KEY" \
    --target-handle "@owner/my-agent" \
    --message "Run the daily report"
```

## Test plan
- [x] Manually tested against live platform (agent auth mode)
- [x] Verified peer lookup with `--verbose` logging
- [x] Fixed handle matching (API returns handles without `@` prefix)
- [ ] Add unit tests with mocked REST client
- [ ] Test user auth mode against live platform

Closes INT-227

### DIFF
--- a/src/thenvoi/cli/trigger.py
+++ b/src/thenvoi/cli/trigger.py
@@ -205,87 +205,90 @@ async def run(args: argparse.Namespace) -> str:
 
     client = AsyncRestClient(api_key=args.api_key, base_url=args.rest_url.rstrip("/"))
 
-    # Step 1: Look up the target peer by handle
-    logger.info("Looking up peer with handle: %s", args.target_handle)
-    peer = await find_peer_by_handle(client, args.target_handle, args.auth_mode)
-    if not peer:
-        raise ValueError(
-            f"Target agent with handle '{args.target_handle}' not found. "
-            "Verify the handle is correct and that you have access to this peer."
-        )
-    logger.info("Found peer: %s (id=%s)", peer["name"], peer["id"])
-
-    # Step 2: Create a new chatroom
-    logger.info("Creating chatroom...")
     try:
-        if args.auth_mode == "agent":
-            chat_response = await client.agent_api_chats.create_agent_chat(
-                chat=ChatRoomRequest(),
-                request_options=DEFAULT_REQUEST_OPTIONS,
+        # Step 1: Look up the target peer by handle
+        logger.info("Looking up peer with handle: %s", args.target_handle)
+        peer = await find_peer_by_handle(client, args.target_handle, args.auth_mode)
+        if not peer:
+            raise ValueError(
+                f"Target agent with handle '{args.target_handle}' not found. "
+                "Verify the handle is correct and that you have access to this peer."
             )
-        else:
-            chat_response = await client.human_api_chats.create_my_chat_room(
-                chat=CreateMyChatRoomRequestChat(),
-                request_options=DEFAULT_REQUEST_OPTIONS,
-            )
-    except ApiError as e:
-        raise RuntimeError(_format_api_error(e, "create chatroom")) from e
-    room_id = chat_response.data.id
-    logger.info("Created chatroom: %s", room_id)
+        logger.info("Found peer: %s (id=%s)", peer["name"], peer["id"])
 
-    # Step 3: Add the target agent as a participant
-    # Step 4: Send the message mentioning the target agent
-    # Wrapped in try/except so we log the orphan room ID on partial failure.
-    try:
-        logger.info("Adding %s to chatroom...", peer["name"])
-        if args.auth_mode == "agent":
-            await client.agent_api_participants.add_agent_chat_participant(
-                chat_id=room_id,
-                participant=ParticipantRequest(participant_id=peer["id"]),
-                request_options=DEFAULT_REQUEST_OPTIONS,
-            )
-        else:
-            await client.human_api_participants.add_my_chat_participant(
-                chat_id=room_id,
-                participant=ParticipantRequest(participant_id=peer["id"]),
-                request_options=DEFAULT_REQUEST_OPTIONS,
-            )
-        logger.info("Added participant: %s", peer["name"])
+        # Step 2: Create a new chatroom
+        logger.info("Creating chatroom...")
+        try:
+            if args.auth_mode == "agent":
+                chat_response = await client.agent_api_chats.create_agent_chat(
+                    chat=ChatRoomRequest(),
+                    request_options=DEFAULT_REQUEST_OPTIONS,
+                )
+            else:
+                chat_response = await client.human_api_chats.create_my_chat_room(
+                    chat=CreateMyChatRoomRequestChat(),
+                    request_options=DEFAULT_REQUEST_OPTIONS,
+                )
+        except ApiError as e:
+            raise RuntimeError(_format_api_error(e, "create chatroom")) from e
+        room_id = chat_response.data.id
+        logger.info("Created chatroom: %s", room_id)
 
-        logger.info("Sending message...")
-        mention = ChatMessageRequestMentionsItem(
-            id=peer["id"],
-            handle=peer["handle"],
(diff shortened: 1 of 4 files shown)


src/thenvoi/cli/trigger.py
```diff
@@ -0,0 +1,328 @@
...
+    client: AsyncRestClient,
+    handle: str,
+    auth_mode: str,
+) -> dict[str, str] | None:
+    """
+    Paginate through peers to find one matching the given handle.
+
+    Returns dict with 'id', 'name', 'handle' or None if not found.
+    """
+    handle = handle.lstrip("@").lower()
```
darvell-thenvoi: This only normalizes the caller input. The comparison a few lines below still uses the raw `peer.handle` value from the API, so `@owner/agent` from the backend will not match `owner/agent` from the CLI even though they're the same logical handle. The rest of the SDK already treats `@handle` and `handle` as equivalent. I think both sides need to be normalized here, ideally by reusing the repo's existing handle-normalization convention instead of open-coding a one-sided `lstrip('@')`.

tests/cli/test_trigger.py
```diff
@@ -0,0 +1,631 @@
...
+        client.human_api_peers.list_my_peers.return_value = _make_peers_response([peer])
+
+        result = await find_peer_by_handle(client, "@owner/agent", "user")
+
+        assert result is not None
+        assert result["id"] == "peer-1"
+        client.human_api_peers.list_my_peers.assert_called_once()
+
+    @pytest.mark.asyncio
+    async def test_case_insensitive_matching(self):
```
darvell-thenvoi, line 161: Nice coverage on input normalization, but we're still missing the regression that matters for the backend behavior called out in the PR description. I'd add a test where the API returns `peer.handle = '@owner/agent'` and the CLI input is `@owner/agent` or `owner/agent'`, so we prove both spellings still resolve to the same peer. Right now the tests only prove the input side is normalized.

darvell-thenvoi (changes requested):
I like the overall shape here. The CLI is focused, the agent/user split is straightforward, and the targeted test file is in much better shape than the PR body suggested — I ran `tests/cli/test_trigger.py` locally and it passed cleanly.

I still don't think this is ready to merge yet because handle normalization is only fixed in one direction. The script strips `@` from the input handle, but it compares that against the raw peer handle from the API. If the backend returns `@owner/agent` instead of `owner/agent`, this still fails even though the caller supplied a valid handle. The rest of the SDK already treats `@handle` and `handle` as equivalent, so this CLI should too.

I'd fix that by normalizing both sides (ideally using the shared handle-normalization conventions already in the repo), then add a regression test that proves API responses with an `@`-prefixed handle still match.

src/thenvoi/cli/trigger.py
yael-avioz-thenvoi: Fixed in 972460f — both sides now use the shared `normalize_handle` utility from `thenvoi.runtime.types`. The input handle and each `peer.handle` from the API are both normalized before comparison (lines 140 and 161).

tests/cli/test_trigger.py
yael-avioz-thenvoi: Added in 972460f — two regression tests now cover this:
- `test_matches_when_api_returns_at_prefixed_handle` — API returns `@owner/agent`, input is `@owner/agent`
- `test_matches_when_api_returns_at_prefix_and_input_has_none` — API returns `@owner/agent`, input is `owner/agent` (no `@`)

Both prove that `normalize_handle` is applied to the API side as well.

amit-gazal-thenvoi (changes requested):
Re-reviewed. Handle normalization fix landed cleanly and the regression
tests at tests/cli/test_trigger.py:563-589 cover both spellings against
the new normalize_handle path. Manually ran the agent-mode flow against
prod (peer lookup → create room → add participant → send message): all
4 calls 201, ~1.3s, clean stdout room ID, exit 0.

Requesting changes for one reason: every test in tests/cli/test_trigger.py
is fully mocked. AsyncRestClient is patched at the class level, peer/chat
responses are MagicMock + SimpleNamespace, and asyncio.run is faked.
That means the suite passes even if:

- the backend renames peer.handle / peer.id / response.data.id
- list_agent_peers changes its pagination contract (page/page_size → cursor)
- ChatRoomRequest gains a required field
- ApiError.body shape changes (currently the test constructs
  SimpleNamespace(error=SimpleNamespace(message=...)) — _format_api_error
  silently falls back to "HTTP <code>" if reality differs)
- the endpoint path moves
- the auth header format changes

A CLI built for cron / CI / GitHub Actions is exactly the surface where
mock drift bites quietly: the script keeps "succeeding" (or silently
posting messages without mentions) until someone notices.

Please add tests/integration/test_trigger.py following the existing
pattern (tests/integration/conftest.py + tests/conftest_integration.py):

- @requires_api: agent-mode happy path. Use api_client / shared_agent1_info
  / test_peer_id fixtures. Spawn the CLI via subprocess (or call run()
  directly with an argparse.Namespace), assert the room is created, the
  participant is added, and the message is delivered. Clean up the room
  after (respect is_no_clean_mode).
- @requires_user_api + @requires_multi_agent: user-mode happy path against
  the real human_api_* endpoints. This also closes test-plan checkbox 4
  in the PR description, which is currently unchecked.
- @requires_api: peer-not-found error path against real list-peers (assert
  ValueError + non-zero exit).
- @requires_api: real ApiError shape — trigger one (e.g. malformed
  participant_id or invalid handle) and assert _format_api_error extracts
  the human-readable message rather than falling back to "HTTP <code>".

Skipped in CI by default, runnable locally via .env.test, same as every
other integration suite in this repo. Once those land I'll re-approve.

yael-avioz-thenvoi (commented):
Added integration tests in `tests/integration/test_trigger.py` (f9abd1b) to address your review.

5 tests covering the cases you outlined:

- **Agent-mode happy path** (`@requires_api`): calls `run()` with real credentials, verifies room creation, participant presence, and message delivery. Cleans up via `delete_agent_chat` (respects `is_no_clean_mode`).
- **Peer-not-found error path** (`@requires_api`): hits real `list_agent_peers` with a bogus handle, asserts `ValueError`.
- **`find_peer_by_handle` against real API** (`@requires_api`): resolves a real peer and asserts ID matches.
- **Real `ApiError` shape** (`@requires_api`): sends an invalid `participant_id` to trigger a real 4xx, then asserts `_format_api_error` extracts the structured message (not the generic "HTTP <code>" fallback).
- **User-mode happy path** (`@requires_user_api`): full flow against `human_api_*` endpoints.

All tests follow the existing `conftest_integration.py` pattern (skip markers, `.env.test` credentials, `is_no_clean_mode` cleanup).

darvell-thenvoi (approved):

amit-gazal-thenvoi (approved):
Integration tests address the mock-drift concern. All five requested
cases covered:

- Agent-mode happy path: room creation, participant verification,
  message delivery, cleanup.
- User-mode happy path: human_api endpoints with Enterprise plan skip.
- Peer-not-found: real list_agent_peers with bogus handle → ValueError.
- Real ApiError shape: _format_api_error against actual Fern error body.
- find_peer_by_handle against real API: ID match assertion.

Follows existing integration test patterns (conftest_integration,
.env.test, skip markers, is_no_clean_mode). LGTM.

PR #210, opened 2026-04-06
fix: match participants by handle, name, or ID instead of name only (INT-287)

yael-avioz-thenvoi:
## Summary

- **`thenvoi_add_participant`** and **`thenvoi_remove_participant`** now match by handle, name, or ID instead of display name only — fixing failures when multiple users share the same name or when the LLM passes a handle/ID
- Renamed the tool parameter from `name` to `identifier` across all adapters (core, LangGraph, Pydantic AI, CrewAI, Parlant, Claude SDK) so the LLM knows it can pass handles
- Added `_matches_identifier()` helper that checks handle → name → ID (case-insensitive)

Closes INT-287

## Test plan

- [x] Existing unit tests updated and passing (1507 pass, 7 pre-existing ACP e2e timeouts unrelated)
- [x] New tests added: match by handle, match by ID, already-in-room detection by handle
- [x] All 307 framework conformance tests pass
- [x] Ruff check + format clean


src/thenvoi/runtime/tools.py
```diff
@@ -26,6 +26,16 @@
 logger = logging.getLogger(__name__)
 
 
+def _matches_identifier(entity: dict[str, Any], identifier: str) -> bool:
```
darvell-thenvoi: `_matches_identifier()` still does raw case-insensitive equality on `handle`, `name`, and `id`, but it doesn't normalize `@` prefixes on handles. Elsewhere in this file we already treat `@alice` and `alice` as equivalent in `_resolve_mentions()`. That means the new "match by handle" behavior still depends on one exact handle spelling. If the backend returns `@handle` and the caller passes `handle`, this still misses. I think this helper should normalize handles the same way mention resolution does, and we should add a test that proves `@user-one` and `user-one` both match.

src/thenvoi/runtime/tools.py
```diff
@@ -79,24 +89,32 @@ class SendEventInput(BaseModel):
...
 
 class AddParticipantInput(BaseModel):
-    """Add a participant (agent or user) to the chat room by name.
+    """Add a participant (agent or user) to the chat room.
 
     IMPORTANT: Use thenvoi_lookup_peers() first to find available agents.
     """
 
-    name: str = Field(
+    identifier: str = Field(
```
darvell-thenvoi, line 124: This rename from `name` to `identifier` looks right long-term, but right now it's a hard breaking change. `execute_tool_call()` now validates against `AddParticipantInput(identifier=...)` / `RemoveParticipantInput(identifier=...)`, so older payloads like `{"name": "alice"}` now fail validation instead of being accepted. Direct Python callers using `add_participant(name=...)` / `remove_participant(name=...)` also break immediately. We already have a soft-migration pattern in this module for deprecated tool names, so I think we need the same kind of compatibility shim here for one release cycle, either via a Pydantic alias or by translating `name -> identifier` at the validation boundary.

src/thenvoi/runtime/prompts.py
```diff
@@ -31,7 +31,7 @@
 or any current information you cannot answer directly:
 
 1. Call `thenvoi_lookup_peers()` to find available specialized agents
-2. If a relevant agent exists, call `thenvoi_add_participant(name)` to add them
+2. If a relevant agent exists, call `thenvoi_add_participant(identifier)` to add them (use the handle from lookup_peers)
```
darvell-thenvoi: Updating the prompt contract here is good, but the example a few lines below still teaches `thenvoi_add_participant("Weather Agent")`, which keeps nudging the model back toward display-name usage. Since INT-287 is specifically about name-only matching being fragile, I'd switch that example to a handle-based call too. Also, `src/thenvoi/integrations/claude_sdk/prompts.py` still shows the old add/remove payloads with `name`, so the prompt/example surfaces are not fully in sync yet.

tests/runtime/test_tools.py
```diff
@@ -232,6 +272,36 @@ async def test_add_participant_success(self, mock_rest_client):
         assert result["status"] == "added"
         mock_rest_client.agent_api_participants.add_agent_chat_participant.assert_called_once()
 
+    async def test_add_participant_by_handle(self, mock_rest_client):
```
darvell-thenvoi, line 315: These new tests cover name / handle / ID lookup, which is good, but they still don't reproduce the actual ticket failure mode: two peers with the same display name and different handles/IDs. I'd like at least one test that proves the ambiguous-name case is now handled correctly, otherwise we're proving broader matching without proving the original bug is fixed.

darvell-thenvoi (changes requested):
I like the direction here. Fixing this centrally in `runtime/tools.py` and pushing the schema change through the adapters is the right shape; doing it separately per adapter would have been the wrong layer.

I don't think this is ready to merge yet. The main blocker is that `name -> identifier` landed as a hard public-contract break with no compatibility path. Older tool payloads like `{"name": "alice"}` now fail validation, and direct Python callers using `add_participant(name=...)` / `remove_participant(name=...)` also break immediately.

There's also still a correctness gap in handle matching: `_matches_identifier()` doesn't normalize `@` prefixes, so the new handle-matching path still depends on one exact handle spelling even though the rest of the runtime already treats `@handle` and `handle` as equivalent in mention resolution.

I also want one more pass on the prompt/example surfaces. `src/thenvoi/integrations/claude_sdk/prompts.py` still shows the old add/remove payloads with `name`, and the Parlant examples still teach the old parameter too.

Last thing: the added tests are helpful, but I'd want one explicit duplicate-display-name case to prove the original INT-287 failure mode is actually covered.

src/thenvoi/runtime/tools.py
yael-avioz-thenvoi: Good catch — fixed in e3be3fe. `_matches_identifier()` now strips the `@` prefix on both sides via a shared `_normalize_handle()` helper. Added a test that proves `@user-one` and `user-one` both match regardless of which side has the prefix.

src/thenvoi/runtime/tools.py
yael-avioz-thenvoi: Agreed — added backward compat via Pydantic `AliasChoices("identifier", "name")` on both `AddParticipantInput` and `RemoveParticipantInput`. Old payloads like `{"name": "alice"}` now validate successfully and map to the `identifier` field. `model_dump()` always outputs `identifier`, and the JSON schema advertises `identifier` to new callers. Added tests confirming the legacy `name` key is accepted.

src/thenvoi/runtime/prompts.py
yael-avioz-thenvoi: Fixed both — the weather example in `prompts.py` now uses `thenvoi_add_participant("@john/weather-agent")` instead of the display name, and `claude_sdk/prompts.py` examples now use `"identifier": "@john/weather-agent"` for both add and remove payloads.

tests/runtime/test_tools.py
yael-avioz-thenvoi: Added `test_add_participant_ambiguous_name_resolved_by_handle` — sets up two peers both named "Weather Agent" with different handles (`@alice/weather` and `@bob/weather`), then proves that passing `@bob/weather` as the identifier correctly picks agent-b. This directly reproduces the INT-287 scenario.

darvell-thenvoi (dismissed):
After merge conflicts resolved, approved.

darvell-thenvoi (approved):

PR #211, opened 2026-04-07
feat: add foundation types for SDK normalization (INT-294 Step 1)

darvell-thenvoi:
## Summary
- Add `Capability` and `Emit` string enums to `core/types.py`
- Add `AdapterFeatures` frozen dataclass with frozenset/tuple normalization
- Add `ThenvoiError` exception hierarchy in `core/exceptions.py`

**Part 1 of 6** for INT-294 SDK Cleanup & Normalization.

## Changes
| File | Change |
|------|--------|
| `src/thenvoi/core/types.py` | `Capability`, `Emit` enums + `AdapterFeatures` dataclass |
| `src/thenvoi/core/exceptions.py` | New: `ThenvoiError`, `ThenvoiConfigError`, `ThenvoiConnectionError`, `ThenvoiToolError` |
| `tests/core/test_types.py` | 13 tests for enums and AdapterFeatures |
| `tests/core/test_exceptions.py` | 5 tests for exception hierarchy |

## Test plan
- [x] `uv run pytest tests/core/test_types.py tests/core/test_exceptions.py -v` — 18/18 pass
- [x] Full suite: 2453 passed, same pre-existing failures only
- [x] `uv run ruff check . && uv run ruff format .` — clean
- [x] `uv run pyrefly check` — clean

### DIFF
--- a/src/thenvoi/core/exceptions.py
+++ b/src/thenvoi/core/exceptions.py
@@ -1,0 +1,19 @@
+"""Thenvoi SDK exception hierarchy."""
+
+from __future__ import annotations
+
+
+class ThenvoiError(Exception):
+    """Base for all SDK exceptions."""
+
+
+class ThenvoiConfigError(ThenvoiError):
+    """Configuration or setup problems. Actionable by developer."""
+
+
+class ThenvoiConnectionError(ThenvoiError):
+    """Transport failures (WebSocket, REST). Actionable by ops."""
+
+
+class ThenvoiToolError(ThenvoiError):
+    """Tool execution failures. Actionable by adapter/LLM."""
--- a/src/thenvoi/core/exceptions.py
+++ b/src/thenvoi/core/exceptions.py
@@ -2,13 +2,42 @@
 
 from __future__ import annotations
 
+from collections.abc import Iterable
+
 
 class ThenvoiError(Exception):
     """Base for all SDK exceptions."""
 
 
 class ThenvoiConfigError(ThenvoiError):
-    """Configuration or setup problems. Actionable by developer."""
+    """Configuration or setup problems. Actionable by developer.
+
+    Use ``with_suggestion`` to attach a "did you mean?" hint when the
+    user likely typoed a known parameter or capability name.
+    """
+
+    @classmethod
+    def with_suggestion(
+        cls,
+        message: str,
+        bad_name: str,
+        valid_names: Iterable[str],
+        *,
+        max_distance: int = 2,
+    ) -> "ThenvoiConfigError":
+        """Build an error message with a typo suggestion if one is close enough.
+
+        Args:
+            message: Base error message.
+            bad_name: The unknown / misspelled name the user supplied.
+            valid_names: Known-good names to compare against.
+            max_distance: Maximum Levenshtein distance to consider a match
+                (default 2 — catches single-char typos and small swaps).
+        """
+        suggestion = _closest_match(bad_name, valid_names, max_distance=max_distance)
+        if suggestion is not None:
+            return cls(f"{message} Did you mean {suggestion!r}?")
+        return cls(message)
 
 
 class ThenvoiConnectionError(ThenvoiError):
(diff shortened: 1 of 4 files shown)


jakub-hajto-thenvoi (approved):
## Review: feat: add foundation types for SDK normalization (INT-294 Step 1)

### Non-Blocking Observations (P3)

1. **`core/__init__.py` not updated**: The new types (`Capability`, `Emit`, `AdapterFeatures`) and exceptions (`ThenvoiError`, `ThenvoiConfigError`, etc.) are not re-exported from `core/__init__.py`. Consumers currently import from `thenvoi.core.types` and `thenvoi.core.exceptions` directly, which is fine — but if you want these in the public API surface of `thenvoi.core`, the `__init__.py` and `__all__` should be updated in a later step.

2. **`include_categories` has no existing backing concept**: The `include_categories` field on `AdapterFeatures` references a "category" concept that doesn't exist yet in `runtime/tools.py`. This is fine for a forward-looking foundation type — just noting it for Step 2+ awareness.

3. **`with_suggestion` return type uses string literal**: `-> "ThenvoiConfigError"` instead of `-> Self` (Python 3.11+ `typing.Self`). Since the SDK targets 3.10+, the string-literal forward ref is the right call. If you ever bump minimum to 3.11, consider `Self` for subclass-correct return types.

jakub-hajto-thenvoi (commented):
Double-checked my earlier approval. Core implementation is solid. Adding a few supplementary observations from the second pass — none are blocking.

src/thenvoi/core/types.py
```diff
@@ -12,6 +13,57 @@
...
+    """
+
+    capabilities: frozenset[Capability] = frozenset()
+    emit: frozenset[Emit] = frozenset()
+    include_tools: tuple[str, ...] | None = None
+    exclude_tools: tuple[str, ...] | None = None
+    include_categories: tuple[str, ...] | None = None
+
+    def __post_init__(self) -> None:
+        object.__setattr__(self, "capabilities", frozenset(self.capabilities))
```
jakub-hajto-thenvoi, line 55: **P3: No defensive guard for `capabilities=None`.** `frozenset(None)` raises `TypeError` here. The type hint says `frozenset[Capability]` (not `| None`), so this is technically caller misuse — but a confusing error message. Consider `frozenset(self.capabilities or ())` for a friendlier experience, or documenting the contract. Same applies to `emit` on the next line.

Note: `include_tools`/`exclude_tools` correctly guard with `if ... is not None` on lines 57-60.

src/thenvoi/core/types.py
```diff
@@ -12,6 +13,57 @@
...
+    Custom tools are NOT included -- they are adapter-local because each
+    framework has its own tool type.
+
+    Accepts list/set inputs for convenience; normalizes to frozen types
+    internally.
+    """
+
+    capabilities: frozenset[Capability] = frozenset()
+    emit: frozenset[Emit] = frozenset()
+    include_tools: tuple[str, ...] | None = None
```
jakub-hajto-thenvoi, line 50: **P3: `include_tools` + `exclude_tools` can both be set.** No validation here, but this is fine — PR #212's `filter_tool_schemas()` handles the interaction with clear precedence (include first, then exclude) and has a test (`test_include_and_exclude_combined`). Noting for traceability.

src/thenvoi/core/exceptions.py
```diff
@@ -0,0 +1,87 @@
...
+
+class ThenvoiConnectionError(ThenvoiError):
+    """Transport failures (WebSocket, REST). Actionable by ops."""
+
+
+class ThenvoiToolError(ThenvoiError):
+    """Tool execution failures. Actionable by adapter/LLM."""
+
+
+def _levenshtein(a: str, b: str) -> int:
```
jakub-hajto-thenvoi, line 51: **P3: `_levenshtein` only tested indirectly** through `with_suggestion()`. Direct unit tests for edge cases (empty strings, transpositions, identical strings) would improve confidence. Not blocking since the indirect coverage via `TestConfigErrorWithSuggestion` is decent.

amit-gazal-thenvoi (approved):

darvell-thenvoi:
Addressed in 8b6dbb6:

1. **Defensive guard for `capabilities=None`/`emit=None`** — `__post_init__` now uses `frozenset(self.capabilities or ())` so passing `None` yields an empty frozenset instead of a `TypeError`.
2. **Direct `_levenshtein` tests** — Added `TestLevenshtein` class with 8 edge case tests: identical strings, empty source/target, both empty, single insertion/deletion/substitution, and transposition behavior.
3. **`include_tools` + `exclude_tools` overlap** — Noted for traceability; already handled by `filter_tool_schemas()` precedence in PR #212.

PR #213, opened 2026-04-07
feat: normalize adapter constructors with AdapterFeatures (INT-294 Step 3)

darvell-thenvoi:
## Summary
Normalize all 14 adapter constructors to use the shared `AdapterFeatures` vocabulary.

### Universal changes (all 14 adapters)
- Declare `SUPPORTED_EMIT` and `SUPPORTED_CAPABILITIES` as `ClassVar` sets
- Accept `features=AdapterFeatures()` via `SimpleAdapter`
- Config-based adapters (Codex, Opencode, Letta) auto-build features from config booleans

### Boolean deprecation (11 adapters)
- `enable_memory_tools` / `enable_execution_reporting` emit `DeprecationWarning`
- Passing both booleans and `features=` raises `ThenvoiConfigError`
- Internal reads migrated from booleans to `features` membership checks

### Selective renames (Anthropic + Gemini only)
- `anthropic_api_key` -> `api_key`, `gemini_api_key` -> `api_key`
- `custom_section` -> `prompt`
- New `include_base_instructions` param
- Old names accepted with `DeprecationWarning` for one release

**Part 3 of 6** for INT-294 SDK Cleanup & Normalization.
**Depends on:** Step 2 (#212)

## Test plan
- [x] 2472 tests pass (34 new from Steps 1-3)
- [x] Conformance configs updated for new attribute names
- [x] All pre-commit hooks pass (ruff check, ruff format, pyrefly)
- [ ] Verify deprecated boolean paths still work (covered by existing tests using old params)
- [ ] Verify `ThenvoiConfigError` on conflicting params

PR #214, opened 2026-04-07
feat: normalize AgentTools contracts with serialization boundary (INT-294 Step 4)

darvell-thenvoi:
## Summary
- `execute_tool_call()` is now the single serialization boundary: calls `.model_dump()` on Pydantic results before returning to adapters
- `send_message()` raises `ThenvoiToolError` for missing mentions instead of returning `{"error": "..."}` (execute_tool_call catches it for adapter compatibility)
- `FakeAgentTools` gains seeded data constructor (`participants`, `peers`, `contacts`) and assertion helpers

**Part 4 of 6** for INT-294 SDK Cleanup & Normalization.
**Depends on:** Step 3 (#213)

## Test plan
- [x] 2484 tests pass (12 new from Step 4)
- [x] Existing send_message tests updated from error dict to ThenvoiToolError assertions
- [x] All pre-commit hooks pass

### DIFF
--- a/src/thenvoi/runtime/tools.py
+++ b/src/thenvoi/runtime/tools.py
@@ -558,7 +558,7 @@ class AgentTools(AgentToolsProtocol):
 
     async def send_message(
         self, content: str, mentions: list[str] | list[dict[str, str]] | None = None
-    ) -> dict[str, Any]:
+    ) -> Any:
         """
         Send a message to the current room.
 
@@ -566,9 +566,11 @@ class AgentTools(AgentToolsProtocol):
             content: Message content to send
             mentions: List of participant handles (strings). SDK resolves handles to IDs.
                       Format: @<username> for users, @<username>/<agent-name> for agents.
+                      Passing list[dict[str, str]] is deprecated; use list[str] instead.
 
         Returns:
-            Full API response dict with message details (id, content, sender, etc.)
+            Fern ChatMessage model (Pydantic). Serialized to dict by
+            execute_tool_call() at the adapter boundary.
 
         Raises:
             ValueError: If a mentioned handle is not found in participants
@@ -577,6 +579,15 @@ class AgentTools(AgentToolsProtocol):
             ChatMessageRequest,
             ChatMessageRequestMentionsItem,
         )
+
+        # Deprecation warning for dict-style mentions
+        if mentions and isinstance(mentions[0], dict):
+            warnings.warn(
+                "Passing mentions as list[dict] is deprecated. "
+                "Use list[str] with handles instead.",
+                DeprecationWarning,
+                stacklevel=2,
+            )
 
         resolved_mentions = self._resolve_mentions(mentions or [])
 
@@ -607,14 +618,14 @@ class AgentTools(AgentToolsProtocol):
         )
         if not response.data:
             raise RuntimeError("Failed to send message - no response data")
-        return response.data.model_dump()
+        return response.data
 
     async def send_event(
         self,
         content: str,
         message_type: str,
         metadata: dict[str, Any] | None = None,
-    ) -> dict[str, Any]:
+    ) -> Any:
         """
         Send an event to the current room.
 
@@ -626,7 +637,8 @@ class AgentTools(AgentToolsProtocol):
             metadata: Optional structured data for the event
 
         Returns:
-            Full API response dict with event details
+            Fern ChatEvent model (Pydantic). Serialized to dict by
+            execute_tool_call() at the adapter boundary.
         """
         from thenvoi.client.rest import ChatEventRequest
 
@@ -643,7 +655,7 @@ class AgentTools(AgentToolsProtocol):
         )
         if not response.data:
             raise RuntimeError("Failed to send event - no response data")
-        return response.data.model_dump()
+        return response.data
 
     async def create_chatroom(self, task_id: str | None = None) -> str:
         """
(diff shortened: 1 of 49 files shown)


jakub-hajto-thenvoi (changes requested):
## Review: feat: normalize AgentTools contracts with serialization boundary (INT-294 Step 4)

### Scope: ✅ CLEAN

### Findings

| Pri | Finding | File |
|-----|---------|------|
| **P1** | **Claude SDK `except ValueError` no longer catches `ThenvoiToolError`.** `send_message` now raises `ThenvoiToolError` for empty mentions, but `_build_platform_sdk_tool` at line 222 catches `ValueError` to enrich errors with "Available handles" guidance. `ThenvoiToolError` inherits from `ThenvoiError → Exception` (not `ValueError`), so it falls through to the generic `except Exception` at line 233 — losing the handle suggestions and producing a noisy `logger.exception()` stack trace. The unknown-handle path (`_resolve_mentions` raising `ValueError`) still works. **No test covers this enrichment path.** Fix: catch `(ValueError, ThenvoiToolError)` at line 222, or add `ThenvoiToolError` handling. | `integrations/claude_sdk/tools.py:222` |
| P2 | **`execute_tool_call` doesn't serialize `list[PydanticModel]`.** Only handles single Pydantic models via `model_dump()`. `get_participants()` returns `list[FernModel]` — adapters using `execute_tool_call` (Anthropic, Codex, Gemini, ADK) get `str()` serialization via `default=str`. Docstring at `get_participants()` falsely claims "Serialized to list[dict] by execute_tool_call()". Fix: add list handling after the single-model check. | `runtime/tools.py:1444-1446` |
| P3 | `FakeAgentTools.send_message` doesn't mirror `ThenvoiToolError` validation on empty mentions — tests using it won't exercise the error path. | `testing/fake_tools.py` |
| P3 | `AgentToolsProtocol` return types widened from `dict[str, Any]` to `Any` — acceptable transitionally but loses type safety. | `core/protocols.py` |

The P1 is a silent behavioral regression in the Claude SDK adapter. Requesting changes.

src/thenvoi/runtime/tools.py
```diff
@@ -1499,6 +1370,14 @@ async def execute_tool_call(self, tool_name: str, arguments: dict[str, Any]) ->
 
         try:
             method = getattr(self, definition.method_name)
-            return await method(**arguments)
+            result = await method(**arguments)
+            # Serialize Pydantic models to dicts at the adapter boundary
+            if hasattr(result, "model_dump"):
```
jakub-hajto-thenvoi, line 1375: **P2: List serialization gap.** This handles single Pydantic models but not `list[PydanticModel]`. `get_participants()` returns `list[FernModel]` — adapters using `execute_tool_call` (Anthropic, Codex, Gemini, ADK) get `str()` serialization via `default=str` instead of proper `model_dump()`. The docstring at `get_participants()` falsely claims "Serialized to list[dict] by execute_tool_call()".

Suggested fix:
```python
if hasattr(result, "model_dump"):
    return result.model_dump()
if isinstance(result, list):
    return [
        item.model_dump() if hasattr(item, "model_dump") else item
        for item in result
    ]
return result
```

jakub-hajto-thenvoi (changes requested):
**P1 (not in this PR's diff, but introduced by it):** `src/thenvoi/integrations/claude_sdk/tools.py:222` — `except ValueError` no longer catches the empty-mentions error. `send_message` in `runtime/tools.py` now raises `ThenvoiToolError` (inherits `ThenvoiError → Exception`, NOT `ValueError`). The Claude SDK's `_build_platform_sdk_tool` handler catches `ValueError` to enrich errors with "Available handles" guidance. After this change, `ThenvoiToolError` falls through to the generic `except Exception` at line 233 — losing the enriched handle suggestions and producing a noisy `logger.exception()` stack trace. The unknown-handle path (`_resolve_mentions` raising `ValueError`) still works. **No test covers the Claude SDK enrichment path.**

Fix: catch `(ValueError, ThenvoiToolError)` at line 222, or add a dedicated `except ThenvoiToolError` block.

See inline comment for the P2 finding.

darvell-thenvoi:
Both findings were already addressed in commit 8e982f0:

- **P1 Claude SDK ThenvoiToolError**: `_build_platform_sdk_tool` at line 223 already catches `(ValueError, ThenvoiToolError)`.
- **P2 List serialization**: `execute_tool_call` at lines 1377-1381 already handles `list[PydanticModel]` with per-item `model_dump()`.

No additional changes needed.

jakub-hajto-thenvoi (dismissed):
Both findings addressed in 8e982f0. Verified:
- P1: Claude SDK now catches `(ValueError, ThenvoiToolError)` at line 223 — enriched "Available handles" message preserved
- P2: `execute_tool_call` now handles `list[PydanticModel]` with per-item `model_dump()`

Note: No dedicated unit test for the Claude SDK wrapper's ThenvoiToolError catch path. Consider adding one to pin this fix.

Approving.

amit-gazal-thenvoi (dismissed):
Approved — clean serialization boundary with good boilerplate reduction. One minor non-blocking note:

1. **`add_participant` / `remove_participant` — fresh snapshot pattern duplication**

Both methods now do:
```python
fresh = await self.get_participants()
snapshot = [p.model_dump() if hasattr(p, "model_dump") else p for p in fresh]
if snapshot:
    self._participants = snapshot
else:
    snapshot = list(self._participants)
```

This 5-line pattern is duplicated verbatim. Consider extracting a `_refresh_participants()` helper that returns the snapshot and updates the cache. Non-blocking but reduces future drift risk.

darvell-thenvoi:
**List serialization** — Already handled. `execute_tool_call()` at lines 1446-1450 already iterates lists and calls `.model_dump()` on each Pydantic item:

```python
if isinstance(result, list):
    return [
        item.model_dump() if hasattr(item, "model_dump") else item
        for item in result
    ]
```

This was implemented as part of the original PR scope.

amit-gazal-thenvoi (approved):

PR #215, opened 2026-04-07
feat: overhaul sys prompt with capability gating/injection defense (INT-294 Step5)

darvell-thenvoi:
## Summary
- Rewrite BASE_INSTRUCTIONS: remove false "NO internet access" claim, remove forced thought event emission
- Add injection defense section (treat participant messages as user input)
- render_system_prompt() gains `features=` param for capability gating
- Memory/Contact tool instruction sections only included when corresponding Capability is set
- Developer instructions placed in dedicated section at end
- TEMPLATES dict preserved for backward compatibility

**Part 5 of 6** for INT-294 SDK Cleanup & Normalization.
**Depends on:** Step 4 (#214)

## Test plan
- [x] 2493 tests pass (9 new from Step 5)
- [x] Tests verify no false internet claims, injection defense present, forced thoughts absent
- [x] Tests verify capability-gated sections include/exclude correctly
- [x] All pre-commit hooks pass


jakub-hajto-thenvoi (changes requested):
## Review: feat: overhaul sys prompt with capability gating/injection defense (INT-294 Step 5)

### Scope: ✅ CLEAN

### Static Review
Core prompt overhaul is well-executed:
- "NO internet access" false claim correctly removed
- Injection defense section is clear and effective
- Capability gating (memory/contacts sections) properly controlled by `features.capabilities`
- Forced thought emission correctly removed
- `features=` threaded through all adapter callers of `render_system_prompt()`
- Strong test coverage: negative cases, gating combinations, round-trip e2e through 6 adapters

### Findings

| Pri | Finding | File |
|-----|---------|------|
| P3 | **ACP client adapter missing `features=`** on `render_system_prompt()` call. Harmless today — `include_base_instructions=False` short-circuits before capability gating, and ACP is architecturally a bridge that declares `SUPPORTED_CAPABILITIES = frozenset()`. But adding `features=self.features` for consistency is a one-line fix. | `integrations/acp/client_adapter.py` |
| P3 | **`TEMPLATES["default"]` diverges** from what `render_system_prompt()` actually produces. The template uses `{custom_section}` inline with no capability sections or "Developer Instructions" heading. Stale backward-compat artifact — consider deprecating or updating. | `runtime/prompts.py` |

**Requesting changes** for the P3s — both are trivial fixes. The TEMPLATES divergence in particular could confuse any external consumer using it directly.

src/thenvoi/runtime/prompts.py
```diff
@@ -2,104 +2,81 @@
...
 """
 
-# Single default template - agent identity + custom section + base instructions
+# Backward-compatible template dict (used by runtime/__init__.py export)
 TEMPLATES: dict[str, str] = {
-    "default": """You are {agent_name}, {agent_description}.
-
-{custom_section}
-"""
+    "default": "You are {agent_name}, {agent_description}.\n\n{custom_section}\n"
```
jakub-hajto-thenvoi: **P3: `TEMPLATES["default"]` diverges from `render_system_prompt()` output.** This template uses `{custom_section}` inline between identity and `BASE_INSTRUCTIONS`, with no capability-gated sections (MEMORY_SECTION, CONTACT_SECTION) and no "Developer Instructions" heading. But `render_system_prompt()` (below) now builds the prompt dynamically with capability gating and places custom_section under a dedicated heading. Any external consumer using `TEMPLATES["default"].format(...)` directly will get a prompt that doesn't match what the SDK actually produces. Consider deprecating this dict or updating it to match.

jakub-hajto-thenvoi (changes requested):
**P3 (not in this PR's diff):** `src/thenvoi/integrations/acp/client_adapter.py:411` — `render_system_prompt()` call is missing `features=self.features` param. Currently harmless because `include_base_instructions=False` short-circuits before capability gating, and ACP is architecturally a bridge with `SUPPORTED_CAPABILITIES = frozenset()`. But adding `features=self.features` for consistency is a one-line fix.

See inline comment for the other finding.

darvell-thenvoi:
Addressed review feedback in 1c90b0f:

- **P3 ACP client features=**: Verified already present at line 415 (`features=self.features`).
- **P3 TEMPLATES divergence**: Updated the comment block to mark `TEMPLATES` as deprecated and note that it does not include the capability-gated sections produced by `render_system_prompt()`. Consumers should call `render_system_prompt()` directly.

jakub-hajto-thenvoi (dismissed):
Findings addressed in 1c90b0f. Verified:
- ACP client adapter confirmed to already pass `features=self.features`
- TEMPLATES comment updated to note consumers should call `render_system_prompt()` directly (doesn't explicitly say "deprecated" but the guidance is clear enough)

Approving.

amit-gazal-thenvoi (dismissed):

darvell-thenvoi:
**TEMPLATES dict divergence** — Already addressed with a deprecation comment (added in prior round) at line 77-80:

```python
# Backward-compatible template dict — DEPRECATED.
# This static template does NOT include capability-gated sections (MEMORY_SECTION,
# CONTACT_SECTION) that render_system_prompt() now produces dynamically.
# Prefer calling render_system_prompt() directly.
```

Full removal is blocked by `LangGraphAdapter.prompt_template` which exposes template customization as a public constructor parameter.

amit-gazal-thenvoi (approved):

PR #237, opened 2026-04-14
fix: resolve WebSocket disconnect loop for agents with many rooms (INT-326)

darvell-thenvoi:
## Summary

- Parallelize room channel joins via `asyncio.gather` so the heartbeat task isn't starved during startup/reconnection
- Wrap each room join in try/except so one unauthorized or broken room doesn't abort all subsequent subscriptions
- Wire `on_reconnect`/`on_disconnect` callbacks from `PHXChannelsClient` through `WebSocketClient` to `ThenvoiLink`
- On reconnect, clear stale `_subscribed_rooms` and emit `ReconnectedEvent` so `RoomPresence` re-derives room state from the API instead of replaying stale PHX-level subscriptions
- Paginate through all pages of `GET /agent/chats` — was only fetching the first page (default page_size=20), silently missing rooms for agents in 20+ rooms
- Per-room error isolation: failed room joins are logged with the room ID and skipped, remaining rooms continue

## Context

Agents with accumulated rooms entered a disconnect/reconnect loop. Root cause: sequential room joins blocked the event loop long enough to starve the 30s heartbeat, causing the server's 45s transport timeout to kill the connection. On reconnect, the PHX client replayed all stale subscriptions, hitting the same timeout. A fresh account with zero rooms worked fine because startup completed instantly.

The platform's `agent_rooms:{agent_id}` channel only notifies about participant add/remove, not new messages, so there's no way to lazily subscribe to rooms on demand. A platform-side `message_pending` event would enable that in the future — for now, parallel subscribe-to-all is the correct approach.

## Verified against production (51 rooms on app.thenvoi.com)

**Pagination bug:** The old code fetched only the first page with default `page_size=20`. With 51 rooms on the agent, **31 rooms were silently dropped** — the agent would never receive messages in those rooms.

**Subscription timing (RTT to prod ~98ms):**

| Approach | 51 rooms | 150 rooms (estimated) |
|----------|----------|-----------------------|
| Sequential (old) | ~10s | ~30s (misses first heartbeat) |
| Parallel (new) | ~100ms | ~100ms |

At 150+ rooms the sequential approach would exceed the 30s heartbeat interval, guaranteeing the server's 45s timeout kills the connection before a heartbeat fires. The parallel approach completes in a single RTT regardless of room count.

## Test plan

- [x] Unit tests pass (1477 passed, pre-existing integration test error unrelated)
- [x] Ruff check + format + pyrefly all pass
- [x] Pagination verified against prod: 51 rooms returned correctly vs 20 with old code
- [x] API response shape verified: `metadata.total_pages` exists and is integer
- [ ] Verify with an agent that has many rooms: startup should complete without disconnect loop
- [ ] Verify reconnection: kill the WS connection mid-session, agent should reconnect and re-derive rooms from API
- [ ] Verify one bad room doesn't kill others: if a room join fails, remaining rooms should still subscribe


yael-avioz-thenvoi (commented):
Code review — two issues found with the reconnection handler. Pagination + parallel subscribe look solid.

src/thenvoi/platform/link.py
```diff
@@ -252,6 +276,23 @@ async def unsubscribe_agent_contacts(self) -> None:
...
+        """Handle PHX client reconnection.
+
+        Clears stale subscription tracking and emits a ReconnectedEvent
+        so RoomPresence can re-derive room state from the server.
+        The PHX client already re-subscribes to agent-level channels
+        (agent_rooms, agent_contacts) via _resubscribe_topics().
+        Room-level channels are re-derived by RoomPresence.
+        """
+        logger.info("WebSocket reconnected — clearing stale room subscriptions")
+        self._subscribed_rooms.clear()
```
yael-avioz-thenvoi: **Bug: `_subscribed_rooms` out of sync after reconnect — causes channel leaks**

`PHXChannelsClient._resubscribe_topics()` already re-subscribes all room-level topics (`chat_room:X`, `room_participants:X`) with original callbacks *before* calling `on_reconnect`. So by the time this line runs, every room topic is already subscribed at the PHX level.

Then when `RoomPresence._handle_reconnect` → `_subscribe_to_existing_rooms` → `subscribe_room` → `join_chat_room_channel` → `subscribe_to_topic`, PHXChannelsClient raises `PHXTopicError("Topic already subscribed")`. `subscribe_room` catches it and returns early — so the room is **never added to `_subscribed_rooms`**.

After reconnect, `_subscribed_rooms` is empty even though all rooms are subscribed at the PHX level.

**Consequences:**
- `unsubscribe_room` checks `room_id not in self._subscribed_rooms` → no-ops → **channel subscription leak** on room removal
- 1 warning log per room per reconnect ("Failed to join chat_room:...")

Messages still work because the original PHX callbacks are intact. But channels leak over reconnect+removal cycles.

**Suggested fix:** Since PHXChannelsClient already handles topic re-subscription, `_on_reconnected` should re-populate `_subscribed_rooms` from `PHXChannelsClient._subscription_callbacks` (or from the room list fetched by `_handle_reconnect`) instead of trying to re-subscribe at the PHX level. Alternatively, after `_subscribe_to_existing_rooms` completes, sync `_subscribed_rooms` with `self.rooms` from presence.

src/thenvoi/runtime/presence.py
```diff
@@ -249,6 +252,32 @@ async def _handle_room_left(
...
+        """
+        logger.info("Handling reconnection — re-subscribing to rooms from API")
+        old_rooms = self.rooms.copy()
+        self.rooms.clear()
+
+        if self.auto_subscribe_existing:
+            await self._subscribe_to_existing_rooms()
+
+        # Notify left for rooms we were in but are no longer present
+        gone_rooms = old_rooms - self.rooms
```
yael-avioz-thenvoi: **Minor: gone rooms not unsubscribed from PHX channels**

This loop calls `on_room_left` (which destroys the execution) but doesn't call `self.link.unsubscribe_room(room_id)` for gone rooms. Their PHX channel subscriptions (`chat_room:X`, `room_participants:X`) stay open, receiving events that get dropped.

Note: even if this call were added, it would currently no-op because `_subscribed_rooms` is empty after reconnect (see comment on `link.py:289`). Both issues share the same root cause — `_subscribed_rooms` tracking is broken after reconnect.

src/thenvoi/platform/link.py
darvell-thenvoi: Fixed in a420ce5. I verified the installed PHX client re-subscribes saved topics before calling our reconnect hook, so clearing `_subscribed_rooms` here was the bug. `_on_reconnected` now preserves room tracking and only emits `ReconnectedEvent`, and room reconciliation moved to presence.

src/thenvoi/runtime/presence.py
darvell-thenvoi: Fixed in a420ce5. Reconnect handling now unsubscribes gone rooms via `self.link.unsubscribe_room(room_id)` before firing `on_room_left`, and because `_subscribed_rooms` is preserved across reconnect, those PHX channel leaves no longer no-op.

yael-avioz-thenvoi (approved):

PR #242, opened 2026-04-14
refactor: split ACP subprocess runtime from outbound bridge

darvell-thenvoi:
## Summary
- split outbound ACP into a Thenvoi bridge layer plus a reusable ACP runtime/profile layer while keeping `ACPClientAdapter` as the public entry point
- isolate runtime-specific ACP behavior behind thin profiles and keep Thenvoi MCP hosting as adapter policy (not generic SDK architecture)
- add explicit architecture documentation and a runnable ACP bridge architecture example to satisfy INT-284 scope
- align ACP extension tests with the runtime profile boundary and document bridge-vs-runtime ownership in adapter code

## Test plan
- [x] `uv run pytest tests/integrations/acp/test_client_adapter.py tests/integrations/acp/test_client_runtime.py tests/converters/test_acp_client.py -v`
- [x] `uv run ruff check src/thenvoi/integrations/acp/client_adapter.py src/thenvoi/integrations/acp/client_runtime.py src/thenvoi/integrations/acp/client_profiles.py src/thenvoi/integrations/acp/client_types.py src/thenvoi/integrations/acp/__init__.py tests/integrations/acp/test_client_adapter.py tests/integrations/acp/test_client_runtime.py examples/acp/08_acp_bridge_architecture.py`
- [x] `uv run pyrefly check`
- [x] `uv run python -m py_compile examples/acp/08_acp_bridge_architecture.py`


src/thenvoi/integrations/acp/client_adapter.py
```diff
@@ -143,83 +76,78 @@ def __init__(
...
-        self._ctx: (
-            AbstractAsyncContextManager[tuple[ACPConnectionProtocol, object]] | None
-        ) = None
-        self._stop_lock = asyncio.Lock()  # Guards stop() to prevent TOCTOU race
+        self._runtime = ACPRuntime(
+            command=self._command,
+            env=self._env,
+            auth_method=self._auth_method,
+            client_factory=ThenvoiACPClient,
+            spawn_process=lambda client, *args, **kwargs: spawn_agent_process(
```
amit-gazal-band, line 85: **Must fix — drop this lambda.** It's a no-op wrapper: `spawn_agent_process` is already the default `spawn_process` in `ACPRuntime.__init__`. Pass nothing and the runtime picks it up. Dead indirection.

amit-gazal-thenvoi (commented):
Must-fix and should-fix items from review.

src/thenvoi/integrations/acp/client_adapter.py
```diff
@@ -143,83 +76,78 @@ def __init__(
...
+        # internal callers.
+        #
+        # Architecture note (INT-284): this adapter is the Thenvoi bridge layer
+        # (room/session mapping, system-context bootstrapping, Thenvoi MCP policy,
+        # and platform event emission). ACP subprocess/session lifecycle lives in
+        # ACPRuntime. These properties exist only to preserve test surface during
+        # migration and can be removed once tests stop reaching into internals.
+
+    @property
+    def _conn(self) -> ACPConnectionProtocol | None:
```
amit-gazal-band: **Must fix — don't ship transitional shims on `_`-prefixed attrs.** These getters/setters (`_conn`, `_client`, `_ctx`, `_agent_mcp_transport`) forward to `self._runtime` purely to keep existing tests passing. A refactor PR should update the tests alongside the refactor. Property access over a private name that silently mutates another object's internals is a trap for the next reader. Please migrate the tests in this PR and delete the shims.

src/thenvoi/integrations/acp/client_adapter.py
```diff
@@ -602,19 +420,8 @@ def _rehydrate(self, room_id: str, history: ACPClientSessionState) -> None:
...
-                    raise RuntimeError(
-                        "ACP client not initialized. Call on_started first."
-                    )
-
-            conn = self._conn
-
-        if conn is None:
-            raise RuntimeError("ACP client connection dropped before prompt")
-        return conn
+        if self._conn is None and self._ctx is None and self.agent_name:
```
amit-gazal-band: **Should fix — log races outside the lock.** Prior impl held `_stop_lock` around this check. Now `self._conn`/`self._ctx` are read lock-free and we log "Respawning…" before delegating. Another coroutine can flip state between the check and `runtime.ensure_connection()`, so the log can fire without an actual respawn. Move the log into `ACPRuntime.start()` (or inside the runtime's lock) so it reflects what actually happened.

src/thenvoi/integrations/acp/client_profiles.py
```diff
@@ -0,0 +1,113 @@
...
+        self,
+        method: str,
+        params: dict[str, object],
+    ) -> dict[str, object]: ...
+
+    async def ext_notification(
+        self,
+        method: str,
+        params: dict[str, object],
+        session_chunks: MutableMapping[str, list[CollectedChunk]],
```
amit-gazal-band: **Should fix — inverted layering.** Passing `MutableMapping[str, list[CollectedChunk]]` lets profiles mutate the runtime's internal buffer directly. Cleaner contract: profile returns `list[CollectedChunk]` (or `None`), runtime appends to its own buffer. Keeps the profile stateless and the runtime as sole owner of `_session_chunks`.

src/thenvoi/integrations/acp/client_types.py
```diff
@@ -1,281 +1,26 @@
...
-            Extracted text, or empty string.
-        """
-        content = getattr(update, "content", None)
-        if content is None:
-            return ""
-        text = getattr(content, "text", None)
-        if text is None and isinstance(content, dict):
-            text = content.get("text", "")
-        return str(text) if text else ""
+        super().__init__(profile=CursorACPClientProfile())
```
amit-gazal-band: **Should fix — hard-coding Cursor profile for all outbound ACP.** `ThenvoiACPClient` forces `CursorACPClientProfile()` for every runtime — Codex, Claude Code, Gemini, etc. Preserves prior behavior so not a regression, but now that profiles exist this should be adapter-configurable (e.g., `ACPClientAdapter(..., profile=...)`), defaulting to `NoopACPClientProfile` with Cursor opt-in.

src/thenvoi/integrations/acp/client_adapter.py
amit-gazal-band: Retracting this — the lambda is load-bearing. Tests patch `thenvoi.integrations.acp.client_adapter.spawn_agent_process` (e.g., test_client_adapter.py:184). The lambda defers name resolution to call time so the patch is respected; passing `spawn_agent_process` directly would bind at `__init__` and bypass the patch.

That said, this is tangled with the shim issue above: if tests moved to injecting `spawn_process` via a constructor arg (or patched `client_runtime.spawn_agent_process` instead), the lambda could go. Not a blocker for this PR.

amit-gazal-thenvoi (changes requested):
See my comments please.
Great documentation BTW!

src/thenvoi/integrations/acp/client_adapter.py
darvell-thenvoi: Kept the lambda per your follow-up. The load-bearing part is still there, but the bigger issue is fixed in `ed25c0a`: the private adapter shims are gone and the ACP tests now hit `self._runtime` directly.

src/thenvoi/integrations/acp/client_adapter.py
darvell-thenvoi: Done in `ed25c0a`. I removed the `_conn` / `_client` / `_ctx` / `_agent_mcp_transport` forwarding shims from `ACPClientAdapter` and migrated the ACP tests to use `_runtime` directly.

src/thenvoi/integrations/acp/client_adapter.py
darvell-thenvoi: Done in `ed25c0a`. The spawn/respawn log now lives in `ACPRuntime.start(respawn=...)`, and `ensure_connection()` calls that path under the runtime lock instead of logging from the adapter side.

src/thenvoi/integrations/acp/client_profiles.py
darvell-thenvoi: Done in `ed25c0a`. `ACPClientProfile.ext_notification()` now returns `list[CollectedChunk]`, and the runtime is the only owner/writer of `_session_chunks`.

src/thenvoi/integrations/acp/client_types.py
darvell-thenvoi: Done in `ed25c0a`. `ThenvoiACPClient` no longer hard-wires Cursor behavior; `ACPClientAdapter` now accepts `profile=...`, defaults to no profile, and the Cursor examples opt in explicitly.

amit-gazal-thenvoi (approved):

PR #245, opened 2026-04-16
fix: agent misses messages after completing a turn and going idle (INT-333)

darvell-thenvoi:
## Summary

- **Root cause (platform)**: The platform's delivery worker never initiated delivery for the message (`delivery_status` block absent on the API response) — a separate platform-side bug.
- **Root cause (SDK, this PR)**: Once Phase 2 of `_process_loop` started, `queue.get()` blocked indefinitely with no timeout and no fallback. Any platform-side miss or WS reconnect gap left the agent permanently deaf until an external nudge.

### `runtime/execution.py`
- **`_process_loop` Phase 2**: replaced bare `await queue.get()` with `asyncio.wait_for(queue.get(), timeout=config.idle_resync_seconds)`. On `TimeoutError`, calls `_resync_pending_messages()` and continues. This limits the worst-case silent stall to `idle_resync_seconds` (default: 60s).
- **`_resync_pending_messages()`** (new method): polls `GET /next` in a loop until it returns `None`, processing each message through the same `_process_backlog_message` path as startup sync. No startup-specific sync-point tracking — just a clean catch-up.
- **`request_resync()`** (new method, also on `Execution` Protocol): puts a `_ResyncRequest` sentinel in the queue so the Phase 2 loop wakes immediately without waiting for the idle timeout. Used by the reconnect path.
- **`_ResyncRequest`** (new sentinel class): a lightweight object pushed into the queue to distinguish a reconnect-triggered resync from a real platform event.
- **Idle-state debug log**: logs queue size before each `queue.get()` so future silent stalls are diagnosable without a timeout having to fire first.

### `runtime/types.py`
- **`SessionConfig.idle_resync_seconds`** (new field, default `60`): controls how long Phase 2 waits before re-polling `/next`. Operators can lower this for faster recovery or raise it to reduce REST traffic in high-room-count deployments.

### `runtime/presence.py`
- **`RoomPresence.on_reconnected`** (new callback): fires at the end of `_handle_reconnect` (inside a `finally` so it fires even when there are no new rooms to subscribe). Wrapped in try/except so a callback error doesn't suppress the reconnect path.

### `runtime/runtime.py`
- **`AgentRuntime._on_reconnected()`** (new method): iterates all active `ExecutionContext`s and calls `request_resync()` on each. Wired to `presence.on_reconnected` in `__init__`.

## How this fixes the incident

| Time | Before this PR | After this PR |
|------|---------------|---------------|
| 09:45:13 | `mark_processed` → loops back to `await queue.get()` | Same |
| 09:45:13+ | Blocked indefinitely, no logs | Blocked, but with 60s timeout |
| 09:45:55 | Reviewer's message never pushed by platform | Same (platform bug) |
| 09:46:13 | Still blocked | Still blocked (heartbeat tick) |
| ~09:46:13 | Nothing | `TimeoutError` → `_resync_pending_messages()` → `/next` returns the message → agent processes it |
| 10:01+ | Still stuck | Agent has processed the message ✓ |

## Test plan

- [x] `uv run ruff check .` — clean
- [x] `uv run ruff format .` — clean
- [x] `uv run pyrefly check` — 0 errors (1 suppressed for the sentinel `type: ignore`)
- [x] `uv run pytest tests/runtime/ tests/framework_conformance/ -q` — 753 passed, 32 skipped
- [ ] Manual: deploy to staging devSquad, verify agent recovers within 60s after a platform delivery miss
- [ ] Manual: verify reconnect-triggered resync fires immediately after WS drop/reconnect

## Related

- INT-333 (this ticket)
- Platform team to separately investigate: missing `delivery_status` block on the lost message (delivery worker never enqueued the job)


src/thenvoi/runtime/presence.py
```diff
@@ -260,74 +261,98 @@ async def _handle_reconnect(self) -> None:
...
-            )
-        else:
-            logger.info("Subscribed to %s rooms during reconnect", succeeded)
+        finally:
+            # Notify callers so they can resync /next for messages missed during downtime
+            if self.on_reconnected:
+                try:
+                    await self.on_reconnected()
+                except asyncio.CancelledError:
+                    pass  # Don't resync if we're being cancelled/shutdown
```
amit-gazal-band: Swallowing `CancelledError` here breaks structured concurrency — if the surrounding task is cancelled mid-callback, this `pass` suppresses the signal and the shutdown can stall. Re-raise instead:

```python
except asyncio.CancelledError:
    raise
```

The `logger.warning` branch below already handles the "callback misbehaved" case.

src/thenvoi/runtime/execution.py
```diff
@@ -850,6 +906,63 @@ async def _get_next_message(self) -> PlatformMessage | None:
...
+        WebSocket sync-point marker. Called:
+        - After idle timeout in Phase 2 (platform may have missed a push)
+        - After WebSocket reconnect (messages arrived during downtime)
+        """
+        logger.debug(
+            "ExecutionContext %s: Re-polling /next for missed messages", self.room_id
+        )
+        caught_up = 0
+        try:
+            while True:
```
amit-gazal-band, line 923: `while True` has no iteration cap and no cooperative cancellation check. On a pathological backlog (or a replay loop), this can run for a long time, starving the room's ability to react to `stop()` and leaving operators wondering if the agent is stuck. Suggest logging progress every 100 messages so it's visibly alive:

```python
if caught_up and caught_up % 100 == 0:
    logger.info(
        "ExecutionContext %s: Still catching up, %d messages processed so far",
        self.room_id, caught_up,
    )
```

src/thenvoi/runtime/types.py
```diff
@@ -83,6 +83,11 @@ class SessionConfig:
     max_context_messages: int = 100
     max_message_retries: int = 1  # Max attempts per message before permanently failing
     enable_context_hydration: bool = True  # Whether to fetch history from platform API
+    # Phase 2 idle timeout (seconds) before re-polling /next as a safety net.
+    # Lower values recover faster from missed WS pushes but generate more REST traffic.
+    # With N rooms, each resync fires N parallel /next polls. Default 60s balances
+    # recovery speed against REST load for typical single-agent deployments.
+    idle_resync_seconds: int = 60
```
amit-gazal-band: Two concerns on this field:

- Type should be `float`, not `int` — lets tests use sub-second values (e.g. `0.01`) without brittleness, and no reason to forbid fractional seconds in prod.
- No validation: a negative or zero value makes `asyncio.wait_for` raise immediately, turning Phase 2 into a tight loop hammering `/next`. Validate `> 0` in `__post_init__`, or clamp.

amit-gazal-thenvoi (commented):
Three follow-ups from review.

amit-gazal-thenvoi (changes requested):
Requesting changes on the three inline follow-ups above — blocker is the `CancelledError` swallow in `presence.py`. The other two (unbounded `while True` progress logging, `idle_resync_seconds` type + validation) should land together.

darvell-thenvoi (commented):
Addressed all three items in d7d8210:

1. `presence.py` — CancelledError is now re-raised to preserve structured concurrency; the `logger.warning` branch still handles misbehaving callbacks.
2. `execution.py` — resync catch-up loop logs progress every 100 messages so pathological backlogs stay visibly alive.
3. `types.py` — `idle_resync_seconds` is now `float` with `__post_init__` rejecting values `<= 0`, so Phase 2 can't be turned into a REST hot loop.

Tests in `tests/runtime/test_resync.py` updated to match (CancelledError propagates, validation rejects zero). Ready for another look.

amit-gazal-thenvoi (approved):
👑

PR #251, opened 2026-04-19
INT-340: disable CrewAI tool cache + propagate AgentTools participant changes

alon-zino-rosenblum-thenvoi:
## Summary
- Override `cache_function` on all 17 Thenvoi CrewAI tool classes plus the dynamic `CustomCrewAITool` wrapper so CrewAI's `CacheHandler` no longer memoizes outputs of room-scoped tools (ContextVar-based room identity never enters the cache key, so caching silently leaked results across rooms).
- Add an optional `ctx` backref to `AgentTools` so `add_participant` / `remove_participant` propagate to `ExecutionContext._participants`, keeping the next preprocessing pass in sync before the WebSocket `participant_added` event arrives.

## Linear issue
INT-340 — https://linear.app/thenvoi/issue/INT-340/crewaiadapter-tools-are-cached-by-crewai-cache-function-unset-stateful

## Changes
- `src/thenvoi/adapters/crewai.py` — new module-level `_never_cache` helper; `cache_function: Callable = _never_cache` set on `CustomCrewAITool`, `SendMessageTool`, `SendEventTool`, `AddParticipantTool`, `RemoveParticipantTool`, `GetParticipantsTool`, `LookupPeersTool`, `CreateChatroomTool`, `ListContactsTool`, `AddContactTool`, `RemoveContactTool`, `ListContactRequestsTool`, `RespondContactRequestTool`, `ListMemoriesTool`, `StoreMemoryTool`, `GetMemoryTool`, `SupersedeMemoryTool`, `ArchiveMemoryTool`.
- `src/thenvoi/runtime/tools.py` — `AgentTools.__init__` accepts keyword-only `ctx: ExecutionContext | None`; `from_context` threads the ctx backref; `add_participant` / `remove_participant` call `ctx.add_participant` / `ctx.remove_participant` (idempotent, so WebSocket re-delivery is safe).
- `tests/adapters/test_crewai_cache.py` — new regression suite exercising the **real** CrewAI SDK (bypasses the mocked `BaseTool` fixture used elsewhere): asserts every tool's `cache_function` returns `False`, the custom tool wrapper disables caching, and the override is not CrewAI's always-true default.
- `tests/runtime/test_tools.py` — 5 new tests covering ctx backref storage, ctx propagation on add/remove, and graceful no-op when ctx is absent.

## Test results
- Scoped verification: `uv run pytest tests/adapters/test_crewai_cache.py tests/runtime/test_tools.py tests/adapters/test_crewai_adapter.py tests/framework_conformance/ tests/framework_configs/ --no-cov -q` — **473 passed, 32 skipped**.
- Full unit suite: **2561 passed, 139 skipped**.
- `uv run ruff check .` and `uv run pyrefly check` both clean.

## Limitations
- The existing "already_in_room" early-return in `AgentTools.add_participant` does not sync ctx when `self._participants` already contains the peer but `ctx._participants` is stale; WebSocket `participant_added` remains the recovery path for that edge case. Out of scope for this fix per plan.

### DIFF
--- a/src/thenvoi/adapters/crewai.py
+++ b/src/thenvoi/adapters/crewai.py
@@ -15,7 +15,7 @@ import logging
 import threading
 import warnings
 from contextvars import ContextVar
-from typing import ClassVar, Any, Coroutine, Literal, Type, TypeVar
+from typing import Callable, ClassVar, Any, Coroutine, Literal, Type, TypeVar
 
 from pydantic import BaseModel, Field, field_validator
 
@@ -42,6 +42,19 @@ from thenvoi.runtime.tools import get_tool_description
 logger = logging.getLogger(__name__)
 
 T = TypeVar("T")
+
+
+def _never_cache(*_args: Any, **_kwargs: Any) -> bool:
+    """cache_function override for every Thenvoi CrewAI tool.
+
+    Thenvoi tools reach the current room via the _current_room_context
+    ContextVar, not via tool arguments. CrewAI's CacheHandler keys entries
+    by (tool_name, input_string) — room identity never enters the key —
+    so caching would leak results across rooms. Returning False here
+    disables CrewAI's per-tool cache write unconditionally.
+    """
+    return False
+
 
 # Module-level state for nest_asyncio patch.
 # See module docstring for important notes about global process impact.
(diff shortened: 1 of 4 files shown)


darvell-thenvoi:
Superseded by #252

PR #250, opened 2026-04-19
fix: upgrade crewai to 1.14.2, resolve parlant conflict

vlad-thenvoi:
## Summary

- Upgrade crewai from 0.80–1.6.x to **1.14.2** (latest)
- Resolve transitive dependency conflict between crewai and parlant
- Remove unnecessary `protobuf>=6.33.5` pins from a2a extras

## Root Cause

crewai 1.14.2 and parlant >=3.1.0 have **incompatible** transitive dependencies that cannot coexist in one environment:

| Package | `opentelemetry-sdk` requirement |
|---|---|
| crewai 1.14.2 | `~=1.34.0` (>=1.34.0, **<1.35.0**) |
| parlant >=3.1.0 (latest: 3.3.1) | **>=1.37.0** |

This is a structural upstream conflict — no version combination resolves it.

A secondary blocker was explicit `protobuf>=6.33.5` pins in the `a2a` and `a2a_gateway` extras. These were unnecessarily strict — `a2a-sdk` only requires `protobuf>=5.29.5` transitively, which is compatible with crewai's `protobuf<6.0` ceiling.

## Changes

- **`pyproject.toml`**:
  - Pin `crewai==1.14.2` in `crewai` and `dev` extras
  - Update `openai>=2.0.0` in crewai extra (crewai 1.14.2 requirement)
  - Remove `protobuf>=6.33.5` from `a2a`/`a2a_gateway` extras
  - Declare `[tool.uv] conflicts` for crewai/parlant mutual exclusion
  - Create `dev-parlant` extra for isolated parlant testing
  - Remove parlant from `dev` extra
- **`AGENTS.md`**: Document the conflict, extras layout, and CI instructions
- **`uv.lock`**: Regenerated (361 packages, clean resolution)

## CI Impact

Parlant adapter tests now need a separate step:
```bash
uv sync --extra dev-parlant
uv run pytest tests/adapters/test_parlant_adapter.py tests/integrations/parlant/ -v
```

## Test plan

- [x] `uv lock` resolves cleanly (361 packages)
- [x] `uv sync --extra dev` installs crewai 1.14.2
- [x] `uv sync --extra crewai` works standalone
- [x] `uv sync --extra parlant` works standalone
- [x] Pre-commit hooks pass
- [ ] CI pipeline passes (crewai adapter tests)
- [ ] CI pipeline passes (parlant adapter tests with `dev-parlant`)

PR #252, opened 2026-04-19
fix: disable CrewAI tool caching, sync participant state, fix pyrefly errors

vlad-thenvoi:
## Summary

- **CrewAI 1.14.2 upgrade**: Resolves dependency conflict between crewai and parlant (opentelemetry-sdk version clash). Declares `[tool.uv] conflicts` for mutually exclusive extras, creates separate `dev-parlant` extra.

- **Bug 1 — CrewAI tool caching breaks multi-room agents**: CrewAI's `CacheHandler` caches by `(tool_name, input_string)`, which is not room-scoped. Since `room_id` lives in a `ContextVar` (not in tool input), identical calls in different rooms return stale cached results. Fixed by setting `cache_function` to always return `False` on all 18 `BaseTool` subclasses.

- **Bug 2 — `AgentTools._participants` diverges from `ExecutionContext`**: `AgentTools.from_context()` copies participants into a local list. Mutations via `add/remove_participant` never sync back to `ExecutionContext`, so the next turn rebuilds `AgentTools` from stale data. Fixed by storing a `_ctx` backref and syncing changes back.

- **Pyrefly fixes (0 errors)**: Updated ACP server to match latest `acp` SDK (`AuthMethod` -> `AuthMethodAgent`, fixed `prompt`/`set_config_option` signatures, implemented abstract `close_session`). Excluded parlant files from pyrefly (conflict group).

- **CI fix**: Gate parlant conformance `on_started` test when parlant SDK is not installed (skip instead of fail).

## Test plan

- [x] All unit tests pass (743 conformance+adapter, 2787 total)
- [x] `IN_CI=true` conformance + drift detection pass (0 failures)
- [x] `uv run ruff check` + `uv run ruff format` — clean
- [ ] E2E test with multi-room CrewAI agent to verify no cross-room caching
- [ ] E2E test with participant add/remove across turns

Supersedes PR #250.


src/thenvoi/integrations/acp/server.py
```diff
@@ -13,13 +13,18 @@
 )
 from acp.schema import (
     AgentCapabilities,
+    AudioContentBlock,
     AuthenticateResponse,
-    AuthMethod,
+    AuthMethodAgent,
```
darvell-thenvoi, line 18: Blocking: this import breaks against the dependency range declared in `pyproject.toml`. In this environment `agent-client-protocol==0.8.1` exposes `AuthMethod`, not `AuthMethodAgent`, so `uv run pytest tests/integrations/acp/test_server.py -q` fails during collection with `ImportError: cannot import name 'AuthMethodAgent' from 'acp.schema'`. Please either pin ACP to a version that provides `AuthMethodAgent` or make this code compatible with both SDK shapes and cover that with a test.

darvell-thenvoi (commented):
I found one blocking compatibility issue while reviewing locally.

src/thenvoi/integrations/acp/server.py
darvell-thenvoi: Nevermind, https://github.com/thenvoi/thenvoi-sdk-python/pull/250 fixes this. Block merge until that's in.

vlad-thenvoi: @darvell-thenvoi please re-review

darvell-thenvoi (approved):
@vlad-thenvoi  LGTM. Approved for merge. 

@amit-gazal-thenvoi We should keep an eye on the parlant/crewai dependency problem and resolve that nicely ASAP when appropriate versions of the libraries come out.

PR #98, opened 2026-02-01
feat: Add Letta adapter with persistent memory and MCP integration

vlad-thenvoi:
## Summary

Adds a production-ready Letta adapter that enables stateful AI agents with persistent memory on the Thenvoi platform.

Documentation [PR](https://github.com/thenvoi/fern/pull/136) 

## Operating Modes

| Mode | Description | Use Case |
|------|-------------|----------|
| `PER_ROOM` | One Letta agent per room, complete memory isolation | Business agents, data analysis, sensitive contexts |
| `SHARED` | One agent across all rooms, per-room context via Conversations API | Personal assistants, cross-room memory |

## Key Capabilities

- **Persistent Memory**: Agents remember conversations across sessions and restarts
- **Memory Consolidation**: Automatic cleanup on graceful shutdown (Ctrl+C) or room deletion
- **State Recovery**: Room-to-agent mappings persisted to disk, survives restarts
- **MCP Tool Execution**: Platform tools executed via MCP server (send_message, add_participant, etc.)
- **Room Lifecycle**: Handles room_deleted events for proper cleanup

## Configuration Options

```python
LettaConfig(
    # Mode
    mode=LettaMode.SHARED,              # SHARED or PER_ROOM
    
    # Server
    base_url="http://localhost:8283",   # Letta server URL
    mcp_server_url="http://localhost:8002/sse",  # MCP server for tools
    
    # Model
    model="openai/gpt-4o",              # LLM model
    embedding_model="openai/text-embedding-3-small",
    
    # Agent
    persona="You are a helpful assistant.",
    shared_agent_id=None,               # Reuse existing agent (SHARED mode)
    
    # Tools
    mcp_tools=["create_agent_chat_message", "list_agent_peers", ...],
    letta_base_tools=["memory", "conversation_search", "archival_memory_insert", "archival_memory_search"],
    
    # MCP Server Registration
    mcp_server_name="thenvoi",
    mcp_server_type="sse",              # or "http"
    
    # Limits
    api_timeout=30,
    tool_execution_timeout=30,
    max_reasoning_steps=10,
    summary_max_length=150,
    room_id_suffix_length=8,
    
    # Memory Block Defaults
    initial_participants_text="No participants yet...",
    initial_room_contexts_text="No room contexts yet...",
)
```

## CLI Arguments

```bash
uv run python examples/run_agent.py --example letta_shared \
    --letta-url http://localhost:8283 \
    --letta-model openai/gpt-4o \
[...]

PR #257, opened 2026-04-21
fix: harden graceful shutdown against signal races and task cancellation (INT-327)

yael-avioz-thenvoi:
## Summary

Closes [INT-327](https://linear.app/thenvoi/issue/INT-327/fix-harden-graceful-shutdown-against-signal-races-and-task). Fixes three robustness gaps in `runtime/shutdown.py` surfaced during audit of closed PR #98 (Letta adapter). General runtime fixes, not Letta-specific.

- **Signal guard runs too late:** Moved the `_shutting_down` / `_shutdown_task` guard to the top of `_handle_signal`, before the `on_signal` callback and `_shutdown_event.set()`. Honors the documented "subsequent signals are ignored" contract and prevents non-idempotent user callbacks (e.g. alerting) from firing twice.
- **`agent.stop()` not shielded from outer-task cancellation:** Wrapped with `asyncio.shield(...)` so a cancellation propagating through the outer awaiter (e.g. a parent `run()` being cancelled) doesn't interrupt cleanup mid-flight. Caveat: `asyncio.run()` teardown calls `_cancel_all_tasks()` which cancels the inner shielded task too — shield protects from the outer awaiter, not full loop teardown. The context-manager / await-stop patterns remain the correct choice for guaranteed completion on process exit.
- **`CancelledError` not caught in `_shutdown()`:** `except Exception` misses `CancelledError` (it inherits from `BaseException` in Py3.8+), so cancelled shutdowns surfaced only as unraisable task-exception warnings. Added an explicit `except asyncio.CancelledError` branch that logs and re-raises. It deliberately does **not** kick off a second concurrent `agent.stop` — `AgentRuntime.stop` iterates `executions.keys()` with no in-flight guard, and two concurrent stops would race on per-room teardown. The shielded inner `agent.stop` continues running in the background.

## Test plan

- [x] New: second signal during shutdown does not re-invoke `on_signal` and does not re-set the shutdown event
- [x] New: cancelling `_shutdown` mid-flight lets `agent.stop()` complete (shield verification, via `asyncio.wait_for`)
- [x] New: `CancelledError` in `_shutdown` is logged and re-raised exactly once (no second concurrent stop)
- [x] All 23 tests in `tests/runtime/test_shutdown.py` pass
- [x] `uv run ruff check .` clean on modified files
- [x] `uv run ruff format .` clean on modified files
- [x] `uv run pyrefly check src/thenvoi/runtime/shutdown.py` clean (no new errors)

### DIFF
--- a/tests/runtime/test_shutdown.py
+++ b/tests/runtime/test_shutdown.py
@@ -221,17 +221,13 @@ class TestGracefulShutdownHandler:
             "agent.stop was cancelled mid-flight; asyncio.shield is missing"
         )
 
-    async def test_shutdown_handles_cancelled_error(self, mock_agent):
-        """_shutdown should log and run best-effort cleanup when cancelled."""
+    async def test_shutdown_reraises_cancelled_error(self, mock_agent):
+        """_shutdown should log and re-raise CancelledError without a second stop."""
         stop_calls: list[float] = []
 
         async def tracking_stop(*, timeout):
             stop_calls.append(timeout)
-            # First call (the shielded one) raises CancelledError to simulate
-            # the shield being bypassed; second call is the best-effort cleanup.
-            if len(stop_calls) == 1:
-                raise asyncio.CancelledError()
-            return True
+            raise asyncio.CancelledError()
 
         mock_agent.stop = AsyncMock(side_effect=tracking_stop)
         shutdown = GracefulShutdown(mock_agent, timeout=5.0)
@@ -239,8 +235,10 @@ class TestGracefulShutdownHandler:
         with pytest.raises(asyncio.CancelledError):
             await shutdown._shutdown()
 
-        # First call uses the configured timeout, second is the best-effort quick stop.
-        assert stop_calls == [5.0, 0.0]
+        # Only one call — asyncio.shield keeps the original agent.stop running
+        # in the background on real cancellation; starting a concurrent second
+        # stop would race on AgentRuntime's per-room executions teardown.
+        assert stop_calls == [5.0]
 
 
 class TestGracefulShutdownWaitForShutdown:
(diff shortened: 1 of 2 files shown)


tests/runtime/test_shutdown.py
```diff
@@ -152,6 +152,92 @@ async def test_shutdown_handles_exception(self, mock_agent):
...
+
+        mock_agent.stop = AsyncMock(side_effect=tracking_stop)
+        shutdown = GracefulShutdown(mock_agent, timeout=5.0)
+
+        with pytest.raises(asyncio.CancelledError):
+            await shutdown._shutdown()
+
+        # Only one call — asyncio.shield keeps the original agent.stop running
+        # in the background on real cancellation; starting a concurrent second
+        # stop would race on AgentRuntime's per-room executions teardown.
```
amit-gazal-band: Nit: this comment describes production behavior, not what this test exercises. Here `agent.stop` raises `CancelledError` directly, so nothing is running in the background — `stop_calls == [5.0]` just verifies we didn't start a second stop. The real shield mechanic is covered by `test_shutdown_shields_agent_stop_from_cancellation`. Suggest splitting the comment so it only talks about the single-call assertion, or move the production note to the `except CancelledError` block in `shutdown.py` where it applies.

amit-gazal-thenvoi (dismissed):
LGTM. Three fixes correctly address the gaps from INT-327:

1. Signal guard moved above side effects — honors the documented "subsequent signals ignored" contract and protects non-idempotent `on_signal` callbacks.
2. `asyncio.shield` around `agent.stop()` — prevents outer-task cancellation from interrupting cleanup mid-flight. Inline caveat about `_cancel_all_tasks()` teardown is accurate.
3. Explicit `except asyncio.CancelledError` branch — logs + re-raises without kicking off a second concurrent stop (verified against `AgentRuntime.stop` which iterates `executions.keys()` with no in-flight guard).

Test coverage solid: three new tests exercise each gap with specific assertions, including a real shield test via `asyncio.wait_for`. One minor inline nit below — non-blocking.

amit-gazal-thenvoi (approved):
Looks great

PR #261, opened 2026-04-21
fix(websocket): include threadId in interrupt payload [INT-323]

darvell-thenvoi:
## Summary
- include `threadId` alongside `turnId` when the Codex adapter sends `turn/interrupt` after a timeout
- tighten the timeout regression so it asserts the exact JSON-RPC interrupt payload

## Test plan
- [x] `uv run pytest tests/adapters/test_codex_adapter.py -k \"test_turn_timeout_sends_interrupt_and_clean_error\"`


amit-gazal-thenvoi (approved):
LGTM

PR #262, opened 2026-04-21
fix(websocket): resync surviving rooms after reconnect [INT-324]

darvell-thenvoi:
## Summary
- forward a `ReconnectedEvent` to surviving rooms after reconnect reconciliation so existing executions are notified
- re-run `/next` synchronization in `ExecutionContext` when a room receives that reconnect signal
- serialize reconnect resync through the execution loop to avoid overlapping replay with normal room processing
- increase the dedupe cache window from 5 to 500 entries so reconnect-heavy rooms keep a realistic duplicate suppression window
- add reconnect regressions for both room propagation and execution resync behavior

## Test plan
- [x] `uv run pytest tests/test_session_sync.py tests/runtime/test_presence.py -q`


amit-gazal-thenvoi (approved):
Important!

darvell-thenvoi:
@amit-gazal-thenvoi merge conflict fixed, could you please re-approve?

amit-gazal-band (approved):

PR #264, opened 2026-04-22
feat: add HumanTools and ToolDefinition.surface field (INT-349)

darvell-thenvoi:
## Summary

Phase 1 of INT-338: extend the SDK tool registry with a `surface` field and add a `HumanTools` class alongside `AgentTools`, so `thenvoi-mcp` (and any other adapter) can consume one source of truth for both scopes.

Closes INT-349. Part of INT-338.

## Changes

- `src/thenvoi/runtime/tools.py`:
  - `ToolDefinition.surface: Literal["agent", "human"] = "agent"` (default preserves every existing entry's behavior).
  - `iter_tool_definitions(*, surface="agent", include_memory=False, include_contacts=True)` — three filters composed as independent predicates. **The `surface` default is pinned to `"agent"`** so existing callers (`claude_sdk`, `opencode`, `acp`) that pipe the result straight into an `AgentTools`-shaped MCP backend don't silently gain `HumanTools` entries. Pass `surface=None` explicitly to opt in to the union view. Memory/contact filters apply uniformly across both surfaces via parallel `HUMAN_MEMORY_TOOL_NAMES` and `HUMAN_CONTACT_TOOL_NAMES` sets.
  - `HumanTools` class — stateless per credential; **28 methods** (one per entry in the Phase 1 human-tool mapping table minus `delete_my_agent`, see below), each wrapping a `rest.human_api_*` call. Mirrors today's `thenvoi-mcp` human handler behavior field-for-field: same validation, same request-object construction, **error strings returned verbatim** (`"Error: Either contact_id or handle must be provided"`, `"Error: recipients cannot be empty"`, `"Error: Not found: ..."`, `"Error: At least one field ..."`, and the `"Added participant: <id>"` / `"Removed participant: <id>"` success strings); `delete_user_memory` wrapped to a `{"deleted": True, "id": ...}` payload. **Write methods pass `request_options=DEFAULT_REQUEST_OPTIONS`** (429 retry + timeouts) to match the `AgentTools` convention; reads follow the `AgentTools` sibling pattern (no options).
  - 28 new human tools registered into `TOOL_DEFINITIONS` with `surface="human"` (agents × 2, chats × 3, contacts × 9, messages × 2, participants × 3, memories × 6, profile × 2, peers × 1).
  - `TOOL_MODELS` and `ALL_TOOL_NAMES` stay agent-only so existing callers/adapters (`claude_sdk`, `framework_conformance`, etc.) aren't silently widened.
- `src/thenvoi/runtime/mcp_server.py`: `build_thenvoi_mcp_tool_registrations()` and `build_resolved_thenvoi_mcp_tool_registrations()` pin `surface="agent"` when they call `iter_tool_definitions` and — as defense-in-depth — drop non-agent entries (with a warning log) when `tool_definitions` is explicitly passed. `LocalMCPServer` remains agent-only.
- `src/thenvoi/runtime/__init__.py`: export `HumanTools`.

## Testing

- `tests/runtime/test_tool_definitions_surface.py` — `surface` defaults to `"agent"`; **new regression test pins `iter_tool_definitions()` with no args to agent-only** (C1); set-equality asserts for `surface="agent"` matching the pre-Phase-1 snapshot and `surface="human"` returning only human tools; filter composition (`surface="human", include_memory=True, include_contacts=False` -> only human memory tools; baseline across both surfaces with no memory/contacts still includes `send_message`, `get_my_profile`, `send_my_chat_message`); every `input_model` is a `BaseModel` subclass and every `method_name` resolves on `AgentTools`/`HumanTools` for its surface; `LocalMCPServer` agent-only snapshot for both builder variants; **two new defense-in-depth tests** assert that `build_*_tool_registrations` filters non-agent entries from an explicit mixed `tool_definitions` list and logs a warning per dropped entry.
- `tests/runtime/test_human_tools.py` — one shape-equal test per `HumanTools` method. Each scripts the REST client via `unittest.mock.AsyncMock` and asserts (a) the right `rest.human_api_*` method is awaited, (b) with the arguments today's MCP handler would have produced, (c) the return shape (keys + Python types) matches today's MCP output. Error-path tests assert the exact error string (matches `thenvoi-mcp` byte-for-byte); write-method tests assert `request_options=DEFAULT_REQUEST_OPTIONS` is forwarded.

```
uv run pytest tests/runtime/ -v
= 481 passed, 3 warnings =
```

`uv run pre-commit run --all-files` green (ruff check, ruff format, pyrefly, detect-secrets, commitizen).

## Notes for reviewers

- **C1 regression guard.** The original Phase 1 defaulted `iter_tool_definitions(surface=None)`. That silently widened every caller that pipes the result into an `AgentTools`-shaped MCP backend (`claude_sdk`, `opencode`, `acp` client adapter), which would `AttributeError` at tool-call time on a human tool. Default is now pinned to `surface="agent"`. Call `iter_tool_definitions(surface=None)` explicitly for the union view.
- **MCP error shape preserved.** Phase 1's original draft raised `ValueError` from six validation paths. After team review, the call is to preserve today's MCP error-string shape verbatim (copied from `thenvoi-mcp/src/thenvoi_mcp/tools/human/*.py`) so downstream callers that render the error string to the LLM see no behavior change. Tests now assert against the exact string instead of `pytest.raises`.
- **`delete_my_agent` dropped from Phase 1.** There is no corresponding handler in `thenvoi-mcp` today; shipping the SDK method would widen the observable tool surface past the spec. Follow-up once the MCP side adds it. The underlying `rest.human_api_agents.delete_my_agent` Fern method is still reachable for callers that need it.
- **Input models copied from `thenvoi-mcp/src/thenvoi_mcp/tools/human/*.py` field-for-field.** Those handlers don't actually declare pydantic models — they're plain `@mcp.tool()` functions — so I built the `*Input` models from the handler signatures (names, types, defaults, docstrings). The observable tool surface is intentionally unchanged from today's MCP behavior.
- **`LocalMCPServer` stays agent-only.** Both `build_thenvoi_mcp_tool_registrations` and `build_resolved_thenvoi_mcp_tool_registrations` pin `surface="agent"` and additionally filter explicit mixed `tool_definitions` lists with a warning log, so a future human tool added to the registry cannot leak into an in-process adapter that expects only agent tools.
- **Tests are shape-equal, not byte-equal** (pragmatic tradeoff, documented in the test module docstring). The acceptance criteria mentioned byte-equal fixtures against recorded MCP output, but the MCP handlers serialize with `serialize_response` and produce JSON strings while `HumanTools` returns the raw Fern model; we felt shape-equality on the REST call boundary and response keys/types catches regressions without requiring a fixture-capture step that touches `thenvoi-mcp` before this SDK change lands. **This was reviewed and explicitly approved by the team**; byte-equal regression fixtures are a follow-up.
- **Downstream heads-up for INT-351:** `iter_tool_definitions(surface, include_memory, include_contacts)` is ready for the MCP registrar to consume. Method names on `HumanTools` match the registry's `method_name` for each human definition.

### DIFF
--- a/src/thenvoi/runtime/__init__.py
+++ b/src/thenvoi/runtime/__init__.py
@@ -36,6 +36,7 @@ from .runtime import AgentRuntime
 # Tools
 from .tools import (
     AgentTools,
+    HumanTools,
     TOOL_MODELS,
     ALL_TOOL_NAMES,
     BASE_TOOL_NAMES,
@@ -72,6 +73,7 @@ __all__ = [
     "AgentRuntime",
     # Tools
     "AgentTools",
+    "HumanTools",
     "TOOL_MODELS",
     "ALL_TOOL_NAMES",
     "BASE_TOOL_NAMES",
--- a/src/thenvoi/runtime/mcp_server.py
+++ b/src/thenvoi/runtime/mcp_server.py
@@ -68,6 +68,31 @@ class MCPToolRegistration:
         )
 
 
+def _filter_to_agent_surface(
+    definitions: Sequence[ToolDefinition],
+) -> list[ToolDefinition]:
+    """Drop non-agent definitions and log a warning for each discarded entry.
+
+    ``build_*_tool_registrations`` wire their execution path through
+    ``AgentTools``; a ``surface="human"`` definition in the list would
+    ``AttributeError`` at call time because ``AgentTools`` has no
+    ``HumanTools`` methods. Rather than propagate the error, quietly filter
+    and warn so a regression in a caller is observable but not fatal.
+    """
+    filtered: list[ToolDefinition] = []
+    for definition in definitions:
+        if definition.surface != "agent":
+            logger.warning(
+                "Dropping non-agent tool definition %r (surface=%r) from MCP "
+                "registrations; LocalMCPServer is agent-only.",
+                definition.name,
+                definition.surface,
+            )
+            continue
+        filtered.append(definition)
+    return filtered
+
+
 def build_thenvoi_mcp_tool_registrations(
     agent_tools: AgentToolsProtocol,
     *,
@@ -80,7 +105,7 @@ def build_thenvoi_mcp_tool_registrations(
     # so a human tool added to the registry never leaks into an adapter
     # expecting only agent tools. Widening is deferred to a future ticket.
     definitions = (
-        list(tool_definitions)
+        _filter_to_agent_surface(list(tool_definitions))
         if tool_definitions is not None
         else [
             definition
@@ -110,7 +135,7 @@ def build_resolved_thenvoi_mcp_tool_registrations(
     """Build MCP registrations that resolve room-scoped tools at call time."""
     # LocalMCPServer stays agent-only — see build_thenvoi_mcp_tool_registrations.
     definitions = (
-        list(tool_definitions)
+        _filter_to_agent_surface(list(tool_definitions))
         if tool_definitions is not None
         else [
             definition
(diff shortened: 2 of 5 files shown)


amit-gazal-thenvoi (approved):
Great work 🥇

PR #270, opened 2026-04-23
fix: fix coding_agents example based on walkthrough review (INT-374)

roi-shikler-thenvoi:
## Summary

Fixes all issues found during hands-on walkthrough of the `examples/coding_agents` tutorial (Linear: INT-374).

- Rewrote README as a proper step-by-step tutorial with platform UI as primary agent creation flow
- Added missing Dockerfiles (`docker/claude_sdk/Dockerfile`, `docker/codex/Dockerfile`, `docker/entrypoint.sh`)
- Fixed Claude SDK container `WORKDIR` to `/workspace/repo` so the agent can access the shared repo volume
- Added `agent_config.yaml` symlink logic in `entrypoint.sh` so `load_agent_config()` finds the config file
- Added termination rules to `planner.md` and `reviewer.md` to prevent infinite acknowledgment loops between agents
- Changed `create_agents.py` description from "E2E test agent" to "Coding team agent"
- Documented `create_agents.py` as optional (Enterprise-only User API key required)
- Added comprehensive troubleshooting section covering all 11 issues found

## Test plan

- [ ] `docker compose build` succeeds with the new Dockerfiles
- [ ] `docker compose up -d` starts both containers
- [ ] Both agents connect to the Band platform via WebSocket
- [ ] Planner responds to @mentions in chat rooms
- [ ] Reviewer responds when @mentioned by the planner
- [ ] Agents stop talking after delivering final status (no infinite loop)
- [ ] README steps are followable from scratch by a new user


darvell-thenvoi (changes requested):
Request changes. The intended prompt changes are good for the peer-room model, but the runnable Docker example does not actually load them.

The planner container starts `run_agent.py` without passing a custom prompt, and `run_agent.py` only feeds Claude SDK from `args.custom_section`; it does not read `AGENT_ROLE` or `/app/config/prompts`. The reviewer has the same issue: compose sets `CODEX_ROLE=reviewer`, but the runner only uses the `--codex-role` CLI arg, and even that loader looks under `examples/codex/prompts`, not this example's mounted prompt directory. That means the new planner/reviewer instructions are effectively dead in the flow this example tells people to run.

I also think the config symlink needs to move. `docker/entrypoint.sh` symlinks `agent_config.yaml` into `$PWD`, and for these containers `$PWD` is `/workspace/repo`. Since that config carries agent API keys, putting it inside the target repo workspace is an avoidable secret exposure and accidental-copy risk for coding agents.

One prompt-level issue after the wiring is fixed: the planner is told to @mention the reviewer for re-review after changes, but the reviewer termination rule says not to respond again unless a human asks for a new review. Those should agree so the example supports the intended re-review loop.

PR #274, opened 2026-04-28
feat: make Claude SDK model and fallback_model optional (Codex parity)

vlad-thenvoi:
## Summary

Make `ClaudeSDKAdapter.model` optional and add `fallback_model`, mirroring how `CodexAdapter` handles model configuration. Removes a stale hardcoded `claude-sonnet-4-5-20250929` default that was overriding the npm `claude` binary's own default even when callers passed nothing.

- `ClaudeSDKAdapter(model=None, fallback_model=None)` is now the default. Both are forwarded directly to upstream `ClaudeAgentOptions` (which already supports `str | None` for both fields).
- When unset, no `--model` flag is sent to the npm `claude` binary, so it picks its own default. Family aliases (`"sonnet"` / `"opus"` / `"haiku"`) are forwarded as-is.
- `examples/run_agent.py` gets a new `--fallback-model` CLI flag and switches `--model` to default `None`. `_DEFAULT_MODELS` is reorganized so `claude_sdk` flows through with `None` and `pydantic_ai`'s de-facto default is preserved explicitly.
- Both docker runners (`docker/claude_sdk/runner.py`, `examples/claude_sdk_docker/runner.py`) drop the stale fallback string and read `fallback_model` from YAML.
- Stale model pins removed from `examples/claude_sdk/01/03/04*.py` and the docker YAML emitters; `02_extended_thinking.py` is kept as the canonical alias demo (`model="opus"`, `fallback_model="sonnet"`).
- Docs (`README.md`, `examples/claude_sdk/README.md`, `examples/claude_sdk_docker/README.md`) updated to use aliases or omit the model.

### Scope note

Configuration-surface parity with Codex only. Codex does runtime discovery via `model/list` on its daemon and tracks a `_selected_model` separate from configured. The npm `claude` binary has no analog, so this PR does **not** add runtime resolution to the Claude SDK adapter; status output reports configured values only (`model: auto` when unset).

## Test plan

- [x] New behavior tests in `TestOnStarted` lock in options forwarding:
  - `test_default_options_have_no_model` — proves no override is sent when caller passes nothing.
  - `test_explicit_model_is_forwarded` — proves `model="opus"` lands in `ClaudeAgentOptions.model`.
  - `test_fallback_model_is_forwarded` — regression guard against forgetting `fallback_model=` in the options construction.
- [x] `uv run pytest tests/adapters/test_claude_sdk_adapter.py tests/integrations/claude_sdk/test_session_manager.py` — 95 passed.
- [x] `uv run pytest tests/framework_conformance/test_config_drift.py tests/framework_conformance/test_adapter_conformance.py -k "claude_sdk or drift"` — 19 passed (including config-drift detection picking up the new `fallback_model` field).
- [x] `uv run ruff check .` and `uv run ruff format .` — clean.
- [x] `uv run pyrefly check` — clean (default scope).
- [ ] Manual smoke against staging/prod: minimal `agent.yaml` (just `agent_id` + `api_key`) connects, `model=auto` shown in logs, npm `claude` CLI runs with no `--model` flag.
- [ ] Manual: `--model opus --fallback-model sonnet` flows into `ClaudeAgentOptions` end-to-end.

### DIFF
--- a/examples/claude_sdk/README.md
+++ b/examples/claude_sdk/README.md
@@ -77,7 +77,7 @@ python examples/claude_sdk/01_basic_agent.py
 ```
 
 Features:
-- Standard Claude Sonnet model
+- npm `claude` binary's default model (no override)
 - Platform tool integration
 - Execution reporting
 
--- a/examples/claude_sdk/README.md
+++ b/examples/claude_sdk/README.md
@@ -51,7 +51,8 @@ from thenvoi import Agent
 from thenvoi.adapters import ClaudeSDKAdapter
 
 adapter = ClaudeSDKAdapter(
-    model="claude-sonnet-4-5-20250929",
+    # Omit `model` to use the npm `claude` binary's default, or pass a
+    # family alias (`"sonnet"` / `"opus"` / `"haiku"`).
     custom_section="You are a helpful assistant.",
 )
 
@@ -101,7 +102,8 @@ Enable extended thinking for complex reasoning tasks:
 
 ```python
 adapter = ClaudeSDKAdapter(
-    model="claude-sonnet-4-5-20250929",
+    model="opus",
+    fallback_model="sonnet",
     max_thinking_tokens=10000,  # Enable extended thinking
     enable_execution_reporting=True,
 )
--- a/README.md
+++ b/README.md
@@ -221,7 +221,10 @@ from thenvoi import Agent
 from thenvoi.adapters import ClaudeSDKAdapter
 
 adapter = ClaudeSDKAdapter(
-    model="claude-sonnet-4-5-20250929",
+    # Omit `model` to use the npm `claude` binary's default, or pass a
+    # family alias (`"sonnet"` / `"opus"` / `"haiku"`) for latest-of-family.
+    model="opus",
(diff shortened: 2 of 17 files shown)


examples/claude_sdk/README.md
```diff
@@ -51,7 +51,8 @@ from thenvoi import Agent
 from thenvoi.adapters import ClaudeSDKAdapter
 
 adapter = ClaudeSDKAdapter(
```
darvell-thenvoi, line 53: Nit: Only because I think we will completely forget about this, can you also change the nearby Features bullet for the basic example from `Standard Claude Sonnet model` to `the default model currently configured for Claude Code in this environment`? The example now omits `model`, so that bullet is stale.

darvell-thenvoi (changes requested):
@vlad-thenvoi Safe to approve otherwise; requesting this tiny docs fix so it doesn't get forgotten.

vlad-thenvoi:
Done in ed2efe6 — changed the bullet to "npm `claude` binary's default model (no override)". Thanks for catching it.

darvell-thenvoi (approved):
Looks good.

PR #290, opened 2026-05-03
fix(integrations): drop AGPL-3 codex-app-server-sdk dep, remove transport="sdk"

amit-gazal-band:
## Summary

- Drop AGPL-3 third-party `codex-app-server-sdk` dependency from the `codex` extra and dev group.
- Remove `CodexAdapterConfig.transport="sdk"` and all SDK-bridge helpers (`_build_sdk_client`, `_sdk_request_scope`, `_handle_sdk_server_request`, `_sdk_request_context`).
- Delete `src/thenvoi/integrations/codex/sdk_client.py` and the five SDK-only test classes in `tests/adapters/test_codex_adapter.py`.

## Why

Security scanner flagged `codex-app-server-sdk@0.2.0` as Critical: the package is AGPL-3 licensed (network-copyleft, propagates to SaaS users) and is maintained by a third party (`github.com/emsi/codex-app-server-client`), despite our docstring labelling it "official". We also depended on its private `_request_raw` API, which had no stability contract.

The `stdio` and `ws` transports already implement the same `_CodexClientProtocol` interface and speak the same JSON-RPC, so every "deep Codex" feature added in INT-226 (streaming, approvals, plans, diffs, token usage, slash commands) keeps working without the SDK transport. `examples/codex/01_basic_agent.py:73` already rejects `transport="sdk"`; no production config uses it.

Closes [INT-444](https://linear.app/thenvoi/issue/INT-444/drop-agpl-3-codex-app-server-sdk-dependency).

## Test plan

- [x] `uv run ruff check .`
- [x] `uv run ruff format .`
- [x] `uv run pyrefly check`
- [x] `uv run pytest tests/adapters/test_codex_adapter.py tests/converters/test_codex.py tests/framework_conformance/` — 538 passed
- [x] `uv run pytest tests/ --ignore=tests/integration/ --ignore=tests/e2e/` — 3078 passed (only failures are in `tests/integration/` and `tests/integrations/acp/test_e2e_*`, all needing live network/CLI infra unrelated to this change)
- [x] `rg "codex-app-server-sdk|codex_app_server|CodexSdkClient" src/ tests/ pyproject.toml uv.lock` returns nothing


darvell-thenvoi:
Fair enough, LGTM. 

I would suggest we keep https://github.com/openai/codex/tree/main/sdk/python in mind whenever it gets out of 'experimental'.

darvell-thenvoi (approved):

PR #291, opened 2026-05-05
fix(websocket): keep room deletion handler optional

darvell-thenvoi:
## Summary
- Make `WebSocketClient.join_room_participants_channel` backward-compatible by defaulting `on_room_deleted` to an async no-op.
- Add regression coverage for callers that omit the room deletion handler.
- Keep explicit `room_deleted` handlers routed and typed as before.

## Test plan
- `uv run pytest tests/websocket/test_client.py -v`
- `uv run ruff check . && uv run ruff format --check . && uv run pyrefly check`
- `uv run pytest tests/*.py tests/adapters tests/bridge tests/cli tests/client tests/compat tests/converters tests/core tests/docker tests/docs tests/example_agents tests/examples tests/framework_configs tests/framework_conformance tests/github_actions tests/integrations tests/platform tests/preprocessing tests/prompt_policies tests/prompts tests/runtime tests/testing tests/websocket -q`

Fixes INT-450.

### DIFF
--- a/src/thenvoi/client/streaming/client.py
+++ b/src/thenvoi/client/streaming/client.py
@@ -99,6 +99,10 @@ class RoomDeletedPayload(BaseModel):
     id: str
 
 
+async def _noop_room_deleted(_: RoomDeletedPayload) -> None:
+    return None
+
+
 class ParticipantAddedPayload(BaseModel):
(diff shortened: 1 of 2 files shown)


amit-gazal-band (approved):
Great fix.

PR #293, opened 2026-05-05
fix(integrations): require send message mentions

darvell-thenvoi:
## Summary
- Make CrewAI `thenvoi_send_message` mentions required and non-empty in the tool args schema.
- Validate CrewAI send-message args before platform calls, while still accepting JSON-string mention arrays for compatibility.
- Apply the same non-empty mentions requirement to the shared runtime `SendMessageInput` used by other tool surfaces.
- Add regression coverage for required schema fields, missing mentions, empty mentions, and runtime validation.

## Test plan
- `uv run python - <<'PY' ...` verified CrewAI and runtime generated schemas both require `mentions` with `minItems: 1`.
- `uv run pytest tests/integrations/test_crewai_tools.py tests/adapters/test_crewai_adapter.py tests/runtime/test_tool_definitions.py tests/runtime/test_tools.py -q`
- `uv run ruff check . && uv run ruff format --check . && uv run pyrefly check`
- Broad explicit non-live suite was also attempted; it hit an unrelated transient local MCP HTTP/SSE teardown failure. The isolated failing MCP test passed immediately on rerun.

Fixes INT-453.

### DIFF
--- a/src/band/adapters/crewai_flow.py
+++ b/src/band/adapters/crewai_flow.py
@@ -995,6 +995,11 @@ class SideEffectExecutor:
                 content=content,
                 mentions=mentions or None,
             )
+        except BandToolError:
+            # send_message rejected the call itself (e.g. missing mentions).
+            # Propagate the original message so the LLM learns the real cause.
+            await self._record_indeterminate(side_effect_key=side_effect_key)
+            raise
         except Exception:  # noqa: BLE001
             logger.warning(
                 "send_message failed for final side effect %s",
@@ -1380,14 +1385,22 @@ class CrewAIFlowSubCrewReporter:
                 content=content,
                 mentions=mentions or None,
             )
-        except Exception:  # noqa: BLE001
+        except BandToolError:
+            # send_message rejected the call itself (e.g. missing mentions).
+            # Record the indeterminate side effect but propagate the original
+            # message so the LLM learns the real cause and can retry.
+            await self._executor._record_indeterminate(side_effect_key=key)
+            raise
+        except Exception as exc:  # noqa: BLE001
             logger.warning(
                 "send_message failed for sub-Crew side effect %s",
                 key,
                 exc_info=True,
             )
             await self._executor._record_indeterminate(side_effect_key=key)
-            raise BandToolError(f"send_message failed for sub-Crew side effect {key}")
+            raise BandToolError(
+                f"send_message failed for sub-Crew side effect {key}"
+            ) from exc
         message_id = (
             getattr(message_response, "id", None)
             if message_response is not None
--- a/src/band/integrations/langgraph/langchain_tools.py
+++ b/src/band/integrations/langgraph/langchain_tools.py
@@ -13,6 +13,7 @@ from typing import Any
 
 from langchain_core.tools import StructuredTool
 
+from band.core.exceptions import BandToolError
 from band.core.protocols import AgentToolsProtocol
 from band.core.tool_filter import filter_tool_schemas
 from band.core.types import AdapterFeatures, Capability
@@ -113,9 +114,15 @@ def agent_tools_to_langchain(
         ) -> Any:
             try:
                 return await tools.execute_tool_call(_tool_name, kwargs)
+            except BandToolError as error:
+                # Deliberate, LLM-facing tool errors (e.g. the unified
+                # missing-mention guidance) must reach the transcript so the
+                # model can correct itself. These messages are authored for
+                # the LLM and contain no sensitive detail.
+                return f"Error executing {_tool_name}: {error}"
             except Exception:
-                # Tool errors feed back into the LLM transcript and may be
-                # relayed to chat. Keep the message generic; the full
+                # Unexpected failures feed back into the LLM transcript and may
+                # be relayed to chat. Keep the message generic; the full
                 # traceback (with paths, DB strings, tokens, etc.) only
                 # lives in the agent log.
                 logger.exception("Error executing platform tool %s", _tool_name)
--- a/src/band/integrations/parlant/tools.py
+++ b/src/band/integrations/parlant/tools.py
@@ -149,12 +149,11 @@ def create_parlant_tools(features: AdapterFeatures | None = None) -> list[Any]:
             return ToolResult(data="Error: No tools available in current context")
 
         try:
-            # Parse mentions from comma-separated string
+            # Parse mentions from comma-separated string. Empty-mention
+            # enforcement is centralized in AgentTools.send_message so the
+            # LLM gets the same message (including the send_event fallback)
+            # on every surface.
             mention_list = [m.strip() for m in mentions.split(",") if m.strip()]
-            if not mention_list:
-                logger.warning("[Parlant Tool] send_message: No mentions provided")
-                return ToolResult(data="Error: At least one mention is required")
-
             logger.info("[Parlant Tool] Sending message to: %s", mention_list)
             await tools.send_message(content, mention_list)
             # Mark that we sent a message via the tool (so adapter doesn't duplicate)
--- a/src/band/runtime/tools.py
+++ b/src/band/runtime/tools.py
@@ -77,6 +77,16 @@ class ToolDefinition:
 # --- Tool input models (single source of truth for schemas) ---
 
 
+#: Single source of truth for the "no mentions" rejection message. Every surface
+#: that enforces the ≥1-mention rule (runtime send_message, CrewAI, Parlant) must
+#: surface this exact text so the LLM always learns it can fall back to
+#: band_send_event for non-addressed status or observability.
+SEND_MESSAGE_REQUIRES_MENTION_MESSAGE = (
+    "band_send_message requires at least one mention. "
+    "Use band_send_event for non-addressed status or observability."
+)
+
+
 class SendMessageInput(BaseModel):
     """Send a message to the chat room.
 
@@ -88,7 +98,6 @@ class SendMessageInput(BaseModel):
     content: str = Field(..., description="The message content to send")
     mentions: list[str] = Field(
         ...,
-        min_length=1,
         description=(
             "List of participant handles to @mention. At least one required. "
             "For users: @<username> (e.g., '@john'). "
@@ -1228,16 +1237,10 @@ class AgentTools(AgentToolsProtocol):
         resolved_mentions = self._resolve_mentions(mentions or [])
 
         # Validate mentions are not empty — API requires ≥1 mention.
-        # Return a helpful error so the LLM can retry with proper mentions.
+        # Return a helpful error so the LLM can retry with proper mentions
+        # or fall back to band_send_event for non-addressed status.
         if not resolved_mentions:
-            participant_names = [
-                p.get("handle") or p["name"] for p in self._participants
-            ]
-            raise BandToolError(
-                "At least one mention is required. "
-                f"Available participants: {participant_names}. "
-                "Please retry with mentions specifying who this message is for."
-            )
+            raise BandToolError(SEND_MESSAGE_REQUIRES_MENTION_MESSAGE)
 
         logger.debug("Sending message to room %s", self.room_id)
 
--- a/src/band/runtime/tools.py
+++ b/src/band/runtime/tools.py
@@ -88,6 +88,7 @@ class SendMessageInput(BaseModel):
     content: str = Field(..., description="The message content to send")
     mentions: list[str] = Field(
         ...,
+        min_length=1,
         description=(
             "List of participant handles to @mention. At least one required. "
             "For users: @<username> (e.g., '@john'). "
--- a/tests/adapters/test_crewai_flow_subcrew_reporter.py
+++ b/tests/adapters/test_crewai_flow_subcrew_reporter.py
@@ -1,0 +1,90 @@
+"""Regression tests for CrewAIFlowSubCrewReporter error propagation.
+
+When the centralized ``AgentTools.send_message`` rejects a sub-Crew visible
+send (e.g. missing mentions), the reporter must propagate the original
+``BandToolError`` so the LLM learns the real cause, instead of wrapping it in
+the generic "send_message failed for sub-Crew side effect" message.
+"""
+
+from __future__ import annotations
+
+import sys
+from typing import Any
+from unittest.mock import AsyncMock, MagicMock
+
+import pytest
+
+
+@pytest.fixture(autouse=True)
+def _mock_crewai(monkeypatch: pytest.MonkeyPatch):
+    fake = MagicMock()
+    fake_flow_module = MagicMock()
+    fake_flow_module.Flow = type("Flow", (), {})
+    fake.flow = MagicMock()
+    fake.flow.flow = fake_flow_module
+    monkeypatch.setitem(sys.modules, "crewai", fake)
+    monkeypatch.setitem(sys.modules, "crewai.flow", fake.flow)
+    monkeypatch.setitem(sys.modules, "crewai.flow.flow", fake_flow_module)
+    yield
+
+
+from band.adapters.crewai_flow import (  # noqa: E402
+    CrewAIFlowSubCrewReporter,
+    SideEffectExecutor,
+)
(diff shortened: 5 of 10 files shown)


amit-gazal-band (changes requested):
We shouldn't enforce non-empty `mentions` at the SDK level. The intent is to have observability at the platform level: when an LLM forgets to provide mentions, that signal needs to surface in platform metrics/logs so we can measure mention compliance across agents and frameworks.

In the current PR state, empty/missing-mention tool calls never reach the platform:

- **CrewAI**: `BaseTool._validate_kwargs` runs `args_schema.model_validate(kwargs)` before `_run`. `min_length=1` on `_SendMessageInput.mentions` raises `ValidationError`, hint returned to the LLM, `_run` never executes, no REST call.
- **Runtime (other adapters)**: `execute_tool_call` calls `validate_tool_arguments(SendMessageInput, ...)`. Same `min_length=1` raises, error string returned to the LLM, `tools.send_message` never invoked, no REST call.

Net effect: zero platform visibility into bad LLM behavior, and the platform's validation layer becomes redundant.

### Proposal: split the two layers

- **Tool schema (LLM-facing):** keep `mentions` in the JSON schema's `required` array and keep the "at least one required" description, so the LLM is told the field is mandatory.
- **Pydantic runtime (SDK-facing):** drop `min_length=1` and switch to `Field(default_factory=list, ...)` so missing/empty mentions pass SDK validation and reach the platform. Let the platform return the validation error, which we surface back to the LLM as the tool result.

In pydantic v2 we can decouple them via `json_schema_extra`:

```python
class SendMessageInput(BaseModel):
    model_config = ConfigDict(
        json_schema_extra={"required": ["content", "mentions"]}
    )
    content: str = Field(..., description="...")
    mentions: list[str] = Field(
        default_factory=list,
        description="At least one required. ...",
    )
```

Same treatment for `_SendMessageInput` in `src/thenvoi/integrations/crewai/tools.py`. Net effect: LLM still sees a required field, but every empty/missing-mention call is observable on the platform.

darvell-thenvoi:
@amit-gazal-thenvoi 

Fixed in `ad30229` and smoke tested against the platform.

The SDK now keeps `mentions` in the LLM-facing schema `required` list, but no longer exports `minItems` or blocks missing/empty mentions in SDK validation. Empty/missing mentions now reach the platform and the platform returns the validation error.

Smoke test target:

```text
Platform: https://app.thenvoi.com
Agent id: dfadca0c-7955-4e72-92ab-f3405998c5ea
Agent identity: OK, name="Demo"
Room: 1684dd8c-018b-4085-bb70-86685d43a422
```

Direct SDK call:

```python
await AgentTools(room_id, client, participants=[]).send_message(
    "smoke empty mentions <uuid>",
    mentions=[],
)
```

Output:

```text
sdk_local_validation_blocked=false
platform_received_request=true
platform_rejected=true
status=422
request_id=02c4b865170c0235d7f33191451a112c
body.error.code=validation_error
body.error.details={"/message/mentions": ["Array length 0 is smaller than minItems: 1"]}
```

Runtime `execute_tool_call` with omitted mentions:

```python
await tools.execute_tool_call(
    "thenvoi_send_message",
    {"content": "smoke missing mentions <uuid>"},
)
```

Output:

```text
sdk_validation_error=false
platform_validation_error=true
request_id=c0d9a87cae4ea0fab790f8a6db4a65a1
```

Runtime `execute_tool_call` with empty mentions:

```python
await tools.execute_tool_call(
    "thenvoi_send_message",
    {"content": "smoke empty mentions <uuid>", "mentions": []},
)
```

Output:

```text
sdk_validation_error=false
platform_validation_error=true
request_id=ca43632335b4652b1bcdb3926b10e050
```

Local verification also passed:

```text
81 passed focused mention/schema tests
246 passed affected test files
ruff check passed
ruff format --check passed
pyrefly check passed
```

amit-gazal-band:
@darvell-thenvoi sorry for the back-and-forth on this — after running this end-to-end against dev I want to walk back my earlier "let the platform validate it" framing. The empirical results changed my mind. Posting a full writeup so the next reviewer doesn't have to re-derive any of this.

### tl;dr

- The original INT-453 problem (gpt-4o ignoring `mentions`) is fixed by **type migration + `min_items=1` in the SDK**, not by removing SDK validation.
- The "platform observability" concern I raised earlier was already solved by the existing `tool_call`/`tool_result` event sink — it just needs `enable_execution_reporting=True` on the adapter.
- Recommended direction: restore `Field(..., min_items=1)` on both `_SendMessageInput` and `SendMessageInput`, restore the runtime guard at `runtime/tools.py:1231`, and restore the parlant early-return.

### Why this regressed in the first place

The fix already existed once. INT-336 (`073622c`, 2026-04-26, by Yael) migrated CrewAI's `_SendMessageInput.mentions` from `str` to `list[str]` and added a bracketed-handle-string parser, specifically to break this gpt-4o loop. The commit message describes our exact symptom verbatim:

> "gpt-4o-mini emits the `mentions` arg as the bare-list string `"[@yael.avioz/test2]"`... validator only normalized lists, so the field validator and tool runner both fell back to `[]` ... leaving the agent in a tight retry loop and never producing a response."

INT-437 phase 0 (`d62dc4b`, 2026-04-27, "extract shared CrewAI tool wrappers") branched from main *before* INT-336 merged (their merge-base `c7188a18` does not contain INT-336). The refactor moved the tool definitions into the new shared `integrations/crewai/tools.py` file with the *pre-INT-336* `str` schema. When the refactor merged, the new file shadowed dev's fix — no Git conflict because the code moved files. So PR 293 isn't introducing a new fix; it's restoring INT-336.

### How it should reach the platform

With `mentions: list[str] = Field(..., min_items=1, ...)`:

1. **LLM-side** — Pydantic emits `minItems: 1` and includes `mentions` in `required`. CrewAI's OpenAI native provider already sets `strict: true` on every tool (`crewai/llms/providers/openai/completion.py:1583`), and `ensure_all_properties_required` forces `mentions` into the `required` array (`crewai/utilities/pydantic_schema_utils.py:421`). So gpt-4o sees a strict, required, `array<string>` field. (Note: OpenAI strict mode doesn't actually enforce `minItems` server-side — see [OpenAI docs](https://learn.microsoft.com/en-us/azure/foundry/openai/how-to/structured-outputs#unsupported-type-specific-keywords) — but the combination of structured signal + description + tool error is what nudges the model.)
2. **SDK-side** — If the LLM still sends `mentions=[]`, Pydantic's `min_items=1` raises `ValidationError`. CrewAI's `BaseTool._validate_kwargs` catches and returns the schema hint to the LLM. For runtime adapters, `validate_tool_arguments` returns an error string from `execute_tool_call`. Either way the LLM gets a clear "you must provide mentions" signal.
3. **Platform-side observability** — When the adapter is configured with `enable_execution_reporting=True` (or `features=AdapterFeatures(emit={Emit.EXECUTION})`), a `tool_call` event is POSTed to `/events` *before* tool execution and a `tool_result` event after. Both fire even when the SDK rejects, so the platform sees: (a) the bad args the LLM tried, (b) the error string that came back. Verified empirically against `app.thenvoi.com` with both anthropic and crewai adapters.
4. **LLM self-correction** — Once the LLM sees the tool error, it retries with proper mentions. Verified in dev: gpt-4o and gpt-4o-mini both converge on the first retry with the minimal patch; without it they loop until CrewAI's `max_iter` cap.

### Why I misled you on dropping SDK validation

I thought the cleanest fix was "let the platform own the rule" — drop SDK enforcement, push the empty-mention `POST /messages` through, let the platform 4xx, surface the 4xx as the tool error. Your `ad30229` does exactly that and the smoke test proves it works at the platform layer.

But two things I didn't fully appreciate when I posted that:

1. **The original INT-453 bug isn't about validation enforcement, it's about schema signaling.** gpt-4o's loop was driven by the `mentions: str = Field(default="[]")` schema saying "this is an optional string defaulting to empty". With that schema, neither approach (SDK guard or platform 4xx) makes the LLM stop sending `[]` — it just keeps retrying. The fix is the *type migration*, which both the dev branch (after `d62dc4b` merge) and PR 293's current state still have, but the missing piece was the `min_items=1` constraint that signals "non-empty required" via `minItems` in the tool schema. I tested this on dev: with PR 293's `default_factory=list` + json_schema_extra approach, gpt-4o would have nothing forcing the LLM-visible schema to express "non-empty". With `Field(..., min_items=1)`, the schema explicitly says `"minItems": 1` and Pydantic enforces it.

2. **Platform observability was never structurally missing — only off by default.** I conflated "events are off in the agent config we ran" with "platform can't see the violation". With `enable_execution_reporting=True`, every adapter (anthropic, crewai) already POSTs `tool_call` + `tool_result` events to `/events` regardless of where validation happens. We confirmed this end-to-end on dev: platform UI shows the bad-call args and the rejection. Removing SDK validation to "expose" violations was solving a problem that didn't exist — the existing event sink already did that job. The only adapter actually missing observability is **Parlant** (`SUPPORTED_EMIT = frozenset()`), which is a separate gap not addressed by either approach to this PR.

### Proposed direction

Closer to my original PR review minus the bad framing. Concretely:

```python
# src/thenvoi/integrations/crewai/tools.py
class _SendMessageInput(BaseModel):
    content: str = Field(..., description="The message content to send")
    mentions: list[str] = Field(
        ...,
        min_items=1,
        description=(
            "List of participant handles to @mention. At least one required. "
            "For users: @<username> (e.g., '@john'). "
            "For agents: @<username>/<agent-name> (e.g., '@john/weather-agent')."
        ),
    )

    @field_validator("mentions", mode="before")
    @classmethod
    def normalize_mentions(cls, v):
        if isinstance(v, str):
            try:
                return json.loads(v)
            except json.JSONDecodeError:
                return v
        return v
```

```python
# src/thenvoi/runtime/tools.py — restore Field(...) and the runtime guard
mentions: list[str] = Field(..., min_items=1, description=...)
```

Restore `runtime/tools.py:1231` `if not resolved_mentions: raise ThenvoiToolError(...)` (the "Available participants: [...]" hint is high-signal for the LLM — Claude self-corrects on it in 1 retry).

Restore `integrations/parlant/tools.py` early-return on empty mentions.

Drop the `json_schema_extra=_require_mentions_in_schema` callbacks — `Field(...)` already puts `mentions` in `required`, and `min_items` already emits `minItems`. The callback approach is needed only when you want schema signal without Pydantic enforcement, which is no longer our goal.

Optional separate cleanup: the `inspect` import + custom-tools `_run` refactor in this PR isn't related to INT-453 — worth either splitting into its own PR or calling out in the description so reviewers can evaluate it independently.

### Empirical results (dev, 2026-05-07)

Tested CrewAI + gpt-4o, gpt-4o-mini, and `anthropic/claude-sonnet-4-5-20250929`:

- **dev as-is** (`mentions: str`): gpt-4o and gpt-4o-mini loop indefinitely; Claude self-corrects in 1 retry.
- **minimal patch** (str → list[str] + min_items=1 + matching validator/`_run`): all three models reach platform on first try (Claude) or after a single resolver-format error retry (both gpt-4o variants). No loop.

The `tool_call`/`tool_result` `/events` posts were observed in all cases when `Emit.EXECUTION` was enabled, confirming the platform observability path works without removing SDK guards.

amit-gazal-band (changes requested):
@darvell-thenvoi
See my previous comment. I looked into it deeply and apparently I mislead you. Claude wrote the comment, but believe me that it is accurate 😉
Please align with that and I'll approve.
Sorry for the back and forth...

darvell-thenvoi:
@amit-gazal-thenvoi aligned this with the later review direction in 92567c2.

What changed:
- restored SDK-side non-empty mention validation for `SendMessageInput`
- restored CrewAI non-empty mention schema/runtime validation while keeping bare-list string normalization
- restored a Parlant early return for empty mention strings
- added regression coverage for schema `minItems`, missing/empty mentions, and CrewAI bare-list string parsing

Validation:
- `uv run pytest tests/runtime/test_tools.py::TestAgentToolsSendMessage tests/runtime/test_tools.py::TestToolInputModels tests/integrations/test_crewai_tools.py::TestToolSetComposition`
- `uv run ruff check src/thenvoi/runtime/tools.py src/thenvoi/integrations/crewai/tools.py src/thenvoi/integrations/parlant/tools.py tests/runtime/test_tools.py tests/integrations/test_crewai_tools.py`
- commit hooks: hardcoded secrets, ruff check, ruff format, pyrefly, commitizen

Re-requested review.

darvell-thenvoi:
@amit-gazal-thenvoi re-requested review on latest head eb5583f8846be0a965bcf803bd0000ee1b334408. The latest branch state is aligned with your later direction: SDK-side non-empty mention validation is restored, CrewAI bracketed-handle list parsing is restored, the runtime guard is restored, Parlant keeps its early-return path, and execution-reporting coverage proves bad tool calls still emit tool_call/tool_result events when enabled.\n\nCurrent GitHub Actions on eb5583f8846be0a965bcf803bd0000ee1b334408 are all green: lint, packaging, test (3.11), test (3.12), test-parlant (3.11), test-parlant (3.12), and Validate PR Title.

darvell-thenvoi:
Rechecked this branch against the latest requested direction from Amit. No new commit was needed on top of current head `eb5583f`: it already restores SDK-level non-empty mention validation.

Current behavior verified in the diff:
- `SendMessageInput.mentions` is required with `min_length=1`.
- `AgentTools.send_message()` rejects empty resolved mentions before REST and returns a helpful `ThenvoiToolError` with available handles.
- CrewAI `_SendMessageInput.mentions` is `list[str]` with `min_length=1`, while still normalizing JSON-string lists like `'["@john"]'`.
- Parlant keeps the early return for empty parsed mentions.

Validation:
- `uv run pytest tests/runtime/test_tools.py tests/runtime/test_tool_definitions.py tests/adapters/test_crewai_adapter.py tests/integrations/test_crewai_tools.py -k 'mention or SendMessageInput or send_message' -q --no-cov` -> `39 passed`
- `uv run ruff check src/thenvoi/runtime/tools.py src/thenvoi/integrations/crewai/tools.py src/thenvoi/integrations/parlant/tools.py tests/runtime/test_tools.py tests/adapters/test_crewai_adapter.py tests/integrations/test_crewai_tools.py` -> passed

darvell-thenvoi:
> *This was generated by AI during triage.*

Updated the branch to remove the stale Parlant assertion left over from the earlier platform-validation direction. Parlant test coverage now matches the current SDK-side behavior: empty mentions are rejected before `tools.send_message` is called.

Validation:
- `uv run pytest tests/integrations/test_crewai_tools.py tests/adapters/test_crewai_adapter.py tests/runtime/test_tool_definitions.py tests/runtime/test_tools.py tests/runtime/test_mcp_server.py tests/integrations/parlant/test_tools.py -q --no-cov` -> `249 passed`
- `uv run ruff check tests/integrations/parlant/test_tools.py && uv run ruff format --check tests/integrations/parlant/test_tools.py` -> passed
- Commit hooks passed, including hardcoded secrets, ruff, format, pyrefly, and commitizen.

amit-gazal-band (changes requested):
Requesting changes. Ran all surfaces with empty mentions to confirm:

1. The same "no mentions" condition produces different LLM-facing messages depending on the surface: runtime send_message says "requires at least one mention. Use thenvoi_send_event..."; CrewAI's basic tool inherits that (good); Parlant returns its own "Error: thenvoi_send_message requires at least one mention." from a duplicate pre-check (drops the send_event hint); and the crewai_flow reporter catches the ThenvoiToolError and re-raises "send_message failed for sub-Crew side effect {key}", so the LLM never learns mentions were the problem. Enforcement is correctly centralized in AgentTools.send_message — the messages aren't. Suggest one shared message constant, dropping Parlant's pre-check, and not swallowing ThenvoiToolError in the flow reporter's except.

2. In the CrewAI SendMessageTool, _SendMessageInput.model_validate runs inside execute, so a ValidationError is caught by the generic except in _execute_tool and returned to the LLM as a raw multi-line Pydantic dump (e.g. "1 validation error for _SendMessageInput
content
  Field required ... https://errors.pydantic.dev/..."). The runtime path returns "Invalid arguments for thenvoi_send_message: content: Field required". Catch ValidationError separately and format it the same way.

PR #294, opened 2026-05-06
feat(integrations): update LangGraph SDK support

darvell-thenvoi:
## Summary

Updates the LangGraph integration to track the LangChain 1.x / LangGraph 1.x
release line, fixes how the system prompt is delivered, repairs tool-call
history reconstruction so multi-turn tool flows survive checkpointer rehydrate,
emits `Emit.EXECUTION` tool events in the shared shape so the LangGraph adapter
matches the rest of the SDK, and refreshes `examples/langgraph/` so each example
actually runs against production.

Linear: [INT-445](https://linear.app/thenvoi/issue/INT-445) — "Update LangGraph to its latest version"

## Why

LangChain 1.x replaced `langgraph.prebuilt.create_react_agent` with
`langchain.agents.create_agent`, restructured tool-call/result message
shapes, and tightened a number of internal assumptions. Without this PR the
adapter still imports the old prebuilt path, the converter mismatches tool
calls under `gpt-5.4-mini`'s parallel-tool behaviour, and the system prompt
is appended *after* base SDK plumbing under "## Developer Instructions" —
which buries developer personas behind ~2k tokens of SDK boilerplate.

The LangGraph adapter also fell behind the shared SDK contract for tool
plumbing: it carried 7 hand-rolled platform-tool wrappers (drifting from
`AddParticipantInput`, memory/contact gating, etc.), emitted raw stream
events as tool history, and could not opt into `Emit.EXECUTION` because the
event shapes did not match the shared parsers other adapters use.

## Dependency bumps (`uv.lock` / `pyproject.toml`)

| Package | Before | After |
|---|---|---|
| `langchain` | 1.2.15 | 1.2.17 |
| `langchain-anthropic` | 1.4.1 | 1.4.3 |
| `langchain-core` | 1.3.0 | 1.3.3 (+ new `langchain-protocol` 0.0.15 dep) |
| `langchain-openai` | 1.1.14 | 1.2.1 |

`pyproject.toml` adds `langchain>=1.0.0` to the `a2a_gateway_demo` extra so the
demo orchestrator can import `langchain.agents.create_agent` directly.

### `src/thenvoi/adapters/langgraph.py`
- Switch the simple-pattern path from `langgraph.prebuilt.create_react_agent`
  to `langchain.agents.create_agent` (the supported entry point on 1.x).
- Drop the `THENVOI_SYSTEM_PROMPT_CONFIG_KEY` escape hatch. The adapter now
  has exactly one prompt-delivery path: render in `on_started`, prepend a
  single `("system", rendered_prompt)` on bootstrap, let the checkpointer
  carry it forward. We deliberately do **not** pass `system_prompt=` to
  `create_agent` — the bootstrap-once + checkpointer pattern matches every
  other Band adapter and avoids two competing sources of system context.
- `_bootstrapped_rooms` becomes an `OrderedDict` with FIFO eviction at the
  warn threshold so long-lived processes don't leak unbounded room IDs.
- Bootstrap marking now happens **after** the stream completes successfully,
  so a failed run no longer poisons the cache and re-bootstraps cleanly on
  retry.
- Add `Emit.EXECUTION` to `SUPPORTED_EMIT` and emit `tool_call` /
  `tool_result` events with the shared SDK payload shape
  (`{name, args|output, tool_call_id}`) instead of dumping raw LangGraph
  stream events. Opt in via `AdapterFeatures(emit={Emit.EXECUTION})` or the
  `enable_execution_reporting=True` deprecation shim (parallel to
  `enable_memory_tools`).
- Replace `except Exception: pass` in the error-reporting fallback with
  `logger.exception(...)` so failed error reports remain observable.

### `src/thenvoi/converters/langchain.py`
- Switch tool-event parsing to the shared
  `thenvoi.converters._tool_parsing.parse_tool_call` /
  `parse_tool_result` helpers used by every other adapter. Pre-fix LangGraph
  parsed its own raw `on_tool_start`/`on_tool_end` event blobs; post-fix it
  consumes the same `{name, args|output, tool_call_id}` envelope the new
  adapter emits, with a legacy fallback for events written before this PR.
- `_pop_matching_tool_call` matches on `tool_call_id` first and falls back
  to most-recent name match. Under `gpt-5.4-mini`'s parallel tool calls the
  old earliest-name-wins logic paired wrong tool_calls with wrong
  tool_results, which the model then saw as orphaned `ToolMessage`s and
  refused to recover from.
- `_extract_tool_call_id` (legacy fallback path) handles current LangChain
  structured output shapes (`dict` with `tool_call_id` / nested `messages` /
  nested `output`, attribute access for `ToolMessage` objects), with the
  legacy string-repr regex as a final fallback.
- Together: tool-call/tool-result pairs stay paired across history
  rehydrate, preserving the strict role alternation the LLM API requires.

### `src/thenvoi/integrations/langgraph/langchain_tools.py`
- Rebuild the LangGraph platform-tool wrappers from the shared
  `TOOL_DEFINITIONS` registry (`iter_tool_definitions`) instead of hand-
  maintaining one wrapper per tool. This drops ~250 lines of duplication and
  keeps LangGraph in lockstep with the canonical SDK tool schemas — including
  `thenvoi_add_participant(identifier, role="member")` (room-role parameter
  preserved unchanged), contact/memory gating, defaults, and validation.
- Route wrapper execution through `AgentTools.execute_tool_call()` so
  validation and result serialization use the same boundary as the other
  schema-driven integrations.
- Hub-room callers force `include_contacts=True` regardless of the adapter's
  normal capability gate, matching CrewAI's behaviour.

### `src/thenvoi/integrations/langgraph/graph_tools.py`
- Modernise typing to PEP 604 / PEP 585.
- Fix the docstring example to import from `standalone_calculator` (the
  current module) instead of the non-existent
  `standalone_calculator_graph`, and to construct via `LangGraphAdapter`
  directly instead of the removed `create_langgraph_agent` helper.
- Add `Annotated[..., InjectedToolArg]` plumbing so per-call config can be
  injected without leaking into the LLM-visible schema.

### `src/thenvoi/runtime/prompts.py`
- `custom_section` now leads the rendered prompt; agent identity follows;
  `BASE_INSTRUCTIONS` follows; capability-gated `MEMORY_SECTION` /
  `CONTACT_SECTION` come last. Previously `custom_section` was appended at
  the end under "## Developer Instructions" — ~2k tokens after the
  identity line — which let SDK plumbing dominate over the developer's
  persona/role.

  Before:
  ```
  You are TestBot, A test agent.

  ## Environment ... (BASE_INSTRUCTIONS, ~2k tokens) ...

  ## Developer Instructions

  Focus on Python programming.
  ```
  After:
  ```
  Focus on Python programming.

  You are TestBot, A test agent.

  ## Environment ... (BASE_INSTRUCTIONS) ...
  ```

### `examples/prompts/characters.py`
- Tom and Jerry prompts now open with explicit identity ("You are Tom, the
  cat from Tom & Jerry. ... never break character") instead of opening
  with gameplay meta-rules ("## How to Use Thoughts"). Verified live:
  both stay in character on `gpt-5.4-mini` after this change.

### `examples/a2a_gateway/demo_orchestrator/agent.py`
- Standalone orchestrator (does **not** use `LangGraphAdapter`; builds its
  own graph for the A2A demo). Migrated from
  `langgraph.prebuilt.create_react_agent(prompt=...)` to
  `langchain.agents.create_agent(system_prompt=...)`. The `prompt=` →
  `system_prompt=` rename is required by LangChain 1.x; `system_prompt=`
  is the real parameter on `create_agent` (`inspect.signature` confirms).
  This is consistent with the SDK rule — the rule is "`LangGraphAdapter`
  doesn't pass `system_prompt=` because the adapter manages its own
  bootstrap-once delivery"; standalone consumers of `create_agent` are
  free to use the public API as designed.
- `response_format` now passes the `ResponseFormat` model directly and the
  format instructions are concatenated into `system_prompt` (the 1.x API
  no longer accepts `(prompt, schema)` tuples for `response_format`).

### `examples/langgraph/`
- All eight existing examples (`01`–`08`) updated for the post-refactor
  adapter constructor.
- `05_rag_as_tool.py`: prompt updated to use participant IDs (not handles)
  in `mentions=[...]`, matching the platform contract.
- New `09_research_ops_orchestrator.py`: a real `graph_factory` example
  that builds a `StateGraph` with `MessagesState`, a `ToolNode` for
  platform tools, and two delegated subgraphs (calculator + SQL) wrapped
  via `graph_as_tool`. First example exercising the `graph_factory=` path
  end-to-end.
- README rewritten around the current contract.
- `agent_config.yaml.example` gains a `research_ops_agent` stanza.
- `docker-compose.yml` adds `langgraph-09-research-ops`.

### Tests
- `tests/adapters/test_langgraph_adapter.py`: bootstrap-once invariant,
  `create_agent` integration, tool-call rehydrate, deprecation shim for
  `enable_memory_tools` and `enable_execution_reporting`, multi-turn
  checkpointer carry-forward, and the new `Emit.EXECUTION` payload shape.
- `tests/converters/test_langchain.py`: parallel-tool-call matching against
  the new shared payload shape, structured `ToolMessage` shapes, legacy
  string-repr fallback.
- `tests/integrations/langgraph/test_langchain_tools.py`: the registry-
  driven wrappers route through `execute_tool_call`, schema includes the
  full `AddParticipantInput` (identifier + role), hub-room override, and
  error propagation.
- `tests/integrations/langgraph/test_graph_tools.py`: subgraph wrapper
  semantics.
- `tests/test_prompts.py` updated for the new ordering.

## Local validation

| Check | Command | Result |
|---|---|---|
| Focused LangGraph + converter + crew + runtime + pydantic_ai suites | `uv run pytest tests/integrations/langgraph/test_langchain_tools.py tests/adapters/test_langgraph_adapter.py tests/converters/test_langchain.py tests/runtime/test_tools.py tests/integrations/test_crewai_tools.py tests/adapters/test_crewai_adapter.py -q` | 232 passed |
| Lint | `uv run ruff check .` | passed |
| Format | `uv run ruff format --check ...` | passed |
| Type | `uv run pyrefly check` | 0 errors |
| Lockfile | `uv lock --check` | passed |
| Docker config | `docker compose config --services` | new `langgraph-09-research-ops` listed alongside the other LangGraph services |

## Live production validation — raw API records

Each example was booted as its own subprocess against
`https://app.thenvoi.com`, then driven by a real user message: the user
creates a fresh chat room, adds the agent, posts
`@<agent-handle> <prompt>`, and the test polls
`GET /api/v1/human/chats/{room}/messages` for the persisted record.
Model: `gpt-5.4-mini`.

### Round-trip A — `02_custom_tools` (custom `calculate` tool)

- Room: `7aea78e8-5c19-4128-b5be-4cd380b664d7`
- Trigger message: `0797c95c-e28d-4661-bab5-7f99856235be`
- User prompt: `@darv/claude-code What is 12 * 11? Use the calculate tool, then send the answer.`
- Agent text reply: `@[[<user>]] The result of 12 * 11 is 132.`

| inserted_at UTC | message_id | type | content (truncated) |
|---|---|---|---|
| 00:31:51.815 | `0797c95c…` | text (user) | `@[[…agent…]] What is 12 * 11? Use the calculate tool, then send the answer.` |
| 00:31:52.849 | `bdb41bcc…` | tool_call | `name=calculate input={"operation":"multiply","left":12,"right":11} run_id=019dff…` |
| 00:31:52.950 | `520a6caa…` | tool_result | `name=calculate output="content='Result: 132.0' tool_call_id='call_AHFOWOHe…'"` |
| 00:31:53.625 | `ed2ba7ec…` | tool_call | `name=thenvoi_send_message input={"content":"The result of 12 * 11 is 132.","mentions":["@darv"]}` |
| 00:31:53.750 | `4c0bb3f2…` | text | `@[[…user…]] The result of 12 * 11 is 132.` |
| 00:31:53.860 | `cce97869…` | tool_result | `name=thenvoi_send_message output="id='4c0bb3f2-2f5d-48de-9782-b6135034db2e' recipients=[…handle='darv'…]"` |

This exercises the converter fix: `calculate` and `thenvoi_send_message`
execute back-to-back, each `tool_call` is paired with exactly one
`tool_result` carrying the matching `tool_call_id`, and the platform
`text` message ID matches the ID returned by `thenvoi_send_message`'s
response object. Pre-fix, the parallel-tool matcher mispaired these and
the next turn's rehydrate would orphan a `ToolMessage`.

### Round-trip B — `06_delegate_to_sql_agent` (graph-as-tool subgraph)

- Room: `0f0422e2-ab78-47ca-909d-11adcb9d4b3a`
- Trigger message: `0d645590-5f5c-4577-bbd6-08c189cd23e4`
- User prompt: `@darv/claude-code Use the SQL tool to count rows in the Employee table; report the number.`
- Agent text reply: `@[[<user>]] The Employee table contains 8 rows.`

| inserted_at UTC | message_id | type | content (truncated) |
|---|---|---|---|
| 00:31:16.669 | `0d645590…` | text (user) | `@[[…agent…]] Use the SQL tool to count rows in the Employee table; report the number.` |
| 00:31:18.441 | `3a4c8d99…` | tool_call | `name=database_assistant input={"messages":[{"role":"user","content":"Count the number of rows in the Employee table"}]}` |
| 00:31:19.568 | `81426498…` | tool_call | `name=sql_db_list_tables input={} run_id=019dffd8-9dd2-…` |
| 00:31:19.656 | `049d1b03…` | tool_result | `name=sql_db_list_tables output="content='Album, Artist, Customer, Employee, Genre, Invoice, InvoiceLine, MediaType, Playlist, PlaylistTrack, Track'"` |
| 00:31:20.355 | `7a2410cb…` | tool_call | `name=sql_db_schema input={"table_names":"Employee"}` |
| 00:31:20.440 | `95f240f5…` | tool_result | `name=sql_db_schema output="content='\\nCREATE TABLE \"Employee\" (…)'"` |
| 00:31:21.438 | `81466d6e…` | tool_call | `name=sql_db_query input={"query":"SELECT COUNT(*) FROM Employee;"}` |
| 00:31:21.523 | `47f07a53…` | tool_result | `name=sql_db_query output="content='[(8,)]' tool_call_id='call_8pl2gO7p6bplzv63e…'"` |
| 00:31:22.112 | `d48d0dce…` | tool_result | `name=database_assistant output="content='The Employee table contains 8 rows.' tool_call_id='call_nJNu3a2HxJycGhcGbKX1K7z9'"` |
| 00:31:23.440 | `d7c722d1…` | tool_call | `name=thenvoi_send_message input={"content":"The Employee table contains 8 rows.","mentions":["@darv"]}` |
| 00:31:23.587 | `4de78049…` | text | `@[[…user…]] The Employee table contains 8 rows.` |
| 00:31:23.676 | `f55c0071…` | tool_result | `name=thenvoi_send_message output="id='4de78049-…' recipients=[…handle='darv'…]"` |

This proves the `graph_as_tool` integration end-to-end: the LLM picks
`database_assistant`, the wrapped subgraph runs four nested SQL tool
calls, the subgraph's final state is formatted back into a single
`database_assistant` `tool_result`, and only then does the orchestrator
issue the platform `thenvoi_send_message`. Every nested tool call is
correctly paired in the rehydrated history.

### Round-trip C — bootstrap-once invariant

Captured against `https://app.thenvoi.com` on commit `923035c` (the
`THENVOI_SYSTEM_PROMPT_CONFIG_KEY` removal):

- Room `8b319c00-cea6-4abf-8e52-b52fdd57e93c`: rendered system prompt
  1787 chars (`contains_agent_name=True`, `contains_custom=True`); adapter
  chose `thenvoi_send_event`; observed via REST `fetch_room_context`:
  `tool_call(on_tool_start thenvoi_send_event)` →
  `thought("Hello everyone! I'm here to assist…")` →
  `tool_result(on_tool_end success=True)`.
- Room `306ac939-3049-44dd-b57a-6465a06b6232`: turn 1 bootstrap leaves
  the checkpointer at `total=5, system=1, system_len=1727`; turn 2
  (`is_session_bootstrap=False`) leaves it at `total=9, system=1` — the
  adapter does **not** prepend a second system message; the checkpointer
  carries the original `SystemMessage` forward.

### All 9 examples — summary

For completeness, every numbered example was driven by the same
human-in-the-loop harness on `gpt-5.4-mini`:

| # | Example | Room | Reply (excerpt) |
|---|---|---|---|
| 01 | `01_simple_agent` | `c90f0559-e7bb-4ccd-b064-0db850368f9c` | "Hey Darvell — not much on my side..." |
| 02 | `02_custom_tools` | `7aea78e8-5c19-4128-b5be-4cd380b664d7` | "The result of 12 * 11 is 132." |
| 03 | `03_custom_personality` | `e8532c7b-3964-4a2e-a447-643976e5c73a` | "Ahoy matey!..." (pirate persona) |
| 04 | `04_calculator_as_tool` | `0e633abb-cc06-47be-9991-db192c5a2402` | "7 + 8 is 15" |
| 05 | `05_rag_as_tool` | `c47135fe-fb33-4acd-94b6-a04ea2b30268` | LangGraph summary via `rag_tool` |
| 06 | `06_delegate_to_sql_agent` | `0f0422e2-ab78-47ca-909d-11adcb9d4b3a` | "The Employee table contains 8 rows." |
| 07 | `07_tom_agent` | `85c78be5-24e8-49e8-bf19-caaf2aa9da45` | "Meow! I'm Tom — clever, sneaky, whisker-twitching cat..." |
| 08 | `08_jerry_agent` | `a3ab9098-da95-4ba5-a340-861302656056` | "Squeak! Jerry at your service..." |
| 09 | `09_research_ops_orchestrator` | `4cc79b27-2d4e-4071-a00e-e98ae4d6f6b9` | "Quick status: I'm online and ready..." |

## Test plan

- [x] `uv run pytest tests/integrations/langgraph/test_langchain_tools.py tests/adapters/test_langgraph_adapter.py tests/converters/test_langchain.py tests/runtime/test_tools.py tests/integrations/test_crewai_tools.py tests/adapters/test_crewai_adapter.py -q` (232 passed)
- [x] `uv run pytest tests/test_prompts.py -v`
- [x] `uv run ruff check .` / `uv run ruff format --check ...` / `uv run pyrefly check`
- [x] `uv lock --check`
- [x] `docker compose config --services` includes `langgraph-09-research-ops`
- [x] Live human-in-the-loop chat on `https://app.thenvoi.com` for all 9 examples
- [x] Raw REST API records captured for the two most tool-heavy examples (02, 06) and inlined above


bastizol-wq:
On “feat(integrations): update LangGraph SDK support”: you can get a lot of safety by doing a deterministic pre-flight before the agent calls the API.

If you already have OpenAPI/MCP metadata, you can make an allow/warn/block decision deterministically (and log it) before execution.

ATCP sits in front of tool execution: evaluate first, then decide whether the agent can call.

ATCP overview: https://www.sp-ai.io/atcp/

Curious what you’re using today for tool allowlists / safety receipts — happy to compare notes.

ed-lepedus-thenvoi (changes requested):
# Review: PR #294

## TL;DR
The advanced-pattern reference `examples/langgraph/09_research_ops_orchestrator.py` ships a `nonlocal compiled_graph` cache that pins the first room's tool wrappers permanently. Confirmed live against `app.band.ai` (real registered agent + real OpenAI LLM, one agent across two rooms): every room's `thenvoi_send_message` dispatches to the first room. The simple pattern (`llm=` + `checkpointer=`) is unaffected. Other PR claims — system-prompt persistence, parallel tool-call pairing, bootstrap re-entry, registry-derived tools — verified clean.

## Methodology
1. Static review of diff + working tree (31 files, +1430/-559).
2. Six SDK-level experiments mixing mocked `AgentToolsProtocol` instances with real `langgraph.StateGraph` / `ToolNode` / `InMemorySaver`.
3. Live reproduction on production:
   - Registered a temporary agent via `human_api_agents.register_my_agent`.
   - Ran it with `LangGraphAdapter(graph_factory=…)` mirroring example 09's caching pattern.
   - Drove from the user side: created two rooms, added the agent to both, sent identical greetings.
   - Fetched both rooms' messages via `human_api_messages.list_my_chat_messages` and compared.
   - Cleaned up: agent deletion, env-only credential handling.

## Findings

### 🔴 Cached-graph factory dispatches to wrong room (live-confirmed)
`examples/langgraph/09_research_ops_orchestrator.py:101-124` caches the compiled graph behind `nonlocal compiled_graph`. The graph captures `band_tools` (whose `StructuredTool` closures hold the per-room `AgentToolsProtocol`) at compile time. On subsequent messages the cached graph still has room A's wrappers, so tool calls from room B dispatch to room A.

Live result (one user, one agent, two rooms, identical greetings):
```
ROOM A: 2 agent text messages   ← second one triggered from room B
ROOM B: 0 agent text messages   ← agent never replied here
```
Same failure mode reproduced in unit + integration tests.

### 🟡 Concurrent-bootstrap race (latent)
Two interleaved `on_message(is_session_bootstrap=True)` calls for the same room can both pass `room_id not in self._bootstrapped_rooms` (the gate-set is at line 257-258, after streaming). Both will inject hydrated history. Reproduced in unit test by holding the first stream mid-completion; whether the runtime ever produces concurrent same-room `on_message` calls in production I didn't fully verify — flagging as latent, not confirmed-in-prod.

### 🟢 Verified clean
- Simple-pattern factory rebuilds per message → multi-room routes correctly.
- Bootstrap-once + checkpointer carry-forward: turn 2 sees the `SystemMessage`; no second prepend, no loss.
- `on_cleanup` correctly resets bootstrap tracking → re-bootstrap rehydrates history.
- `_pop_matching_tool_call` pairs out-of-order parallel tool results by `tool_call_id`.
- Drift test enforces `iter_tool_definitions` import in `langchain_tools.py`.

### 🟢 Style / minor
- `src/thenvoi/adapters/langgraph.py` lines 9, 86, 91, 137 use `typing.List[Any]` instead of `list[Any]` (project CLAUDE.md prefers PEP 585).
- Error reporting forwards raw exception text to user-visible surfaces (`adapters/langgraph.py:272`, `integrations/langgraph/langchain_tools.py:58`) — could leak DB strings / tokens / paths. Consider a redacted user-facing message + `logger.exception(...)` for the trace.
- `runtime/prompts.py:81-86` `TEMPLATES` is marked DEPRECATED but still serves as the default `template=`. Either remove or actually `DeprecationWarning` on access.
- `_extract_tool_call_id` recursion lacks a depth bound. Practical risk minimal.

## Recommendations

### Required before merge
1. **Fix `examples/langgraph/09_research_ops_orchestrator.py`.** Pick one:
   - *(simplest)* Drop the `nonlocal compiled_graph` cache; rebuild every call. One-line change, matches the simple-pattern factory.
   - Cache topology only; thread tools through `RunnableConfig` so wrappers resolve at tool-call time.
   - Stopgap: prominent docstring warning that the cache makes the example single-room only, plus a tracked follow-up issue.
2. **Add a multi-room conformance test** for the advanced `graph_factory` pattern (two rooms, two `AgentTools` mocks, cached factory).

### Optional
3. Sweep `List[Any]` → `list[Any]` in `adapters/langgraph.py`.
4. Add a unit test for cross-turn system-prompt persistence (currently only live-validated).
5. Redact exception details in user-visible error paths.

## Verdict
**Request changes.** The PR's load-bearing changes are high-quality — the converter parallel-tool fix is real and tested, the bootstrap-once invariant is correct, the registry-derived wrappers are a meaningful cleanup. But the canonical advanced-pattern example demonstrates a silent multi-room correctness bug that's easy to copy. Fix the example (or document the constraint), and this is otherwise ready.

darvell-thenvoi:
Thanks @ed-lepedus-thenvoi — the live two-room reproduction caught the real bug.

Pushed `9f46c10`:

**Required fix #1 — `examples/langgraph/09_research_ops_orchestrator.py`**

Took the simplest option: dropped the `nonlocal compiled_graph` cache, the factory rebuilds the graph every call. The `InMemorySaver` checkpointer (the load-bearing piece across turns) stays outside the inner closure; the `StateGraph` recompile is cheap. Added a docstring note explaining *why* the rebuild has to happen there: each room's `band_tools` closures need to land in *that* call's `ToolNode`.

**Required fix #2 — multi-room conformance test**

Added `TestGraphFactoryMultiRoom::test_factory_receives_distinct_tools_per_room` in `tests/adapters/test_langgraph_adapter.py`. The test drives two `on_message` calls with two distinct `AgentToolsProtocol` mocks against the same adapter and asserts:

1. `graph_factory(band_tools)` is called once per room (not cached at the adapter layer either).
2. Each call gets its own `StructuredTool` wrappers — the lists are distinct objects.
3. Invoking room A's wrapper dispatches to room A's `tools.execute_tool_call`, never room B's.

This pins the adapter-level contract regardless of how an example handles caching, so a future cached `graph_factory` example would still get caught at the unit-test level once the user follows the proper pattern. Would have failed against the cached pattern from the original PR.

**Skipping for now (will follow up):**

- 3, 4, 5 (typing sweep, system-prompt cross-turn unit test, exception-redaction in user-visible paths) — happy to take these in a follow-up; calling them out so they don't get lost.

`uv run pytest tests/adapters/test_langgraph_adapter.py tests/integrations/langgraph/ tests/converters/test_langchain.py --no-cov -q` → 60 passed. Lint, format, pyrefly clean.

darvell-thenvoi:
Pulled the rest in — `85c9412`:

**3 (typing sweep) — ✅**
Switched the remaining `List[Any]` annotations in `src/thenvoi/adapters/langgraph.py` to PEP 585 `list[Any]` (constructor + simple-pattern factory). Dropped the unused `List` import.

**4 (cross-turn system-prompt unit test) — ✅**
Added `TestSystemPromptCrossTurn` in `tests/adapters/test_langgraph_adapter.py` with two cases:

1. *Mocked-graph contract*: bootstrap turn 1 prepends exactly one `("system", rendered_prompt)`; non-bootstrap turn 2 (same room) prepends NO system message — the adapter relies on the checkpointer.
2. *End-to-end with a real `InMemorySaver`*: a tiny `StateGraph` echo-node records what `SystemMessage`s it sees. After turn 2 (non-bootstrap), the node observes the *same* `SystemMessage` content as turn 1, proving the checkpointer (not the adapter) carries it forward.

So the contract from the PR description ("`is_session_bootstrap=False` → adapter does not prepend a second system message; checkpointer carries the original `SystemMessage`") is now pinned at the unit-test level, not just live-validated against `app.band.ai`.

**5 (redact exception detail in user paths) — ✅**

- `src/thenvoi/adapters/langgraph.py:269` — `except Exception:` now sends a generic `"Internal error while processing message; see agent logs."` payload via `tools.send_event`. Switched from `logger.error(..., exc_info=True)` to `logger.exception(...)` for the traceback.
- `src/thenvoi/integrations/langgraph/langchain_tools.py:58` — wrapper errors now return `"Error executing {tool}: see agent logs."` and emit the trace via `logger.exception(...)`. Added a module-level `logger = logging.getLogger(__name__)`.
- Both spots have tests pinning the redaction:
  - `test_langchain_tools.py:148`: asserts the result string AND that the original exception text (`"platform unavailable"`) does NOT appear in it.
  - `test_langgraph_adapter.py::TestErrorHandling::test_reports_error_on_graph_failure`: asserts `send_event` was called with `message_type="error"` AND that the raw `"Graph error!"` text is NOT in `content`.

**Status**

- Tests: `tests/adapters/test_langgraph_adapter.py tests/integrations/langgraph/ tests/converters/test_langchain.py` → **62 passed** (was 60; +2 new system-prompt tests).
- `ruff check` / `ruff format` / `pyrefly check`: clean.

Should be ready for another look.

darvell-thenvoi:
nvm — updated to hit 3, 4, 5 as well. See `85c9412` and the comment above.

eric-descourtis-thenvoi:
# Review of PR #294 — consolidated findings

Verified against PR head `85c94127` via static review, runtime reproduction, and 130K+ Hypothesis-generated examples. Severity ratings re-checked against authoritative API docs after first pass — see "What I corrected" at the bottom.

## Findings, ordered by true severity

| # | Severity | Finding | Reachability |
|---|---|---|---|
| 1 | 🔴 **HIGH** for `ChatAnthropic` / 🟠 MEDIUM for OpenAI | `on_cleanup` doesn't clear checkpointer state — room re-entry duplicates SystemMessage in stored state | Persistent checkpointer + process restart, OR room kick+re-invite, OR platform reissuing `room_added` |
| 2 | 🟠 MEDIUM | Static-graph (`graph=`) and `graph_factory=` paths now silently get a SystemMessage prepended; customers pre-injecting their own get the same `ChatAnthropic` `ValueError` | First-call for affected customers; loud and immediate |
| 3 | 🟠 LOW–MEDIUM | `parse_tool_call` / `parse_tool_result` crash with `AttributeError` on non-dict JSON (5 adapters affected, pre-existing in `_tool_parsing.py` but PR newly routes LangGraph through it) | Low under normal SDK message shapes; only hits malformed/legacy persisted history or adversarial peers |
| 4 | 🟡 LOW | `_extract_tool_call_id` infinite-recurses on cyclic dict/list structures | Essentially zero in practice (JSON-derived data is acyclic) |
| 5 | 🟡 LOW | `TEMPLATES["default"]` is exported public API; `template=` parameter accepted; **neither is consulted** inside `render_system_prompt` | Anyone calling the documented param gets silent no-op |
| 6 | 🟡 LOW | `Emit.EXECUTION` gating untested in the off-state — gate-removal regression wouldn't be caught | Future regression risk only |
| 7 | 🟡 LOW | `enable_execution_reporting=True` deprecation shim untested (memory shim has a test) | Future regression risk only |

---

## 1. 🔴 HIGH — Bootstrap re-entry corrupts checkpointer state

`on_cleanup` pops `_bootstrapped_rooms[room_id]` but does not clear the checkpointer's state for that thread. The author left an explicit `# Future graph_factory-specific cleanup (e.g. checkpointer) goes here` comment acknowledging this gap.

When a `room_added` arrives for the same `room_id` after cleanup, a fresh `ExecutionContext` sets `is_session_bootstrap=True`, the bootstrap gate passes (`room_id not in _bootstrapped_rooms`), and the adapter re-prepends `("system", prompt) + history` into a checkpointer state that already contains them. `add_messages` appends rather than deduping by content (verified from [LangGraph source](https://github.com/langchain-ai/langgraph/blob/main/libs/langgraph/langgraph/graph/message.py): same content + different auto-generated IDs → both appended), so the persisted state ends up with two SystemMessages mid-conversation.

**Why this is HARD-failure for Anthropic users**, not soft cache degradation:

- Anthropic's Messages API has [no `system` role in the messages array at all](https://platform.claude.com/docs/en/api/messages) — system context goes in the top-level `system=` parameter only.
- LangChain's `ChatAnthropic._format_messages` enforces this by raising `ValueError("Received multiple non-consecutive system messages")` *before* the API is even called.
- That string is the **exact** error message [`tests/adapters/test_langgraph_adapter.py:572-575`](https://github.com/thenvoi/thenvoi-sdk-python/blob/85c94127/tests/adapters/test_langgraph_adapter.py#L572-L575) quotes verbatim. The existing `test_no_extra_system_messages_with_history_and_participants` test isn't paranoia — it's guarding against a documented LangChain crash mode (see [langchain-ai/langchain#20835](https://github.com/langchain-ai/langchain/issues/20835)).

For OpenAI users the same scenario degrades silently: prefix caching breaks at the duplicate-insertion point and the model sees a confused conversation.

### Runtime verification

Real `LangGraphAdapter` + real `StateGraph(MessagesState)` + shared `InMemorySaver`, simulating `room_added → bootstrap → cleanup → room_added → bootstrap`:

```
[L1] After first bootstrap: 4 total, 1 SystemMessage
  [0] SystemMessage: 'You are TestBot, Test bot...'
  [1] HumanMessage:  'prior turn from platform history'
  [2] AIMessage:     'prior assistant turn'
  [3] HumanMessage:  '[Alice]: Hello from lifecycle 1'

--- on_cleanup(room_id) ---
[Cleanup] Checkpointer state after cleanup: 4 total, 1 SystemMessage
[Cleanup] _bootstrapped_rooms: {}

[L2] After second bootstrap: 8 total, 2 SystemMessage
  [0..3] (untouched)
  [4] SystemMessage   ← duplicate
  [5] HumanMessage    ← duplicate of [1]
  [6] AIMessage       ← duplicate of [2]
  [7] HumanMessage:   '[Alice]: Hello from lifecycle 2'
```

### Why existing tests don't catch this

- `test_no_duplicate_system_prompt_on_re_bootstrap` covers re-bootstrap *while still tracked* (the gate works).
- `test_warns_on_large_bootstrapped_rooms` covers eviction-warning but not eviction + re-bootstrap.
- `TestSystemPromptCrossTurn` (added in `85c9412`) covers turn-1-bootstrap → turn-2 *non-bootstrap* in the same `ExecutionContext`.
- `test_cleanup_resets_bootstrap_tracking` only asserts the membership pop.
- No test exercises `on_message(bootstrap=True) → on_cleanup → on_message(bootstrap=True)` for the same `room_id` with a shared checkpointer instance.

@ed-lepedus-thenvoi's "🟢 `on_cleanup` correctly resets bootstrap tracking → re-bootstrap rehydrates history" bullet validates the gate-pop and that rehydration *fires*; this finding is about what happens to the checkpointer state when rehydration fires *onto surviving thread state*. Different scenario, not contradictory.

### Suggested fix

LangGraph 1.x's `BaseCheckpointSaver` exposes `adelete_thread` (verified present and properly overridden on `InMemorySaver`):

```python
async def on_cleanup(self, room_id: str) -> None:
    self._bootstrapped_rooms.pop(room_id, None)
    if self.checkpointer:
        try:
            await self.checkpointer.adelete_thread(room_id)
        except (AttributeError, NotImplementedError):
            pass
```

Consider an opt-out flag (`clear_checkpointer_on_cleanup: bool = True`) for users who want to retain state across re-entries deliberately.

<details>
<summary><code>verify_finding_1.py</code> — re-runnable</summary>

```python
import asyncio
from datetime import datetime, timezone
from unittest.mock import AsyncMock, MagicMock

from langchain_core.messages import AIMessage, HumanMessage, SystemMessage
from langgraph.checkpoint.memory import InMemorySaver
from langgraph.graph import END, START, MessagesState, StateGraph

from thenvoi.adapters.langgraph import LangGraphAdapter
from thenvoi.core.types import PlatformMessage


def build_graph(checkpointer):
    async def passthrough(state: MessagesState):
        return {"messages": []}
    builder = StateGraph(MessagesState)
    builder.add_node("passthrough", passthrough)
    builder.add_edge(START, "passthrough")
    builder.add_edge("passthrough", END)
    return builder.compile(checkpointer=checkpointer)


def make_message(content, msg_id, room_id):
    return PlatformMessage(
        id=msg_id, room_id=room_id, content=content,
        sender_id="user-1", sender_type="User", sender_name="Alice",
        message_type="text", metadata={},
        created_at=datetime.now(timezone.utc),
    )


async def inspect(checkpointer, room_id):
    snap = await checkpointer.aget_tuple({"configurable": {"thread_id": room_id}})
    return [] if snap is None else snap.checkpoint["channel_values"].get("messages", [])


async def main():
    room_id = "room-X"
    checkpointer = InMemorySaver()
    graph = build_graph(checkpointer)
    adapter = LangGraphAdapter(graph=graph)
    await adapter.on_started("TestBot", "Test bot for verification")
    tools = MagicMock()
    tools.send_message = AsyncMock()
    tools.send_event = AsyncMock()
    history = [
        HumanMessage(content="prior turn from platform history"),
        AIMessage(content="prior assistant turn"),
    ]
    await adapter.on_message(
        msg=make_message("Hello from lifecycle 1", "msg-1", room_id),
        tools=tools, history=history, participants_msg=None, contacts_msg=None,
        is_session_bootstrap=True, room_id=room_id,
    )
    s1 = await inspect(checkpointer, room_id)
    print(f"L1: {len(s1)} msgs, {sum(isinstance(m, SystemMessage) for m in s1)} system")
    await adapter.on_cleanup(room_id)
    await adapter.on_message(
        msg=make_message("Hello from lifecycle 2", "msg-2", room_id),
        tools=tools, history=history, participants_msg=None, contacts_msg=None,
        is_session_bootstrap=True, room_id=room_id,
    )
    s2 = await inspect(checkpointer, room_id)
    print(f"L2: {len(s2)} msgs, {sum(isinstance(m, SystemMessage) for m in s2)} system")
    assert sum(isinstance(m, SystemMessage) for m in s2) == 1, \
        f"Expected 1 SystemMessage; got {sum(isinstance(m, SystemMessage) for m in s2)}"


if __name__ == "__main__":
    asyncio.run(main())
```

`AssertionError: Expected 1 SystemMessage; got 2`

</details>

---

## 2. 🟠 MEDIUM — Static-graph and `graph_factory=` paths silently get a prepended SystemMessage

`test_static_graph_with_participants_msg` flipped from `len(system_msgs) == 0` to `== 1`. Pre-PR these patterns were a carve-out; the user's compiled graph fully owned its system prompt. Post-PR, the adapter prepends `("system", prompt)` regardless.

**Customer running a compiled graph that already injects its own SystemMessage now hits the same `ChatAnthropic.ValueError` as Finding 1**, but on the very first call.

### Runtime verification

```python
async def first_node(state):
    prepended = [SystemMessage(content="CUSTOMER OWN SYSTEM PROMPT")] + list(state["messages"])
    seen.append(prepended)
    return {"messages": []}

# ...build graph, hand to LangGraphAdapter(graph=graph), call on_message(is_session_bootstrap=True)
```

Result against `85c94127`:

```
What the customer's first node sees: 3 messages, 2 SystemMessage
  [0] SystemMessage: 'CUSTOMER OWN SYSTEM PROMPT'
  [1] SystemMessage: 'You are TestBot, Test bot.  ## Environment  ...'
  [2] HumanMessage:  '[Alice]: hi'
```

The new docstring (`adapters/langgraph.py:50-62`) says "your graph should also read state['messages']" — that documents the new contract but doesn't migrate existing users.

### Suggested mitigations

1. **Detect a leading SystemMessage in user state and skip the prepend** (non-destructive, transparent for both old and new patterns).
2. Or expose `prepend_system_prompt: bool = True` as an opt-out for advanced graphs.
3. At minimum: call this out as a breaking change in the release notes / CHANGELOG.

---

## 3. 🟠 LOW–MEDIUM — `parse_tool_call` / `parse_tool_result` crash on non-dict JSON (5 adapters)

Surfaced by 20K-example Hypothesis fuzz (P5/P6 in PBT pass): **15,341 counterexamples across 8 distinct shapes** crashed `convert()` with `AttributeError: 'list'/'int'/'NoneType'/'bool'/'str' object has no attribute 'get'`.

`src/thenvoi/converters/_tool_parsing.py:50-54` and `:90-94`:

```python
try:
    event = json.loads(content)
except json.JSONDecodeError:
    logger.warning(...)
    return None

tool_call_id = event.get("tool_call_id")  # AttributeError if event is list/int/None/bool/str
```

`json.loads('"hello"')`, `json.loads('[1,2,3]')`, `json.loads('null')`, etc. are all valid JSON returning non-dict types. The shared parsers don't guard.

### Affected adapters

`gemini.py`, `google_adk.py`, `anthropic.py`, `langchain.py`, `pydantic_ai.py` — all import `parse_tool_call` / `parse_tool_result`. Pre-existing bug in `_tool_parsing.py` (introduced in PR #94), but **PR #294 newly routes LangGraph through it**.

### Asymmetric robustness within this PR

The legacy parser path the PR keeps around (`_parse_legacy_tool_call`) **does** have the guard:

```python
if not isinstance(event, dict):
    logger.warning("Skipping non-object tool_call: %s", content[:100])
    return None
```

So the modern path the PR routes through is *less* robust than the legacy fallback. Net regression.

### Reachability

Low under normal SDK message shapes — Thenvoi-emitted tool events are dict-shaped. Hits in practice would require: history written by older / non-conforming SDK versions, or adversarial peer messages if the platform doesn't validate content shape server-side. `convert()` is called unguarded at `src/thenvoi/core/simple_adapter.py:101`, so any AttributeError propagates up through `on_event`.

### Suggested fix (2 lines per parser)

```python
def parse_tool_call(content: str) -> ParsedToolCall | None:
    try:
        event = json.loads(content)
    except json.JSONDecodeError:
        return None
    if not isinstance(event, dict):
        logger.warning("Skipping non-object tool_call: %s", repr(content[:100]))
        return None
    # ... rest unchanged
```

Mirror the guard the legacy parser already has. Add a one-line PBT regression:

```python
@given(st.one_of(st.lists(st.integers()), st.integers(), st.booleans(), st.none(), st.text()))
def test_parse_tool_call_handles_non_dict_json(payload):
    assert parse_tool_call(json.dumps(payload)) is None
```

---

## 4. 🟡 LOW — `_extract_tool_call_id` infinite-recurses on cyclic structures

PBT generated cycles aren't possible (Hypothesis only emits acyclic data), but four hand-crafted cycle shapes all crash with `RecursionError`:

```python
a = {"messages": []}; a["messages"].append(a)
b = {}; b["output"] = b
lst = []; lst.append(lst)
c = {"messages": [{"output": None}]}; c["messages"][0]["output"] = c
```

All depths up to 300 of acyclic nesting succeed. Reachability is essentially zero (JSON-derived data is acyclic), but the SDK has no defense.

### Suggested fix

Bound recursion with an `id()` set:

```python
@classmethod
def _extract_tool_call_id(cls, output, _seen=None):
    if _seen is None:
        _seen = set()
    if id(output) in _seen:
        return None
    _seen.add(id(output))
    # ...recursive calls pass _seen
```

---

## 5. 🟡 LOW — `TEMPLATES["default"]` is dead public API

`src/thenvoi/runtime/prompts.py:81-86` exports a `TEMPLATES` dict that **is not consulted** inside `render_system_prompt` (which builds parts directly). The `template: str = "default"` parameter is accepted and documented but never used — passing `template="totally_made_up"` returns identical output. 200/200 PBT cases confirm.

The DEPRECATED comment acknowledges divergence; the export stays in `runtime/__init__.py:56,92`.

### Suggested fix

- Remove `TEMPLATES` from the export and drop `template` from `render_system_prompt`'s signature, or
- Honor `template` by actually consulting `TEMPLATES`.

The current state is doc-vs-code drift.

---

## 6. 🟡 LOW — `Emit.EXECUTION` gating untested in the off-state

Existing tests `test_handle_tool_start_event` / `test_handle_tool_end_event` were updated to add `features=AdapterFeatures(emit=frozenset({Emit.EXECUTION}))`. No companion test runs the same handler with default features and asserts `mock_tools.send_event.assert_not_awaited()`. Removing the gate entirely (a real refactor risk) would not be caught.

### Suggested fix

```python
@pytest.mark.asyncio
async def test_does_not_emit_when_execution_feature_off(self, mock_tools, mock_llm, mock_checkpointer):
    adapter = LangGraphAdapter(llm=mock_llm, checkpointer=mock_checkpointer)
    event = {"event": "on_tool_start", "name": "x", "run_id": "r", "data": {"input": {}}}
    await adapter._handle_stream_event(event, "room-1", mock_tools)
    mock_tools.send_event.assert_not_awaited()
```

---

## 7. 🟡 LOW — `enable_execution_reporting=True` shim untested

`test_enable_memory_tools_shim_enables_memory_capability` covers the memory side; no equivalent for `enable_execution_reporting`.

```python
@pytest.mark.asyncio
async def test_enable_execution_reporting_shim_enables_execution_emit(self, ...):
    with pytest.warns(DeprecationWarning):
        adapter = LangGraphAdapter(llm=mock_llm, checkpointer=mock_checkpointer,
                                   enable_execution_reporting=True)
    assert Emit.EXECUTION in adapter.features.emit
```

---

## Property-based test pass — what held

130K+ Hypothesis examples across 10 properties against `85c94127`. Properties that **held cleanly** (high-confidence clean):

- **P1** `_extract_tool_call_id` never crashes on bounded acyclic JSON-like blobs (max_leaves=80) — 0/20K
- **P3** `_extract_tool_call_id` finds embedded `tool_call_id` at any depth 1-300 — 0/3K
- **P4** `_pop_matching_tool_call` invariants — always returns a `pending` entry (popped, length -1) or `None`; matches by `tool_call_id` first, then most-recent name (LIFO) — 0/20K
- **P5** `convert()` produces valid `AIMessage(tool_calls) → ToolMessage` pairings on well-formed events — 0/20K (failures all in malformed-input bucket above)
- **P7** `render_system_prompt` invariants: identity always present, `custom_section` always leads, no `## Developer Instructions` regression header, capability sections appear iff capability enabled — 0/20K
- **P9** `agent_tools_to_langchain` capability gating: `CHAT_TOOL_NAMES` always present, contacts iff `include_contacts OR is_hub_room`, memory iff `include_memory_tools` — 0/200
- **P10** Bootstrap state machine membership: across arbitrary sequences of bootstrap/non_bootstrap/cleanup, `_bootstrapped_rooms.keys()` always equals `{rooms whose last action was a successful bootstrap}` (eviction not exercised — bounded action count keeps us under threshold 1000) — 0/3K

Repro for crash buckets: `pbt_findings_v2.py` runs in ~5 minutes against PR head.

---

## What I corrected after homework

After my first pass I framed the cleanup re-entry issue as "loud, recoverable" cache degradation. That was wrong:

- Anthropic's Messages API has **no `system` role at all** in the messages array — system context goes in the top-level `system=` parameter only ([docs](https://platform.claude.com/docs/en/api/messages)).
- LangChain's `ChatAnthropic._format_messages` raises `ValueError("Received multiple non-consecutive system messages")` *before* the API is even called ([source](https://github.com/langchain-ai/langchain/blob/master/libs/partners/anthropic/langchain_anthropic/chat_models.py)).
- That string is the **exact** error quoted in `tests/adapters/test_langgraph_adapter.py`. The author wrote that test against a known LangChain crash mode (see [langchain-ai/langchain#20835](https://github.com/langchain-ai/langchain/issues/20835)).

So Finding 1 is a **hard `ValueError` for `ChatAnthropic` users**, not soft cache degradation. The trigger conditions (persistent checkpointer + process restart, OR room kick+re-invite) aren't exotic — every production deployment with a SQLite/Redis checkpointer hits this on any process restart.

Sources:
- [Anthropic Messages API — no system role in messages array](https://platform.claude.com/docs/en/api/messages)
- [LangGraph `add_messages` — same content + different IDs both appended](https://github.com/langchain-ai/langgraph/blob/main/libs/langgraph/langgraph/graph/message.py)
- [LangChain issue #20835 — ChatAnthropic can't handle multiple system messages](https://github.com/langchain-ai/langchain/issues/20835)
- [`ChatAnthropic._format_messages` source — raises `ValueError`](https://github.com/langchain-ai/langchain/blob/master/libs/partners/anthropic/langchain_anthropic/chat_models.py)
- [OpenAI prompt caching — exact prefix matching](https://developers.openai.com/api/docs/guides/prompt-caching)

---

## Recommendation

- **Fix 1** before merging if any users target `ChatAnthropic` (which the SDK's `langchain-anthropic` lockfile pin says is supported). The fix is small (2-line `on_cleanup`).
- **Fix 2** at the same time, or document as a CHANGELOG breaking change with migration guidance.
- **Fix 3**'s 2-line `isinstance(event, dict)` guard while you're in the parsers — it removes a class of crash for ~zero cost and aligns with the legacy parser's existing defense.
- **4–7** can land in follow-ups.

ed-lepedus-thenvoi caught the most user-impacting issue (the cached-graph multi-room dispatch); these are second-pass findings on top of that work.

darvell-thenvoi:
@ed-lepedus-thenvoi re-requested review. The cached graph example issue was fixed after your review in 9f46c10, and 85c9412 added the follow-up hardening/coverage for the remaining LangGraph findings.

darvell-thenvoi:
@ed-lepedus-thenvoi pushed f0f82ec4e899d4d371cffcb8b3756cdf60a4e2c6 and re-requested review. The advanced research_ops example already rebuilds the graph per call so room-bound Band tool wrappers are not cached across rooms; I added an adapter-level multi-room regression test that drives two rooms through graph_factory and asserts each call receives that room’s converted platform tools and thread_id.\n\nValidation on f0f82ec4e899d4d371cffcb8b3756cdf60a4e2c6:\n- local: uv run --extra dev --extra langgraph pytest tests/adapters/test_langgraph_adapter.py -> 35 passed\n- local: ruff check/format --check on LangGraph adapter, test, and example passed\n- GitHub Actions: lint, packaging, test (3.11), test (3.12), test-parlant (3.11), test-parlant (3.12), Validate PR Title all passed.

darvell-thenvoi:
Updated for Eric's second-pass review in `5288b02`.

What changed:
- Fixed the checkpointer re-entry issue: `on_cleanup()` now clears the room's LangGraph checkpointer thread, so `room_added -> cleanup -> room_added` does not duplicate hydrated history/SystemMessages.
- Changed advanced `graph=` / `graph_factory=` prompt behavior: Band system-prompt injection is no longer implicit for advanced graphs, because those callers may already manage their own system messages. Advanced graphs can opt in with `inject_system_prompt=True`; the research ops example now does that explicitly.
- Hardened malformed tool-event parsing: non-object JSON payloads now return `None` instead of raising, and malformed `args` falls back to `{}`.
- Guarded malformed LangGraph stream events before calling `.get()`.
- Made `render_system_prompt(template=...)` actually use the public `TEMPLATES` mapping and raise on unknown templates.
- Added the low-risk coverage from findings 6/7: execution emit off-state and `enable_execution_reporting=True` shim.

Validation:
- `uv run --extra dev --extra langgraph pytest tests/adapters/test_langgraph_adapter.py tests/converters/test_tool_parsing.py tests/test_prompts.py -q --no-cov` -> `79 passed`
- `uv run ruff check src/thenvoi/adapters/langgraph.py src/thenvoi/converters/_tool_parsing.py src/thenvoi/runtime/prompts.py tests/adapters/test_langgraph_adapter.py tests/converters/test_tool_parsing.py tests/test_prompts.py examples/langgraph/09_research_ops_orchestrator.py` -> passed
- `uv run ruff format --check src/thenvoi/adapters/langgraph.py src/thenvoi/converters/_tool_parsing.py src/thenvoi/runtime/prompts.py tests/adapters/test_langgraph_adapter.py tests/converters/test_tool_parsing.py tests/test_prompts.py examples/langgraph/09_research_ops_orchestrator.py` -> passed
- `uv run pyrefly check` -> `0 errors`
- Commit hooks passed: hardcoded secrets, ruff, format, pyrefly, commitizen.

nir-singher-band:
Resolves [INT-508](https://linear.app/thenvoi/issue/INT-508/langgraph-adapter-emits-tool-events-unconditionally-ignoring-user-emit) as well

nir-singher-band (changes requested):
## QA Validation — LangGraph Adapter Update

Ran a full end-to-end QA pass against the updated LangGraph adapter on `app.band.ai` using `gpt-4o`.

### Testing Method

The validation covers two layers:

**Core examples (01–03)** — Each LangGraph example is spawned as a real agent process, connected to the platform over WebSocket, and exercised through three scenario types:
- A multi-turn conversation (greeting, domain Q&A, context follow-up, platform tool use, farewell)
- A rehydration cycle (SIGINT → restart → verify the agent re-joins the same room and recalls prior context)
- A context isolation check (new room created to confirm no cross-room leakage)

**Capability scenarios** — Purpose-built agents with specific SDK features enabled (memory, contacts, execution emit) are tested against targeted workflows:
- Memory lifecycle: store → list → get → cross-room recall → supersede → archive
- Contact strategies: DISABLED (ignore), CALLBACK (auto-approve via programmatic filter), HUB_ROOM (LLM decides per request)
- Execution emit: verify `tool_call` / `tool_result` events are visible on the platform
- Concurrent rooms: 3 rooms with distinct secrets, verify no cross-room leakage

Each agent is registered with its own identity on the platform — no shared agent IDs across tests.

---

### Results

#### Core Examples

| Example | Conversation | Rehydration | Isolation |
|---------|:---:|:---:|:---:|
| 01_simple_agent | PARTIAL | FAIL | PASS |
| 02_custom_tools | PARTIAL | FAIL | PASS |
| 03_custom_personality | PASS | FAIL | PASS |

**Conversation** works well — agents respond coherently, use platform tools (`get_participants`, `lookup_peers`), and maintain context within a session. The pirate personality agent (03) passed all 6 steps cleanly. Intermittent timeouts on steps 2–3 for examples 01 and 02 caused PARTIAL; these appear to be LLM latency rather than adapter issues.

**Rehydration** fails across all examples. The agent reconnects to rooms after restart, but replays old messages and loses context:

1. **Message replay on restart:** After a kill/restart cycle, the agent re-answers every message it already responded to in the previous session. In the 01_simple_agent test, 6 unsolicited replay messages appeared within 15 seconds of restart — one for each prior message in the room. Confirmed manually across multiple chat rooms ([chat 8e386602](https://app.band.ai/chat/8e386602-acfe-4b68-89b8-b425b05ec590)).

2. **Pending messages ignored:** A message sent while the agent is down ("What word did I ask you to remember?") never gets a response. The agent processes the old replayed messages instead and never reaches the pending one.

3. **Context lost after restart:** Asking the agent to summarize the pre-restart conversation produces no response. The agent has no memory of anything discussed before the restart.

**To reproduce:** Start any LangGraph example, send a few messages, kill the process (SIGINT), send a new message while the agent is down, restart the process. Observe: (a) the agent re-answers all old messages, (b) the message sent while it was down gets no reply, (c) asking about pre-restart context gets no response.

**Context isolation** passes perfectly — new rooms start with a clean slate, no cross-room bleed.

#### Capability Scenarios

| Scenario | Status | Detail |
|----------|:---:|--------|
| Memory tools | N/A | Memory is disable for my user ATM |
| Contact DISABLED | PASS | Request stays pending, zero room activity — correct behavior. |
| Contact CALLBACK | PASS | Programmatic callback fires, auto-approves matching handles, broadcast notification delivered to room. No LLM invocation (as expected for callback strategy). |
| Contact HUB_ROOM | PARTIAL | Friendly request approved, empty request handled gracefully. Spam request stayed pending — LLM didn't reject within the timeout window. Timing-sensitive. |
| Execution emit | PASS | `tool_call` and `tool_result` events visible on platform for both `get_participants` and `lookup_peers`. Event counts correct (4 calls, 4 results). |
| Concurrent rooms | PARTIAL | Room creation and agent subscription work correctly across 3 simultaneous rooms. Secret recall failed in rooms 1–2 due to the same platform subscription limits affecting memory. Room 3 passed. |

---

### Key Findings

**What works well:**
- Agent startup, WebSocket subscription, and room management are solid across all examples
- Platform tool use (`get_participants`, `lookup_peers`) works reliably
- Execution emit delivers tool events to the platform correctly
- Contact strategies (all three modes) function as designed
- Context isolation between rooms is clean
- Custom personality/system prompts are respected throughout conversations

**Issues to address:**
- Rehydration is broken — on restart, the agent replays responses to all prior messages, ignores messages sent while it was down, and loses all conversation context. The agent needs to distinguish already-handled messages from genuinely new ones when reconnecting to a room.

darvell-thenvoi:
@nir-singher-thenvoi  

TL;DR: 

Fixed up, also fixed another edge-case discovered if the agent dies mid-tool call that would mean the agent is just broken forever. Repeated your method of testing, tested against prod, worked as expected. `features=` was also updated to be the primary surface instead of `include_tools` and `include_memories` as an extra change.

Details:

I treated the restart failure as two bugs showing up in the same flow: we were reopening old work on reconnect, and a fresh LangGraph process was not getting enough prior context back from Band to behave like the same agent.

The replay part came down to delivery status. On restart, an old `message_created` can show back up for a message the agent has already processed. Calling `mark_processing()` on that is not harmless; it creates a new processing attempt and effectively makes completed work active again. The runtime now checks delivery status before claiming WebSocket-delivered work. If Band already says this agent processed that message, we skip it before `mark_processing()` and before the adapter runs.

The other side is the message sent while the agent is down. I intentionally kept `/messages/next` as the source of truth there. A pending message can also appear in hydrated context, so treating "present in context" as "already handled" would recreate the missed-message bug. Context delivery status is only used to suppress processed replay events; `/next` is still what tells us there is real work to answer.

The context part was the LangChain transcript we rebuild on restart. A fresh `InMemorySaver` has no process-local LangGraph memory, so it has to get a usable conversation back from Band. The converter now preserves the agent's own visible replies as `AIMessage` when they are normal chat turns, while still preserving LangChain's required `AIMessage(tool_calls) -> ToolMessage` ordering around active tool-call segments.

I ran a live restart smoke for the exact shape you called out and ran it against production with a temporary agent, temporary room, fresh `InMemorySaver`, and a random uuid, it did the following:

- start the agent and ask it to remember `REHYDRATE_<uuid>`
- stop the agent
- send "what nonce did I ask you to remember?" while it is down
- restart with a fresh adapter/checkpointer
- watch for replayed old answers, the down-time answer, and a second restart with no new user message

Result from the live run:

```text
RESULT langgraph_restart_rehydration=PASS down_message_id=c40074a9-0ec8-40dd-b5e2-4d28572fd1ce nonce_prefix=REHYDRATE_6f0a8a24
ASSERTIONS no_replay_burst=True pending_down_message_answered_once=True recalled_pre_restart_context=True second_restart_no_new_reply=True
```

I also covered the deterministic pieces locally: processed WebSocket replays do not get claimed again, stale WebSocket payloads are checked against hydrated delivery status, `/next` messages still execute even when they are present in context, same-id WebSocket/`/next` delivery only runs once locally, LangGraph receives hydrated prior user + prior own-agent turns on bootstrap, `features=AdapterFeatures(...)` now drives the LangGraph tool surface, and the tool wrappers still go through the canonical Band tool boundary.

nir-singher-band:
## QA Validation — LangGraph PR #294 (second pass)

**Verdict: This PR passes QA.** The LangGraph adapter upgrade, refactored components, and converter changes are validated across the full scenario suite. All 9/9 core scenarios PASS.

**Date:** 2026-05-29 | **Platform:** app.thenvoi.com | **Model:** gpt-4o
**Harness:** `feat/qa-test-harness` branch (merged into working branch for testing) — automated core scenarios + expanded scenario scripts

---

### Core Scenarios (3 examples x 3 scenarios)

| Example | A: Basic conversation | B: Rehydration | C: Context isolation | Overall |
|---|:---:|:---:|:---:|:---:|
| 01_simple_agent | PASS (6/6) | PASS (7/7) | PASS (4/4) | **PASS** |
| 02_custom_tools | PASS (6/6) | PASS (7/7) | PASS (4/4) | **PASS** |
| 03_custom_personality | PASS (6/6) | PASS (7/7) | PASS (4/4) | **PASS** |

All three examples start cleanly, connect, and handle multi-turn conversation. Rehydration works end-to-end: agent restarts, processes the message sent while down, recalls prior conversation correctly, and produces no unsolicited replay messages. Context isolation is clean — no cross-room leakage.

---

### Expanded Scenarios

| Scenario | Status | Pass rate | Notes |
|---|:---:|:---:|---|
| E: Memory tools | PARTIAL | 7/8 | Store / list / get / supersede / archive all pass. Cross-room recall (E4) fails. |
| F1: Contact DISABLED | **PASS** | 3/3 | Request stays pending, zero contact messages in room. |
| F2: Contact CALLBACK | **PASS** | 3/4 | Whitelist auto-approve fires correctly, no LLM invocation. Broadcast step partial. |
| F3: Contact HUB_ROOM | PARTIAL | 5/7 | Hub room created, friendly request approved. Spam + empty pending within window. |
| G: Execution emit | PARTIAL | 2/3 | `lookup_peers` and count checks pass. First emit step missed in poll window. |
| I: Concurrent rooms | **PASS** | 3/3 | Three rooms with distinct secrets — no cross-room leakage. |

---

### Divergences from Expected Behavior

**E4 — Cross-room memory recall (FAIL)**
Expected: Agent in a new room proactively recalls memories stored in a different room by calling `thenvoi_list_memories`.
Actual: Agent claims no memories exist without calling the memory tool. The memories are present on the platform — the agent simply doesn't invoke the tool unprompted.
Assessment: Prompt/LLM behavior. The agent needs explicit instruction or a user prompt referencing memories to trigger the tool call. Not an SDK or converter bug.
Note INT-515 and INT-516 are tracking similar issues.

**F3 steps 5 + 7 — Spam and empty contact requests stayed pending (FAIL)**
Expected: The hub room LLM processes the spam request ("FREE BITCOIN!...") and the empty-message request within 30 seconds, producing reject or unknown decisions.
Actual: Both remained in `pending` status at the 30-second poll boundary. The friendly collaboration request (step 3) was approved correctly, confirming the hub room pipeline works end-to-end.
Assessment: Hub room LLM processing latency on subsequent requests after the first. The SDK correctly delivered all three requests to the hub room. This is LLM throughput under sequential request load, not an SDK issue.

**G1 — `thenvoi_get_participants` emit check (PARTIAL)**
Expected: `tool_call` and `tool_result` events for `thenvoi_get_participants` appear in the platform events feed.
Actual: Not detected in the first emit-check step. However, the aggregate count check (step 3) found 3 tool_call + 3 tool_result events, confirming the adapter is emitting correctly across the full conversation.
Assessment: Poll-window race condition in the test harness. The events are being emitted — the total count confirms it — but the first targeted check polls before the specific events arrive. Not an SDK regression.

---

### Chat Rooms (for platform inspection)

| Scenario | Room ID |
|---|---|
| F1: Contact DISABLED | `a48d4027-daee-41d9-93c7-744c1e5383f9` |
| F2: Contact CALLBACK | `fdf622b1-ffa5-446c-a26b-8ee0b4227bcf` |
| F3: Contact HUB_ROOM | `57548410-959c-41f5-aee6-1bb16394ef56` |

---

All core scenarios pass cleanly. The expanded scenario divergences are LLM behavior (E4, F3), broadcast delivery timing (F2), and test harness poll windows (G1) — none trace back to adapter or converter bugs in this PR.

✌️

nir-singher-band:
@darvell-thenvoi @amit-gazal-thenvoi I'm approving this PR as functionally it passes, this is without the actual code review. There are a lot of changes and @amit-gazal-thenvoi should sanity validate them at the least.

nir-singher-band (approved):

PR #296, opened 2026-05-10
chore(ci): add dev-parlant CI job for parlant adapter coverage

amit-gazal-band:
## Summary
- Add `test-parlant` matrix job (3.11, 3.12) installing `--extra dev-parlant`
- Run parlant-targeted tests: adapter, converter, integration, conformance `-k parlant`
- Run pyrefly scoped to parlant sources (full-tree check fails in dev-parlant venv due to missing langgraph/anthropic/etc.)

Closes [INT-344](https://linear.app/thenvoi/issue/INT-344/choreci-add-dev-parlant-ci-job-for-parlant-adapter-test-coverage).

## Test plan
- [x] Local `uv sync --extra dev-parlant` + targeted pytest: 90 passed, 4 skipped
- [x] Local scoped pyrefly: 0 errors
- [ ] CI `test-parlant (3.11)` green
- [ ] CI `test-parlant (3.12)` green
- [ ] Existing `lint`, `test`, `packaging` jobs still green

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -94,6 +94,10 @@ jobs:
       run: uv sync --extra dev-parlant
 
     - name: Run parlant tests
+      env:
+        # dev-parlant venv lacks langgraph/anthropic/crewai/etc.; tolerate
+        # missing framework adapter/converter configs instead of failing fast.
+        THENVOI_ALLOW_MISSING_FRAMEWORKS: "1"
       run: |
         uv run pytest \
           tests/adapters/test_parlant_adapter.py \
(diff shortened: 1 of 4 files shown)


darvell-thenvoi (approved):
Approving. Splitting Parlant into its own CI job is the right shape here because the normal dev extra cannot include it without dependency conflicts, and this still exercises the adapter, converter, integration, and conformance paths instead of just adding a dependency install check.

The `THENVOI_ALLOW_MISSING_FRAMEWORKS` escape hatch is scoped to this job, while the regular test matrix keeps the strict framework-config behavior. Optional later: add a tiny guard that asserts Parlant is actually present in the selected adapter/converter configs so this job cannot silently drift.

PR #299, opened 2026-05-13
fix(examples): add darter config key for codex/opencode examples [INT-463]

nir-singher-band:
## Summary

- Add `darter` config entry to `agent_config.yaml.example` as the default `AGENT_KEY` for codex and opencode examples, so new users don't hit a missing-key error when following the template
- Add a Codex CLI login check to `examples/codex/01_basic_agent.py` that detects when `codex` is not installed or not logged in, and prompts the user to authenticate instead of crashing with an opaque error

## Test plan

- [ ] Verify `agent_config.yaml.example` contains the new `darter` section under "Codex / OpenCode Examples"
- [ ] Run `uv run examples/codex/01_basic_agent.py` without Codex installed and confirm a clear error message
- [ ] Run with Codex installed but not logged in and confirm the login prompt appears
- [ ] Run with Codex installed and logged in and confirm the agent starts normally

Closes INT-463


nir-singher-band:
Closes INT-477 as well

amit-gazal-band (approved):

PR #304, opened 2026-05-14
fix(integrations): validate Parlant integration [INT-446]

darvell-thenvoi:
## Summary
- Fix the generic Parlant runner and support example to use the current `ParlantAdapter(server=..., parlant_agent=...)` API with OpenAI-backed Parlant and Band platform tools.
- Harden Parlant E2E coverage by normalizing Band credential precedence, using isolated Parlant ports, bounding teardown, removing hidden flaky reruns, and targeting the real user participant for tool-response prompts.
- Update the Parlant README surfaces and add `parlant_validation_report.md` with sanitized live API/session proof for every standalone Parlant example.

## Live proof
- Production Band URLs were used with `OPENAI_BASE_URL` and `OPENAI_API_BASE` unset.
- Provider proof was sanitized: `OPENAI_API_KEY_present=True`, `OPENAI_API_KEY_shape=sk`; exact local credential-store paths and secrets are not included.
- Live Parlant E2E smoke passed: `tests/e2e/adapters/test_parlant.py::TestParlantE2E::test_smoke_responds_to_message` -> `1 passed in 43.43s`.
- Live Parlant tool execution passed: `tests/e2e/adapters/test_parlant.py::TestParlantE2E::test_tool_execution_send_message` -> `1 passed in 47.13s`.
- Standalone examples were run as their actual scripts and each proof records agent identity, live room ID, trigger message ID, response message ID, response sender ID, and token assertion:
  - `examples/parlant/01_basic_agent.py` -> `response_contains_token=True`
  - `examples/parlant/02_with_guidelines.py` -> `response_contains_token=True`
  - `examples/parlant/03_support_agent.py` -> `response_contains_token=True`
  - `examples/parlant/04_tom_agent.py` -> distinct Tom agent identity, `response_contains_token=True`
  - `examples/parlant/05_jerry_agent.py` -> distinct Jerry agent identity, `response_contains_token=True`

## Test plan
- [x] `uv run ruff check examples/run_agent.py examples/parlant/03_support_agent.py tests/e2e/adapters/test_parlant.py`
- [x] `uv run ruff format --check examples/run_agent.py examples/parlant/03_support_agent.py tests/e2e/adapters/test_parlant.py`
- [x] `uv run pytest tests/adapters/test_parlant_adapter.py tests/converters/test_parlant.py tests/integrations/parlant/test_tools.py tests/framework_conformance/test_adapter_conformance.py tests/framework_conformance/test_converter_conformance.py -k parlant -v --no-cov` -> `89 passed, 4 skipped, 232 deselected`
- [x] Live E2E smoke against production Band with real OpenAI provider
- [x] Live E2E tool execution against production Band with real OpenAI provider
- [x] Live standalone smoke for all five Parlant example scripts with persisted Band room/message IDs in the validation report
- [x] `git diff --check`
- [x] Pre-commit hooks on both commits: hardcoded secret detector, Ruff, format, pyrefly, commitizen

## Notes
- `dev-parlant` and normal `dev` are separate dependency forks because Parlant and CrewAI currently conflict through OpenTelemetry SDK constraints. Parlant runtime/live checks were run in `dev-parlant`; pre-commit pyrefly was run in the normal `dev` fork.


darvell-thenvoi:
Merged current `origin/dev` into this branch in `921d09a` to clear the PR's dirty merge state.

Conflict resolution:
- `examples/parlant/03_support_agent.py` kept the `parlant_tools = create_parlant_tools()` initialization because `setup_support_agent(server, parlant_tools)` passes those tools into the escalation guideline. Dropping it would break the support example.

Validation after the merge:
- `uv run ruff check examples/parlant/03_support_agent.py examples/run_agent.py tests/e2e/adapters/test_parlant.py` -> passed
- `uv run ruff format --check examples/parlant/03_support_agent.py examples/run_agent.py tests/e2e/adapters/test_parlant.py` -> passed
- `uv run --extra dev-parlant pytest tests/adapters/test_parlant_adapter.py tests/converters/test_parlant.py tests/integrations/parlant/test_tools.py -q --no-cov` -> `58 passed`
- Merge commit hooks passed: hardcoded secrets, ruff, format, pyrefly, commitizen.

Local untracked validation artifacts (`parlant_validation_report.md`, `reason/`) were left uncommitted.

amit-gazal-band (changes requested):
In general, the Parlant integration currently doesn't include the system prompt (not as prompt and not as guidelines). We need to tackle this before we can say it is validated.

darvell-thenvoi:
@amit-gazal-thenvoi Resolved, and then some.

You were right about the gap. So, I fixed it and tightened the rest of the adapter surface while I was in there, because the same issue showed up in a few adjacent places:

- Parlant platform tools now come from the canonical Band tool registry instead of a second hand-written Parlant implementation. That means the Parlant tool surface tracks the same names, params, required fields, descriptions, contact tools, and memory tools as the rest of the SDK.
- `additional_tools` now works for Parlant using the same `CustomToolDef` tuple format as the other adapters. Custom tool args go through the normal validation path, and validation errors are returned to Parlant as model-visible tool errors.
- Contact and memory tools are now capability-gated through `AdapterFeatures`. Empty explicit features hide contacts/memory, `CONTACTS` exposes contact tools, `MEMORY` exposes memory tools, and include/exclude/category filters apply to the generated Parlant tool entries.
- Parlant response failures no longer look like successful message processing. If session creation, event lookup, timeout/no-response, empty final content, or platform delivery fails before a message is actually delivered, `on_message()` raises so the SDK can treat the source message as failed instead of processed.
- `thenvoi_send_message` only marks the message delivered after the canonical platform tool call succeeds. A failed send inside the Parlant tool wrapper no longer tricks the adapter into thinking the turn was handled.
- `Emit.EXECUTION` now reports tool execution from the tool wrappers themselves, in real time. `tool_call` is emitted before the tool runs, `tool_result` after, sharing a stable `tool_call_id` for correlation. This started as a post-hoc drain of Parlant's session log, which produced two real bugs in prod: tool events landed *after* the reply in the room timeline, and the same call could be re-reported across `wait_for_more_events` iterations. Real-time emission fixes both — ordering is now correct relative to the actual side effects, and each invocation reports exactly once.
- Worth calling out as a tradeoff: only adapter-registered tools (built-ins + `additional_tools`) are reported. Raw `@p.tool` tools attached directly to a guideline outside `additional_tools` bypass the wrapper and are not reported. In practice everything routes through `create_parlant_tools`, so this is an edge case that deserves a doc note rather than a regression — the supported shape for custom logic is now `additional_tools`.
- I did not add reasoning/thought emit support, because Parlant 3.3.1 exposes message/tool/status/custom session events but not a real thought/reasoning stream. Claiming it would be fake support and I tried to think of a clever way around it but everything turned out too hacky.
- The generic runner now handles the Parlant/CrewAI dependency conflict honestly and wires contacts correctly when enabled: both runtime `contact_config` and model-visible `Capability.CONTACTS` are set.
- CI gating in both dependency forks is now correct: Parlant-specific adapter/integration tests skip cleanly in the plain `test` job (no Parlant installed) and run for real in `test-parlant`; the stale tool-name drift test was rewritten to verify dynamic registry derivation plus behavioral coverage instead of grepping literal tool-name strings; and the new generic-runner test resolves `examples/run_agent.py` via a path relative to the test file instead of a hardcoded absolute path.

The behavior proof is now aimed at the adapter contract rather than example shape:

- Prompt behavior: starting the adapter installs the rendered prompt and custom section into Parlant as a `MATCH_ALWAYS` guideline whose `description` carries the full prompt. Parlant's `prompt_builder` injects matched-guideline descriptions into the model prompt, so the SDK prompt actually reaches the model — verified at the Parlant source layer, not just at the renderer. The guideline is installed exactly once per adapter lifecycle.
- Tool surface behavior: generated Parlant tools match `iter_tool_definitions(...)` for names, parameter names, required fields, list typing for `mentions`, contact/memory capability gates, include/exclude filters, and generated `additional_tools`.
- Tool execution behavior: generated Parlant wrappers execute through the session-bound `AgentToolsProtocol` and call `execute_tool_call(...)` with canonical Band tool names and canonical argument shapes.
- Delivery behavior: successful `thenvoi_send_message` marks delivery, failed `thenvoi_send_message` does not, missing session tools return a model-visible error, invalid JSON/dict args do not call the platform tool, and dispatcher exceptions are surfaced back to Parlant as tool errors.
- Execution reporting behavior: the wrapper emits `tool_call` immediately before the tool runs and `tool_result` immediately after, with one stable `tool_call_id` linking the pair. Platform tools (`thenvoi_send_message`, `thenvoi_send_event`) are skipped because they already create Band-visible effects. With emit off, the tool still runs but no execution events are sent. An ordering oracle in the unit tests asserts the sequence is `["tool_call", "execute", "tool_result"]` for every invocation, which is the regression test for the prod bug above.

For validation, the following was tested in prod:

- Live platform behavior: the E2E suite created fresh Band rooms, started an in-process Parlant 3.3.1 server, connected the SDK agent to Band, sent real trigger messages through Band, and then verified the durable Band room history.
  - smoke: the agent was mentioned in Band and produced a user-visible text response in the room.
  - platform send tool: the agent was asked to reply with a generated `PINEAPPLE-...` token, and Band room history contained a text message with that exact token.
  - real-time tool execution reporting: the agent was asked to call a Band-registered custom tool `nativeecho`, wired via `additional_tools=[(NativeEchoInput, native_echo_handler)]`, with a generated `NATIVE-...` token. Band room history contained all four expected artifacts in correct chronological order: the trigger message, then `tool_call` for `nativeecho` with `args.code == token`, then `tool_result` for `nativeecho` with `output.echo == token` and the same `tool_call_id`, then the final text reply containing the token. The earlier draft of this test registered the tool as a raw `@p.tool` attached to a guideline; that path is intentionally not covered by the new wrapper-based reporting, so the test now exercises the supported `additional_tools` shape instead.

So the PR now covers the thing you called out and the adjacent adapter-contract behavior that was previously too loose. It is ready for another review.

nir-singher-band (changes requested):
The different paralant examples should no logner set the system prompt as a description, nor should they contain additional guidelines on platform tools use. Those should all be inherited from the adapter setup (the always match rule)

darvell-thenvoi:
@nir-singher-thenvoi Updated the examples, something I want to note though, the second example (with guidelines) we have for parlant though is pretty bad though in concept, since the guidelines we have in it boils down to essentially "just act good as a band agent". It would be nice if we could scope out some work for example rewriting/enhancement sometime soon that focus heavily on what external developers intend to be doing with the SDK.

nir-singher-band (approved):
**TL;DR:** Independently ran the multi-adapter QA harness against this PR's Parlant code on **production** (`app.thenvoi.com`). The Parlant integration is in good shape — every core flow passes, plus multi-agent coordination and execution-event emission. The handful of PARTIALs trace to a **platform memory-API 404** and **Parlant NLP latency**, not to the adapter/tools rewrite. The example-hygiene cleanup I'd have flagged is **already in this PR** — confirmed via an independent pass (details below).

## Scorecard

| Scenario | Result | Primary room |
|---|---|---|
| A — Basic conversation (tools, context) | ✅ PASS | `3f0a2484-3e28-486c-82e3-860bc9383494` |
| B — Rehydration (SIGINT → restart, recall, no replay) | ✅ PASS | `3f0a2484-3e28-486c-82e3-860bc9383494` |
| C — Context isolation (new room, no leak) | ✅ PASS | `6ca1489e-19da-4c36-836d-8ca12171ce4e` |
| D — Multi-participant (direct + relay-back, both directions) | ✅ PASS | `d08380a9…`, `375ecbbb…` |
| E — Memory lifecycle | ◑ PARTIAL | `bde2b3bb-b128-495e-9ef2-86dbca7e38bf` |
| F1 — Contacts DISABLED | ✅ PASS | `b657cc2b-d775-48fd-9057-1f8b68feb77f` |
| F2 — Contacts CALLBACK | ✅ PASS | `c265a500-7b12-4664-9459-8540188c6dac` |
| F3 — Contacts HUB_ROOM | ◑ PARTIAL | `bd3228a1-b359-4991-b440-9de118a0950c` |
| G — Execution emit (tool_call/tool_result) | ✅ PASS | `1296e91b-8bbd-4a45-aea3-ede5f753d143` |
| I — Concurrent rooms (no cross-room leak) | ◑ PARTIAL | `c873835b-37f1-40a6-863b-1e347ba0d19a` |

(Per-scenario room IDs for every step are in the harness reports under `tests/qa/reports/`.)

## Confirmed working (Parlant adapter / PR #304)
- Messaging + `@mention`-gated replies; `get_participants`, `lookup_peers`, `send_message`, thoughts.
- **Rehydration**: graceful shutdown, clean restart, recalls pre-restart context, **zero replayed/duplicate messages** (the highest-risk class — clean).
- **Context isolation**: fresh room = clean slate, no leakage.
- **Multi-agent coordination** with langgraph, both directions and both patterns: direct delivery (A delegates → B answers the user) and relay-back (A asks B → B replies → A relays).
- **Execution emit**: `tool_call`/`tool_result` events posted when `Emit.EXECUTION` is set (7/7).
- **Contact strategies** all wired: DISABLED ignores (pending), CALLBACK whitelist auto-approves with no LLM call, HUB_ROOM routes to the agent's LLM (approved a genuine request).
- Memory tools are exposed, invoked, and their errors surfaced gracefully.

## PARTIALs — root causes (not adapter regressions)
- **Memory get/supersede/archive → HTTP 404 `not_found`** (drives E, and I's room 1). `store_memory` succeeds and returns an ID, but reads/mutations by that ID 404 against production. This is the **platform memory API / canonical tool** (consistent with the known PLT-915 note), reproduced identically through the Parlant wrapper. → suggest a separate platform/tooling ticket.
- **F3 spam/empty stayed pending**: the hub-room LLM correctly **approved** the friendly request but didn't act on the spam/empty variants within the 30s window — Parlant's per-turn NLP latency, not a logic failure.
- **F2 broadcast** sub-check: auto-approve + "no LLM invocation" passed; only the `[Contacts]` broadcast message check missed (it lands in the agent's other rooms, not the scenario room) — cosmetic.

## Example hygiene — already handled in this PR (confirmed)
While validating, I independently trimmed the Parlant examples (`01–05`) to **persona-only** — moving the Band platform contract (tool list, `@mention` format, delegation/relaying) and all tools out of the agent `description`/manual guidelines, relying solely on the adapter's always-on `MATCH_ALWAYS` guideline. Comparing afterward: **this PR already does the same cleanup** — neither version references `create_parlant_tools`, both are persona-only, and the PR additionally updates the examples' `README.md`. The two passes converge on the same architecture; the only differences are cosmetic (naming, comments). **No action needed** — flagging to confirm it's the right call. It matters functionally, not just for hygiene: removing the duplicated guidelines/tools cut Parlant's per-turn evaluation load enough to flip core **A/B from PARTIAL → PASS**. *(The shared `examples/prompts/characters.py` for Tom/Jerry still references tools — left untouched since it's shared across 6 adapters; worth a glance but out of scope for Parlant.)*

## Recommendation
- **Adapter/tools rewrite looks solid to merge** from a QA standpoint — all core + multi-agent + emit + contact-strategy behavior verified live.
- File a follow-up for the **memory read/mutate 404** (platform side).
- **Rebase onto `dev` and re-validate before merge** — `dev` has moved on since this PR's base (see footnote).

---
**Validation setup (read this):** Run via the QA harness on branch `feat/qa-test-harness`, against **production** with the real OpenAI-backed Parlant NLP. PR #304's `adapters/parlant.py` and `integrations/parlant/tools.py` were overlaid and verified **byte-identical to the current PR head** (`b359288`); the venv loads `thenvoi` editable from the worktree `src/`, so the agents ran PR #304's adapter + tools. **Caveat:** only the Parlant files were overlaid — not a full PR-branch checkout. PR #304 is **behind `dev`**: its base predates both a large `dev` refactor (`runtime/oneshot.py` removed; `runtime/execution.py`/`langgraph`/`langchain` reworked) and the since-merged **INT-488 dependency restructuring (#319)** (parlant folded into the default `dev` extra, crewai isolated behind a lazy import, `phoenix-channels` bumped). So this validates **PR #304's Parlant code running on current `dev`** (a useful forward-compat signal), not the PR branch as-is. Recommend a **rebase onto `dev` + quick re-validation** before merge; PR #304's own `tests/` were not run in this pass. New Parlant expanded harness coverage (E/F/G/I scripts + a rebuilt multi-directional Scenario D) was added during this pass.

nir-singher-band:
@amit-gazal-thenvoi please review again

amit-gazal-band (approved):

PR #310, opened 2026-05-19
fix(sdk): surface websocket disconnect reasons [INT-330]

darvell-thenvoi:
## Summary
- Join `agent_control:{agent_id}` on SDK websocket connect and handle terminal `supersede` events.
- Record the platform disconnect reason, surface it through `ThenvoiLink.last_disconnect_reason` / `WebSocketDisconnectedEvent`, and disable Phoenix auto-reconnect before the subsequent empty close.
- Add HTTP websocket upgrade error parsing for the platform JSON contracts for 400, 409, 429, and 503 while leaving empty 403 responses generic.

## Platform contract evidence
- `agent_control:{agent_id}` can push `supersede` before the old socket is closed.
- `supersede` includes `reason`, `message`, `retryable`, `retry_after`, `target_socket_id`, and nullable `correlation_id`; the SDK records this payload and does not infer a reason from the later empty Phoenix disconnect.
- Upgrade failures are HTTP/JSON responses, not channel events: 409 `connection_conflict`, 429 `too_many_requests` with `retry_after` / `Retry-After`, 400 `invalid_on_conflict`, and 503 `tracking_failed`. Empty 403 remains generic/auth.

## Test plan
- `pytest tests/websocket/test_client.py tests/platform/test_link.py` — 72 passed
- `ruff check src/thenvoi/client/streaming/client.py src/thenvoi/client/streaming/__init__.py src/thenvoi/platform/event.py src/thenvoi/platform/link.py tests/websocket/test_client.py tests/platform/test_link.py` — passed
- `ruff format --check src/thenvoi/client/streaming/client.py src/thenvoi/client/streaming/__init__.py src/thenvoi/platform/event.py src/thenvoi/platform/link.py tests/websocket/test_client.py tests/platform/test_link.py` — passed
- `pyrefly check src/thenvoi/client/streaming/client.py src/thenvoi/platform/event.py src/thenvoi/platform/link.py` — passed
- GitHub CI — passed
- Live production smoke with a temporary registered agent: opened two Python SDK `ThenvoiLink` connections for the same agent and verified the first received `WebSocketDisconnectedEvent` with `reason=session.already_connected`, `retryable=false`, `retry_after=30`, `target_socket_id=agent_socket:<agent_id>`, nullable `correlation_id`, `last_disconnect_reason` set, and `is_connected=false`.

## Notes
- A raw websocket `on_conflict=reject` production probe did not deterministically produce a live 409 because production routing can place duplicate reject connects on different platform nodes. The 409/429/400/503 parsing remains covered by behavior tests against the platform JSON contract.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -23,7 +23,7 @@ classifiers = [
 
 dependencies = [
     "thenvoi-client-rest==0.0.7",
-    "phoenix-channels-python-client>=0.1.5",
+    "phoenix-channels-python-client>=0.2.1",
     "python-dotenv>=1.2.2",
     "pyyaml>=6.0",
     "cryptography>=46.0.5",
--- a/src/thenvoi/client/streaming/__init__.py
+++ b/src/thenvoi/client/streaming/__init__.py
@@ -9,7 +9,6 @@ Usage:
 from thenvoi.client.streaming.client import (
     WebSocketClient,
     WebSocketDisconnectReason,
-    WebSocketUpgradeError,
     MessageCreatedPayload,
     RoomAddedPayload,
     RoomRemovedPayload,
@@ -24,6 +23,7 @@ from thenvoi.client.streaming.client import (
     ContactRemovedPayload,
     SupersedePayload,
 )
+from thenvoi.client.streaming.errors import WebSocketUpgradeError
 
 __all__ = [
     "WebSocketClient",
--- a/src/thenvoi/client/streaming/__init__.py
+++ b/src/thenvoi/client/streaming/__init__.py
@@ -8,6 +8,8 @@ Usage:
 
 from thenvoi.client.streaming.client import (
     WebSocketClient,
+    WebSocketDisconnectReason,
+    WebSocketUpgradeError,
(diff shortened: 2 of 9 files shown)


amit-gazal-band (changes requested):
Two things to address before merge.

1. Duplicated disconnect-reason state. Both WebSocketClient._last_disconnect_reason and ThenvoiLink._last_disconnect_reason hold the same value — _on_supersede writes to the link's field and also calls self._ws.record_terminal_disconnect(reason), which stores it again on the client. Pick one owner (suggest the client, with the link's property delegating to self._ws.last_disconnect_reason) so there's a single source of truth.

2. No test for the __aenter__ upgrade-error wrapping. from_exception is unit-tested directly, but the WebSocketClient.__aenter__ glue that catches the PHX handshake exception, calls from_exception, and raises the wrapped error (or re-raises the original when from_exception returns None) is uncovered. Add a test that injects an InvalidStatus into a mocked PHX client and asserts WebSocketUpgradeError propagates, plus one for the passthrough case.

amit-gazal-band (approved):

PR #313, opened 2026-05-22
fix(integrations): dedupe Claude SDK send_message MCP retries

darvell-thenvoi:
## Summary

User-visible symptom: a single Claude SDK-driven LLM turn renders **two identical chat messages** in the room (and the customer is billed for two LLM calls). The duplicate appears *after* the `Complete` event, with logs showing one `Sending query` and one `Complete` for the turn.

The underlying trigger is event-loop starvation: under sustained Claude CLI load the Python event loop is starved of Phoenix WS heartbeats, the WS reconnects many times in a few minutes, and along the way Claude re-emits the same `band_send_message` MCP tool call for one LLM-intended send via several paths:

- MCP transport retry when the in-process handler takes too long to ack.
- Session resume after a Phoenix WS reconnect, while the previous response is still being streamed.
- Claude CLI producing a new turn after the original `Complete` event has already fired.

The event-loop starvation itself is being addressed separately in infrastructure and is not in this repo.

This PR is the **SDK-side defense** for the same failure: a tool-call MCP server should not produce two side effects for one logical call, no matter why the duplicate arrived.

## What changed

- New `src/band/integrations/claude_sdk/dedup_tools.py` containing `DedupingAgentTools`, a wrapper for the per-room `AgentToolsProtocol` that:
  - dedupes `send_message(content, mentions)` calls by `(content, frozenset(mentions))` inside a 30 s TTL window;
  - returns the cached result verbatim for any repeat;
  - forwards every other method through `__getattr__` so the wrapper does not silently become a stale interface when `AgentToolsProtocol` grows new methods;
  - serializes concurrent calls with an `asyncio.Lock` so two racing duplicates can never both miss the cache and both POST;
  - bounds the cache to 64 entries per room with LRU eviction;
  - logs a `WARNING` on every dedup hit so the failure mode stays visible in production logs.
- `ClaudeSDKAdapter.on_message` wraps the incoming `tools` with `DedupingAgentTools` (unless the operator opts out via `send_message_dedup_ttl_seconds=0`) and stores the wrapped instance in `self._room_tools[room_id]`. The MCP backend (`_create_mcp_backend`) already passes `get_tools=self._room_tools.get`, so every MCP tool invocation routes through the shim transparently — no MCP backend changes needed.
- New constructor parameter `send_message_dedup_ttl_seconds: float = DEFAULT_DEDUP_TTL_SECONDS` on `ClaudeSDKAdapter`, validated `>= 0`.

### Cross-turn persistence

The dominant pattern is a duplicate tool call landing **after** the original turn's `Complete` event — i.e. on a *subsequent* `on_message`. `SimpleAdapter` builds a fresh `AgentTools` per inbound message, so a wrapper rebuilt per call would drop the cache and let that post-`Complete` duplicate through. `DedupingAgentTools.update_inner()` swaps just the inner reference under the lock, preserving the cache across the turn boundary. The adapter only calls `update_inner` when the tools identity actually changed, to avoid a needless lock acquisition.

### Operator-debuggable warnings

`DedupingAgentTools` takes an optional `label` (the adapter passes `room_id`) that is interpolated into the dedup-hit warning. Without it the log line is indistinguishable across a multi-room tenant during a dedup storm.

### Failure-path semantics pinned

A failure of the inner `send_message` must not poison the cache — otherwise a transient platform 5xx would silently swallow every subsequent send for the next `ttl_seconds`. New tests pin that contract so a future cache refactor cannot regress it without making the change visible.

## Why this is scoped to ClaudeSDKAdapter only

The shim is intentionally not in `runtime/tools.AgentTools`. Other adapters (Anthropic, LangChain, Pydantic AI, Codex, etc.) drive HTTP/LLM calls directly from the runtime task and do not exhibit this failure mode. Penalising every adapter with a dedup window because one framework runs a heavy subprocess that can re-issue MCP calls would be wrong.

## What this PR is **not** trying to fix

- It does not fix the event-loop starvation. That stays owed to the infra team.
- It does not bump `ExecutionContext._max_processed_ids`. That was already raised to 500 on `dev`.
- It does not change `ClaudeSessionManager` or the WS resync path. Those already invalidate dead subprocess sessions correctly.

## Test plan

- [x] `tests/integrations/claude_sdk/test_dedup_tools.py`: keying (content, mention order, dict-vs-string mentions, None-vs-empty), TTL expiry via monkeypatched `time.monotonic`, lock serialization under two concurrent duplicate sends, LRU bound, transparent passthrough of `send_event` / `add_participant` / `participants`, `update_inner` cache survival and in-flight-send lock semantics, dedup-hit warning logging, `label` interpolation in warnings, and failure semantics (inner-raise propagates, cache not poisoned, recovery POST runs, post-recovery duplicates dedup).
- [x] `TestSendMessageDedupWiring` in `tests/adapters/test_claude_sdk_adapter.py`: default wrapping in `on_message`, `ttl=0` opt-out, negative-ttl `ValueError`, end-to-end dedup through the stored room-tools entry, cross-turn wrapper persistence with inner swap, cross-room isolation (identical sends in two rooms POST twice), `on_cleanup` evicts the wrapper, and `update_inner` is skipped on identity match.
- [x] `uv run pytest tests/integrations/claude_sdk/ tests/adapters/test_claude_sdk_adapter.py -q` — 126 passed.
- [x] `uv run ruff check` clean.
- [x] `uv run ruff format --check` clean.
- [x] `uv run pyrefly check` clean.

## Risk

Low. The wrapper is opt-out (set `send_message_dedup_ttl_seconds=0` to bypass). The 30 s TTL is shorter than any reasonable human user re-asking for the same content. Cache is per-room, bounded, and uses `asyncio.Lock`, so no cross-room interference or memory growth. Per-room isolation is locked down by `test_distinct_rooms_get_distinct_wrappers` so a future `_room_tools` refactor cannot silently turn the shim into a tenant-wide message suppressor.

### DIFF
--- a/src/band/adapters/claude_sdk.py
+++ b/src/band/adapters/claude_sdk.py
@@ -502,10 +502,10 @@ class ClaudeSDKAdapter(SimpleAdapter[ClaudeSDKSessionState]):
                 await existing.update_inner(tools)
                 tools = cast(AgentToolsProtocol, existing)
             else:
-                wrapped: Any = DedupingAgentTools(
+                wrapper = DedupingAgentTools(
                     tools, ttl_seconds=self.send_message_dedup_ttl_seconds
                 )
-                tools = wrapped
+                tools = cast(AgentToolsProtocol, wrapper)
                 self._room_tools[room_id] = tools
         else:
             self._room_tools[room_id] = tools
--- a/src/band/adapters/claude_sdk.py
+++ b/src/band/adapters/claude_sdk.py
@@ -61,6 +61,10 @@ from band.integrations.mcp.backends import (
 )
 from band.integrations.claude_sdk.session_manager import ClaudeSessionManager
 from band.integrations.claude_sdk.prompts import generate_claude_sdk_agent_prompt
+from band.integrations.claude_sdk.dedup_tools import (
+    DEFAULT_DEDUP_TTL_SECONDS,
+    DedupingAgentTools,
+)
 from band.runtime.custom_tools import CustomToolDef
 from band.runtime.tools import (
     ALL_TOOL_NAMES,
@@ -199,6 +203,7 @@ class ClaudeSDKAdapter(SimpleAdapter[ClaudeSDKSessionState]):
         max_pending_approvals_per_room: int = 50,
         approval_authorized_senders: set[str] | None = None,
         features: AdapterFeatures | None = None,
+        send_message_dedup_ttl_seconds: float = DEFAULT_DEDUP_TTL_SECONDS,
     ):
         """
         Initialize the Claude SDK adapter.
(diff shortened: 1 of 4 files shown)


amit-gazal-band (approved):
Great. Please just remove references to Linear and/or internal processes (INT-502 for example)

PR #314, opened 2026-05-24
fix: enable multi-agent orchestration through dumb bridge (agentcore) [INT-506]

amit-gazal-band:
## Summary
- Bridge rewritten as a dumb WS→HTTP/AgentCore forwarder with N-identity support. No Band logic in the bridge anymore — all platform semantics (mention parsing, message construction, lifecycle) live in the container's SDK.
- AgentCore container is now an SDK-driven FastAPI app: receives forwarded events, runs `AnthropicAdapter`, honors `mark_processing` / `mark_processed` / `get_next_message` so concurrent invocations don't duplicate work.
- Three demo agents (`@weather`, `@math`, `@personal_assistant`) coordinate in a single Band room and answer the INT-506 acceptance prompt end-to-end.

## What changed

**Bridge** (`thenvoi-bridge/`)
- New `BridgeConfig.agents: list[AgentConfig]` and tagged-union `Target` (`HTTPTarget | AgentCoreTarget`); replaces scalar single-agent config.
- New `HTTPForwarder` and `AgentCoreForwarder` (boto3 SigV4).
- Per-room `asyncio.Lock` in `AgentRunner` so events for the same room serialize through the forwarder; cross-room throughput unaffected.
- Deleted: `MentionRouter`, `Handler` protocol, `SessionStore`, `handlers/` (`AgentCoreHandler`, `LangChainHandler`), `langchain_agent_server.py`, `echo_server.py`, and their tests.
- Deploy templates updated: single `THENVOI_BRIDGE_AGENTS` secret; IAM wildcard for `runtime/*`.

**Container** (`examples/agentcore/`)
- `agentcore_llm_server.py`: full rewrite from MCP shim to FastAPI app (`/invocations`, `/ping`).
- Builds `PlatformMessage` / `AgentTools` / `AgentInput` per invocation; calls `adapter.on_event` directly.
- Lifecycle dance per invocation: `get_next_message` → `mark_processing` → adapter → `mark_processed` → drain (`mark_processing`+`mark_processed` per stale msg). Mirrors `ExecutionContext` without the long-running process loop.
- `Emit.EXECUTION` on by default so `tool_call` / `tool_result` show as platform events.
- Dockerfile uses `uv sync --no-editable` so the SDK is properly installed into `.venv`.
- New `pyproject.toml` extra: `agentcore_runtime` (fastapi + uvicorn).

**Demo + docs** (`examples/agentcore/`)
- `agents/{weather,math,personal_assistant}.txt` — system prompts for the three roles.
- `agent_config.yaml.example` and `README.md` covering build → ECR push → AgentCore Runtime registration → bridge config.

## Tests
- [x] **102 unit tests pass** — 78 bridge + 24 container. `pyrefly` clean (0 errors), `ruff check`/`format` clean.
- [x] Regression coverage: same-room serialization, different-room overlap, lifecycle marks in expected order, skip on `no_pending` / `already_processed`, drain with full `processing → processed` transition, `mark_failed` on adapter exception.
- [x] **Live end-to-end verified** against `Thenvoi-Dev` (account `603630702366`, us-east-1). Three AgentCore Runtimes deployed; bridge connected as three identities; user asks "what is the temperature difference now, in percents, between Tel Aviv and Warsaw?" → PA recruits weather & math, asks each, synthesizes the final answer. CloudWatch confirmed the lifecycle path (sibling invocations hit `get_next_message → 204` and skipped).
- [ ] **E2E test scaffold** at `tests/e2e/agentcore/test_three_agent_orchestration.py` (gated by `E2E_TESTS_ENABLED + AGENTCORE_DEMO_PA_AGENT_ID`). Skips by default; needs persistent demo deployments to run in CI.

## Known limitations / follow-ups
- The container delegates the full lifecycle dance to the SDK's `OneShotInvoker` (`src/thenvoi/runtime/oneshot.py`) rather than re-implementing it — request/response hosts (AgentCore Runtime, Lambda) call SDK code instead of duplicating `ExecutionContext`. The container is now a thin FastAPI transport.
- AgentCore Runtime caps sessions at 15-min idle / 8-hr max; `runtimeSessionId` is derived from `room_id` so each Band room gets its own session.
- Drain loop is capped at 50 messages defensively.

## Net diff
~3.2k insertions / ~7.5k deletions — substantial net delete from removing the bridge's Band-aware layers.

Closes [INT-506](https://linear.app/thenvoi/issue/INT-506).


thenvoi-bridge/bridge_core/bridge.py
```diff
@@ -504,224 +279,346 @@ async def _connect_and_consume(self) -> None:
...
+                    "Agent %s: failed to unsubscribe from room %s",
+                    self._config.agent_id,
+                    event.room_id,
+                    exc_info=True,
+                )
+
+        # Dedup message events by message id (reconnect may redeliver).
+        if isinstance(event, MessageEvent) and event.payload is not None:
+            msg_id = getattr(event.payload, "id", None)
+            if msg_id and self._is_duplicate(msg_id):
```
darvell-thenvoi: `_is_duplicate()` records the message id before the forward succeeds. If `self._forwarder.forward(payload)` raises or times out, the id stays in `_processed_message_ids`; on reconnect `_rehydrate_backlog()` can fetch the same still-open message and immediately drop it as a duplicate.

Please make the dedupe check non-mutating until after a successful forward. Something like:

```python
msg_id = getattr(event.payload, "id", None)
if msg_id and msg_id in self._processed_message_ids:
    logger.debug("Agent %s: skipping duplicate message %s", self._config.agent_id, msg_id)
    return

...

async with room_lock:
    await self._forwarder.forward(payload)

if msg_id:
    self._remember_processed_message(msg_id)
```

where `_remember_processed_message()` contains the current bounded-cache insertion/eviction logic. That way a failed forward stays retryable through WS redelivery or rehydration. Please also add a regression test where the first forward raises, `_rehydrate_backlog()` sees the same message id, and the second forward is attempted.

thenvoi-bridge/bridge_core/bridge.py
```diff
@@ -504,224 +279,346 @@ async def _connect_and_consume(self) -> None:
...
+            logger.warning(
+                "Agent %s: failed to fetch existing rooms",
+                self._config.agent_id,
+                exc_info=True,
+            )
         return []
 
-    async def _handle_event(self, event: object) -> None:
-        """Dispatch a platform event to the appropriate handler.
+    async def _rehydrate_backlog(self, room_ids: list[str]) -> None:
```
darvell-thenvoi, line 318: The rehydration comment says `/next` also returns messages stuck in `processing`, but the SDK’s recovery path from #294 handles those through `get_stale_processing_messages()` because `/next` skips them. This AgentCore bridge path only calls `get_next_message()`, so a container crash after `mark_processing()` but before `mark_processed()` still will not be replayed by bridge reconnect rehydration.

The fix needs to cover both sides of this path. The bridge should fetch stale `processing` messages and forward them as recovery nudges, but `OneShotInvoker` also needs to accept those forwarded messages by checking `get_stale_processing_messages()` during its claim step. If the bridge forwards a stale `processing` message while the container still only verifies against `/next`, the container will reject it because `/next` skips that message.

Concretely, bridge rehydration should do something like:

```python
stale_messages = await self._link.get_stale_processing_messages(room_id)
for msg in stale_messages:
    await self._safe_handle_event(self._backlog_event(msg))

next_msg = await self._link.get_next_message(room_id)
if next_msg is not None:
    await self._safe_handle_event(self._backlog_event(next_msg))
```
and the one-shot claim path should recognize the stale case too:

```
next_msg = await self._link.get_next_message(room_id)
if next_msg is None or next_msg.id != msg_id:
    stale_messages = await self._link.get_stale_processing_messages(room_id)
    if not any(msg.id == msg_id for msg in stale_messages):
        return {"status": "no_pending" if next_msg is None else "already_processed", "message_id": msg_id}
```
Then it can call mark_processing() and continue normally.

thenvoi-bridge/bridge_core/forwarder.py
```diff
@@ -0,0 +1,170 @@
...
+            # We discard the response — the bridge holds no Band logic.
+            body = response.get("response") or response.get("body")
+            if body is not None:
+                try:
+                    body.read(_MAX_RESPONSE_BYTES)
+                finally:
+                    body.close()
+
+        try:
+            await asyncio.wait_for(
```
darvell-thenvoi, line 160: This timeout only stops the bridge from waiting. It does not necessarily stop the boto call already running inside the background thread.

So if AgentCore is slow, this can happen:

1. The bridge sends message A for room 123.
2. The boto call keeps running.
3. `wait_for(...)` times out.
4. The bridge logs a timeout and unlocks room 123.
5. The bridge can now send message B for the same room, even though message A may still be running in AgentCore.

That means the per-room lock is no longer actually protecting the room from concurrent AgentCore invocations.

I’d make the boto client timeout match the bridge timeout, so the underlying call ends when the bridge thinks it ended. Something like:

```python
from botocore.config import Config

self._client = boto3.client(
    "bedrock-agentcore",
    region_name=self._target.region,
    config=Config(
        connect_timeout=min(self._target.timeout, 10),
        read_timeout=self._target.timeout,
        retries={"max_attempts": 1},
    ),
)
```

If we keep `to_thread`, we should also bound those calls with a semaphore or dedicated executor so repeated slow AgentCore calls cannot pile up background threads.

thenvoi-bridge/bridge_core/config.py
```diff
@@ -0,0 +1,221 @@
...
+        return cls(**kwargs)
+
+
+class ReconnectConfig(BaseModel):
+    """Per-agent reconnection backoff."""
+
+    initial_delay: float = 1.0
+    max_delay: float = 60.0
+    multiplier: float = 2.0
+    jitter: float = 0.5
```
darvell-thenvoi, line 197: This config reads like `jitter` is a magnitude, but the runner treats it as a boolean. Here `ReconnectConfig` exposes it as a float:

```python
jitter: float = 0.5
```

and validates only that it is non-negative. But the backoff code only checks whether it is greater than zero in bridge.py:

```
if self._reconnect.jitter > 0:
    sleep_time = random.uniform(0, delay)
else:
    sleep_time = delay
```

That means 0.25, 0.5, and 1.0 all produce full jitter over [0, delay]. We should do the following:

```
jitter = min(self._reconnect.jitter, 1.0)
sleep_time = delay * (1 - jitter) + random.uniform(0, delay * jitter)
```

thenvoi-bridge/bridge_core/bridge.py
```diff
@@ -476,26 +259,18 @@ async def _connect_and_consume(self) -> None:
...
                     if isinstance(e.__cause__, StopAsyncIteration) or isinstance(
                         e.__context__, StopAsyncIteration
                     ):
                         break
                     raise
                 next_fut = None
 
-                # Fire handler as a background task so we can pull the next
-                # event immediately without waiting for the handler to finish.
+                # Fire forward as a background task so we keep pulling events.
```
darvell-thenvoi, line 289: This creates a new task for every WebSocket event with no cap.

The per-room lock helps once the task is running, but it does not stop tasks from piling up. If one room gets a burst of events while the target is slow, the bridge can create a lot of tasks that all sit around waiting on the same room lock. Across rooms, it can also fan out a lot of HTTP/AgentCore calls at once.

Rehydration has the same shape: it does an unbounded `gather()` over every existing room.

I’d add a bounded queue or at least a semaphore. For example:

```python
self._forward_semaphore = asyncio.Semaphore(config.max_concurrent_forwards)

async def _safe_handle_event(self, event: object) -> None:
    async with self._forward_semaphore:
        await self._handle_event(event)
```

and use the same limit around rehydration nudges.

src/thenvoi/runtime/oneshot.py
```diff
@@ -0,0 +1,438 @@
...
+        Raises:
+            OneShotEnvelopeError: envelope is missing ``room_id`` or
+                ``payload.id`` for a ``message_created`` event.
+            RuntimeError: ``startup()`` was not called first.
+        """
+        if not self._started:
+            raise RuntimeError("OneShotInvoker.startup() not called")
+
+        event_type = body.get("event_type")
+        if event_type != "message_created":
```
darvell-thenvoi, line 196: I think maybe we should handle room_removed/room_deleted and forward them to the Anthropic adapter here. 

```python
if event_type in {"room_removed", "room_deleted"}:
    room_id = body.get("room_id") or (body.get("payload") or {}).get("id")
    if room_id:
        await self._adapter.on_cleanup(room_id)
    return {"status": "cleaned_up", "event_type": event_type, "room_id": room_id}
```

I can't shake the feeling I'm misunderstanding something though and this isn't required.

examples/agentcore/README.md
```diff
@@ -0,0 +1,238 @@
+# AgentCore demo — three agents orchestrating in a Band room
+
+A working example of three AWS Bedrock AgentCore-hosted agents that
+coordinate inside a single Thenvoi (Band) chatroom: `@weather`, `@math`, and
+`@personal_assistant`. The personal assistant adds the other two when it
+needs them, asks each a targeted question, reads their replies, and posts
+a final answer back to the user. No human relay between agents.
+
+This demo proves the bridge's dumb-pipe model (INT-506).
```
darvell-thenvoi: Remove the linear ticket number/reference, ideally just remove this line entirely.

examples/agentcore/README.md
```diff
@@ -0,0 +1,238 @@
...
+```
+
+## Step 6 — Run the bridge
+
+```bash
+# Install bridge deps
+uv sync --extra bridge_agentcore
+
+# Start the bridge
+uv run python -m bridge_core
```
darvell-thenvoi: This command does not work from the repo root. `bridge_core` lives under `thenvoi-bridge/` and is not installed as a package, so `uv run python -m bridge_core` fails with `No module named bridge_core` unless the cwd/PYTHONPATH is set up first.

I’d make this block only show commands that work from the context the README gives:

```bash
PYTHONPATH=thenvoi-bridge uv run python -m bridge_core
# or
uv run python examples/agentcore/run_agentcore.py
```

or add `cd thenvoi-bridge` before `python -m bridge_core`.

pyproject.toml
```diff
@@ -100,20 +100,26 @@ acp = [
...
+    # Deps for the container hosted by Bedrock AgentCore Runtime
+    # (examples/agentcore/agentcore_llm_server.py). Combine with the
+    # adapter extra of your choice (e.g. anthropic).
+    "fastapi>=0.110",
+    "uvicorn>=0.29",
 ]
 google_adk = [
     "google-adk>=1.0.0,<2",
 ]
-bridge_langchain = [
```
darvell-thenvoi, line 114: The `bridge_langchain` extra is removed from `pyproject.toml`, but the root README still advertises it in the bridge extras list.

I’d update the README in the same PR: remove `bridge_langchain` and mention only the extras that still exist for this path (`bridge`, `bridge_agentcore`, `agentcore_runtime`). Otherwise `pip install ...[bridge_langchain]` is still documented but no longer exists.

src/thenvoi/runtime/oneshot.py
```diff
@@ -0,0 +1,438 @@
...
+        # 1. Self-message filter — Thenvoi echoes the agent's own outbound
+        # messages back on its WS subscription, which the bridge forwards here.
+        if (
+            payload.get("sender_type") == "Agent"
+            and payload.get("sender_id") == self._agent_id
+        ):
+            return {"status": "skipped_self", "message_id": msg_id}
+
+        # 2. Verify the triggering message is the next open one for this agent.
+        next_msg = await self._link.get_next_message(room_id)
```
darvell-thenvoi, line 246: `get_next_message()` returns `None` for two very different cases: there is no pending message, or the REST call failed. In this path those get treated the same.

That means a temporary platform/API/network failure becomes `{"status": "no_pending"}`. The HTTP request succeeds, so the bridge thinks the event was handled and will not retry, even though the message may still be open on the platform.

I’d add modify get_next_message to be more strict for this claim step: return `None` only for a real 204/no-content response, and raise on anything else.

```python
async def get_next_message(self, room_id: str) -> PlatformMessage | None:
    try:
        response = await self.rest.agent_api_messages.get_agent_next_message(
            chat_id=room_id,
            request_options=DEFAULT_REQUEST_OPTIONS,
        )
    except ApiError as e:
        if e.status_code == 204:
            return None
        raise

    if response.data is None:
        return None
    return _platform_message_from_response(response.data, room_id)
```

Then `OneShotInvoker` can correctly fail so "I could not check the claim" fails the invocation instead of looking the same as "there was nothing to do."

darvell-thenvoi (changes requested):
Adjacent bridge docs still point at the removed handler-era model: `thenvoi-bridge/.env.example` documents `AGENT_MAPPING`, `LANGCHAIN_URLS`, `AGENTCORE_RUNTIME_ARN`, `SESSION_TTL`, and `HANDLER_TIMEOUT`; `thenvoi-bridge/Dockerfile` / `docker-compose.yml` still tell users to override CMD to register handlers. Since this PR deletes the handler/router/session architecture, we should remove them too.

thenvoi-bridge/bridge_core/bridge.py
amit-gazal-band: Done. Split into check-only + `_remember_processed_message` after successful forward. Regression test `test_failed_forward_is_retryable_on_rehydration`.

thenvoi-bridge/bridge_core/bridge.py
amit-gazal-band: Re-checked against the platform code and I don't think this is needed — `/next` already covers stuck-processing messages.

The chain: `messages_controller.ex:458` (`def next`) → `Chat.get_next_actionable_message` (`chat.ex:3865`) → `list_messages_for_participant(..., delivery_status: nil, page_size: 1)`. With `delivery_status_filter == nil`, the SQL at `chat.ex:1645-1666` is:

```
?->'delivery_status'->? IS NULL
OR COALESCE(?->'delivery_status'->?->>'status', '') != 'processed'
```

i.e. "include the message if there's no delivery_status entry for this participant, **or** the status is anything except `processed`." So `delivered`, `processing` (stuck), and `failed` all pass through. The separate `"pending"` branch at `chat.ex:1668-1672` exists specifically to exclude active `processing` attempts — which would be redundant if the default did the same thing. Platform comments agree: `messages_controller.ex:464` "Includes stuck/processing messages for crash recovery" and `chat.ex:3850` "Processing messages (stuck/crashed - supports crash recovery)."

The misleading SDK docstring on `get_stale_processing_messages` ("the /next endpoint skips them") is what threw us off — that was authored back when the method was added but doesn't reflect what `/next` actually does. I updated it in this PR so the next reader doesn't get tripped up the same way.

So no code change here — bridge rehydration and OneShot's claim step keep calling `/next` only, since one call covers both backlog and stuck-message reclaim. If I missed something on the platform side, happy to revisit.

thenvoi-bridge/bridge_core/forwarder.py
amit-gazal-band: Done. `_get_client` now builds boto3 with `Config(connect_timeout=min(timeout,10), read_timeout=timeout, retries={"max_attempts": 1})`. Test patches `boto3.client` and asserts on the requested `Config`. Concurrency-cap half handled in #5.

thenvoi-bridge/bridge_core/config.py
amit-gazal-band: Done. Extracted to `_backoff_sleep_seconds`, formula is `delay * (1 - jitter) + uniform(0, delay * jitter)` with `min(jitter, 1.0)`. Test covers zero/partial/full jitter.

thenvoi-bridge/bridge_core/bridge.py
amit-gazal-band: Done. `BridgeConfig.max_concurrent_forwards` (default 32) + `asyncio.Semaphore` in `_safe_handle_event`. Per-agent scope. Test fires 10 tasks against a blocking forwarder with `Semaphore(3)`, asserts peak=3.

src/thenvoi/runtime/oneshot.py
amit-gazal-band: Done. `room_removed` / `room_deleted` now call `adapter.on_cleanup` (try/except, returns `cleaned_up`). Added inline note on the other event types staying `ignored`. Four tests.

examples/agentcore/README.md
amit-gazal-band: Done.

examples/agentcore/README.md
amit-gazal-band: Done.

pyproject.toml
amit-gazal-band: Done.

src/thenvoi/runtime/oneshot.py
amit-gazal-band: Done. `get_next_message` returns `None` only for 204; everything else propagates. Wrapped the OneShot drain (only unwrapped caller). Tests for 204 / non-204 / transport in `tests/platform/test_link.py`.

darvell-thenvoi (approved):

PR #319, opened 2026-05-26
fix(sdk): output_type crash, missing anthropic dep INT-488

nir-singher-band:
## Summary

- **Fix pydantic-ai output_type crash**: newer pydantic-ai versions reject `output_type=None` — use `output_type=str` and align return type annotation
- **Add anthropic to pydantic-ai extra**: required by `pydantic-ai[anthropic]` examples
- **Isolate crewai into dedicated CI job**: crewai conflicts with both parlant (opentelemetry-sdk) and pydantic-ai (pydantic version pin) — swap parlant into `dev` extra, create `dev-crewai` extra for isolated testing
- **Skip crewai tests in main CI job**: all crewai-specific tests run in `test-crewai` job; main job no longer fails on partially-mocked crewai

## CI layout change

| Before | After |
|--------|-------|
| `dev` includes crewai, excludes parlant | `dev` includes parlant + pydantic-ai, excludes crewai |
| `test-parlant` job (isolated) | `test-crewai` job (isolated) |
| crewai_flow tests fail in main job via mocking | crewai tests skip in main job, pass in `test-crewai` |

## Test plan

- [x] `test-crewai` job passes on both Python 3.11 and 3.12
- [x] Main `test` job has zero non-pre-existing failures
- [x] `lint` and `packaging` jobs pass
- [x] pydantic-ai adapter and converter tests pass
- [x] Parlant tests run in main job (no longer isolated)


amit-gazal-band (changes requested):
Two real fixes plus the dependency restructure look good. Before merging, I'd like to address three things:

1. **Add a regression test for the actual INT-488 fix.** The 3-line `output_type=str` change isn't asserted anywhere — a future "cleanup" back to `None` would only be caught if a conformance test happens to hit pydantic-ai-slim 1.87+. A one-liner in `tests/adapters/test_pydantic_ai_adapter.py` is enough:

   ```python
   with patch("thenvoi.adapters.pydantic_ai.Agent") as MockAgent:
       adapter._create_agent()
       assert MockAgent.call_args.kwargs["output_type"] is str
   ```

2. **Use `pydantic-ai-slim[anthropic]` instead of pinning raw `anthropic`** in the `pydantic-ai` extra. The idiomatic pattern lets pydantic-ai pin the compatible anthropic range itself; pinning raw `anthropic>=0.75.0` risks version skew the next time pydantic-ai bumps its provider surface.

3. **Make the crewai adapter import lazy** so the 11 new `pytestmark = pytest.mark.skipif(not _HAS_CREWAI, ...)` guards can be deleted.

   The reason these guards are needed at all is that `src/thenvoi/adapters/crewai.py` has a top-level `try: from crewai import ... except ImportError: raise ImportError(...)`. That hard-fails at module-import time when crewai isn't in the venv (which is now every CI job except `test-crewai`). `src/thenvoi/adapters/parlant.py` handles this the right way per `.claude/rules/05-optional-dependencies.md` — `TYPE_CHECKING`-only imports at module level, real imports inside `__init__`/methods — which is why parlant tests never needed these guards.

   If the crewai adapter follows the parlant pattern, all 11 skip-pytestmark blocks in this PR can go away, the `test-crewai` job keeps working (since real crewai is installed there), and the codebase converges on one optional-dep pattern instead of two.

nir-singher-band:
Thanks for the review — all three addressed in `74c5dfc`.

**1. Regression test for the `output_type=str` fix** — added `test_create_agent_uses_str_output_type` in `tests/adapters/test_pydantic_ai_adapter.py`, which patches `Agent` and asserts `call_args.kwargs["output_type"] is str`. A future revert to `None` now fails this unit test directly, independent of which pydantic-ai-slim version CI happens to run.

**2. `pydantic-ai-slim[anthropic]` instead of a raw pin** — the `pydantic-ai` extra now uses `pydantic-ai-slim[anthropic]>=1.56.0` and the standalone `anthropic>=0.75.0` pin is gone; `uv.lock` relocked accordingly. anthropic now comes transitively from pydantic-ai's own provider extra, so no more skew risk.

**3. Lazy crewai import → guards deleted** — `adapters/crewai.py` now follows the parlant `TYPE_CHECKING` pattern (real `from crewai import …` moved into `on_started` / `_get_context` / `create_crewai_tools`); `nest_asyncio` and `BaseTool` imports in `integrations/crewai/` are lazy too. With the hard import gone, **11 module-level `skipif(not _HAS_CREWAI)` guards were removed**. Two narrow guards were kept intentionally, since those tests genuinely need real crewai:
- `test_crewai_flow_phase3.py` — re-scoped from a module guard to a `@requires_nest_asyncio` decorator on just the two tests that drive the real sync→async bridge (`run_async` → `nest_asyncio.apply()`, which can't be mocked).
- `test_capability_gating_e2e.py` — kept its guards but fixed the patch targets to `crewai.Agent` / `crewai.LLM` (the lazy refactor removed those module-level names from the adapter).

`tests/framework_configs/adapters.py` gained `skip_on_started_conformance` when crewai is absent, since `on_started` still does a runtime `from crewai import …`.

Verified both CI jobs locally:
- **main (`dev`)** — 2570 passed, 0 failed (the now-unguarded crewai unit tests run here via mocking).
- **test-crewai (`dev-crewai`)** — 248 passed; pyrefly on crewai sources clean.
- ruff check + format clean.

✌️

amit-gazal-band (approved):

PR #329, opened 2026-05-28
ci: add promote dev to main workflow and CI/CD docs

amit-gazal-band:
## Summary

Adds a one-click **Promote Dev to Main** workflow and documents the repo's CI/CD flow, mirroring the model used in `thenvoi-website` but with a single promotion hop (`dev → main`, no `staging`).

## Changes

- **`.github/workflows/promote-dev-to-main.yml`** — manual `workflow_dispatch` action that:
  - Mints a token via the local `./.github/actions/GithubToken` action
  - Counts commits ahead (`origin/main..dev`); exits cleanly if `main` is already up to date
  - Builds the PR body **from git history** (commit list, `git diff --stat`, referenced PR numbers, metadata footer) — no external AI / no new secrets
  - Opens a `dev → main` PR titled `chore: promote dev to main` (valid Conventional Commits, so it clears the `Validate PR Title` required check), assigned to the triggerer
  - Supports a **`dry_run`** toggle that prints the description to the run summary without creating a PR
  - Never merges — review + rulesets still gate the merge
- **`docs/ci-cd-workflows.md`** — documents `feature → dev → main`, the branch rulesets, all workflows, and how to run the promote action.

## Notes

- `main` uses **strict** required status checks, so after a release-please commit lands on `main`, `dev` must be brought up to date before the next promotion PR can merge. This is documented in the new doc; no automated `main → dev` sync was added (per request).

### DIFF
--- a/.github/workflows/promote-dev-to-main.yml
+++ b/.github/workflows/promote-dev-to-main.yml
@@ -3,8 +3,6 @@ name: Promote Dev to Main
 run-name: Promote dev to main by @${{ github.actor }}
 
 on:
-  push:
-    branches: [ci/promote-dev-to-main-workflow]   # TEMP test trigger — remove before merge
   workflow_dispatch:
     inputs:
       dry_run:
@@ -102,7 +100,7 @@ jobs:
           cat pr_body.md
 
       - name: Create Pull Request
-        if: steps.check_promotion.outputs.needs_promotion == 'true' && inputs.dry_run == false && github.event_name != 'push'   # TEMP guard — never create a PR during push test
+        if: steps.check_promotion.outputs.needs_promotion == 'true' && inputs.dry_run == false
         env:
           GH_TOKEN: ${{ steps.app_token_generator.outputs.token }}
         run: |
--- a/.github/workflows/promote-dev-to-main.yml
+++ b/.github/workflows/promote-dev-to-main.yml
@@ -23,6 +23,12 @@ jobs:
       pull-requests: write
 
     steps:
+      - name: Checkout repository
+        uses: actions/checkout@v6
+        with:
+          fetch-depth: 0
+          ref: dev
+
       - name: Generate GitHub App Token
         id: app_token_generator
         uses: ./.github/actions/GithubToken
@@ -30,13 +36,6 @@ jobs:
           app_id: ${{ secrets.APP_ID }}
           installation_id: ${{ secrets.INSTALLATION_ID }}
           private_key: ${{ secrets.APP_PRIVATE_KEY }}
-
-      - name: Checkout repository
-        uses: actions/checkout@v6
-        with:
-          fetch-depth: 0
-          token: ${{ steps.app_token_generator.outputs.token }}
-          ref: dev
 
       - name: Fetch main
         run: git fetch origin main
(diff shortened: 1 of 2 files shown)


.github/workflows/promote-dev-to-main.yml
```diff
@@ -0,0 +1,129 @@
...
+
+jobs:
+  promote:
+    runs-on: ubuntu-latest
+    permissions:
+      contents: write
+      pull-requests: write
+
+    steps:
+      - name: Generate GitHub App Token
```
darvell-thenvoi, line 32: I compared this against the promotion workflows in `thenvoi-website` and I think there's a small difference here that might make this fail before checkout.

Website uses `thenvoi/thenvoi-envs/.github/actions/GithubToken@main`, so GitHub can fetch that action before checking out the repo. This one uses `./.github/actions/GithubToken`, which is a local action, so I don't think the runner has it available until after `actions/checkout` runs.

Could we either use the same external `thenvoi-envs` action as website, or do an initial checkout of `main` / the default branch before this step, mint the app token from that trusted copy of the action, then checkout `dev` with the app token after?

darvell-thenvoi (dismissed):

.github/workflows/promote-dev-to-main.yml
amit-gazal-band: Good catch — you were right. The local `./.github/actions/GithubToken` ran before `actions/checkout`, so the action wasn't on disk yet and it failed before checkout.

Fixed by reordering to match `release.yml`: checkout first, then mint the token. I also dropped `token:` from the checkout — only `gh pr create` needs the app token (it gets it via `GH_TOKEN`), and the git steps here are read-only.

Validated with a live run by temporarily adding a branch-scoped `push` trigger (since `workflow_dispatch` isn't available until the workflow lands on `main`). Checkout → Generate GitHub App Token → Fetch main all passed in order, and it correctly reported "No promotion needed" (dev is currently in sync with main). The temp trigger has been reverted.

nir-singher-band (approved):
## Validation
Ran the workflow with `dry_run: true` (run `26709409234`) — ✅ success. `dev` is currently 0 ahead / 3 behind `main`, so it correctly took the **no-promotion-needed** path and the dry-run summary did *not* fire on an empty body. I separately replayed the `Generate PR description` bash against a non-empty range: it renders cleanly (`exit 0`, no `pipefail` trip), and PR-number extraction works.

## Suggestions (non-blocking)

1. **Link the new doc from `CONTRIBUTING.md` in this PR.** As-is, `docs/ci-cd-workflows.md` ships orphaned — nothing references it. `CONTRIBUTING.md`'s *Release Process* section is a 3-line stub that this doc directly expands, and *Development Workflow* stops at "open a PR against `dev`" without mentioning promotion. Suggested add to *Release Process*:
   ```markdown
   Code flows `feature → dev → main`; merges to `main` trigger the PyPI release.
   For the full branch model, rulesets, and the **Promote Dev to Main** action, see
   [CI/CD Workflows Guide](docs/ci-cd-workflows.md).
   ```

2. **Doc — broaden the "dev falls behind" note.** It's attributed solely to the release-please commit, but the **merge-commit strategy itself** leaves `dev` behind `main`, so a `main → dev` sync is needed before *every* promotion once `dev` has new work — not only after a release. (Confirmed live: the 3 commits `main` currently leads by are the release-please release commit plus two merge commits.) The guidance is correct; just the cause is understated.


✌️

PR #331, opened 2026-05-31
feat: Slack adapter for remote Thenvoi agents (INT-461)

amit-gazal-band:
## Summary

Adds a first-class **Slack adapter** to the SDK so a remote Thenvoi agent can be exposed as a Slack-native AI App, mounted into the developer's existing web service. Implements Phase 1 (MVP) of [INT-461](https://linear.app/thenvoi/issue/INT-461/slack-integration).

**Architecture — wrapping, not gateway.** `SlackAdapter` wraps an `inner: SimpleAdapter` (the brain — Anthropic, LangGraph, etc.) and decorates `on_message`. One Thenvoi service identity, one process.

- Two outbound tools: `thenvoi_send_message` (real peer message) and `slack_send_message` (plain reply to the bound Slack thread).
- Thread context backfill: refetches `conversations.replies` per turn (no caching) so the brain sees full thread history.
- Session rehydration from the bootstrap `task` event binding.
- HTTP (Starlette/FastAPI) **and** Socket Mode transports.

## What's included (Steps 0–12)

- HMAC signature verification + URL verification, retry idempotency
- Per-thread Thenvoi room mapping, peer reply → Slack threading
- Status indicators (`assistant.threads.setStatus`) + Block Kit `plan`/`task` tool-call visibility
- Slack thread history backfill (`channels:history` / `groups:history`)
- Session rehydration on bootstrap
- Framework mount helpers + Socket Mode transport
- Manifest template, getting-started example (`examples/slack/01_basic_bot.py`), README section
- Test suite (~3.4k lines): signature, retry, server, socket, block kit, history converter, status, framework mounts, metrics

## Still open (follow-up)

- **Step 13** — mirror Slack context into the Thenvoi room as `thought` events for audit timelines
- **Step 14** — fresh-install validation pass (known: `delayed_events_enabled: true` set manually in the Slack GUI)

## Known limitation

If a Slack event arrives before any WS message after a restart, rehydration races and a duplicate Thenvoi room is created; per-turn backfill keeps replies correct. Eager-startup rehydration is the proper fix (not yet scoped).

## Test plan

- [ ] `uv run pytest tests/integrations/slack/ -v`
- [ ] `uv run ruff check . && uv run ruff format . --check`
- [ ] `uv run pyrefly check`
- [ ] Manual: restart-bridge backfill/rehydration smoke (`examples/slack/_dev_bridge.py`)


src/thenvoi/integrations/slack/adapter.py
```diff
@@ -0,0 +1,1187 @@
...
+                is_session_bootstrap=is_new_room,
+                room_id=room_id,
+            )
+        finally:
+            await self._set_status(slack_client, channel, thread_ts, "")
+
+    # ── Room management ──────────────────────────────────────────────
+
+    @staticmethod
+    def _thread_key(channel: str, thread_ts: str) -> str:
```
darvell-thenvoi, line 817: `SlackAdapter` supports multiple `SlackApp`s in one process, and `SlackRoomBinding` already carries `app_slug`, but the actual room lookup key is only `channel:thread_ts`. That feels like the wrong side of the trust boundary to shave off the app/workspace identity.

The failure mode is pretty straightforward:

- app A gets a Slack thread at `C123:1700.0001`
- `_get_or_create_room()` creates room R1 and stores `_thread_to_room["C123:1700.0001"] = R1`
- app B, in another Slack workspace, happens to see the same workspace-scoped channel/thread tuple
- `_get_or_create_room()` returns R1 before it looks at app B
- replies and plan/status updates go out through app A's Slack client because the room binding still points at app A

Impact: Messages end up in the wrong channel/workspace. 

Suggested fix: make the key include `app.slug`, and add a regression test with two `SlackApp`s using the same channel/thread IDs and getting different Band rooms.

src/thenvoi/integrations/slack/adapter.py
```diff
@@ -0,0 +1,1187 @@
...
+    # ── Slack thread history backfill ────────────────────────────────
+
+    async def _fetch_thread_history(
+        self,
+        *,
+        slack: AsyncWebClient,
+        channel: str,
+        thread_ts: str,
+        exclude_ts: str,
+    ) -> list[dict[str, Any]]:
```
darvell-thenvoi, line 1027: `_fetch_thread_history()` calls anything with `bot_id` or subtype `bot_message` a bot message. If it has `blocks`, it gets dropped as progress UI. If it’s plain text, it gets turned into:

- `sender_name = self.agent_name`
- `sender_type = "Agent"`
- `role = "assistant"`

The important bit is the adapter writes over the Slack sender identity before any framework converter gets to check it. So, converters that check `sender_name == self._agent_name` will think this is our own prior assistant turn, because `_fetch_thread_history()` has already set `sender_name` to `self.agent_name` for every plain-text bot message.

That means another Slack bot or webhook in the thread can end up in the model context as this bridge agent's own prior answer. The current tests even encode that shape with an arbitrary `bot_id` like `B999`, so this isn't just an untested corner.

Proposed fix: resolve/cache the current app's bot identity, probably via Slack `auth.test` or equivalent bot metadata, and only map messages from that bot id to assistant history. Other bots/webhooks should either be dropped or included as external context with a clear `slack-bot:<id/name>` sender.

src/thenvoi/integrations/slack/adapter.py
```diff
@@ -0,0 +1,1187 @@
...
+        finally:
+            await self._set_status(slack_client, channel, thread_ts, "")
+
+    # ── Room management ──────────────────────────────────────────────
+
+    @staticmethod
+    def _thread_key(channel: str, thread_ts: str) -> str:
+        return f"{channel}:{thread_ts}"
+
+    async def _get_or_create_room(
```
darvell-thenvoi, line 820: Every Slack event gets scheduled as its own background task. `_get_or_create_room()` checks `_thread_to_room`, then awaits `create_agent_chat`, then awaits `_emit_context_event`, and only after both awaits does it write the mapping.

So two events for the same Slack thread can both miss the map, both create rooms, both emit bootstrap events, and then race on which room gets stored. The loser is orphaned on the platform with a bootstrap event and no future traffic.

Impact: duplicate rooms for one Slack thread, split audit, and inconsistent `is_session_bootstrap` behavior.

Fix I’d want: per-thread `asyncio.Lock`, or a second check after the create call plus cleanup if another task won. If you can find somewhere we could have an idempotency key for chat creation, even better (I don't think platform has this though). Add a two-task regression test for the same `app/channel/thread_ts`.

pyproject.toml
```diff
@@ -97,6 +97,11 @@ acp = [
     "starlette>=0.40.0",
     "uvicorn>=0.32.0",
 ]
+slack = [
```
darvell-thenvoi, line 100: The `slack` extra installs `slack-sdk`, `starlette`, and `uvicorn`, but not `aiohttp`. Socket Mode imports `slack_sdk.socket_mode.aiohttp.SocketModeClient`, and `socket.py` already has a runtime ImportError saying Socket Mode requires `aiohttp`.

This is easy to miss locally because the dev/bridge extras do install `aiohttp`. A user installing only `thenvoi-sdk[slack]` can follow the default Socket Mode path and hit the missing dependency.

src/thenvoi/integrations/slack/block_kit.py
```diff
@@ -0,0 +1,150 @@
...
+    """
+    stripped = name
+    for prefix in ("thenvoi_", "slack_"):
+        if stripped.startswith(prefix):
+            stripped = stripped[len(prefix) :]
+            break
+    return stripped.replace("_", " ").strip().capitalize() or name
+
+
+def render_plan_blocks(plan: PlanState) -> list[dict[str, Any]]:
```
darvell-thenvoi, line 104: `render_plan_blocks()` emits:

- one header block
- one divider if there are tasks
- one section block per task

So for N tasks, we render `N + 2` blocks. At 49 tasks, that's 51 blocks, and Slack can reject the `chat.postMessage` / `chat.update`. The adapter catches and logs the failure, so the final reply might still happen, but the progress UI can disappear or stop updating.

Fix I’d want: cap this at 50 blocks, reserve space for an overflow summary, and add a regression test with 49+ tasks.

Slack doc link: Slack says messages can include up to 50 blocks: https://docs.slack.dev/reference/block-kit/blocks/

src/thenvoi/integrations/slack/__init__.py
```diff
@@ -0,0 +1,28 @@
+"""Slack integration for Thenvoi SDK.
+
+Exposes one or more Thenvoi peers as Slack-native AI agents.
+
+Example:
+    from thenvoi import Agent
+    from thenvoi.integrations.slack import SlackAdapter, SlackApp
+
+    adapter = SlackAdapter(
```
darvell-thenvoi, line 9: The module docstring example is copy-paste broken in two ways.

It shows this:

```python
adapter = SlackAdapter(
    apps=[
        SlackApp(
            slug="recruit",
            signing_secret="...",
            bot_token="xoxb-...",
            target_peer_slug="recruit-assistant",
        ),
    ],
    rest_url="https://app.thenvoi.com",
    api_key="...",
)
```

But `SlackApp` is defined in `src/thenvoi/integrations/slack/types.py:33-36` as only:

```python
slug: str
bot_token: str
signing_secret: str = ""
app_token: str = ""
```

And `SlackAdapter.__init__` is defined in `src/thenvoi/integrations/slack/adapter.py:341-356` with required keyword-only `inner`:

```python
def __init__(
    self,
    *,
    inner: SimpleAdapter[Any],
    apps: list[SlackApp],
    ...
) -> None:
```

I also ran the docstring constructor shape directly and it fails exactly how the code read suggests:

```text
SlackApp failed: SlackApp.__init__() got an unexpected keyword argument 'target_peer_slug'
SlackAdapter failed: SlackAdapter.__init__() missing 1 required keyword-only argument: 'inner'
```

Impact: `help(thenvoi.integrations.slack)` / autodoc gives a copy-paste example that fails immediately and teaches the wrong mental model. This adapter is a wrapper around an inner brain, not a peer-slug gateway.

Fix I’d want: replace that docstring with the real wrapping shape from `examples/slack/01_basic_bot.py`.

Maybe something like:
```
"""Slack integration for the Thenvoi SDK.

Wraps an inner framework adapter so one Thenvoi agent can receive Slack
Events API traffic and reply back into the originating Slack thread.

Example:
    from thenvoi import Agent
    from thenvoi.adapters import AnthropicAdapter
    from thenvoi.integrations.slack import SlackAdapter, SlackApp

    brain = AnthropicAdapter(model="claude-sonnet-4-6")

    slack = SlackAdapter(
        inner=brain,
        apps=[
            SlackApp(
                slug="recruit",
                signing_secret="...",
                bot_token="xoxb-...",
            ),
        ],
        rest_url="https://app.thenvoi.com",
        api_key="...",
    )

    agent = Agent.create(adapter=slack, agent_id="...", api_key="...")

    # HTTP transport: mount slack.router into your ASGI app, for example:
    # app.mount("/slack", slack.router)
    await agent.run()
"""
```

darvell-thenvoi (changes requested):

amit-gazal-band:
Moved to https://github.com/thenvoi/thenvoi-sdk-python/pull/342

PR #333, opened 2026-05-31
fix(integrations): stop posting "Invalid response from LLM call" error on every reply

nir-singher-band:
> Linear: [INT-795](https://linear.app/thenvoi/issue/INT-795/crewai-adapter-posts-invalid-response-from-llm-call-none-or-empty)

## Problem

CrewAI agents post **`Error: Invalid response from LLM call - None or empty.`** as an error event alongside essentially **every reply**, even though the actual answer is delivered correctly. Observed live in QA rooms (e.g. 11 error events in one basic-conversation room):

```
👤 "What is the capital of France?"
❌ [Agent/error] Error: Invalid response from LLM call - None or empty.
🤖 [Agent/text]  "The capital of France is Paris."
```

## Root cause

`ValueError("Invalid response from LLM call - None or empty.")` is raised by **CrewAI 1.14.3 itself** (`crewai/utilities/agent_utils.py`) when its ReAct loop produces an empty *final answer*. This adapter has the agent reply through the `thenvoi_send_message` **tool**, so the agent finishes its turn via the tool and then returns an empty final answer. The adapter's blanket `except Exception` caught that `ValueError`, posted it as an error event, and re-raised — once per turn.

Not a regression from the model or any recent refactor: the `crewai==1.14.3` pin is unchanged on `dev`.

## Fix

- **`integrations/crewai/tools.py`** — add a per-turn `ReplyTracker`; `_execute_tool` flips it to `replied=True` when `thenvoi_send_message` succeeds (shared by reference so it survives CrewAI's sync→async tool boundary).
- **`adapters/crewai.py`** — carry the tracker via a ContextVar + into `_process_message`. When CrewAI raises *after* a reply was already delivered, log it as non-fatal and skip the error event / re-raise. **Genuine no-response failures (no reply delivered) still surface exactly as before.**

## Tests

- Adapter: empty final answer after a successful reply is suppressed (no error event, no raise); genuine kickoff failure still reports + raises.
- Tools: a successful `thenvoi_send_message` marks the tracker; a failed send does not.

Full crewai suite green (259 passed); full unit suite green (non-live).

## Context

Surfaced by the INT-488 QA harness run. Independent of PR #319 (INT-488) — separate branch off `dev` to keep that PR focused.

✌️


amit-gazal-band (changes requested):
Nice catch on the root cause — the ReplyTracker approach is clean. One change before merge: the guard in `_process_message`'s `except Exception as e:` suppresses **any** exception once a reply went out, not just CrewAI's empty-final-answer ValueError.

Suggest narrowing it to the known case:

```python
if (
    reply_tracker is not None
    and reply_tracker.replied
    and isinstance(e, ValueError)
    and "Invalid response from LLM call" in str(e)
):
    logger.warning(...)
    return
```

As written, a genuine failure after the first successful `thenvoi_send_message` (downstream error, a bug in later processing, a second tool call throwing) gets swallowed with only a warning and no error event. Matching the specific ValueError keeps real failures visible.

Worth adding a test pinning the "genuine error after a reply still reports + raises" path too.

nir-singher-band:
Good call — narrowed the guard in `d475f5e`. It now suppresses only CrewAI's specific `ValueError("Invalid response from LLM call ...")` after a reply; any other exception (or a different ValueError) still reports an error event and propagates.

```python
if (
    reply_tracker is not None
    and reply_tracker.replied
    and isinstance(e, ValueError)
    and "Invalid response from LLM call" in str(e)
):
    logger.warning(...)
    return
```

Also added a parametrized test (`test_genuine_error_after_reply_still_reports_and_raises`) covering both a non-ValueError and an unrelated ValueError raised after a reply — both must report + raise. Full crewai adapter suite green (70 passed).

amit-gazal-band (approved):

PR #335, opened 2026-06-01
fix: force REST Responses API to avoid WebSocket 401 [INT-501]

amit-gazal-band:
## Problem

The Codex agent fails with **401 Unauthorized** when the Codex CLI connects to `wss://api.openai.com/v1/responses`. The same API key works for REST `POST /v1/responses` (HTTP 200) but fails over WebSocket. Affects API-key auth (`sk-proj`, `sk-svcacct`). Fixes [INT-501](https://linear.app/thenvoi/issue/INT-501/codex-agent-fails-to-authenticate-websocket-401-on-responses-api).

## Root cause (verified against a live `sk-svcacct` key)

Codex defaults to the **WebSocket transport** for the OpenAI Responses API, which appears to be account-gated and returns 401 even when REST works. Worse, Codex's automatic WS→HTTPS fallback **drops the `Authorization` header** ([openai/codex#15492](https://github.com/openai/codex/issues/15492)), so the fallback 401s too:

```
ERROR codex_api::endpoint::responses_websocket: failed to connect to websocket:
  HTTP error: 401 Unauthorized, url: wss://api.openai.com/v1/responses
...
unexpected status 401 Unauthorized: Missing bearer or basic authentication in header,
  url: https://api.openai.com/v1/responses
```

## Fix

`docker/codex/entrypoint.sh` now injects a custom model provider pinned to the HTTP/REST Responses transport into the generated `config.toml`:

```toml
model_provider = "openai_rest"

[model_providers.openai_rest]
name = "OpenAI (REST)"
base_url = "https://api.openai.com/v1"
env_key = "OPENAI_API_KEY"
wire_api = "responses"
supports_websockets = false
```

- **Gated on `OPENAI_API_KEY`** — only applies to API-key auth; OAuth/`auth.json` installs are untouched.
- **Toggle:** `CODEX_DISABLE_WEBSOCKET=false` restores Codex's default WebSocket path.
- **Idempotent**, and respects an operator-supplied `model_provider` (adds only the provider definition in that case).
- Source-home mode (`CODEX_USE_SOURCE_HOME=true`) is left alone, since that mount may be read-only.

> Note: setting `supports_websockets=false` on the built-in `openai` provider is not enough — its REST fallback drops auth, and the override has historically been ignored ([openai/codex#13103](https://github.com/openai/codex/issues/13103)). The custom provider keeps the bearer header on the REST path.

## Verification

Against a live `sk-svcacct` key with `codex-cli 0.135.0`:
- Default provider → `responses_websocket … 401` (reproduces the bug).
- The exact config the entrypoint generates → successful REST completion (`pong`, `rc=0`), no WS 401.

Also added `tests/integration/test_codex_responses_transport.py` — an opt-in reproduction/regression test (skips in CI; runs only with `CODEX_REPRO_ENABLED=true`, `OPENAI_API_KEY`, and a runnable `codex` binary).

## Out of scope

The Dockerfile still uses `@openai/codex@latest` (no version pin in this PR).

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -490,6 +490,7 @@ Replace `<extra>` with the appropriate framework extra (e.g., `langgraph`, `anth
 - Use async/await everywhere in async codebases; use `AsyncMock` for testing async methods
 - Catch `pydantic.ValidationError` separately from generic `Exception`
 - Use `raise ValueError(...)` for missing required config, not `logger.error()` + `sys.exit()`
+- Never put issue-tracker references in code — no Linear issue IDs (e.g. `INT-123`), Linear URLs, or ticket numbers in comments, docstrings, or strings. Explain the *why* in plain terms instead. (Branch names, commit messages, and PR descriptions may reference issues.)
 
 ## Pre-Commit Checklist
 
--- a/docker/codex/entrypoint.sh
+++ b/docker/codex/entrypoint.sh
@@ -75,8 +75,8 @@ fi
 # that endpoint returns 401 Unauthorized even though REST POST /v1/responses
 # succeeds (HTTP 200) with the same key, and Codex's WS->HTTPS fallback drops
 # the auth header (openai/codex#15492) so it 401s too. Route Codex through a
-# custom provider pinned to the HTTP/REST Responses transport instead. See
-# INT-501 (verified against a live sk-svcacct key).
+# custom provider pinned to the HTTP/REST Responses transport instead
+# (verified against a live sk-svcacct key).
 #
 # Gated on OPENAI_API_KEY because the REST provider authenticates via env_key;
 # OAuth/ChatGPT logins (auth.json) use a different transport and are left alone.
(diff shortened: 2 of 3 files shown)


darvell-thenvoi (approved):

yhan1-godaddy:
I think this force you to use `OPENAI_API_KEY` defined in the env. It does not use `OPENAI_API_KEY` defined in `~/.codex/auth.json`

amit-gazal-band:
@yhan1-godaddy 
Good catch - you're right that the REST provider authenticates from the `OPENAI_API_KEY` env var (`env_key = "OPENAI_API_KEY"`), and the whole override is gated on that env var being set, so it won't pick up a key that lives only in `~/.codex/auth.json`.

That's intentional for this fix: it's scoped to env-key auth in the Docker entrypoint, which is the normal way the containerized agent is authenticated. When `OPENAI_API_KEY` isn't in the env, `force_rest_responses_api` returns early and the config is left untouched — so nothing regresses for `auth.json`/OAuth setups, they just keep Codex's default transport.

The one case it doesn't help is an API key provided *only* via a mounted `auth.json` with no env var (and without `CODEX_USE_SOURCE_HOME=true`) — that path would still hit the WS-401. If that's the setup you're running, we'd happily take a PR to bridge `auth.json` → env for API-key auth. Contributions welcome!

yhan1-godaddy:
Thanks for the explanation! Awesome fix btw. :)
So my use case (litellm) is just to use `~/.codex/auth.json` to store  `OPENAI_API_KEY` instead of putting it into the env. I stopped using env var once I learned this work.

Now `OPENAI_API_KEY ` env var is just used for agents I build in project folders, which gets injected using .env files.

PR #336, opened 2026-06-01
fix(sdk): replace sunset Codex models [INT-799]

darvell-thenvoi:
## Summary
- Swaps the SDK off the sunset `gpt-5.2` and `gpt-5.3-codex` model IDs and moves those references to `gpt-5.5`.
- Makes Codex auto-selection respect our configured model order first, so `gpt-5.5` wins when it is visible and `gpt-5.4-mini` is the next fallback.
- Keeps the docs, examples, Docker default, and model-related tests in sync with that behavior.

## Test plan
- `uv run ruff check .`
- `uv run pyrefly check`
- `uv run --extra dev python -m pytest tests/adapters/test_codex_adapter.py tests/integrations/codex/test_adapter_e2e.py tests/adapters/test_opencode_adapter.py tests/example_agents/twenty_questions_arena/test_arena_prompts.py -q`
- `git diff --check`
- `rg -i "gpt[- ]5\\.2|gpt[- ]5\\.3[- ]codex|gpt-52|gpt-53|gpt-5\\.5-spark" .` returned no matches

I also tried the broad `tests/` run with integration/e2e ignores, but collection still picked up a live platform fixture in `tests/integration/test_a2a_gateway_context.py` and failed on `httpx.ConnectError: All connection attempts failed`. The focused model-related suite above passes.

### DIFF
--- a/docker/codex/entrypoint.sh
+++ b/docker/codex/entrypoint.sh
@@ -152,7 +152,7 @@ bootstrap_codex_home() {
     ' "${source_config}" > "${runtime_config}"
   else
     cat > "${runtime_config}" <<EOF
-model = "${CODEX_MODEL:-gpt-5.3-codex}"
+model = "${CODEX_MODEL:-gpt-5.5}"
 approval_policy = "never"
 sandbox_mode = "danger-full-access"
 EOF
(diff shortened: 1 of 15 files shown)


nir-singher-band (commented):
@darvell-thenvoi I'm not sure having fallback models at all is the right design, as we discussed the agent builder will either be using a default or hard-code theirs, in either case you don't want to replace their model even if the agent failing is the alternative

nir-singher-band (approved):

PR #341, opened 2026-06-03
fix(sdk): use ephemeral ports for Parlant servers

darvell-thenvoi:
## What changed

All five Parlant examples (`01_basic_agent.py`–`05_jerry_agent.py`) now start their in-process servers on ports reserved from the OS instead of Parlant's fixed defaults:

```python
ports = reserve_server_ports()
async with p.Server(
    port=ports.port,
    tool_service_port=ports.tool_service_port,
    nlp_service=p.NLPServices.openai,
) as server:
    ...
```

`reserve_server_ports()` is new, in `src/band/integrations/parlant/ports.py`. The baseline E2E toolkit already had this logic privately as `parlant_server.py::_reserve_two_ports`; it's lifted into the SDK so the examples (which can't import from `tests/`) and the toolkit share one definition, and the toolkit now imports it.

The README gains a "Running two agents locally (Tom and Jerry)" section.

## Why

Parlant's integrated tool service binds a fixed default port (`8818`) during `Server.__aenter__`. Starting a second Parlant agent on the same machine fails while the first holds it — running Tom and Jerry side by side, the second dies with `SystemExit: 1` out of uvicorn startup.

**Why not `port=0`?** Parlant doesn't support it: it string-formats the number it's given into URLs rather than reading it back off the bound socket. At the pinned v3.3.2:

- `tool_service_port=0` — the built-in tool service registers as [`url=f"http://{host}:{port}"`](https://github.com/emcie-co/parlant/blob/v3.3.2/src/parlant/sdk.py#L5560) → `http://127.0.0.1:0`, and [`kind == "sdk"`](https://github.com/emcie-co/parlant/blob/v3.3.2/src/parlant/core/services/tools/service_registry.py#L334) resolves to an HTTP `PluginClient`. Every `@p.tool` is [routed there](https://github.com/emcie-co/parlant/blob/v3.3.2/src/parlant/sdk.py#L304), so tool calls dial port 0 — including `band_send_message`, i.e. the agent can't reply at all.
- `port=0` — the readiness poll targets [`http://localhost:0/healthz`](https://github.com/emcie-co/parlant/blob/v3.3.2/src/parlant/sdk.py#L4232) and swallows `ConnectError` in a 1s retry loop, so `ready` never sets and [`finally: await health_check_task`](https://github.com/emcie-co/parlant/blob/v3.3.2/src/parlant/sdk.py#L4225) never returns. Ctrl-C hangs.

Reserving instead: bind two sockets at once (so the OS can't return the same port twice), read the assigned numbers, close them, and hand Parlant real integers to re-bind. The close→rebind gap is the standard ephemeral-port reservation race — far smaller than the collision risk of two fixed ports.

## Validation

Static:

```sh
uv run ruff check . && uv run ruff format --check .
uv run pyrefly check
uv run pytest tests/ --ignore=tests/integration/ --ignore=tests/e2e/   # 4071 passed, 81 skipped
```

Behavioural — asserted against Parlant's own `ServiceRegistry` that the reserved port reaches the registered URL:

```
RESERVED   -> port=49751 tool_service_port=49752
REGISTERED -> http://127.0.0.1:49752          # not :0
CONNECTED  -> tool service is live on 49752
GET http://localhost:49751/healthz "HTTP/1.1 200 OK"
```

That health check is the exact call `port=0` left retrying forever; shutdown returned cleanly rather than hanging.

End-to-end — Tom and Jerry run concurrently against the live platform, on distinct pairs (`49791/49792` and `49801/49802`), holding a real conversation. 92 and 80 tool-path sends respectively, zero connection errors:

> **Jerry:** Quadruple-layer soufflé, @tom? Now you're speaking my language…
> **Tom:** Keep those whiskers twitching, @jerry—

`band_send_message` over HTTP to the plugin server is precisely what `tool_service_port=0` would have broken.

## Notes

Rebased onto current `main`, squashed to one commit.

Earlier revisions of this branch carried a large Parlant rewrite plus generic-runner/docs changes; those either landed on `main` independently or were superseded, so this is scoped to the port fix alone.

**The `opentelemetry-resourcedetector-gcp` commit has been dropped.** It's no longer needed on two counts: `main` already declares that dependency (landed independently after this branch's base), and upstream has since published a stable `1.14.0` after the yanked `1.13.0`, so a fresh lockfile-free resolve of `[dev]`/`[google_adk]` succeeds without any declaration. Verified by building a wheel with the declaration removed and dry-running `[dev]` in a clean venv — resolves to stable `1.14.0`. The `packaging` CI job passes on this branch, which carries neither fix.

Closes INT-530


examples/parlant/01_basic_agent.py
```diff
@@ -124,7 +124,11 @@ async def main() -> None:
     if not rest_url:
         raise ValueError("BAND_REST_URL environment variable is required")
     # Start Parlant server with OpenAI (requires OPENAI_API_KEY env var)
-    async with p.Server(nlp_service=p.NLPServices.openai) as server:
+    async with p.Server(
+        port=0,
+        tool_service_port=0,
```
AlexanderZ-Band: `port=0` isn't supported by Parlant — it string-formats the port into URLs and never reads it back from the bound socket (all refs at our pinned v3.3.2):

- `tool_service_port=0`: the built-in tool service registers as [`url=f"http://{host}:{port}"`](https://github.com/emcie-co/parlant/blob/v3.3.2/src/parlant/sdk.py#L5560) → `http://127.0.0.1:0`, and [`kind == "sdk"`](https://github.com/emcie-co/parlant/blob/v3.3.2/src/parlant/core/services/tools/service_registry.py#L334) resolves to an HTTP `PluginClient`. Every `@p.tool` is [routed to that service](https://github.com/emcie-co/parlant/blob/v3.3.2/src/parlant/sdk.py#L304), so tool calls dial port 0.
- `port=0`: the readiness poll targets [`http://localhost:0/healthz`](https://github.com/emcie-co/parlant/blob/v3.3.2/src/parlant/sdk.py#L4232) and swallows `ConnectError` in a 1s retry loop, so `ready` never sets and [`finally: await health_check_task`](https://github.com/emcie-co/parlant/blob/v3.3.2/src/parlant/sdk.py#L4225) never returns — Ctrl-C hangs.

The mechanism already exists: `tests/e2e/baseline/toolkit/parlant_server.py::_reserve_two_ports` reserves two real ephemeral ports, closes them, and passes the integers — "which Parlant re-binds itself". Suggest lifting it into `src/band/integrations/parlant/` (examples can't import from `tests/`) so examples and the baseline toolkit share one definition, and logging the chosen ports — the fixed 8800 UI is otherwise unreachable.

examples/parlant/04_tom_agent.py
```diff
@@ -52,7 +52,11 @@ async def main() -> None:
         raise ValueError("BAND_REST_URL environment variable is required")
 
     # Load Tom's credentials from agent_config.yaml
-    async with p.Server(nlp_service=p.NLPServices.openai) as server:
+    async with p.Server(
+        port=0,
+        tool_service_port=0,
```
AlexanderZ-Band: Concrete fallout of `tool_service_port=0` here: the guideline below attaches `parlant_tools` and tells Tom to reply with `band_send_message`, but that call goes over HTTP to the built-in tool service registered at `http://127.0.0.1:0`. Tom can't reply at all — worth running Tom+Jerry through one actual reply before merging.

examples/parlant/README.md
```diff
@@ -146,6 +146,20 @@ uv run python examples/parlant/03_support_agent.py
 
 > **Note:** The config loader looks for `agent_config.yaml` in the current working directory. Running from a subdirectory will cause a `FileNotFoundError`.
 
+### Running two agents locally (Tom and Jerry)
+
+Each Parlant agent starts its own in-process server. The examples pass
+`port=0` and `tool_service_port=0` so the OS assigns free ports, which lets
```
AlexanderZ-Band: This documents behaviour that doesn't hold: with `tool_service_port=0` the two-agent run starts fine and then dies at the first tool call (see the note on `01_basic_agent.py`). Needs updating once the ports come from a real reservation.

pyproject.toml
```diff
@@ -137,6 +137,12 @@ agentcore_runtime = [
 ]
 google_adk = [
     "google-adk>=1.0.0,<2",
+    # google-adk pulls opentelemetry-exporter-gcp-trace, which only depends on
+    # pre-release (alpha) builds of opentelemetry-resourcedetector-gcp. A plain
+    # install excludes pre-releases by default and the one stable build (1.13.0)
+    # was yanked upstream, leaving no solution — so request it explicitly with a
+    # pre-release specifier to allow the alpha.
+    "opentelemetry-resourcedetector-gcp>=1.11.0a0",
```
AlexanderZ-Band: Two things on this line.

**The floor is inert** — only the `a0` suffix does work (it satisfies uv's pre-release opt-in). Verified with `uv pip install --dry-run "<wheel>[extra]"` against a probe package:

```
>=1.0a0      -> resourcedetector-gcp==1.12.0a0
>=1.5.0a0    -> resourcedetector-gcp==1.12.0a0
>=1.11.0a0   -> resourcedetector-gcp==1.12.0a0
```

`1.11.0a0` is tighter than anything upstream asks (`opentelemetry-exporter-gcp-trace` wants `==1.*,>=1.5.0dev0`; google-adk wants `>=1.9.0a0,<2`), so it buys nothing and adds a future conflict surface. Prefer mirroring upstream: `>=1.5.0a0`.

**Don't let anyone "fix" this by excluding the yank instead.** With `[tool.uv] constraint-dependencies = ["opentelemetry-resourcedetector-gcp!=1.13.0"]`, uv silently backtracks `opentelemetry-exporter-gcp-trace` to 1.4.0 rather than erroring — a wrong resolution with no signal. Declaring it here, as you did, is the only fix that travels with the published package.

**The comment above needs rewording** — it points the next reader at the wrong resolver. pip already resolves this: `pip install --dry-run "google-adk>=1.0.0,<2"` picks `opentelemetry-resourcedetector-gcp-1.12.0a0`, because [PEP 440 permits pre-releases when the specifier itself carries one](https://peps.python.org/pep-0440/#handling-of-pre-releases) and google-adk requests `>=1.9.0a0`. The failure is uv-specific: [uv accepts pre-releases only for a *direct* dependency with a pre-release specifier, or when *all* published versions are pre-releases](https://docs.astral.sh/uv/concepts/resolution/#pre-release-handling) — and the [yanked stable 1.13.0](https://pypi.org/project/opentelemetry-resourcedetector-gcp/1.13.0/) ("breaks imports") defeats that second rule.

AlexanderZ-Band (changes requested):
Dep commit looks right (one wording fix inline). The Parlant half needs rework: `port=0` / `tool_service_port=0` aren't supported by Parlant — it formats the port into URLs instead of reading the bound socket, so tool calls dial port 0 and shutdown hangs. Details inline. The dep commit is fine on its own — worth splitting it out so it can land now and unblock the packaging job.

src/band/integrations/parlant/ports.py
```diff
@@ -0,0 +1,73 @@
...
+    Logged so that a refused connection or a stuck listener can be traced back to a
+    known pair — with the ports varying per run, there is otherwise nothing to match
+    against ``lsof`` output or a Parlant error.
+    """
+    with (
+        socket.socket(socket.AF_INET, socket.SOCK_STREAM) as api,
+        socket.socket(socket.AF_INET, socket.SOCK_STREAM) as tool_service,
+    ):
+        # Both are held open at once so the OS cannot hand back the same port twice.
+        api.bind(("127.0.0.1", 0))
```
AlexanderZ-Band: Reserve the API port on the host Parlant will actually bind, and keep the tool service on loopback:

```suggestion
def reserve_server_ports(host: str = "") -> ServerPorts:
    """Reserve two distinct free ports and log them.

    ``host`` is where Parlant's API server will bind — the ``host`` given to
    ``p.Server``, all interfaces (``""``) by default. That port is reserved on the
    same host, because a wildcard bind loses to a port already held on *any*
    interface and a loopback-only reservation would not have seen it. The tool
    service is always reserved on ``127.0.0.1``: Parlant hardcodes that host for it.

    Reserving on all interfaces costs nothing in exposure — the socket is never
    ``listen()``ed, so it accepts no connections; a SYN gets an RST.

    Logged so that a refused connection or a stuck listener can be traced back to a
    known pair — with the ports varying per run, there is otherwise nothing to match
    against ``lsof`` output or a Parlant error.
    """
    with (
        socket.socket(socket.AF_INET, socket.SOCK_STREAM) as api,
        socket.socket(socket.AF_INET, socket.SOCK_STREAM) as tool_service,
    ):
        # Both are held open at once so the OS cannot hand back the same port twice.
        api.bind((host, 0))
```

tests/integrations/parlant/test_ports.py
```diff
@@ -0,0 +1,27 @@
...
+def test_reserved_ports_are_free_for_the_caller_to_bind() -> None:
+    """The reservation is released — Parlant itself re-binds the numbers."""
+    ports = reserve_server_ports()
+
+    with contextlib.ExitStack() as stack:
+        for port in ports:
+            bound = stack.enter_context(
+                socket.socket(socket.AF_INET, socket.SOCK_STREAM)
+            )
+            bound.bind(("127.0.0.1", port))
```
AlexanderZ-Band, line 31: This one passes on "`bind` didn't raise", so it reads unfinished — and it's inherently racy (a released ephemeral port can be taken before the rebind), which is worth stating so a future flake isn't mistaken for a real regression:

```suggestion
    # Racy in principle: a released ephemeral port can be taken before the
    # rebind. That gap is the reservation's known, accepted trade-off.
    with contextlib.ExitStack() as stack:
        for port in ports:
            bound = stack.enter_context(
                socket.socket(socket.AF_INET, socket.SOCK_STREAM)
            )
            bound.bind(("127.0.0.1", port))
            assert bound.getsockname()[1] == port, (
                f"reserved port {port} should be bindable by the caller"
            )
```

AlexanderZ-Band (commented):
Approach is right, and the `port=0` analysis checks out against the pinned parlant (`sdk.py:4232` formats `self.port` into the healthz URL; `:5560` formats `tool_service_port` into the registered tool URL; the health-poll task is created in `__aexit__`, so the API port really is unbound inside the `async with` body).

One substantive thing plus two doc nits.

**API port is reserved on the wrong host.** Parlant binds the tool service on a hardcoded `127.0.0.1` (`sdk.py:5540`), so a loopback reservation is exactly right there. But the API server binds `self.host` — `"0.0.0.0"` by default — and a loopback reservation does not imply that port is free for a wildcard bind:

| holder | later bind | result |
|---|---|---|
| `0.0.0.0:P` | `127.0.0.1:P` | EADDRINUSE |
| `192.168.1.5:P` | `127.0.0.1:P` | succeeds |
| `192.168.1.5:P` | `0.0.0.0:P` | EADDRINUSE |

So a port held on some non-loopback interface passes the reservation and then fails Parlant's bind. The toolkit's old `_reserve_two_ports` used `bind(("", 0))`, which asks the OS for a port free on every interface — the guarantee the API server actually needs. This matters most for the one consumer being rewired: `running_parlant_server`'s teardown waits on `server.ready`, which only fires once uvicorn binds `0.0.0.0:port` and answers `/healthz`, so the docstring's "in an agent that stays inside the `async with` body the API port is never bound at all" is true for the examples but not for the toolkit.

Also, the reason given for declining a wildcard reservation doesn't hold: the socket is never `listen()`ed, so it accepts nothing — a SYN gets an RST. There's no network exposure to trade away. Suggestion inline.

**Two things a suggestion can't reach:**

- `tests/e2e/vscode/server.py:36` still says it mirrors `parlant_server._reserve_two_ports`, which this PR deletes — repoint it at `band.integrations.parlant.ports.reserve_server_ports`.
- `src/band/integrations/parlant/__init__.py` now *says* to pass `reserve_server_ports()`, but the snippet below it (line 13) is still bare `async with p.Server() as server:` — same for the untouched `src/band/adapters/parlant.py:45` docstring and `examples/parlant/README.md:38`. Updating the snippets beats prefacing them with a cross-reference.

AlexanderZ-Band (dismissed):
Approving — the fix is correct and the `port=0`-doesn't-work analysis holds up against the pinned parlant. Nothing here blocks merge.

The suggestions in my previous review still stand, in priority order:

1. **API port reservation host** (`ports.py`) — reserve it on the host `p.Server` will bind (wildcard by default), not loopback; a port held on a non-loopback interface passes a loopback reservation and then fails Parlant's `0.0.0.0` bind. Matters for `running_parlant_server`, whose teardown depends on that bind succeeding. Click-to-commit suggestion is inline; I checked the patched file is `ruff format`/`ruff check` clean and that a wildcard rebind of the reserved port works.
2. **Stale ref** — `tests/e2e/vscode/server.py:36` still points at `parlant_server._reserve_two_ports`, deleted here.
3. **Snippets** — `integrations/parlant/__init__.py:13`, `adapters/parlant.py:45` and `examples/parlant/README.md:38` still show bare `p.Server()` while the surrounding prose says to pass reserved ports.
4. **Test assertion** (`test_ports.py`) — second test passes on "`bind` didn't raise"; suggestion inline.

1 is the only one I'd want before merge; 2–4 are fine as a follow-up.

PR #342, opened 2026-06-07
feat: Slack adapter for remote Band agents [INT-461]

amit-gazal-band:
## Summary

Adds a first-class **Slack adapter** so a remote Band agent can be exposed as a Slack-native AI App. Implements Phase 1 (MVP) of [INT-461](https://linear.app/thenvoi/issue/INT-461).

This **supersedes #331** — same feature, rebased onto the Band rebrand on `dev` and with all of that PR's review feedback incorporated. #331 can be closed in favour of this branch.

**Architecture — wrapping, not gateway.** `SlackAdapter` wraps an `inner: SimpleAdapter` (the brain — Anthropic, LangGraph, etc.) and decorates `on_message`. One Band service identity, one process.

- Two outbound tools: `band_send_message` (real peer message) and `slack_send_message` (plain reply to the bound Slack thread).
- Thread context backfill: refetches `conversations.replies` per turn (no caching) so the brain sees full thread history.
- Session rehydration from the bootstrap `task` event binding.
- HTTP (Starlette/FastAPI) **and** Socket Mode transports.

## Review fixes carried over from #331

| Area | Fix |
|------|-----|
| Room routing | Lookup key now includes `app.slug` — two SlackApps sharing a `channel:thread_ts` tuple map to distinct rooms; no cross-workspace reply routing |
| Thread history | Own `bot_id` resolved via `auth.test`; only this bridge's plain-text replies map to assistant history. Other bots/webhooks are kept as external context with a `slack-bot:<id>` sender |
| Room creation race | Per-thread `asyncio.Lock` + re-check collapses concurrent events for one thread onto a single room (no duplicate/orphaned rooms) |
| Packaging | `aiohttp` added to the `[slack]` extra (Socket Mode imports it) |
| Block Kit | `render_plan_blocks` capped at Slack's 50-block limit with an overflow summary |
| Docs | Package docstring example corrected to the real `inner=...` wrapping shape |

Each fix has a regression test (multi-app isolation, foreign-bot, two-task concurrency, 49-task block cap, etc.).

## Supporting SDK change

`AgentTools.execute_tool_call_structured` + `ToolCallOutcome` give the Slack progress UI a machine-readable `ok` flag instead of sniffing error-string prefixes. `execute_tool_call` now delegates to it (behaviour-preserving).

## What's included

- HMAC signature verification + URL verification, retry idempotency
- Per-thread Band room mapping, peer reply → Slack threading
- Status indicators (`assistant.threads.setStatus`) + Block Kit `plan`/`task` tool-call visibility
- Slack thread history backfill (`channels:history` / `groups:history`)
- Session rehydration on bootstrap; Slack context mirroring into the Band room
- Framework mount helpers + Socket Mode transport
- Manifest template, getting-started example (`examples/slack/01_basic_bot.py`), README section
- `slack` adapter + converter excluded from framework conformance (transport bridge / metadata-only converter)

## Test plan

- [x] `uv run pytest tests/integrations/slack/ -v` — 140 passed
- [x] `uv run pytest tests/framework_conformance/ tests/framework_configs/` — 338 passed
- [x] `uv run ruff check . && uv run ruff format . --check`
- [ ] Manual: restart-bridge backfill/rehydration smoke (`examples/slack/_dev_bridge.py`)

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,11 +1,11 @@
 # Band Python SDK
 
 <p align="center">
-  <img src="https://raw.githubusercontent.com/band-ai/band-sdk-python/main/assets/band-readme-banner.png" alt="Band" width="100%">
+  <img src="https://raw.githubusercontent.com/thenvoi/thenvoi-sdk-python/main/assets/band-readme-banner.png" alt="Band" width="100%">
 </p>
 
 <div align="center">
-  <a href="https://github.com/band-ai/band-sdk-python/actions"><img src="https://img.shields.io/badge/CI-passing-brightgreen" alt="CI"></a>
+  <a href="https://github.com/thenvoi/thenvoi-sdk-python/actions"><img src="https://img.shields.io/badge/CI-passing-brightgreen" alt="CI"></a>
   <a href="https://docs.band.ai"><img src="https://img.shields.io/badge/docs-band.ai-blue" alt="Docs"></a>
   <a href="https://discord.gg/gvMYpB9eAY"><img src="https://img.shields.io/badge/Discord-join%20chat-5865F2?logo=discord&logoColor=white" alt="Discord"></a>
   <a href="https://pypi.org/project/band-sdk/"><img src="https://img.shields.io/pypi/v/band-sdk.svg" alt="PyPI version"></a>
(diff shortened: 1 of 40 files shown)


darvell-thenvoi (approved):

darvell-thenvoi:
Approved, probably change what git repo the example points to but we should be moving to that repo soon.

PR #344, opened 2026-06-08
build(deps): unvendor Fern REST client and testing package

darvell-thenvoi:
## Summary

Replaces the vendored `band_rest` copy of `thenvoi_rest` and the vendored `band_testing` fixtures with the published packages, so the SDK consumes them as ordinary dependencies instead of shipping copies in `src/` after an accident was introduced in renaming from Thenvoi to Band.

- `thenvoi-client-rest==0.0.7` (imported as `thenvoi_rest`)
- `thenvoi-testing-python==0.1.4` (imported as `thenvoi_testing`)

## Changes

- Delete `src/band_rest/` (the alias shim) and the vendored `src/band_testing/` tree.
- Add `thenvoi-client-rest==0.0.7` to runtime deps; drop the now-transitively-provided `httpx`/`pydantic`/`pydantic-core`/`typing-extensions` pins from direct deps.
- Add `thenvoi-testing-python==0.1.4` to the `dev` and `dev-crewai` extras.
- Narrow `[tool.setuptools.packages.find]` include to `band`/`band.*` and drop the `src/band_rest/**` pyrefly exclude.
- Repoint all `band_rest` imports in `src/band` and tests to `thenvoi_rest`, and `band_testing` to `thenvoi_testing`.
- Update the two examples that imported `band_rest` to import `thenvoi_rest`; remove the obsolete repo-root `sys.path` hack in `examples/claude_sdk_docker/create_agents.py` (the client is an installed dependency now).

## Why

A mistake was left in the rename PR, this resolves it.

## Testing

- [x] Unit tests pass: `2892 passed, 222 skipped` (`uv run pytest tests/ --ignore=tests/integration --ignore=tests/e2e`). No import errors from the unvendoring.
- [x] Pre-commit checks pass (secret scan, `ruff check`, `ruff format`, `pyrefly`, commitizen) — ran on commit.
- [x] `uv run ruff check .` and `uv run ruff format --check .` clean (451 files).
- [ ] Integration/E2E require live platform credentials; the integration errors seen in a combined run are pre-existing credential/collection-isolation skips (each integration file skips cleanly when run on its own), not caused by this change.

## Checklist

- [x] Tests updated to import the published packages
- [ ] Documentation updated as needed


amit-gazal-band (changes requested):
Please `remove` the band directory and put its content directly under `src`

amit-gazal-band (approved):

PR #362, opened 2026-06-15
feat(sdk): adapters report boolean activity (working) state to the platform [INT-831]

amit-gazal-band:
## Summary

Implements **INT-831** — every adapter automatically reports a **boolean working/not-working** keep-alive signal to the platform for each remote-agent reasoning cycle (per agent + room), with **no agent-business-logic changes and no per-adapter code**. Wired once in the runtime so all 16 adapters get it for free.

- `working: true` when a reasoning cycle starts
- refreshed on a keep-alive cadence (~3s) while the cycle is active
- `working: false` immediately when the cycle ends (complete / error / interrupt)
- platform TTL (≤10s) is the backstop for abnormal ends (crash / hang / disconnect)

Boolean-only payload — no status enum, tool name, `detail`, or free text.

## Design

`ExecutionContext` brackets the adapter invocation (`_on_execute`) for **message-driven cycles only** via `_execute_message_cycle` (`start()` → try → `finally stop()`):

- **WS path** (`_process_event`): wrapped only for `MessageEvent`; participant add/remove and `ReconnectedEvent` are excluded (housekeeping, not reasoning).
- **Backlog `/next` path** (`_process_backlog_message`): always wrapped.
- **Hub room excluded** via the reporter's `enabled` flag (`room_id == hub_room_id`) — it's an internal contact-triage room with no peer-facing "Reasoning…" surface (so synthetic contact messages, which only occur there, are excluded as a consequence).

Components:
- **`BandLink.report_activity(room_id, working)`** — uses the `band_rest` generated `agent_api_activity.report_agent_chat_activity`; per-POST timeout (2s, `max_retries=0`); all failures swallowed → `False` (never raises into the loop); repeated failures debounced to one warning + a recovery info.
- **`WorkingStateReporter`** (`src/band/runtime/working_state.py`) — per room: immediate `true` + keep-alive refresh; **`asyncio.Lock` serializes all sends** so they can't overlap on the wire; **shielded `false`-send** so it lands even mid-cancellation; optional `max_working_state_seconds` cap pauses pings (TTL clears) without ever cancelling reasoning; never raises.
- **`SessionConfig`** knobs: `enable_working_state` (kill-switch, default `True`), `working_keep_alive_seconds` (3), `working_request_timeout_seconds` (2), `max_working_state_seconds` (None), with a `__post_init__` guard (cadence>0, cadence<TTL/2, timeout<cadence, max>0).

## Dependency note

This branch sits on the **`band-client-rest 0.0.10` (`band_rest`) migration** (merged from `chore/bump-band-client-rest-0.0.10`), which ships the activity endpoint this feature reports against. If that bump lands on `dev` independently, this PR's diff shrinks to just the INT-831 changes.

**Rejected alternative (paper trail):** a standalone `httpx.AsyncClient` in `BandLink` — that was the workaround for the old `thenvoi-client-rest==0.0.7` gap (no activity namespace). Superseded by the generated method; the real-client guard test fails loudly if that namespace ever regresses, at which point the fallback is the documented fix.

## Failure guarantee (honest mapping)

Keep-alive is a **liveness** signal. Normal completion / exception / interrupt → immediate `false` (finally). Loop-block / process-death → TTL clears. A **hung-but-connected live await** is not ≤10s-achievable from the SDK alone (requires framework introspection, out of scope) — mitigated by requiring adapter request timeouts; documented as a limitation for the platform side (PLT-945).

## Tests (TDD; each step reviewed to consensus with an architect)

- `tests/platform/test_link.py::TestReportActivity` — real-client method guard, true/false args, error swallowing (404/401/422/network), failure-log debounce, per-POST timeout.
- `tests/runtime/test_working_state.py` — immediate true, keep-alive cadence, **mid-cycle-cancel asserts false** (non-skippable gate), **true-before-false ordering** (non-skippable gate), failure swallowed, no duplicate keep-alive, cap pause, disabled, validation.
- `tests/runtime/test_execution_working_state.py` — true/false around `MessageEvent`, false on exception, participant/`ReconnectedEvent` excluded, hub-room excluded, backlog path, room isolation, config guards.

Touched-area unit suite: **740 passed**; ruff + pyrefly clean.

## Follow-ups (not in this PR)

- Gated E2E observing the real `agent_executions:{room_id}` broadcast flip (the one test proving the live platform contract).
- Optional tighter UX: drop the indicator at reply-send (and/or a declared per-adapter "reasoning span") so it clears the instant the reply lands rather than after the model's trailing wrap-up turn — discussed, deferred.

### DIFF
--- a/src/band/platform/link.py
+++ b/src/band/platform/link.py
@@ -102,6 +102,11 @@ class BandLink:
         # Durable terminal disconnect reason for the current connection lifecycle.
         self._last_disconnect_reason: WebSocketDisconnectReason | None = None
 
+        # Debounce flag for activity-report failures: keep-alive runs at a few
+        # seconds per room, so a down endpoint would otherwise flood the log on
+        # every refresh. Log the first failure and the recovery, suppress repeats.
+        self._activity_report_failing = False
+
     @property
     def is_connected(self) -> bool:
         return self._is_connected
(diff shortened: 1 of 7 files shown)


amit-gazal-band:
https://github.com/user-attachments/assets/84209d7f-18d6-4e22-a705-754e1fbc96ce

PR #365, opened 2026-06-16
fix: require mentions and surface room handles on send failure

AlexanderZ-Band:
## Summary

When a `band_send_message` call fails mention validation, the agent had nothing useful to retry with — the error didn't tell it which handles actually exist in the room. This PR makes failed send-message calls return the room's real participant handles (excluding the agent itself) across **all** integrations, so the LLM can self-correct instead of repeating the schema placeholder.

It also makes the CrewAI `send_message` schema require a non-empty `mentions` value, and stops the prompts from teaching the model to send empty mentions (which directly contradicts the "≥1 mention required" rule).

Self is identified via a new `agent_id` accessor threaded from `ExecutionContext` through `AgentTools` — an agent cannot @mention itself, so listing its own handle only misleads the LLM.

## Changes

- **Shared helpers** (`runtime/tools.py`): consolidate the duplicated "available mention handles" logic into `available_mention_handles`, `available_mention_participant_labels`, and the error enrichers `append_mention_handles_hint` / `append_available_mention_handles` (single source for the hint format string; idempotent and null-safe).
- **CrewAI** (`integrations/crewai/tools.py`): require non-empty `mentions`; drop the local handle helpers and reuse the shared enricher on send failures.
- **Claude SDK** (`integrations/claude_sdk/tools.py`, `prompts.py`): wire `agent_id` into `AgentTools` so it filters self like the other adapters; reuse the shared hint formatter; remove prompt examples that instructed empty `mentions: []`.
- **Parlant** (`integrations/parlant/tools.py`): enrich both the empty-mentions and send-failure errors with available handles.
- **LangGraph** (`integrations/langgraph/langchain_tools.py`): surface `BandToolError` / `ValueError` messages (which already carry the handle hint) instead of a generic error.
- **MCP server** (`runtime/mcp_server.py`): enrich `band_send_message` validation errors with available handles on both the direct and resolved registration paths.
- **Runtime** (`runtime/execution.py`, `runtime/tools.py`): expose `agent_id` and use it to exclude self from the empty-mention error and unknown-handle error.

## Related Issues

Relates to INT-453

## Testing

- [x] Unit tests pass — added/updated coverage for the shared helper (self/null-handle exclusion), the unknown-handle error, and the per-integration enrichment paths (crewai, parlant, langgraph, mcp_server). Full suite green (892 passed).
- [x] Pre-commit checks pass (`ruff check`, `ruff format`, `pyrefly check`)
- [x] Integration tests pass (if applicable)

## Checklist

- [x] Documentation updated as needed

### DIFF
--- a/src/band/adapters/letta.py
+++ b/src/band/adapters/letta.py
@@ -49,7 +49,7 @@ EVERY response MUST include at least one tool call. Specifically:
 Just responding with plain text like this.
 
 ## CORRECT:
-Call band_send_message(content="Your response", mentions=["@user-id"])
+Call band_send_message(content="Your response", mentions=["@username"])
 
 If you respond without calling `band_send_message`, the user sees NOTHING.
 
--- a/src/band/adapters/letta.py
+++ b/src/band/adapters/letta.py
@@ -49,7 +49,7 @@ EVERY response MUST include at least one tool call. Specifically:
 Just responding with plain text like this.
 
 ## CORRECT:
-Call band_send_message(content="Your response", mentions=["@username"])
+Call band_send_message(content="Your response", mentions=["@user-id"])
 
 If you respond without calling `band_send_message`, the user sees NOTHING.
 
--- a/src/band/runtime/tools.py
+++ b/src/band/runtime/tools.py
@@ -75,6 +75,55 @@ def _matches_identifier(entity: dict[str, Any] | Any, identifier: str) -> bool:
     return False
 
 
+def available_mention_handles(
+    participants: list[dict[str, Any] | Any],
+    agent_id: str | None = None,
+) -> list[str]:
+    """Return room handles this agent may mention, excluding itself."""
+    return [
+        handle
+        for participant in participants
+        if (handle := _entity_field(participant, "handle"))
+        and (agent_id is None or _entity_field(participant, "id") != agent_id)
+    ]
+
+
+def available_mention_participant_labels(
+    participants: list[dict[str, Any] | Any],
+    agent_id: str | None = None,
+) -> list[str]:
+    """Return mentionable participant handles, falling back to names."""
+    return [
+        label
+        for participant in participants
+        if (agent_id is None or _entity_field(participant, "id") != agent_id)
+        and (
+            label := _entity_field(participant, "handle")
+            or _entity_field(participant, "name")
+        )
+    ]
+
+
+def append_mention_handles_hint(error: str, handles: list[str]) -> str:
+    """Append a retryable handles hint to a tool error when handles are known."""
+    if not handles or "Available handles:" in error:
+        return error
+    return (
+        f"{error}. Available handles: {handles}. Use participant handles from the list."
+    )
+
+
+def append_available_mention_handles(
+    error: str,
+    participants: list[dict[str, Any] | Any],
+    agent_id: str | None = None,
+) -> str:
+    """Append retryable mention handles to a tool error when available."""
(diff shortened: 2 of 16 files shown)


src/band/adapters/letta.py
```diff
@@ -49,7 +49,7 @@
 Just responding with plain text like this.
 
 ## CORRECT:
-Call band_send_message(content="Your response", mentions=["@user-id"])
+Call band_send_message(content="Your response", mentions=["@username"])
```
AlexanderZ-Band:
```suggestion
Call band_send_message(content="Your response", mentions=["@user-id"])
```

src/band/runtime/tools.py
```diff
@@ -75,6 +75,39 @@ def _matches_identifier(entity: dict[str, Any] | Any, identifier: str) -> bool:
...
+    ]
+
+
+def append_mention_handles_hint(error: str, handles: list[str]) -> str:
+    """Append a retryable handles hint to a tool error when handles are known."""
+    if not handles or "Available handles:" in error:
+        return error
+    return (
+        f"{error}. Available handles: {handles}. Use participant handles from the list."
+    )
```
amit-gazal-band, line 108: **Possible double-listing of handles on the empty-mentions path (CrewAI + MCP server)**

`append_mention_handles_hint`'s idempotency guard only checks for the literal `"Available handles:"`:

```python
if not handles or "Available handles:" in error:
    return error
```

But `AgentTools.send_message` raises the empty-mentions error with a *different* marker — `"Available participants: {available}."`. So when that `BandToolError` flows through CrewAI's `_exec` enricher and the MCP server's `_enrich_send_message_error`, the guard doesn't match and a second `"Available handles: [...]"` clause gets appended. The LLM then sees the same handle list twice under two labels.

Verified against the branch source on both paths. CrewAI `send_message(mentions="[]")` → `mention_list = []` → `send_message([])` raises the `BandToolError`, which matches `isinstance(e, (ValueError, BandToolError))`, so the enricher runs. Actual output (identical for the MCP resolved path via `registration.execute({"mentions": []})`):

```
At least one mention is required. Available participants: ['@john']. Please retry with
mentions specifying who this message is for.. Available handles: ['@john']. Use
participant handles from the list.
```

Note the doubled list and the `for.. Available handles:` artifact from concatenating onto a message that already ended in a period. (Self is correctly excluded from both lists.)

Parlant is unaffected (it short-circuits the empty case with its own message and enriches once). The current tests miss it because the CrewAI test mocks the error string without the `Available participants:` segment and both tests only assert substring presence.

Suggest unifying on a single marker — e.g. have `send_message` emit `Available handles:` too (ideally a shared constant), or make the guard recognize both phrasings.

amit-gazal-band (changes requested):

src/band/runtime/tools.py
```diff
@@ -1973,7 +2031,9 @@ def _resolve_mentions(
                 participant = id_to_participant.get(identifier)
 
             if not participant:
-                available_handles = list(handle_to_participant.keys())
+                # Offer only real, mentionable handles to retry with: @-prefixed,
+                # excluding self and handle-less participants (not the raw lookup keys).
+                available_handles = self.available_mention_handles()
                 raise ValueError(
                     f"Unknown participant '{identifier}'. "
                     f"Available handles: {available_handles}"
```
amit-gazal-band: Please use `_AVAILABLE_HANDLES_MARKER`

amit-gazal-band (approved):

PR #372, opened 2026-06-19
test(tests): add L2 context baseline

darvell-thenvoi:
## Summary

- Adds the L2 context-fidelity baseline split: full history, earliest turn retention, chronological ordering, and speaker attribution.
- Adds the gated L2 live workflow and row-specific artifact checks.
- Rebases the L2 slice onto L1 and latest `origin/dev`.

## Verification

- `uv run ruff check .` passed.
- `uv run pyrefly check` passed.
- `uv run pytest tests/framework_conformance/ tests/e2e/test_baseline_artifacts.py tests/e2e/test_e2e_hygiene.py -q -k 'not crewai'` passed: 805 passed, 26 skipped, 98 deselected.
- `BAND_ALLOW_MISSING_FRAMEWORKS=1 uv run pytest tests/adapters/test_crewai_adapter.py tests/adapters/test_crewai_adapter_soak.py tests/adapters/test_crewai_flow_adapter.py tests/adapters/test_crewai_flow_phase3.py tests/adapters/test_crewai_flow_phase4.py tests/adapters/test_crewai_flow_phase5.py tests/adapters/test_crewai_flow_state_source.py tests/converters/test_crewai.py tests/converters/test_crewai_flow.py tests/integrations/test_crewai_tools.py tests/integrations/test_crewai_flow_real_sdk.py tests/framework_conformance/ -q -k crewai` passed: 308 passed, 7 skipped, 704 deselected.

Live provider/platform E2E was not run locally; the L2 live workflow is gated by `E2E_TESTS_ENABLED=true` and `E2E_BASELINE_L2_LIVE=true`.

### DIFF
--- a/.github/workflows/baseline-l2-live.yml
+++ b/.github/workflows/baseline-l2-live.yml
@@ -1,0 +1,80 @@
+name: Band baseline L2 live
+
+on:
+  workflow_dispatch:
+  schedule:
+    - cron: '17 9 * * *'
+
+jobs:
+  baseline-l2-live:
+    runs-on: ubuntu-latest
+    timeout-minutes: 90
+    env:
+      E2E_TESTS_ENABLED: "true"
+      E2E_BASELINE_L2_LIVE: "true"
+      E2E_BASELINE_RUN_ID: ${{ github.run_id }}
+      E2E_BASELINE_ARTIFACT_DIR: .claude/reports/e2e-baseline-artifacts
+      THENVOI_BASE_URL: ${{ vars.THENVOI_BASE_URL }}
+      THENVOI_WS_URL: ${{ vars.THENVOI_WS_URL }}
+      TEST_AGENT_ID: ${{ secrets.TEST_AGENT_ID }}
+      THENVOI_API_KEY: ${{ secrets.THENVOI_API_KEY }}
+      THENVOI_API_KEY_USER: ${{ secrets.THENVOI_API_KEY_USER }}
+      OPENAI_API_KEY: ${{ secrets.OPENAI_API_KEY }}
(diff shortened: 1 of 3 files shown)


AlexanderZ-Band:
**Re: bounded-count / scoped-window reply assertions** — `assert len(burst_replies) == 9` (~L238) and `assert len(recall_replies) == 1` (~L261).

These exact-count gates on **agent output** are a poor fit for a live LLM under test and will flake on *correct* behaviour, because reply count is non-deterministic:

- the model may batch several planted facts into one reply, or split one into several;
- a turn may legitimately emit **0** replies — the agent isn't obligated to call `band_send_message`;
- retries / regeneration can add a duplicate.

So `== 9` goes red when the adapter is working fine, and the failure *looks* like a dropped/duplicated turn when it isn't. This is the same reason the baseline toolkit README bans exact-count / strict-ordering / mandatory-silence assertions and ships only tolerant ones (`assert_at_least`, `assert_contains_any`, `assert_contains_none`).

**Tolerant reformulation that still proves the L2 contract (context fidelity):**
- Barrier on the *last* planted message via `wait_for_processed`. Per-room FIFO already proves every earlier turn was processed — that's the real "no dropped turn" guarantee, without counting replies.
- For recall, `assert_contains_any` the load-bearing tokens (`MARCO` ∧ `LIGHTHOUSE` ∧ `POSTGRESQL`) and use `assert_at_least(1)` instead of `== 1`.

The count that *is* safe to assert is on **platform state**, not agent output — e.g. the rehydration `source_message_counts == 1` oracle (~L409) is exactly right: it counts what the platform persisted, not how the model chose to reply. Keep those; drop the reply-count equalities.

PR #378, opened 2026-06-22
feat: raise a visible UserWarning for unsupported adapter feature flags

AlexanderZ-Band:
## Summary

`SimpleAdapter.on_started` previously only logged a `logger.warning` when an adapter received `AdapterFeatures` flags it doesn't implement (e.g. `emit` values on `ParlantAdapter`). That line is easy to miss — in libraries the logger is often unconfigured or set above `WARNING`, so the misconfiguration silently has no effect.

This now **also** emits a non-fatal `UserWarning` via `warnings.warn` for unsupported `emit`/`capability` values, while keeping the existing log line.

### Why both?

- **`logger.warning`** — durable operator trail: goes to the configured log sink with timestamps/context, survives in prod history. But easily missed when logging is unconfigured or above `WARNING`.
- **`warnings.warn(UserWarning)`** — dev/CI signal: fires regardless of logging config, surfaces in dev/test output and IDEs, and can be promoted to an error (`-W error` / `filterwarnings`) to fail CI on misconfiguration. But deduped per call-site and often silenced in prod.

Each covers the other's blind spot, so the misconfiguration is hard to miss at dev time **and** recorded at runtime.

## Changes

- Refactored the warning logic into a `_warn_unsupported(adapter_name, kind, unsupported)` helper that both logs and emits the `UserWarning`.
- Check stays in `on_started` so `SlackAdapter`'s inner-support mirroring still runs first (no spurious warning).

## Testing

- `tests/core/test_simple_adapter_features.py` — asserts `UserWarning` is raised for unsupported emit/capability, and that supported configs raise no warning.
- `tests/integrations/slack/test_wrapping.py` — verifies the wrapper mirrors inner support before the base check (no spurious `UserWarning`).
- Tests + ruff + pyrefly pass.

Closes INT-880

### DIFF
--- a/src/band/core/simple_adapter.py
+++ b/src/band/core/simple_adapter.py
@@ -3,6 +3,7 @@
 from __future__ import annotations
 
 import logging
+import warnings
 from abc import ABC, abstractmethod
 from typing import Any, ClassVar, Generic, TypeVar, cast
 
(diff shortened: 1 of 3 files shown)


AlexanderZ-Band:
Duplicate — superseded by #379, which targets the correct base branch (dev).

PR #395, opened 2026-06-25
feat: [1] bootstrap E2E testing tools and smoke tests INT-912

AlexanderZ-Band:
## Summary

Adds the first reusable E2E test toolkit for live Band platform tests, plus a small smoke suite to prove it works.

The L0-L4 baseline suite is not included here.

## What's included

- Helpers to create and clean up test agents and rooms.
- Message waiting and drain helpers for stable live tests.
- Tolerant assertions and a simple optional LLM judge.
- Simulated user operations for rooms and participants.
- Shared pytest settings, dependency gates, and fixtures.
- Smoke tests for provisioning, user ops, two-agent chat, and memory recall.

## Testing

- `E2E_TESTS_ENABLED=true uv run pytest tests/e2e/baseline/smoke/ -v`
- `uv run ruff check . && uv run ruff format .`
- `uv run pyrefly check`

## Specification docs

- [Spec parent: INT-522](https://linear.app/thenvoi/issue/INT-522/define-baseline-conformance-suite-specification)
- [Docs PR: product-docs-vault#128](https://github.com/thenvoi/product-docs-vault/pull/128)
- [L0 scenario spec](https://github.com/thenvoi/product-docs-vault/blob/main/Issues/issue-INT-523-l0-test-scenarios-platform-adaptation.md)
- [L1 scenario spec](https://github.com/thenvoi/product-docs-vault/blob/main/Issues/issue-INT-524-l1-test-scenarios-custom-prompt-and-custom-tools.md)
- [L2 scenario spec](https://github.com/thenvoi/product-docs-vault/blob/main/Issues/issue-INT-525-l2-test-scenarios-conversation-context-fidelity.md)
- [L3 scenario spec](https://github.com/thenvoi/product-docs-vault/blob/main/Issues/issue-INT-527-l3-test-scenarios-multi-participant-chat.md)
- [L4 scenario spec](https://github.com/thenvoi/product-docs-vault/blob/main/Issues/issue-INT-528-l4-test-scenarios-rehydration.md)
- [Supporting docs index](https://github.com/thenvoi/product-docs-vault/blob/main/Resources/baseline-conformance/README.md)

## Related

- Linear: [INT-912](https://linear.app/thenvoi/issue/INT-912/bootstrap-e2e-testing-tools-smoke-tests-pre-l0-l4)
- Parent: [INT-911](https://linear.app/thenvoi/issue/INT-911/testing-framework-and-l0-4-implementation)

PR #396, opened 2026-06-25
feat: [2] add tool-call capture and tool_fired assertion INT-913

AlexanderZ-Band:
## Summary

Adds tool-call capture for live E2E tests and a tolerant `tool_fired` assertion.

This builds on the E2E toolkit from #395. It reads persisted `tool_call` events after a turn finishes, so tests can check which tools were called and with what arguments.

## What's included

- `RoomCapture.tool_calls()` for reading tool calls from a room.
- `ToolCalls.assert_fired()` for checking tool name and argument matches.
- `Replies` and `ToolCalls` observation helpers.
- Smoke tests for tool calls, multiple tools, tolerant args, and room/sender/turn isolation.
- Small toolkit refactors around reply waiting and drain behavior.

## Testing

- `E2E_TESTS_ENABLED=true uv run pytest tests/e2e/baseline/ -v -s --no-cov`
- `uv run ruff check . && uv run ruff format .`

## Related

- Linear: [INT-913](https://linear.app/thenvoi/issue/INT-913/e2e-baseline-trajectory-tool-observation-inspection-tool-fired)
- Parent: [INT-911](https://linear.app/thenvoi/issue/INT-911/testing-framework-and-l0-4-implementation)
- Builds on: [INT-912](https://linear.app/thenvoi/issue/INT-912/bootstrap-e2e-testing-tools-smoke-tests-pre-l0-l4)

### DIFF
--- a/tests/e2e/baseline/README.md
+++ b/tests/e2e/baseline/README.md
@@ -13,10 +13,10 @@ quality. They are deterministic by design (no `sleep`, no silence windows).
 |------|------------|
 | `toolkit/provisioning.py` | `ResourceManager` (provision/reap agents + rooms, orphan sweep), `running_provisioned_agent`, `ProvisionedAgent` |
 | `toolkit/user_ops.py` | `UserOps`: act as the test user (send message, create/delete room, add/remove/list participants, list messages/events) |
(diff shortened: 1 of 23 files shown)


AlexanderZ-Band:
[INT-913](https://linear.app/thenvoi/issue/INT-913/2-tool-call-capture-tool-fired-assertion)

amit-gazal-band (approved):
I'm still getting the same error as in #395, but it has nothing to do with this PR. So approved 💯 

Image

PR #397, opened 2026-06-27
feat: [3] add memory and emitted-event observation tools INT-914

AlexanderZ-Band:
## Summary

Adds observation + assertion helpers for emitted events and memory side effects to the live E2E toolkit, mirroring the tool-call capture from #396.

Events (`thought`/`error`/`task`) read back on the same persisted, read-after-`drain` path as tool calls. Memory adds two layers: a call-layer view (which memory tool fired, with what args) plus a store-layer reader (did the record actually land in the backend).

## What's included

- `RoomCapture.thoughts()` / `errors()` / `tasks()` / `events()` and an `Events` base with thin `Thoughts` / `Errors` / `Tasks` subclasses; tolerant `assert_emitted` / `assert_contains_any`.
- `RoomCapture.memory(agent)` returning both layers in one read: `MemoryToolCalls` (`assert_store_called` etc.; memory tools are opted out of the general `tool_calls()` view by default) and a `Memories` store reader (`where` / `assert_stored` / `assert_present` / `assert_none`).
- `MemoryTool` enum validated against the SDK's `MEMORY_TOOL_NAMES`; shared `tolerant_match` and `ContentAssertions` helpers reused across collections; `agent_rest_client` for agent-scoped reads.
- Smoke tests: parametrized adapter (Anthropic/LangGraph) × event-type matrix, plus memory store, subject-scope, supersede lifecycle, and opt-out cases.

## Testing

- `E2E_TESTS_ENABLED=true uv run pytest tests/e2e/baseline/ -v -s --no-cov`
- `uv run ruff check . && uv run ruff format .`

## Related

- Linear: [INT-914](https://linear.app/thenvoi/issue/INT-914/3-memory-emitted-event-observation-tools)
- Parent: [INT-911](https://linear.app/thenvoi/issue/INT-911/testing-framework-and-l0-4-implementation)
- Builds on: [INT-913](https://linear.app/thenvoi/issue/INT-913/e2e-baseline-trajectory-tool-observation-inspection-tool-fired) (#396)

### DIFF
--- a/tests/e2e/baseline/smoke/sample_agents.py
+++ b/tests/e2e/baseline/smoke/sample_agents.py
@@ -1,0 +1,242 @@
+"""Cheap agents and driving instructions for the matrix smokes.
+
+``ADAPTER_BUILDERS`` maps an adapter id to its dependency gate and a builder;
+``adapter_params`` renders it as ``pytest.param``s with the right ``@requires``
+gate, so a matrix test reads ``parametrize("adapter_id", adapter_params())``.
+Only standard tool-loop adapters are here (Anthropic, LangGraph) -- both route
+tool calls through ``execute_tool_call``, so ``band_send_event`` posts a real
+event and ``band_store_memory`` writes a real memory. Adding a framework is a
+single ``ADAPTER_BUILDERS`` entry plus a ``Dep``.
+
+Following ``sample_tools``/``test_tool_calls``, the agent gets a fixed
+role-setting system prompt and the *user message* carries the instruction (with
+the unique marker). Each instruction forces exactly one observable action --
+``band_send_event`` for an event, ``band_store_memory`` for a memory, the only
+way to produce it -- so a precise instruction is the only way to comply.
+"""
+
+from __future__ import annotations
+
+import uuid
+from collections.abc import Callable, Collection
+
+import pytest
+
+from band.adapters.anthropic import AnthropicAdapter
+from band.core.simple_adapter import SimpleAdapter
+from band.core.types import AdapterFeatures, Capability, Emit, MessageType
+from band.core.memory_types import (
+    MemorySegment,
+    MemoryStoreScope,
+    MemorySystem,
+    WorkingLongTermMemoryType,
+)
+
+from tests.e2e.baseline.requires import Dep, requires
+from tests.e2e.baseline.settings import BaselineSettings
+from tests.e2e.baseline.toolkit.observations import MemoryTool
+
+# Fixed role-setter: the actionable instruction (and marker) travels in the user
+# message, exactly like the opaque-tool smokes.
+TOOL_AGENT_SYSTEM_PROMPT = (
+    "You are under test. When the user messages you, do exactly what they ask: "
+    "make the requested tool call(s) with the given arguments and nothing else. "
+    "Do not send a chat message unless explicitly asked."
+)
+
+AgentBuilder = Callable[..., SimpleAdapter]
+
+
+def build_anthropic_agent(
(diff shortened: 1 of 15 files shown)


tests/e2e/baseline/smoke/sample_agents.py
```diff
@@ -0,0 +1,242 @@
...
+        custom_section=TOOL_AGENT_SYSTEM_PROMPT,
+        features=features,
+    )
+
+
+# TODO: this is temporary until we create generic builders in the next PR
+ADAPTER_BUILDERS: dict[str, tuple[Dep, AgentBuilder]] = {
+    "anthropic": (Dep.ANTHROPIC, build_anthropic_agent),
+    "langgraph": (Dep.OPENAI, build_langgraph_agent),
+}
```
amit-gazal-band, line 86: Love this!

tests/e2e/baseline/toolkit/room_capture.py
amit-gazal-band: Every other durable-read method (tool_calls, events, thoughts, errors, tasks, memory) routes through self._require_user_ops(what). drain() at line 354 uses a hand-written if self._user_ops is None: raise RuntimeError(...) block instead. If _require_user_ops is ever updated (error message, logging, sentinel check), drain() silently diverges. One-line fix: replace the block with self._require_user_ops("drain") and use its return value.

tests/e2e/baseline/toolkit/observations/memories.py
```diff
@@ -0,0 +1,194 @@
...
+                    "scope": str(m.scope),
+                    "system": str(m.system),
+                    "type": str(m.type),
+                }
+                for m in self
+            ] or ["<none>"]
+            raise AssertionError(
+                f"expected a stored memory matching {wanted}, but none did; "
+                f"observed: {observed}"
+            )
```
amit-gazal-band, line 156: Consider adding segment and status, to make the it easier to diagnose.

amit-gazal-band (approved):
Image

Still the same issue. Again, as in #396, since this is a #395 issue, this PR is approved.

tests/e2e/baseline/toolkit/room_capture.py
AlexanderZ-Band: Resolved by rebasing this PR onto the updated #396: `drain` is removed entirely (it was the nonce-echo barrier #395 replaced with the delivery-status barrier `wait_for_processed`, which a batched-burst turn defeated ~5/6 of the time). With `drain` gone, every durable-read method (`tool_calls`/`events`/`thoughts`/`errors`/`tasks`/`memory`) routes through `_require_user_ops`, so the divergence you flagged can't happen.

tests/e2e/baseline/toolkit/observations/memories.py
AlexanderZ-Band: Done — `assert_stored`'s observed-records diagnostic now includes `segment` and `status` alongside content/scope/system/type, so a filter mismatch on either is visible in the failure message.

AlexanderZ-Band:
On the recurring settings issue (the localhost endpoint default that made the gate hook error instead of skip): it's fixed at the #395 root and this PR now inherits it — I rebased #397 onto the updated #396, which sits on the fixed #395 base (`BandEndpoints` defaults to the public host, so `BaselineSettings()` no longer raises when the env vars are unset). This branch was also rebased to drop the buggy `drain`/`RoomCapture` mechanism in favor of the delivery-status barrier; the PR diff is now the clean 15-file memory+events port. All 14 event+memory smokes pass live.

PR #402, opened 2026-06-29
feat: [4] adapter discovery + generic provisioning matrix for the baseline harness

AlexanderZ-Band:
## Summary

Lands the live E2E **baseline toolkit**: a registry-driven adapter matrix with dynamic provisioning, event-driven waits, observation helpers, a framework-neutral custom-tool facade, docs, and a registry-derived CI workflow.

## What's included

- **Provisioning & user ops** — provision/reap agents, create/delete rooms, send user messages, manage participants.
- **Deterministic waits** — reply capture + processing barriers wait on real platform activity (delivery status), not sleeps or silence windows.
- **Observation** — replies, tool calls, emitted events, and memory state; tolerant structural assertions plus an LLM-judge fixture for semantic checks.
- **Adapter matrix** — registry auto-discovers the 12 LLM-agent adapters under `src/band/adapters/` (agno, anthropic, claude_sdk, codex, crewai, crewai_flow, gemini, google_adk, langgraph, letta, opencode, pydantic_ai). Non-agent bridges (a2a, a2a_gateway, acp, slack) and parlant are excluded via a documented allow/deny list. `@with_agents` / `@across_adapters` run one scenario body across the matrix with per-cell prompt/features/tools. An unregistered or misconfigured adapter **fails loudly** — it never skips. (Letta is registered but `e2e_pending`, so it is **not** a matrix cell — see _Out of scope_ below.)
- **`ToolSpec`** — define a custom tool once; it translates to Band `CustomToolDef`, pydantic-ai callables, and agno tools, and rejects adapters that can't take local tools.
- **CI** — manual-dispatch GitHub Actions workflow whose lanes are derived from the registry (`core`, `crewai`, `google`, `backends`, `letta`) and scoped via `BAND_E2E_LANE`; adding an adapter to an existing lane needs no workflow edit. The `letta` lane is present but runs no Letta tests yet (placeholder only — see _Out of scope_).
- **Docs** — baseline README playbook, adapter-registration guide, `.env.example`, AGENTS.md.

## Review focus

- Adapter registry + lane scoping (capability filtering, unsupported-tool handling).
- Processing-barrier / reply-capture semantics in busy or multi-agent rooms.
- CI secrets/env wiring per lane.

## Testing

- Live baseline smokes green on the branch (delivery status, room/turn isolation, events, memory, tool calls, processing barriers).
- Unit coverage for WebSocket reply-capture delivery and `ToolSpec.as_callable` signature synthesis.
- `ruff` and `pyrefly` clean.

[...]

PR #408, opened 2026-07-02
feat: baseline peer helpers — UserOps.lookup_peers + PeerActor (Echo peer)

AlexanderZ-Band:
## Summary

Baseline E2E toolkit additions that unblock the L0/L4 scenarios, plus docs. **Test tooling only — no SDK/product changes.** Stacked on the adapter-discovery branch (#402).

Two capabilities: read the *invitable roster* as the user, and *act as a peer agent* (the L0/L4 `Echo` peer). The Echo peer is deliberately **not** a running "bouncer adapter" — a provisioned identity + a single scripted send is more deterministic (an LLM bouncer could race or double-send).

## Changes

- **`UserOps.lookup_peers`** — driver-side mirror of the agent's `band_lookup_peers`; returns `list[Peer]`, `not_in_room=<room_id>` yields the still-invitable set, `page`/`limit` for pagination. Also cleaned `list_messages` to the same direct-typed-call form, dropping the `dict[str, object]` kwargs bag (the module is now pyrefly-clean).
- **`PeerActor` + `ResourceManager.peer(agent)`** (`toolkit/provisioning.py`) — the Agent-API twin of `UserOps`: post one message as a provisioned peer via its own key. So the Echo peer is just `provision_agent("echo")` + `resource_manager.peer(echo).send_message(...)`. Cached per agent id so repeated calls reuse one REST client.
- **`Replies.from_sender(sender_id)`** (`toolkit/observations/replies.py`) — scope a content assertion to one participant's own messages (needed when a peer and the agent under test both post into the same captured room).
- **Tests**
  - `guards/test_user_ops.py` — `lookup_peers` proven with a known owned agent: invitable to a room it's absent from, gone once added (pages the full roster so it's non-vacuous and not page-1-dependent; no LLM).
  - `guards/test_peer_ops.py` — keyless tool smoke: a peer posts as itself; the message lands attributed to the sending peer (not the user).
  - `smoke/behavior/test_peer_actor.py` — meaningful usage: a peer's message triggers and drives a real turn on a live agent, which echoes the peer's marker — asserted on the *agent's own* replies (`from_sender`), so the peer's trigger can't falsely satisfy it.
- **Docs** (`baseline/README.md`) — layout rows for `PeerActor` / `ResourceManager.peer` / `lookup_peers`, plus two "I want to…" entries.

## Related Issues

Relates to INT-930 (L0–L4 Tier-2 implementation) and parent INT-911. `lookup_peers` and the Echo peer are the two toolkit pieces those scenarios needed.

## Testing

- `uv run ruff check .` / `ruff format --check` — clean on all changed files.
- `uv run pyrefly check` — no net-new errors (the only ones on the touched modules are the pre-existing repo-wide `tests.*` import-root warnings).
- Ran a high-effort self code-review; the three findings (false-green behavior test, page-1 flake, per-call client leak) are fixed in this PR.
- New tests are **E2E-gated** (live `core` lane; need `BAND_API_KEY_USER` + `ANTHROPIC_API_KEY`). Verified they collect and pass every static gate; they execute in the live lane, not PR CI.
- [x] Unit tests unaffected (changes are under `tests/e2e/**`, skipped unless `E2E_TESTS_ENABLED`).
- [x] Lint / format / typecheck pass on changed files.

### DIFF
--- a/tests/e2e/baseline/README.md
+++ b/tests/e2e/baseline/README.md
@@ -201,7 +201,7 @@ These three are why the toolkit is shaped the way it is — keep them when exten
 
 | Path | What it is |
 |------|------------|
-| `toolkit/provisioning.py` | `ResourceManager` (provision/reap agents + rooms, orphan sweep, `track_running` reboot-race guard), `AdapterCell` (build / provision / running / run_as — the per-cell lifecycle object behind `agent`/`cell`), `running_agent` (run an already-provisioned identity — enter twice against one identity for a rejoin), `running_provisioned_agent` (provision + run, composes `running_agent`), `ProvisionedAgent` (`.adapter_id` records the cell/slot it came from) |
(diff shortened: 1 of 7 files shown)


amit-gazal-band (approved):
LGTM

PR #409, opened 2026-07-02
feat: Emit.USAGE seam — per-turn token usage capture

AlexanderZ-Band:
Implements **INT-931** — per-turn LLM token/cost capture — as an additive SDK seam, read back in the E2E baseline toolkit, mirroring the execution-reporting pattern.

## SDK
- `TurnUsage` (input / output / cache-read / cache-write) + `Emit.USAGE` in `band.core.types`. **RAW convention:** each field carries the provider's raw value with **no cache folding** — whether cache is already counted inside `input_tokens` is provider-specific, so prompt-size / cost math uses `input + cache_read + cache_write`. **Reasoning tokens:** providers that report them *disjointly* from output (codex, opencode, gemini, google_adk — their own totals are `input + output + reasoning`) fold reasoning into `output_tokens` via `TurnUsage._build(reasoning=…)`, staying consistent with providers that already count reasoning inside output (Anthropic thinking, OpenAI / LangChain completion tokens). `from_object` / `from_mapping` share `_build`.
- `SimpleAdapter.emit_usage()`: shared, best-effort, gated on `Emit.USAGE`, empty-skip.
- **Adapters emitting usage (10):** anthropic, langgraph, pydantic_ai, claude_sdk, agno, google_adk, gemini, opencode, codex, letta (per-room). **N/A:** parlant (server-side execution), crewai_flow (usage lives in user flow internals). **Deferred:** crewai — its `result.usage_metrics` is a cumulative-lifetime counter, not per-turn (separate follow-up).
- Usage rides an accepted `task` event's metadata (`USAGE_METADATA_KEY = "band_usage"`) because the backend rejects an unknown `usage` message_type today; `USAGE_EVENT_TYPE` keeps the flip to a first-class type a one-liner (see INT-933 below).
- `is_usage_event()` discriminator so task-event consumers (ACP `event_converter`, toolkit `Tasks.read`) don't mistake usage records for lifecycle tasks.

## Toolkit
- `Usage` / `UsageRecord` observation + `capture.usage()` + `COST_AGENT` / `COST_MULTI_TURN_AGENT` shapes + `assert_nonzero_input_and_output()` (the L4 gate).
- E2E smokes: `test_usage_recorded_for_a_turn` (one per-turn record, plausible count) and `test_usage_not_cumulative_across_turns` (a long turn then a one-word turn — a per-turn record stays far below the long turn, a cumulative bug rises to ~= it).

## Review fixes (root-caused)
- **Reasoning undercount:** disjoint-reasoning providers (codex / opencode / gemini / google_adk) fold reasoning into output at the shared builder; agno is deliberately *not* folded — a multi-provider aggregator whose `output_tokens` already includes reasoning for some backends (OpenAI) and excludes it for others (Gemini), so no static fold is correct for all (mirrors the RAW cache stance).
- **pydantic benign-path usage:** scoped to this run by message *identity*, not a positional slice — pydantic-ai's `_clean_message_history` merges adjacent same-type requests, so a `len(prior)` boundary slips and dropped the turn's leading response.
- **task pollution:** ACP no longer renders usage as a bogus plan step; `Tasks.read` excludes usage records.
- **error-path usage loss:** adapters that observe usage before a turn can fail (anthropic, gemini, langgraph, google_adk, pydantic_ai, letta) now emit from a `finally`, so a turn that raises after a successful call still reports the tokens it spent (a first-call failure has accumulated nothing, so nothing is emitted). pydantic_ai's two emit sites collapsed into that single exit point, with the captured-messages fallback covering any raise.
- **opencode usage race:** the per-turn usage dict is now turn-owned (`_begin_turn` allocates a fresh dict; the watch task drains the instance it captured, same snapshot pattern as `turn_future`), so a message racing in between turn completion and the drain can no longer wipe the previous turn's usage.

## Cache & error-path semantics
- Cache is reported RAW (no per-adapter fold) — provider-dependent, so a static normalization isn't correct across adapters.
- Usage emits on every turn exit: the happy path, the benign empty-final-response path, and turns that raise (whatever accumulated before the failure; tokens spent are tokens spent). Adapters whose usage only exists in a terminal result (claude_sdk, agno) have nothing to report when that result never arrives.

## Testing
- Unit: per-adapter mapping (incl. reasoning fold), identity scoping, `is_usage_event`, emit gating, loop-summing, mid-loop-failure emission (anthropic), captured-usage emission on a hard raise (pydantic_ai), turn-owned usage dict survives a racing new turn (opencode).
- **Live E2E:** anthropic / claude_sdk / agno / gemini / google_adk / opencode green; langgraph / pydantic_ai / codex blocked only by an OpenAI account quota (429), not the seam.

## Related
- Implements **INT-931** (parent: INT-911, the L0–L4 baseline ladder).
- Follow-up **INT-933** — make token usage a first-class `usage` event (retires the `band_usage`-on-`task` carrier + the consumer filters).
- Follow-up — per-turn crewai usage (its counter is cumulative-lifetime today).

### DIFF
--- a/src/band/adapters/anthropic.py
+++ b/src/band/adapters/anthropic.py
@@ -271,61 +271,62 @@ class AnthropicAdapter(SimpleAdapter[AnthropicMessages]):
 
         # Tool loop - let LLM decide when to stop. Anthropic reports usage
         # per API call, so sum it across every iteration of the loop into one
-        # per-turn TurnUsage (emitted once, after the loop).
+        # per-turn TurnUsage, emitted on every exit via the finally.
         turn_usage = TurnUsage()
-        while True:
-            try:
-                response = await self._call_anthropic(
-                    messages=self._message_history[room_id],
-                    tools=tool_schemas,
-                )
-            except Exception as e:
-                logger.error("Error calling Anthropic: %s", e, exc_info=True)
-                await self._report_error(tools, str(e))
-                raise  # Re-raise so message is marked as failed
-
-            turn_usage = turn_usage + self._usage_from_response(response)
-
-            # Check for tool use
-            if response.stop_reason != "tool_use":
-                # No more tool calls - extract text content if any
-                text_content = self._extract_text_content(response.content)
-                if text_content:
-                    self._message_history[room_id].append(
-                        {
-                            "role": "assistant",
-                            "content": text_content,
-                        }
+        try:
+            while True:
+                try:
+                    response = await self._call_anthropic(
+                        messages=self._message_history[room_id],
+                        tools=tool_schemas,
                     )
-                logger.debug(
-                    "Room %s: Completed with stop_reason=%s",
-                    room_id,
-                    response.stop_reason,
+                except Exception as e:
+                    logger.error("Error calling Anthropic: %s", e, exc_info=True)
+                    await self._report_error(tools, str(e))
+                    raise  # Re-raise so message is marked as failed
+
+                turn_usage = turn_usage + self._usage_from_response(response)
+
+                # Check for tool use
+                if response.stop_reason != "tool_use":
+                    # No more tool calls - extract text content if any
+                    text_content = self._extract_text_content(response.content)
+                    if text_content:
+                        self._message_history[room_id].append(
+                            {
+                                "role": "assistant",
+                                "content": text_content,
+                            }
+                        )
+                    logger.debug(
+                        "Room %s: Completed with stop_reason=%s",
+                        room_id,
+                        response.stop_reason,
+                    )
+                    break
+
+                # Add assistant response with tool_use blocks to history
+                serialized_content = self._serialize_content_blocks(response.content)
+                self._message_history[room_id].append(
+                    {
+                        "role": "assistant",
+                        "content": serialized_content,
+                    }
                 )
-                break
 
-            # Add assistant response with tool_use blocks to history
-            serialized_content = self._serialize_content_blocks(response.content)
-            self._message_history[room_id].append(
-                {
-                    "role": "assistant",
-                    "content": serialized_content,
-                }
-            )
+                # Process tool calls
+                tool_results = await self._process_tool_calls(response, tools)
 
-            # Process tool calls
-            tool_results = await self._process_tool_calls(response, tools)
-
-            # Add tool results to history
-            self._message_history[room_id].append(
-                {
-                    "role": "user",
-                    "content": tool_results,
-                }
-            )
-
-        # Emit the turn's aggregated token usage (no-op unless Emit.USAGE is on).
-        await self.emit_usage(tools, turn_usage)
+                # Add tool results to history
+                self._message_history[room_id].append(
+                    {
+                        "role": "user",
+                        "content": tool_results,
+                    }
+                )
+        finally:
+            # No-op unless Emit.USAGE is on; best-effort, never raises.
+            await self.emit_usage(tools, turn_usage)
 
         logger.debug(
             "Message %s processed successfully (history now has %s messages)",
--- a/src/band/adapters/anthropic.py
+++ b/src/band/adapters/anthropic.py
@@ -22,6 +22,7 @@ from band.core.types import (
     Capability,
     Emit,
     PlatformMessage,
+    TurnUsage,
 )
 from band.converters.anthropic import AnthropicHistoryConverter, AnthropicMessages
 from band.runtime.custom_tools import (
@@ -55,7 +56,7 @@ class AnthropicAdapter(SimpleAdapter[AnthropicMessages]):
         await agent.run()
     """
 
-    SUPPORTED_EMIT: ClassVar[frozenset[Emit]] = frozenset({Emit.EXECUTION})
+    SUPPORTED_EMIT: ClassVar[frozenset[Emit]] = frozenset({Emit.EXECUTION, Emit.USAGE})
     SUPPORTED_CAPABILITIES: ClassVar[frozenset[Capability]] = frozenset(
         {Capability.MEMORY, Capability.CONTACTS}
     )
@@ -268,7 +269,10 @@ class AnthropicAdapter(SimpleAdapter[AnthropicMessages]):
             custom_schemas = custom_tools_to_schemas(self._custom_tools, "anthropic")
             tool_schemas.extend(cast(list[ToolParam], custom_schemas))
 
-        # Tool loop - let LLM decide when to stop
+        # Tool loop - let LLM decide when to stop. Anthropic reports usage
+        # per API call, so sum it across every iteration of the loop into one
+        # per-turn TurnUsage (emitted once, after the loop).
+        turn_usage = TurnUsage()
         while True:
             try:
                 response = await self._call_anthropic(
@@ -279,6 +283,8 @@ class AnthropicAdapter(SimpleAdapter[AnthropicMessages]):
                 logger.error("Error calling Anthropic: %s", e, exc_info=True)
                 await self._report_error(tools, str(e))
                 raise  # Re-raise so message is marked as failed
+
+            turn_usage = turn_usage + self._usage_from_response(response)
 
             # Check for tool use
             if response.stop_reason != "tool_use":
@@ -318,6 +324,9 @@ class AnthropicAdapter(SimpleAdapter[AnthropicMessages]):
                 }
             )
 
+        # Emit the turn's aggregated token usage (no-op unless Emit.USAGE is on).
+        await self.emit_usage(tools, turn_usage)
+
         logger.debug(
             "Message %s processed successfully (history now has %s messages)",
             msg.id,
@@ -353,6 +362,21 @@ class AnthropicAdapter(SimpleAdapter[AnthropicMessages]):
             system=self._system_prompt,
             messages=cast(list[MessageParam], messages),
             tools=tools,
+        )
+
+    @staticmethod
+    def _usage_from_response(response: Message) -> TurnUsage:
+        """Map an Anthropic ``Message.usage`` onto the framework-agnostic TurnUsage.
+
+        Cache fields are optional on the SDK model (absent/None when prompt
+        caching is off), so they are read defensively and default to 0.
+        """
+        usage = response.usage
+        return TurnUsage(
+            input_tokens=usage.input_tokens or 0,
+            output_tokens=usage.output_tokens or 0,
+            cache_read_tokens=getattr(usage, "cache_read_input_tokens", 0) or 0,
+            cache_write_tokens=getattr(usage, "cache_creation_input_tokens", 0) or 0,
         )
 
     # --- Copied from BandAnthropicAgent._extract_text_content ---
--- a/src/band/adapters/anthropic.py
+++ b/src/band/adapters/anthropic.py
@@ -369,14 +369,14 @@ class AnthropicAdapter(SimpleAdapter[AnthropicMessages]):
         """Map an Anthropic ``Message.usage`` onto the framework-agnostic TurnUsage.
 
         Cache fields are optional on the SDK model (absent/None when prompt
-        caching is off), so they are read defensively and default to 0.
+        caching is off); ``from_object`` reads them defensively (missing → 0).
         """
-        usage = response.usage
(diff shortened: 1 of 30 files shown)


amit-gazal-band (commented):
Automated review pass on the Emit.USAGE seam. Overall the design is solid, but several adapters undercount or drop usage in exactly the ways the PR description claims are fixed. Flagging the ones most worth fixing before merge; verified each against the actual code on this branch (not just the diff).

src/band/adapters/anthropic.py
```diff
@@ -318,6 +324,9 @@ async def on_message(
                 }
             )
 
+        # Emit the turn's aggregated token usage (no-op unless Emit.USAGE is on).
+        await self.emit_usage(tools, turn_usage)
```
amit-gazal-band: **Usage is lost when a turn ends in an exception.** `emit_usage` only runs here, after the tool loop's happy-path exit. The `except Exception: ... raise` a few lines up (inside the same loop) re-raises without ever calling `emit_usage`, so any tokens already spent on earlier successful iterations of a multi-call tool loop are silently dropped. This contradicts the PR description's claim of emitting usage on "all turn-exit paths (incl. ... error)". Same pattern in `gemini.py`, `langgraph.py`, and `google_adk.py` (see linked comments). Suggest emitting accumulated `turn_usage` from the `except` block (or a `finally`) before re-raising.

src/band/adapters/gemini.py
```diff
@@ -248,6 +251,8 @@ async def on_message(
                 await self._report_error(tools, str(e))
                 raise
```
amit-gazal-band: Same class of bug as `anthropic.py`: this `raise` skips `await self.emit_usage(tools, turn_usage)` a few lines below, dropping any usage accumulated on prior loop iterations when a later call fails.

src/band/adapters/langgraph.py
```diff
@@ -340,6 +349,9 @@ async def on_message(
                 version="v2",
             ):
                 await self._handle_stream_event(event, room_id, tools)
+                turn_usage = turn_usage + self._usage_from_stream_event(event)
+
+            await self.emit_usage(tools, turn_usage)
```
amit-gazal-band: Same class of bug: `emit_usage` here is inside the `try` covering the `astream_events` loop. Any exception mid-stream routes to the `except Exception:` block further down and re-raises without ever reaching this call, dropping the `turn_usage` accumulated so far.

src/band/adapters/google_adk.py
```diff
@@ -566,6 +577,10 @@ async def on_message(
         finally:
             await runner.close()
 
+        # Emit the turn's aggregated usage (reached only on a clean run — the
+        # except above re-raises). No-op unless Emit.USAGE is on.
+        await self.emit_usage(tools, turn_usage)
```
amit-gazal-band: Same class of bug — the comment right above this line ("reached only on a clean run — the except above re-raises") explicitly acknowledges that this `emit_usage` call is skipped on the exception path, dropping accumulated usage.

src/band/adapters/opencode.py
```diff
@@ -896,6 +896,7 @@ def _begin_turn(self, room_state: _RoomState, *, sender_id: str | None) -> None:
         room_state.assistant_part_types.clear()
         room_state.reported_tool_calls.clear()
         room_state.reported_tool_results.clear()
+        room_state.usage_by_message.clear()
```
amit-gazal-band: **Race: a new turn can clear `usage_by_message` before the previous turn's usage is emitted.** `_finish_turn` (called from the SSE listener on `session.idle`/`session.error`) resolves `turn_future` *and* `turn_release_future` synchronously in the same call — before `_watch_turn_completion` actually runs `_emit_turn_usage`/reads `usage_by_message`. `on_message`'s concurrency guard only checks `turn_future.done()`, so a new message arriving in the same tick that `turn_future` flips to done — but before the watch task drains `usage_by_message` — passes the guard, reaches `_begin_turn`, and clears this dict (line 899) out from under the still-pending read. Net effect: the first turn's usage silently disappears (sums to empty, `emit_usage` no-ops). Consider keying `usage_by_message` per-turn (e.g. include a turn id) instead of clearing a shared per-room dict, or having `_begin_turn` wait for the prior turn's watch task to fully finish.

src/band/adapters/opencode.py
```diff
@@ -1023,6 +1028,49 @@ async def _deliver_fallback_text(self, room_state: _RoomState) -> None:
...
+        tokens = info.get("tokens")
+        if not isinstance(tokens, dict):
+            return TurnUsage()
+        cache = tokens.get("cache")
+        cache = cache if isinstance(cache, dict) else {}
+        flat = {
+            "input": tokens.get("input"),
+            "output": tokens.get("output"),
+            "cache_read": cache.get("read"),
+            "cache_write": cache.get("write"),
```
amit-gazal-band, line 1095: `_usage_from_info`'s `flat` dict only pulls `input`/`output`/`cache_read`/`cache_write`; `tokens.get("reasoning")` (present in the payload shape documented in this function's own docstring) is never read, so reasoning tokens are silently excluded from the emitted `TurnUsage`, undercounting cost for reasoning models routed through OpenCode. (Same gap in `codex.py:_turn_usage`, which never reads `CodexTokenUsage.turn_reasoning_tokens`.) Note this isn't purely an adapter oversight — `TurnUsage` itself (`core/types.py`) has only input/output/cache_read/cache_write dimensions, no reasoning field, so the fix may need to start at the schema level (e.g. fold reasoning into `output_tokens`, matching how most providers bill it).

src/band/adapters/codex.py
```diff
@@ -1537,6 +1538,19 @@ async def _handle_approval_request(
...
+    @staticmethod
+    def _turn_usage(usage: CodexTokenUsage | None) -> TurnUsage:
+        """Map codex's per-turn token deltas onto the framework-agnostic TurnUsage.
+
+        Uses the per-turn deltas (rise from the turn-start anchor), not the
+        thread cumulatives, so each emitted record is this turn's usage —
+        matching what the in-process adapters emit. Codex reports no cache
+        dimension, so those stay 0; a ``None`` usage yields empty (no emit).
+        """
+        return TurnUsage.from_object(
```
amit-gazal-band, line 1555: Only `turn_input_tokens`/`turn_output_tokens` are read here; `CodexTokenUsage.turn_reasoning_tokens` (actively maintained elsewhere in this file) is never included, so reasoning-heavy Codex turns undercount output tokens in the unified `Emit.USAGE` record even though the adapter tracks reasoning precisely already (`codex_turn_reasoning_tokens` on the legacy marker path).

src/band/adapters/pydantic_ai.py
```diff
@@ -613,6 +619,14 @@ async def on_message(
...
         # band tools) so we can tell a productive turn from a genuine no-op below.
         tool_executed = False
+        # pydantic-ai's result.usage() is already summed across the run's model
+        # calls, so it's set once (on the result event), not accumulated.
+        turn_usage = TurnUsage()
+        # Snapshot the history length so the benign-path usage fallback can sum
+        # only THIS run's responses: capture_run_messages() records the passed
+        # message_history + the new turn, so summing the whole list would
+        # double-count every prior turn's usage.
+        history_len = len(self._message_history[room_id])
```
amit-gazal-band: **Off-by-one on the benign-empty-response fallback when both `participants_msg` and `contacts_msg` fire in the same turn.** `history_len` is snapshotted here, but pydantic-ai's internal `_clean_message_history` (invoked when the run starts) merges adjacent `ModelRequest`s that don't set differing `instructions` into one entry. Since `participants_msg`/`contacts_msg` are each appended as separate `ModelRequest`s without `instructions=` a few lines above, a turn where both are non-empty ends up with `captured` one shorter than `history_len` after the merge. On the `UnexpectedModelBehavior` fallback path further down (`captured[history_len:]`), this skips the turn's first `ModelResponse`, dropping its usage from `_usage_from_messages`. Note: commit `a9f6a442` on this branch fixed a related but different bug (double-counting the whole history) — this merge-driven off-by-one is a separate, still-open edge case.

src/band/adapters/opencode.py
```diff
@@ -928,9 +929,13 @@ async def _watch_turn_completion(
                     "OpenCode timed out before completing the turn.",
                     "error",
                 )
+            # Tokens spent before the timeout were still spent — emit them, same
+            # as the success path (best-effort; no-op if none captured).
+            await self._emit_turn_usage(room_state)
             self._release_turn_wait(room_state)
         else:
             await self._deliver_fallback_text(room_state)
+            await self._emit_turn_usage(room_state)
```
amit-gazal-band: Minor cleanup: `_emit_turn_usage(room_state)` + `_release_turn_wait(room_state)` are duplicated identically in both the timeout branch above and this success branch, right above the existing `finally: self._clear_turn_state(...)` that already runs for both. No other exit path skips these two calls today, so hoisting them into the `finally` would remove the duplication safely.

amit-gazal-band (changes requested):
Left some inline comments from Claude. TBH I didn't dig into all of them, but they sound reasonable.

src/band/adapters/anthropic.py
AlexanderZ-Band: Rethought and fixed in 6e169fd: the loop's accumulated usage now emits from a finally, so a turn that fails after a successful call still reports the tokens it spent (a first-call failure has accumulated nothing, so nothing is emitted). Same treatment in gemini/langgraph/google_adk, extended to pydantic_ai and letta for the same semantic. Regression tests added.

src/band/adapters/gemini.py
AlexanderZ-Band: Fixed in 6e169fd with the anthropic thread (finally-based emit).

src/band/adapters/langgraph.py
AlexanderZ-Band: Fixed in 6e169fd with the anthropic thread (finally-based emit).

src/band/adapters/google_adk.py
AlexanderZ-Band: Fixed in 6e169fd with the anthropic thread (emit moved into the existing finally).

src/band/adapters/opencode.py
AlexanderZ-Band: Rethought and fixed in 6e169fd: the usage dict is now turn-owned. _begin_turn allocates a fresh dict and the watch task drains the instance it captured (same snapshot pattern as turn_future), so a racing new turn cannot clear it. Regression test added.

src/band/adapters/opencode.py
AlexanderZ-Band: Fixed in 7e930a4. Confirmed disjoint via OpenCode's own `acp/usage` (`total = input + output + reasoning + cache`): `_usage_from_info` now folds `tokens.reasoning` into `output_tokens`, through a shared `TurnUsage._build(reasoning=…)` kwarg. Same fold applied to codex/gemini/google_adk, which also report reasoning disjointly.

src/band/adapters/codex.py
AlexanderZ-Band: Fixed in 7e930a4. `CodexTokenUsage`'s own total is `input + output + reasoning`, so `_turn_usage` now folds `turn_reasoning_tokens` into `output_tokens` (shared `reasoning=` kwarg).

src/band/adapters/pydantic_ai.py
AlexanderZ-Band: Fixed in cb9f276 — and confirmed the mechanism: `capture_run_messages` holds the *cleaned* history (`messages[:] = _clean_message_history(...)`), which merges adjacent same-type requests, so the positional `history_len` slips. Now scoped by message identity (`_new_run_messages`) — real API ModelResponses keep identity through cleaning. Regression test added.

src/band/adapters/opencode.py
AlexanderZ-Band: Keeping as-is, with a better reason than "cosmetic": that finally also runs when the watch task is cancelled (cleanup, prompt failure), where firing the emit/release calls would be wrong. The per-branch calls are deliberate.

amit-gazal-band (approved):

PR #410, opened 2026-07-03
feat: add GitHub Copilot SDK adapter INT-921

AlexanderZ-Band:
## Summary

Adds a `copilot_sdk` framework adapter + history converter connecting Band agents to the **GitHub Copilot SDK** (`github-copilot-sdk`, imported as `copilot`). The SDK manages the Copilot CLI runtime subprocess internally; the adapter maps each Band room to its own Copilot session on a single shared client, with cross-restart rehydration, native tool bridging, BYOK support, and **room-routed human-in-the-loop `ask_user`**. The branch also hardens the shared runtime with a **single-instance guard** (one process per agent id per host). Multiple review rounds were run on the branch (high-effort code reviews, a documentation fact-check audit, a subprocess-lifecycle review, and a final 10-angle review + sweep over the ask_user/guard work); all findings fixed — several validated empirically against the live runtime before acting.

### Core (`src/band/`)
- **`adapters/copilot_sdk.py`** — `CopilotSDKAdapter` + frozen `CopilotSDKAdapterConfig`:
  - session-per-room with deterministic ids `band-{band_agent_id}-{room_id}` (immutable agent id from the runtime, so agents sharing a host/client/room never resume each other's state; explicit `session_id_prefix` overrides)
  - resume via the persisted session-id task event (marked persisted only on successful send, so transient failures retry); text-injection fallback when resume misses — the injected history **includes the agent's own replies**
  - Band platform + custom tools bridged as native `copilot.Tool` handlers; built-in shell/file tools disabled via the `available_tools` allowlist — the coupling that makes `approve_all` safe
  - `Emit.EXECUTION`/`THOUGHTS` events (off by default); `Capability.MEMORY`/`CONTACTS` gate the band_* tool groups
  - **BYOK**: typed `ProviderConfig` passthrough (OpenAI/Azure/Anthropic; `base_url` required by the runtime)
  - **One client, many sessions**: `client=` shares a caller-owned `CopilotClient` across adapters (borrowed clients are never stopped); `client_factory=` stays the adapter-owned/test seam
  - lifecycle: eager start + fail-fast auth probe with rollback in `on_started`; failed/timed-out turns abort + evict the session; **a crashed CLI heals on the next message** (no cached liveness — `ensure_started` delegates to the client, verified by killing the subprocess live)

### Human-in-the-loop: `ask_user` (`config.ask_user = "room" | <handler> | None`)
- **Room mode (`"room"`, the Band-native path)** — Copilot's built-in `ask_user` tool routes to the people in the room via **turn-splitting**: the question posts as a room message (choices numbered, mentioning the turn's sender), the tool call resolves immediately with an ack that **echoes the rendered question** (so a bare "2" maps unambiguously to choice 2), the turn ends, and the answer arrives as the next room message on the same persisted session. Blocking the turn can never work: Band delivers a room's messages strictly serially (the answer would deadlock behind the very turn awaiting it) and Copilot keeps an unanswered ask pending forever (no timeout, no cancellation, not replayed on resume — verified against SDK source + docs + issue tracker). The turn-split contract is injected as its own system-prompt section via the new `render_system_prompt(extra_sections=…)` — SDK contract text, not developer text. (`integrations/copilot_sdk/room_ask_user.py`)
- **Handler mode (callable)** — answers from outside the room (terminal operator, approval service). `integrations/copilot_sdk/operator_console.py` ships a hardened terminal handler (`OperatorConsole`): refcounted log gating (auto-engages on first question), per-question deadline with graceful no-answer/EOF fallbacks, answer validation honoring `choices`/`allowFreeform` (numeric labels still match by text; out-of-range numbers re-prompt), and a single persistent stdin reader that survives event-loop restarts. Defaults are safe out of the box (answer window < turn timeout).

### Runtime: single-instance guard (`runtime/single_instance.py`)
- Two processes running one agent id corrupt each other (both claim in-flight room messages; stateful adapters resume the same on-disk sessions). `Agent.start()` now claims a per-agent-id `flock` **before the adapter boots anything**, releases on stop/failed start, and the OS drops it on crash. Only `BlockingIOError` maps to "already running"; open/flock failures get distinct actionable messages. **Opt out with `AgentConfig(single_instance=False)`** — this is a deliberate behavior change for anyone intentionally running same-host duplicates. Scope is stated honestly (same temp dir; oneshot relies on server-arbitrated claiming instead).
- `agent.py` gains a weak registry of running agents (`running_agents()`) so harnesses whose failure paths skip unwinding (e.g. pytest-timeout signal kills) can find and stop leaked agents instead of leaving zombies.

### Shared fixes riding along
- `Agent.start()` rolls back adapter resources if the runtime fails to start; `Agent.stop()` invokes the adapter's `cleanup_all()` in a `finally` (fixes the same subprocess leak for claude_sdk)
- Shared helpers extracted: `SELF_REPORTING_TOOL_NAMES` and `format_validation_error` (fixes codex's latent `loc[0]` IndexError)
- Mentions deprecation warning now exempts id-bearing dicts (adapter-supplied ground truth; the deprecation targets LLM-guessed handles)
- pytest collection fix: exact `integration` path-segment match (`tests/integrations/` was silently dropped from CI runs); e2e-marked tests outside `tests/e2e` gated on `E2E_TESTS_ENABLED`

### Tests
- **Unit (107 across the copilot + runtime-guard suites; full suite 3,542)**: fake Copilot client at the `client_factory` seam — reply/tools/resume/eviction/crash-recovery/persist-retry, per-room + per-agent isolation, shared-client ownership, startup-failure rollback, ask_user room bridge (delivery, ack echo, failure degradation, orphaned-dispatch safety), OperatorConsole (matching, timeouts, EOF, log gating, loop restart, shared gates), single-instance guard (contention, release, reap) and its PlatformRuntime wiring. The fake session now **raises on session errors like the real SDK** — which exposed and removed a production-unreachable error branch.
- **Conformance** registered; config-drift green
- **E2E (live, all passing)**: 5 parametrized cells via the shared registry (smoke, tool execution, rejoin rehydration, noisy-room recall, room isolation) on **BYOK Anthropic**, plus four dedicated tests — shared-client lifecycles, resume-miss rehydration, and **both ask_user modes**: `test_copilot_sdk_ask_user.py` (programmatic operator answers with a secret token; reply must relay it) and `test_copilot_sdk_ask_user_room.py` (question posts to the room, the user's next message answers with a secret token, the reply relays it — the full room round trip). E2E infra: leaked-agent reaper (pytest-timeout signal kills skip async unwinding), anchored rerun-except patterns, 240s budgets for multi-boot copilot tests.

### Docs & examples
- `examples/copilot_sdk/`: basic, BYOK Anthropic, Tom & Jerry, memory+contacts, and **`06_ask_user.py`** (room-routed human-in-the-loop); README covers auth resolution, feature flags, both `ask_user` modes with the timeout-pairing rules, the one-process-per-agent guard, models, troubleshooting
- The adapter-building process was distilled into a reusable plan-template + execution playbook (kept as a local skill, not in-repo); ask_user lifecycle lessons folded back into it

## Verification

- ruff / pyrefly clean; full unit suite green (3,542 collected)
- Live-validated against the real Copilot runtime: reply+mention, bridged tool execution with fallback suppression, cross-process resume/recall, resume-miss injection, subprocess-crash recovery, fail-fast auth, Anthropic + OpenAI BYOK, and **both ask_user round trips** (room mode: question → user answer → relayed token; handler mode: operator token relayed)
- Live E2E prereqs: `GITHUB_TOKEN` (Copilot-entitled) + `ANTHROPIC_API_KEY` in `.env.test`; pre-fetch via `python -m copilot download-runtime`

## Update — per-turn token usage (`Emit.USAGE`) + `core` CI lane

Follow-up commits on this branch bring `copilot_sdk` to parity with every other tool-loop adapter and simplify its CI placement.

- **`Emit.USAGE` seam** — copilot was the only tool-loop adapter still missing the per-turn token-usage capture from the `Emit.USAGE` rollout. Copilot streams usage as **per-API-call** `assistant.usage` events, so the adapter sums them across the turn (`TurnState.usage`) and emits one record from `on_message`'s `finally` (fires on **both** success and failure paths). `reasoning_tokens` is **not** folded into `output_tokens` — the providers Copilot relays (OpenAI/Anthropic) already count reasoning inside output, so folding would double-count. `SUPPORTED_EMIT` now advertises `Emit.USAGE`. Adds 5 unit tests: mapper + no-fold regression guard, cross-call summing, feature-off no-op, failure-path emission (usage still emitted when the turn raises), and empty-turn skip.
- **`core` CI lane** — drops the dedicated `copilot` lane; `Dep.COPILOT_GITHUB_TOKEN` now defaults to `Lane.CORE`, and the `copilot` entry is removed from the workflow dropdown, the lane comment, `AGENTS.md`, and the baseline README's CI-lanes table. Trade-off accepted (documented): `core`'s green signal now depends on the Copilot entitlement (`E2E_GITHUB_TOKEN`) and the CLI self-download.
- **Naming** — `_TurnState` → `TurnState`, `_RoomSessionIds` → `RoomSessionIds` (module-private dataclasses read cleaner without the leading underscore; the private `self._turn_state` / `self._session_ids` instance attributes keep theirs).

### Verification (usage work)

- Full unit suite green; ruff / pyrefly clean.
- **Live E2E**: no new test file — the registry-derived usage-smoke fan now exercises copilot automatically the moment the adapter declares `Emit.USAGE`. `test_usage_recorded_for_a_turn[copilot_sdk]` and `test_usage_not_cumulative_across_turns[copilot_sdk]` both **pass live** on BYOK Anthropic, proving `assistant.usage` fires and the per-turn record is non-cumulative across turns.
- Reviewed by a multi-role pass (elegance / integration / logic) plus a high-effort code-review sweep; the one factual nit it found (a docstring citing a sibling SDK usage type) is fixed.

## Related Issues

Fixes INT-921

### DIFF
--- a/.github/workflows/e2e.yml
+++ b/.github/workflows/e2e.yml
@@ -33,6 +33,10 @@ name: E2E (baseline)
 #                  band-mcp), so the lane runs no Letta cells. Own lane (shares the
 #                  `dev` extra) kept defined for the follow-up; see toolkit/adapters.py.
 #
+# CopilotSDKAdapter (BYOK on Anthropic) runs in `core`: the SDK self-downloads its
+# CLI runtime (no setup step), so it needs no isolated lane — it only needs the
+# Copilot-entitled GITHUB_TOKEN, which every lane's job env carries.
+#
 # An unwired backend still fails loudly (the @requires gate reports the missing
 # CLI/server) until its setup block lands. `fail-fast: false`
 # keeps a red lane from blocking the others.
@@ -149,6 +153,11 @@ jobs:
       ANTHROPIC_API_KEY: ${{ secrets.E2E_ANTHROPIC_API_KEY }}
       OPENAI_API_KEY: ${{ secrets.E2E_OPENAI_API_KEY }}
       GOOGLE_API_KEY: ${{ secrets.E2E_GOOGLE_API_KEY }}
+      # A PAT from a GitHub account with Copilot entitlement — NOT the ambient
+      # secrets.GITHUB_TOKEN (a repo-scoped installation token with no such
+      # entitlement). The copilot_sdk adapter (in the `core` lane) needs it;
+      # exposed to every lane's job env so `core` picks it up.
+      GITHUB_TOKEN: ${{ secrets.E2E_GITHUB_TOKEN }}
 
     steps:
     - name: Checkout code
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -4,7 +4,7 @@ This is a Python SDK that connects AI agents to the Band collaborative platform.
 
 ## Core Features
 
-1. Multi-framework support (LangGraph, Anthropic, CrewAI, Claude SDK, Codex, Pydantic AI, Parlant, Gemini, Letta, Google ADK, OpenCode, Agno)
+1. Multi-framework support (LangGraph, Anthropic, CrewAI, Claude SDK, Copilot SDK, Codex, Pydantic AI, Parlant, Gemini, Letta, Google ADK, OpenCode, Agno)
 2. A2A protocol support: Bridge to remote A2A agents and expose Band peers as A2A endpoints
 3. ACP integration: Editor-facing server and subprocess client adapters (Cursor, Codex, Claude Code)
 4. Platform tools for chat, contacts, and memory management
@@ -391,6 +391,7 @@ resolves each in a separate fork.
 - `GOOGLE_API_KEY`: Google API key for Gemini Developer API (for Gemini/Google ADK examples)
 - `GOOGLE_GENAI_USE_VERTEXAI`: Set to `true` to use Vertex AI instead of Gemini Developer API
 - `GOOGLE_CLOUD_PROJECT`: Google Cloud project ID (required when using Vertex AI)
+- `GITHUB_TOKEN`: A token from a GitHub account with Copilot entitlement, for the Copilot SDK adapter's runtime auth (BYOK inference reuses `ANTHROPIC_API_KEY`). Read by the baseline toolkit's `tests/e2e/baseline/settings.py`
 - `E2E_TESTS_ENABLED`: Set to `true` to enable E2E tests (default: disabled)
 - `E2E_LLM_MODEL`: OpenAI model for E2E tests (default: `gpt-5.4-mini`)
 - `E2E_ANTHROPIC_MODEL`: Anthropic model for E2E tests (legacy E2E default: `claude-3-haiku-20240307`; baseline toolkit default: `claude-haiku-4-5` — the baseline judge uses structured outputs, which `claude-3-haiku-20240307` does not support)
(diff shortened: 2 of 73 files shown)


amit-gazal-band (changes requested):
Thoughts are sent twice.
Image

Image

(Or even three times)

AlexanderZ-Band:
Fixed (folded into the adapter commit, `9801c4b`).

Turns out the Copilot CLI fires the same reasoning block 2-3x per tool-loop turn - same `reasoning_id`, identical text (1x with no tools, 2-3x once it starts calling tools, hence "even three times"). We were posting one thought per event, so they piled up.

Now we dedupe by `reasoning_id` (last write wins) so each block shows up once, and distinct blocks stay separate. The repeats come from the CLI itself - not our listener, the SDK dispatch, or sub-agents - so keying on the id it stamps is the right guard.

amit-gazal-band (approved):

PR #411, opened 2026-07-05
feat: L0–L4 baseline live-E2E coverage as matrix scenarios [INT-930]

AlexanderZ-Band:
## Summary

Closes the live end-to-end coverage gaps derived from the L0–L4 conformance specs (INT-523 / INT-524 / INT-525 / INT-527 / INT-528) as scenario-named `@per_adapter` matrix tests that run across the whole adapter registry and reuse the existing baseline toolkit (`tests/e2e/baseline/`). No "levels" concept leaks into code or file names — tests are named by behaviour.

## Changes

**Toolkit additions (the only new plumbing):**
- `Replies.assert_at_most` — the one sanctioned upper-bound runaway guard, localized on `Replies` (not the floors-only `ContentAssertions` base); used only by loop suppression.
- `AdapterCell.run_many` — K co-resident same-adapter instances in one room, started concurrently.
- `running_members` — concurrent-start helper **factored out of** the `agents` fixture so the fixture and `run_many` share one implementation instead of drifting.
- `ReplyCapture.turn_boundary` — server-timestamp `since=` accessor, DRY-promoted from `test_isolation`'s private `_turn_boundary` (both now use it).

**Scenario tests (all matrix-driven, floors-only assertions):**
- **L0** — identity/roster probe; participant management (invite + directed message + remove); loop suppression (peer bounce, no self-dispatch runaway).
- **L1** — custom prompt takes effect and coexists with platform tools.
- **L2** — burst no-drop via delivery status + spanning multi-fact recall.
- **L3** — same-adapter concurrency gate (incl. codex/opencode); peer-initiated delegation + self-recall.
- **L4** — cold-restart dedup asserted model-independently at the delivery layer + restart token split (replay vs new inference).

**Retirements (subsumed, removed in the same change):**
- `test_peer_actor` → folded into loop suppression.
- `test_agent_recalls_earlier_facts` → folded into burst recall.

Plus shared driving glue in `sample_agents` / `sample_tools`, and PR-run unit tests for the new primitives in `framework_conformance/test_toolkit_helpers.py`.

**Fixture-read hardening (follow-up commit):**
- `MarkerPayload.from_node` — one validated read for the `@per_adapter` / `@with_adapters` marker payloads (fail-loud `UsageError` on a missing/malformed marker instead of a cryptic `IndexError`/`AttributeError`; the `isinstance` check makes the payload type real). The four fixture read sites use it.
- `agent_wiring` now rejects the same topology decorator applied more than once (`get_closest_marker` would otherwise silently drop the extra).

## Coverage vs the L0–L4 specs

Scoped to **Tier 2 (live E2E)** — Tier 1 (isolated per-PR model-output-injection tests) is out of scope and not counted. A per-requirement audit (each spec step mapped to a concrete assertion in the merged tests, not taken on faith) counts **16.5 / 18** Tier-2 requirement units as covered-or-partial (covered = 1, partial = 0.5, deferred/out = 0). Two of those meet the requirement's *intent* through a model-independent floor rather than the spec's literal exact-count assertion (L0 no-loop, L2 no-drop — see the reframes below), so the honest headline is a conservative **~80–90%**, not the raw 92%. It runs across the full adapter matrix (each scenario × 7–12 adapters, vs the specs' single `test-agent`).

Per level: **L0 100% · L1 100% · L2 100% · L4 100% · L3 70%.**

| Level | Requirement | Status | Evidence |
|---|---|---|---|
| L0 | S1 identity + roster + invitable peer | ✅ | `test_identity_and_roster` — own name / a member / an invitable out-of-room peer asserted separately |
| L0 | S2 invite (present via API) + directed @mention w/ marker | ✅ | `test_participant_management` — `echo.id in participants` + coupled `mentioning(echo).assert_contains_any([marker])` |
| L0 | S3 no self-triggered send loop | ✅ reframed | `test_loop_suppression` — `assert_at_most(5)` runaway ceiling (spec "≤1" reframed under floors policy) |
| L0 | S4 remove participant (absent via API) | ✅ | `test_participant_management` — `echo.id not in participants` |
| L1 | S1 custom-prompt marker every reply + tool round-trip | ✅ | `test_custom_prompt` (marker both turns) + `test_tool_round_trip` (`assert_fired` + opaque value in reply) |
| L1 | S2 platform tools coexist + prompt active + no fabrication | ✅ | `test_custom_prompt` t2 — member named + marker persists + `assert_contains_none([nonmember])` |
| L2 | S1 burst no turn dropped | ✅ reframed | `test_burst_recall` — per-message `PROCESSED` delivery-status no-drop (N=6; "exactly-N replies" reframed) |
| L2 | S2 spanning recall (early/mid/recent) | ✅ | `test_burst_recall` — early / mid / recent asserted separately over the recall turn |
| L3 | Gate: K=3 same-adapter instances co-reside, each replies | ✅ | `test_concurrent_instances` — `run_many(3)`, per-instance `assert_present` (incl. codex/opencode) |
| L3 | T1 named routing + value relay | ✅ (return-leg soft) | `test_multi_agent_collaboration` (`assert_mentions` both peers, both opaque results land) + `test_peer_delegation` |
| L3 | T2 description-based routing | ❌ DEFERRED | SDK drops participant `description` from the passive roster → nothing to route on (→ INT-943) |
| L3 | T3 multi-hop in correct causal order | ⚠️ PARTIAL | 2-hop delegation chains barriered, but strict causal-*order* assertion omitted (flaky; floors policy forbids strict ordering) |
| L3 | T4 peer-initiated delegation + self-recall + round-trip | ✅ (round-trip soft) | `test_peer_delegation` — coupled peer-initiated mention carrying self-recalled value + delegate responded |
| L4 | History restored on cold restart | ✅ | `test_rehydration_idempotency` / `test_context_recall` (rejoin) — `assert_contains_any([note])` after cold boot |
| L4 | Offline message picked up on boot | ✅ | `test_rehydration_idempotency` — offline `mid` answered by the boot-drain, capture opened before boot |
| L4 | Already-handled message not re-emitted (dedup) | ✅ reframed | `test_rehydration_idempotency` — no fresh `PROCESSING` transition on cold boot (delivery layer, not reply content) |
| L4 | Completed tool call (invite) not replayed | ✅ reframed | same delivery-layer check on the invite's trigger message (direct tool-replay confounded by rehydration → INT-942) |
| L4 | Restart token split: replay-input + new-inference output | ✅ | `test_rehydration_idempotency` — `assert_nonzero_input_and_output()` scoped past the run-1 boundary |

The uncovered/reframed remainder is deliberate, not missed:
- **L3 description-based routing (T2)** — deferred behind a real SDK gap; the agent has no `description` to route on until INT-943 surfaces it.
- **L3 multi-hop causal ordering (T3, "Calc before Greeter")** — the strict-ordering cascade is left out (flaky/expensive on a small model; floors-only policy forbids strict-order assertions). The multi-hop chaining itself is still exercised.
- **L2 exact counts** — "exactly N replies" / "no double-process" can't be gated on non-deterministic models — reframed to delivery-status no-drop (and burst N reduced 9→6).
- **L4 action-level idempotency** ("side-effect not re-fired") is confounded by rehydration replaying history to the model — escalated as INT-942 and re-expressed as sound message-level (delivery-layer) dedup.
- **L0-S1** room-UUID and participant display-name are not asserted (UUIDs get paraphrased; no toolkit read returns a display name), and the L3 T1/T4 return-leg value relay is kept soft/non-gating — both consistent with the floors-only policy.

## Related issues

- Implements INT-930; derived from specs INT-523 / INT-524 / INT-525 / INT-527 / INT-528.
- Filed from the live run: **INT-942** (cold-boot rehydration replays history → a weak model re-executes it) and **INT-943** (surface participant `description` in the passive roster, unblocking description-based routing).

## Testing

- [x] Unit tests pass — `uv run pytest tests/ --ignore=tests/integration/ --ignore=tests/e2e/` → 3480 passed, 57 skipped.
- [x] Lint + type gates — `ruff check`, `ruff format`, `pyrefly check` (0 errors).
- [x] E2E collection clean — 265 baseline tests collect (all wiring/schedulability guards); framework-conformance PR-run suite → 435 passed.
- [x] Live E2E (anthropic, `core` lane) — all 9 new test functions pass; `test_isolation` (incl. the refactored `turn_boundary` path) green.
- [ ] Other lanes (google / crewai / backends) — run per-lane in CI; not exercised locally.

## Checklist

- [x] Documentation updated as needed (plan doc + toolkit docstrings)


amit-gazal-band (approved):
Super elegant!

PR #413, opened 2026-07-06
feat: wire the Letta lane live — self-hosted MCP mode, seeding, matrix flip

AlexanderZ-Band:
## Summary

Letta goes from `e2e_pending` to a first-class matrix adapter: **14 `@per_adapter` cells + 2 supplementary smokes run live** in the `letta` CI lane against a dockerized Letta server, with the adapter self-hosting its Band MCP server in-process (mirroring the opencode backend pattern — per-agent identity, aligned tool names, no extra image or secret). The external band-mcp remains supported as the Letta Cloud/production topology via `LettaMCPConfig(mode="external")`.

## Changes

**Structure (`refactor(letta)`)**
- New `src/band/integrations/letta/` package, following the repo's integrations convention: `mcp.py` (`LettaMCPBridge` — self-hosted server lifecycle, registration with conflict recovery, tool discovery + send-tool-name resolution, bounded release; all the Letta-server quirk handling), `config.py` (`LettaMCPConfig`/`LettaAdapterConfig`/`MCPTransport`, re-exported from `band.adapters.letta` — import paths unchanged), `prompts.py` (enforcement template + send-tool alias tables). The adapter drops from ~1450 to ~1040 lines: platform contract, rooms, turns, agent lifecycle.

**Adapter design fixes**
- Cold-boot history seeding: the converter builds `replay_messages` from platform history; a fresh Letta agent in a room with history is seeded once — the rehydration cells validate real `/context` rehydration instead of passing via Letta's private memory.
- Enforcement prompt, silent-reporting set, and auto-relay detection derived from the registered server's discovered tools (name-agnostic across the SDK's `band_*` and external `create_agent_chat_*` surfaces).
- Auto-relay warns loudly and is disableable (`auto_relay=False` fails loud via an error event) so a dead MCP tool path can't hide behind relayed text.
- `embedding` passed on agent create (Letta's docker server requires it); shared-mode conversations resume instead of being recreated; `delete_agents_on_cleanup` / `consolidate_memory_on_cleanup` knobs; teardown calls bounded by `teardown_timeout_s`.
- New `on_stopped()` adapter hook called by `Agent.stop()` — rooms are not cleaned on shutdown, so adapters release cross-room resources there.

**Letta-server quirks found by live validation, worked around**
- Registrations are soft-deleted, so a name can never be reused → fresh unique name per registration + recovery when a timed-out create conflicts with its own committed row (409).
- Deleting a registration whose server has died wedges Letta's serial MCP sync worker (~60s stall per later registration, minimally reproduced) → deregister on stop but keep the in-process server alive for the process lifetime; ephemeral OS-assigned ports.
- MCP tools are materialized globally **by name**, so same-process co-resident instances cross-wire their sends through the last registrant's server → the concurrent-instances cell excludes Letta with the evidence documented in its module doc.
- The MCP URL guard rejects any hostname resolving to a non-public IP (no official knob) → `setup-letta.sh` patches the validator inside the throwaway container.

**Toolkit / CI**
- `.github/scripts/setup-letta.sh`: docker run pinned to the validated `letta/letta:0.16.8` digest (mutable image + internal patch = pin required), host-gateway alias, guard patch with a **behavioral** verification (in-container interpreter must accept a private-IP MCP URL), health-wait, loopback-only port publish, `LETTA_BASE_URL` export; `letta` lane steps in `e2e.yml` with `always()` container teardown.
- Builder self-hosts by default (`MCP_SERVER_URL` switches to external); `LETTA_EMBEDDING` + `LETTA_MCP_ADVERTISED_HOST` settings; `Dep.LETTA` also gates `OPENAI_API_KEY` for self-hosted servers.
- `e2e_pending` hardened: reason-string form (`str | None`) with an empty-allowlist drift guard; peer validation decoupled from the live registry.
- Placeholder `test_letta.py` replaced by two supplementary smokes (registration health; relay-disabled send-path proof), with explicit pytest-timeout headroom.
- `LocalMCPServer` accepts an explicit non-loopback bind; `create_band_mcp_backend` gains `host`/`port_min`/`port_max` passthrough.

**Codebase-wide standards applied on this branch** (`refactor: pydantic-settings everywhere...`)
- New coding standards in AGENTS.md: config via pydantic-settings `BaseSettings` (never `os.environ` — applied to `docker/letta/runner.py`, the Letta example, and the live integration test; `pydantic-settings` promoted to a core dependency), `match`/`case` over one-value if/elif chains, no underscores in file/class names (`converters/_utils.py` → `helpers.py`, `converters/_tool_parsing.py` → `parsing.py`), and no tautological tests (default-value restatement tests removed).
- `tests/adapters/test_letta_adapter.py` split: MCP wiring/lifecycle → `test_letta_mcp.py`, shared factories → `lettakit.py` (80 tests preserved).

## Related Issues

Relates to INT-944 (parent INT-911). Follow-up filed: INT-946 (band-mcp rebrand: repo home + `BAND_*` base-URL alias — the compose example documents the `THENVOI_BASE_URL` leftover until then).

## Testing

- [x] Unit + conformance suites pass (3522 passed)
- [x] `ruff check` / `ruff format` / `pyrefly` / markdown-docs pass
- [x] Live lane validated against dockerized Letta 0.16.8 + real platform: 14 matrix cells + 2 smokes green, including both L4 rehydration cells and the rejoin-recall cell (cold-boot seeding proven live); the relay-disabled smoke proves replies travel through MCP `band_send_message`
- [x] Live integration probe (`tests/integration/test_letta_live.py`) green — no skips, fail-loud
- [x] Smokes re-validated live after the integrations split and the setup-script hardening
- [x] CI dispatch: Actions → E2E (baseline) → lane `letta`

## Checklist

- [x] Documentation updated as needed


amit-gazal-band (approved):
LGTM

PR #418, opened 2026-07-09
fix: Windows support for band-sdk-python [INT-947]

AlexanderZ-Band:
## Summary

Add **Windows support** to `band-sdk-python` and get the unit + live‑E2E suites green on **both Windows and Linux** (Python 3.11 / 3.12). All fixes are real runtime/adapter/toolkit bugs, not test‑only shims.

## Windows runtime

- **Shutdown deadlock (headline):** on the Windows Proactor loop, a cancel of the per‑room processing loop could be silently lost and hang the process. Two CPython behaviours Linux CI never exercised — `asyncio.wait_for` dropping cancellation ([bpo‑43389](https://bugs.python.org/issue43389)) and a 3.11‑only cancel↔trace‑function defect ([cpython#106749](https://github.com/python/cpython/issues/106749)) — compounded by our own `suppress(CancelledError)`. Fixed by moving the Phase‑2 loop to `async with asyncio.timeout(...)` and replacing `suppress(CancelledError)` with `gather(..., return_exceptions=True)` (`runtime/execution.py`, `working_state.py`, `mcp_server.py`).
- **Paths / files:** `docker/repo_init.py` uses cross‑platform `filelock` (was Unix‑only `fcntl`) and accepts a POSIX‑ or Windows‑absolute `repo.path`; `adapters/codex.py` times turns with `perf_counter()`.
- **Signals:** `runtime/shutdown.py` degrades to `signal.signal()` on the Proactor loop instead of raising `NotImplementedError`.
- **CI:** `ci.yml` runs the unit suite on Windows + Linux (3.11/3.12); `e2e.yml` runs the baseline matrix on both OSes via an `os` dispatch input; backend setup scripts made Git‑Bash portable.

## E2E all‑green (both OSes)

Root‑cause fixes so the live baseline matrix passes on Linux and Windows:

- **e2e toolkit (reply‑capture race):** a reply's `message_created` frame and the delivery‑status `message_updated`→PROCESSED frame are independent, unordered platform events, so asserting on `capture.messages` right after `wait_for_processed` raced the reply — the cross‑adapter "no agent messages captured" flake. Added `wait_for_reply` (waits for the reply frame and returns it), migrated every reply‑asserting site, tightened the `flaky(reruns=2)` masks that hid it, and added a deterministic unit guard.
- **runtime (mentions):** translate `@[[uuid]]` mentions in the current message to `@handle` before the LLM sees them, so an agent stops grabbing its own UUID as a memory `subject_id`.
- **prompts** — state the delivery contract generally (only a `band_send_message` call reaches the room).
- **pydantic_ai** — pass the rendered prompt as `instructions=` (re‑attached every request) so a custom every‑reply rule survives the tool loop; raise validation `retries` 1→3.
- **opencode** — seed a freshly‑created server session with in‑session history (fixes turn‑2 recall).
- **claude_sdk** — isolate the bridged agent from ambient host config (`setting_sources=[]`, configurable) so its tool set is deterministic across machines/runners.
- **agno** — frame the Band contract ahead of the developer's own instructions; surface swallowed run errors so a failed turn is reported and retried instead of silently completing with no reply.
- **e2e harness** — derive the provisioned‑name run‑id from the 24‑char mention‑handle cap (with a PR‑run guard); crewai_flow accepts peer‑initiated turns.

## Related

[INT‑947](https://linear.app/thenvoi/issue/INT-947/sdk-python-sdk-windows-support-validation) — “[SDK] Python SDK — Windows support + validation”.

## Testing

- Unit suite green on Windows + Linux (3.11 / 3.12); `ruff` + `pyrefly` clean.
- Live baseline E2E matrix runs on both OSes via `e2e.yml`.
- `loop_suppression[agno]` / `memory_subject_scope_inferred[agno]` were flaky; root‑caused (reply‑capture race + agno swallowing run errors + mention‑UUID leaking into `subject_id`) and fixed — both passed in the full‑matrix run.

- [x] Unit tests pass (`uv run pytest tests/ --ignore=tests/integration/`)
- [x] Pre‑commit checks pass (`ruff` / `pyrefly`)
- [x] Live E2E baseline on Windows + Linux


src/band/adapters/agno.py
```diff
@@ -575,28 +604,38 @@ async def _resolve_user_tools(self, run_context: Any) -> list[Any]:
...
+        ``description`` -> ``<instructions>`` -> ... -> ``additional_context``, so
+        writing the Band block to ``description`` puts it first -- ahead of the
+        developer's ``instructions`` -- reproducing that ordering.
+
+        The previous approach appended to ``additional_context``, which lands
+        *after* the developer's instructions. That made the anti-injection Security
+        rule the most-recent system content, so even a capable model over-applied it
+        and refused benign peer requests (e.g. a liveness probe from another
+        participant). Prepending (not replacing) preserves any ``description`` the
+        developer set. Called once at startup, before any room runs.
```
amit-gazal-band, line 621: Nice catch

src/band/preprocessing/default.py
```diff
@@ -60,11 +60,14 @@ async def process(
...
-        # Convert to PlatformMessage (typed attribute access, no dict lookups)
+        # Convert to PlatformMessage (typed attribute access, no dict lookups).
+        # Translate @[[uuid]] mention tokens to @handle like history formatting
+        # does: the platform normalizes mentions to uuid form, and a raw uuid in
+        # the turn misleads the LLM (e.g. into using it as a memory subject_id).
         msg = PlatformMessage(
             id=msg_data.id,
             room_id=room_id,
-            content=msg_data.content,
+            content=replace_uuid_mentions(msg_data.content, ctx.participants),
```
amit-gazal-band, line 70: Great catch!

amit-gazal-band (approved):

PR #415, opened 2026-07-08
feat: GitHub Copilot CLI over ACP + TCP transport [INT-945]

AlexanderZ-Band:
## Summary

GitHub Copilot CLI as an **ACP client backend**, plus a **TCP transport** on the generic
ACP client layer. Copilot speaks vanilla ACP, so it rides the existing `ACPClientAdapter`
via a thin `CopilotACPAdapter` — the only net-new SDK capability is TCP. Verified live
against `copilot` 1.0.69 over both stdio and TCP.

Linear: https://linear.app/thenvoi/issue/INT-945/add-github-copilot-cli-integration-via-acp (Implements INT-945)

> Stacked on #410 (`copilot_sdk`) for its `GITHUB_TOKEN` plumbing — rebase onto `dev` after #410 merges.

## Changes

- **TCP transport** — `tcp_spawn_process` connect-only seam + `host`/`port` on
  `ACPClientAdapter` (keyword-only; exactly-one-of `command`/`host`+`port`). Zero
  `ACPRuntime` core change (uses the injectable `spawn_process` seam).
- **`CopilotACPAdapter`** (+ config) — thin `ACPClientAdapter`; flexible auth via `env`
  (env token / stored login / `gh` / BYOK; `github_token` is a convenience). Registered in
  the baseline matrix (`backends` lane, gated on the CLI only like codex); excluded from
  framework-conformance as a bridge.
- **Tests** — DI (`FakeSpawn`) instead of monkeypatching; new `acp_toolkit` (a fake ACP
  agent + `acp_adapter` driver + `Reply` view) drives the adapter over a real in-process
  socketpair — real JSON-RPC, LLM faked: text/thought/tool_call+result/plan relay,
  permission auto-approve, room isolation.
- **Examples** — reorganized into `examples/acp/{servers,clients,copilot_docker,copilot_sandbox}/`;
  two Copilot-in-a-container Docker examples (compose + colocated) over TCP with Band tools via a
  `band-mcp` SSE server (`socat` fronts `copilot --acp` on `0.0.0.0`, since `--port` binds loopback
  only; images build), plus a **Docker-sandbox (`sbx`) example over stdio** — validated live
  (`sbx exec -i … copilot --acp` is byte-clean; host secret-proxy auth, token never enters the VM).
- **Docs** — AGENTS.md ACP section (transports, Copilot backend, auth, Docker examples).

## Testing

- [x] Unit suite (3582 passed); `ruff` / `ruff format` / `pyrefly` clean
[...]

PR #420, opened 2026-07-10
feat: baseline scorecard — N/A reasons on exclude + CI artifact …

AlexanderZ-Band:
## Summary

Baseline E2E scorecard: N/A reasons on exclusions + a merged CI artifact. Plus the `ci-required` aggregate gate that makes the unit/conformance CI a requireable single context (INT-975).

Linear: [INT-820](https://linear.app/thenvoi/issue/INT-820) · [INT-985](https://linear.app/thenvoi/issue/INT-985) · [INT-975](https://linear.app/thenvoi/issue/INT-975)

> **Stacked on #415** for its Windows single-instance guard fix (`msvcrt`) — without it `test (windows-latest, *)` is red. Auto-retargets to `dev` after #415 merges.

## Changes

- **Scorecard** — per-adapter×test grid with N/A + reason on exclusions; per-lane JSON merged into one artifact at session end.
- **`ci-required`** (`ci.yml`) — `needs: [lint, test, test-crewai, packaging]`, `if: always()`, fails unless all succeed. One stable context covering the whole matrix, for branch protection.

## Follow-up (repo-admin, after this reaches `dev`/`main`)

Point both rulesets at `ci-required`: add it on `dev`; on `main` replace the broken `test (3.11)`/`test (3.12)`. Keep `lint`, `packaging`, `Validate PR Title`.

PR #421, opened 2026-07-10
ci: make unit/conformance CI a mandatory gate via ci-required [INT-975]

AlexanderZ-Band:
## Summary

Adds a single aggregate **`ci-required`** job to `ci.yml` so branch protection can require *one stable context* that actually reports — instead of the matrix per-cell names (`test (ubuntu-latest, 3.11)`, …) or the bare `test (3.11)`/`test (3.12)` that never report and sit "Expected — Waiting" forever.

`ci-required` `needs: [lint, test, test-crewai, packaging]`, runs `if: always()`, and fails unless every dependency resolved to `success`. A matrix `needs:` is green only when all its cells passed, so this one context covers the whole OS/Python matrix and survives future matrix changes with no further ruleset edits.

Linear: https://linear.app/thenvoi/issue/INT-975/make-ci-checks-mandatory-reconcile-required-status-contexts-with (Implements INT-975)

> **Stacked on #415** (`int-945`) for its Windows single-instance guard fix (`msvcrt` locking). Without it, `test (windows-latest, *)` is red — the guard was `fcntl`-only and inert on Windows — so a mandatory gate would block every PR. Rebase onto `dev` after #415 merges.

## Changes

- **`ci.yml`** — new `ci-required` aggregate gate job (the only net change on this branch; everything else is inherited from #415).

## Follow-up (needs repo-admin, done separately, after this lands on `dev`/`main`)

Point both rulesets at `ci-required`:
- `dev-branch-protection`: add `ci-required` (dev requires no test context today).
- `main-branch-protection`: replace the broken `test (3.11)`/`test (3.12)` with `ci-required`; keep `lint`, `packaging`, `Validate PR Title`.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -191,3 +191,29 @@ jobs:
         from band.config import load_agent_config
         print('All imports successful')
         "
+
+
+  # Single aggregate gate for branch protection. Branch rulesets require exactly
+  # this one context: `needs` a matrix job is green only when every one of its
+  # OS/Python cells passed, so requiring `ci-required` covers the whole matrix and
+  # survives future matrix changes without re-editing the ruleset. `if: always()`
+  # runs it even when a dependency failed (a skipped/failed dep would otherwise
+  # skip this job and report nothing, leaving the required check stuck "Expected").
+  ci-required:
+    name: ci-required
+    needs: [lint, test, test-crewai, packaging]
+    if: always()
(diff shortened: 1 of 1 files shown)


AlexanderZ-Band:
Duplicate — the ci-required gate is folded into #420 (int-820), which is now rebased onto #415 for the Windows single-instance guard fix.

PR #422, opened 2026-07-12
test: add isolated baseline conformance harness

AlexanderZ-Band:
## Summary
- Add offline baseline conformance harness and Anthropic injection seam.
- Add deterministic scenarios for platform tools, memory, contacts, events, rehydration, failures, isolation, and replay handling.
- Share adapter identity/discovery between offline and E2E baseline suites.

## Validation
- Full offline unit suite passed
- Ruff and Pyrefly passed


tests/baseline/tools.py
```diff
@@ -0,0 +1,181 @@
...
+            case "band_respond_contact_request":
+                action = self._required_text(tool_name, arguments, "action")
+                result = await self.respond_contact_request(
+                    action,
+                    handle=arguments.get("handle"),
+                    request_id=arguments.get("request_id"),
+                )
+                self.contact_requests.append({"action": action, "result": result})
+                return result
+            case "band_list_memories":
```
amit-gazal-band, line 139: `band_list_memories`, `band_get_memory`, `band_supersede_memory`, and `band_archive_memory` hand-roll logic directly against `self.memories` instead of calling `FakeAgentTools.list_memories`/`get_memory`/`supersede_memory`/`archive_memory` (the way `store_memory` above does). The class docstring says tool calls are dispatched "through platform methods" / "exactly as they receive runtime tools" — that's true for `store_memory` but not for these four, so this harness won't catch a divergence between this hand-rolled logic and the real `FakeAgentTools` implementations for memory reads/supersede/archive. Not blocking, but worth wiring through the base class (or a comment noting the intentional divergence) in a follow-up.

amit-gazal-band (approved):
Solid addition — verified the new tests pass locally and ruff is clean. See inline comment for one non-blocking gap worth a follow-up.

tests/baseline/tools.py
AlexanderZ-Band: Declining — delegating to `FakeAgentTools` would reduce fidelity, not increase it.

- `supersede`/`archive`: base `FakeAgentTools.supersede_memory`/`archive_memory` (`src/band/testing/fake_tools.py:270-274`) return `{id, status}`, but the real SDK returns the full memory dict (`response.data`, `src/band/runtime/tools.py:2069-2110`) — which is what this hand-roll already does. The base fake isn't authoritative here; the platform is.
- reads need `BaselineTools`' own state: base `list_memories` returns `[]` and `get_memory` returns a canned memory (`fake_tools.py:216-269`), so delegation can't give the stateful behaviour these scenarios rely on.

Separate follow-up (out of scope here): both fakes' `list_memories` envelope is `{memories, metadata}`, while the real SDK returns `{data, meta}` (`ListAgentMemoriesResponse`, `tools.py:1934-1985`; platform `MemoriesJSON.index`). That's a genuine divergence worth fixing on `FakeAgentTools` itself.

AlexanderZ-Band: Follow-up ticket for the `list_memories` envelope fix: INT-1008 — https://linear.app/thenvoi/issue/INT-1008/fix-list-memories-envelope-in-fakeagenttools-to-match-real-sdk

PR #423, opened 2026-07-12
fix: pair permissions and hide self-reporting ACP tools

AlexanderZ-Band:
## Summary

Fixes ACP room transcript handling for permission requests and self-reporting Band tools.

- Permission requests now emit matched tool_call and tool_result events with a shared call ID and explicit outcome.
- ACP narration for self-reporting Band tools (band_send_message, band_send_event, and standalone MCP message names) is suppressed so visible room output is not duplicated or ordered after its own reply.
- Adds shared name matching and deterministic ACP/MCP behavior coverage.

## Validation

- Ruff check: passed
- Pyrefly check: 0 errors
- ACP and runtime-focused tests: 350 passed
- Full unit suite (rebased onto current dev): 3811 passed

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -191,29 +191,3 @@ jobs:
         from band.config import load_agent_config
         print('All imports successful')
         "
-
-
-  # Single aggregate gate for branch protection. Branch rulesets require exactly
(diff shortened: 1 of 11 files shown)


AlexanderZ-Band:
Superseded by #439, which combines this with #437 into a single PR against dev.

PR #434, opened 2026-07-13
fix: settle freshly-joined peer before first gateway message [INT-969]

AlexanderZ-Band:
## Summary

Adds `new_participant_settle_seconds` (default `3.0`): a pause after a peer joins and before its first message. Warm rooms are unaffected; set `0` to disable.

## Scope

This is a temporary mitigation, not an exactly-once guarantee. The durable fix is a platform-side exclusive processing claim/lease so an in-flight message cannot be re-served.

## Validation

Covers fresh joins, warm-room reuse, zero-delay opt-out, and concurrent requests for one context.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -202,6 +202,15 @@ from band.adapters.a2a_gateway import A2AGatewayAdapter, GatewayServer
 adapter = A2AGatewayAdapter(port=10000)
 ```
 
+`new_participant_settle_seconds` (default `3.0`) pauses after a peer is freshly
+added to a room and before the first message is posted, giving the peer's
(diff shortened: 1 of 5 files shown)


AlexanderZ-Band:
The CodeQL findings flag missing `permissions:` blocks in `.github/workflows/ci.yml` and `e2e.yml`. Those files aren't touched by this PR (it only changes the A2A gateway adapter, its tests, and the docs) — they're pre-existing repo-wide alerts surfaced on the diff. Out of scope for this bugfix; hardening the workflow token permissions should be a separate focused PR so it doesn't ride on an unrelated change. Not blocking here.

PR #437, opened 2026-07-14
fix: cap oversized ACP event content instead of failing the send

AlexanderZ-Band:
## Summary
- The Agent Events API (`POST /agent/chats/{id}/events`) hard-caps `content` at 16384 chars and rejects anything larger with a 422 — the ACP/GitHub Copilot backend hit this when mirroring large `tool_call`/`tool_result` payloads (e.g. a GitHub MCP `get_file_contents` call on a big file), causing the send to fail outright instead of degrading gracefully.
- The 16384 cap isn't an arbitrary number tied to a technical limit — it's a deliberate guardrail. Every event broadcasts over the room's WebSocket channel to every connected socket, so an unbounded `content` string is a fan-out amplification vector. Per the platform source (`thenvoi-platform/lib/thenvoi_com_web/controllers/api/v1/agent/events_controller.ex:34-39`): *"The caps are generous ceilings (no legitimate event hits them), purely to stop pathological payloads from being broadcast."* The Copilot CLI's large tool_result is exactly the legitimate-but-oversized case that guardrail didn't anticipate.
- `AgentTools.send_event` (`src/band/runtime/tools.py`) — the single chokepoint every adapter funnels through — now truncates oversized content to the cap (keeping the head, dropping the tail with a `"... [truncated]"` marker) before sending, rather than throwing.
- Fixes the root cause generically for every adapter, not just ACP, since all `send_event` calls share this one path.

Linear: INT-1014

## Test plan
- [x] `uv run pytest tests/runtime/test_tools.py -k SendEvent -v --no-cov` — new truncation tests pass (pass-through under the cap, oversized content truncated with marker)
- [x] `uv run pytest tests/ --ignore=tests/integration/ --ignore=tests/e2e/ -q --no-cov` — full suite (3800 passed, 75 skipped)
- [x] `uv run pyrefly check`

### DIFF
--- a/src/band/runtime/tools.py
+++ b/src/band/runtime/tools.py
@@ -40,6 +40,29 @@ if TYPE_CHECKING:
 
 logger = logging.getLogger(__name__)
 
+# The Agent Events API enforces a hard cap on event content (see
+# thenvoi-platform's events_controller.ex `@content_max_length`) and rejects
+# anything larger with a 422 before it ever reaches the room. Event content
+# can be arbitrarily large in practice — e.g. an ACP tool_result mirroring a
+# large file, or a raw exception dump — so truncate defensively rather than
+# letting the send fail.
+_EVENT_CONTENT_MAX_LENGTH = 16384
+_EVENT_TRUNCATION_MARKER = "... [truncated]"
+
+
+def _truncate_event_content(content: str) -> str:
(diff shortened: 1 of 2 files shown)


AlexanderZ-Band:
Superseded by #439, which combines this with #423 into a single PR against dev.

PR #439, opened 2026-07-14
fix: pair ACP permissions, hide self-reporting tools, cap oversized event content

AlexanderZ-Band:
## Summary

- **Permission pairing** — a permission request emits a matched `tool_call`/`tool_result` pair (shared call id, explicit outcome), posted only for self-reporting Band tools (`band_send_message`/`band_send_event`) or denied requests. An ordinary approved tool already shows its own call, so no duplicate bubble.
- **Self-reporting suppression + ordering** — ACP narration for self-reporting Band tools is suppressed so a visible room reply isn't duplicated or ordered after its own reply.
- **Streamed `tool_result` collapse** — a call streamed over several `tool_call_update` frames now posts one `tool_result` (keyed by `tool_call_id`, latest reported status wins, readable content-block text preferred over the raw `rawOutput` dict) instead of one bubble per frame. A status-less trailing frame no longer regresses a recorded `completed` (which would otherwise relay the agent's text again).
- **Event content cap** — `AgentTools.send_event` guards content at the shared chokepoint (the 16384-char Agent Events API cap) instead of failing with a 422: oversized content is truncated keeping head + tail, blank content gets a placeholder. ACP `tool_result` text is recovered from `content` blocks when `rawOutput` is unset.

Combines #423 and #437 into one branch/PR against `dev`; both source PRs close as superseded.

## Linear

- [INT-1005](https://linear.app/thenvoi/issue/INT-1005) — orphaned permission `tool_call` + reply-before-narration ordering
- [INT-1014](https://linear.app/thenvoi/issue/INT-1014) — 16384-char event content 422
- Follow-up (backlog, not in this PR): [INT-1093](https://linear.app/thenvoi/issue/INT-1093) — shared tool-result collapser refactor

## Test plan

- [x] `uv run pytest tests/runtime/test_tools.py tests/integrations/acp/ -q --no-cov` (ACP 245 passed, 8 skipped)
- [x] Live re-test on DEV (Copilot ACP): one readable `tool_result` per call, no duplicate permission bubble, no duplicate text reply


amit-gazal-band (commented):
I found an ACP related bug. So I opened this Linear issue - https://linear.app/thenvoi/issue/INT-1095/acp-client-adapter-doesnt-fall-back-to-band-history-when-a-remote.

I want to approve and merge this PR since we need GitHub CLI support in main.
The bug will be fixed separately.

src/band/integrations/acp/room_emitter.py
```diff
@@ -0,0 +1,168 @@
...
+            chunk.chunk_type == ChunkType.TOOL_RESULT
+            and call_id in posting_call_ids
+            and metadata.get("status") == ToolStatus.COMPLETED
+        ):
+            return True
+    return False
+
+
+class RoomTurnEmitter:
+    """Posts one ACP turn's output to a Band room in causal order.
```
amit-gazal-band, line 51: Non-blocker: the causal-order guarantee documented here isn't enforced by the transport. The `acp` package's dispatcher spawns a separate `asyncio.create_task` per incoming notification (`acp/task/dispatcher.py::_dispatch_notification`), so two consecutive `session_update`s run as concurrent tasks — their `send_event` REST calls are in flight simultaneously and can commit on the platform out of order. `_ingest`'s per-session state also interleaves at its internal awaits. The unit tests can't catch this because `FakeAgentTools` resolves without real I/O.

Suggested fix: a per-session `asyncio.Lock` (or single-worker queue) around the ingest+sink path, shared with the denied-permission pair posting in `request_permission`. Fine to do in this PR, or file a Linear issue and fix later — in the meantime consider softening this docstring to "best-effort ordering."

src/band/integrations/acp/client_runtime.py
```diff
@@ -164,74 +248,248 @@ def __init__(self, profile: ACPClientProfile | None = None) -> None:
...
+            best.from_raw,
+            best.echo,
+        )
+
+    async def _finalize(self, session_id: str, chunk: CollectedChunk) -> None:
+        """Buffer a finalized chunk and post it to the session's live sink, if any."""
+        self._session_chunks.setdefault(session_id, []).append(chunk)
+        sink = self._sinks.get(session_id)
+        if sink is not None:
+            await sink(chunk)
```
amit-gazal-band, line 489: Non-blocker: an exception raised here (e.g. a failed `send_event` REST call in the sink) is silently swallowed — the `acp` library wraps notification dispatch in `contextlib.suppress(Exception)` (`acp/connection.py::_run_notification`), with no log line and nothing reaching the supervisor's error handler. Before this PR every room post happened in `on_message`, where a REST failure hit the error-event + respawn path; now mid-turn narration failures vanish while the turn continues looking healthy.

Suggested fix: catch around the sink call here and at minimum log the failure (optionally surface a "narration degraded" marker). Fine to do in this PR, or file a Linear issue and fix later.

amit-gazal-band (approved):
Approving — the live causal-emission design is solid and well tested. Two non-blocking findings inline (both about guarantees the `acp` library's notification dispatch doesn't actually provide); each can be fixed in this PR or tracked in a Linear issue and fixed later.

src/band/integrations/acp/room_emitter.py
AlexanderZ-Band: Fixed in 0eb45d95 — chose to enforce the guarantee rather than soften the docstring. `ACPCollectingClient` now holds a per-session `asyncio.Lock` around the ingest→sink path (`session_update`, `ext_notification`, `flush`) and the permission handler, so the denied pair can't interleave with narration either. Lock waiters wake FIFO and the dispatcher creates notification tasks in wire-arrival order, so causal order holds end to end. Regression tests dispatch updates as concurrent tasks with a yielding sink (the real-I/O suspension your finding pointed at) and fail on the pre-fix code with a 3-way overlap.

src/band/integrations/acp/client_runtime.py
AlexanderZ-Band: Fixed in 0eb45d95 — the sink call is now guarded at its single chokepoint (`_finalize`): a failed narration post is logged via `logger.exception` (with session id + chunk type) and the chunk stays buffered. Policy made explicit in the docstring: mid-turn narration is best-effort-but-logged, while the turn's load-bearing reply still posts from `on_message`'s own task where a failure keeps reaching the error-event + respawn path. Regression test: a raising sink no longer escapes `session_update` (where the acp transport would swallow it) and leaves an ERROR trace.

PR #442, opened 2026-07-16
feat: remote agents honor interrupt/stop/play control signals

amit-gazal-band:
## Summary
- SDK joins `agent_control:{agent_id}` and handles `interrupt`/`stop`/`play` preemptively, out-of-band from the message queue, so a signal can abort a reasoning cycle already in flight (`ExecutionContext`/`AgentRuntime`).
- `interrupt` aborts the in-flight cycle and consumes the message (nothing sent, no replay). `stop` does the same plus durably suppresses new cycles in that room until `play`, which triggers a rehydration-style `/next` catch-up.
- Bridge (`band-bridge`) honors the same signals for one-shot/bridged agents: `interrupt`/`stop` cancel whatever forward call is currently in flight for a room; `play` proactively nudges the room via `/next` instead of waiting for the next natural event.
- Fix: the `ReconnectedEvent` sync path now short-circuits locally on `_stopped`, consistent with the other stopped-guards, instead of making a `/next` call already known to return empty.

## Test plan
- [x] `uv run pytest tests/ --ignore=tests/integration/ --ignore=tests/e2e/` — full unit suite green
- [x] `uv run pytest tests/runtime/ tests/platform/ tests/websocket/ tests/bridge/ tests/test_platform_runtime.py` — control-signal suites green (851 passed)
- [x] `uv run ruff check .` / `uv run ruff format --check .` / `uv run pyrefly check` — clean
- [ ] Live integration tests (`tests/integration/test_control_signals.py`) — require a live platform + `BAND_API_KEY`/`BAND_API_KEY_USER`; not run in CI


band-bridge/bridge_core/bridge.py
```diff
@@ -382,20 +403,33 @@ def _backlog_event(self, msg: PlatformMessage) -> MessageEvent:
...
-                )
+        # Track this task as the in-flight forward for its room so a control
+        # signal (interrupt/stop) can cancel it. Room-less events (e.g.
+        # contact events) aren't tracked — there's nothing room-scoped to
+        # cancel for them. Guarded pop in `finally`: a newer event for the
+        # same room may already have replaced this entry.
+        room_id = getattr(event, "room_id", None)
+        task = asyncio.current_task()
+        if room_id and task is not None:
+            self._active_room_tasks[room_id] = task
```
AlexanderZ-Band: `_active_room_tasks` is set before the semaphore and per-room lock. A second same-room task can overwrite the active forward while it waits, so interrupt cancels the waiter and the real forward continues. Register the task only while it owns the room lock, or track all tasks per room.

band-bridge/bridge_core/bridge.py
```diff
@@ -533,6 +567,94 @@ def _remember_processed_message(self, message_id: str) -> None:
...
+                self._cancel_active_forward(room_id)
+        elif payload.mode == "play":
+            await asyncio.gather(*(self._nudge_room(rid) for rid in room_ids))
+
+    def _cancel_active_forward(self, room_id: str) -> None:
+        """Cancel the in-flight forward task for a room, if any.
+
+        Best-effort: this stops the bridge from waiting and releases the room
+        lock immediately. It does not guarantee the remote invocation itself
+        stops — ``AgentCoreForwarder``'s underlying boto3 call runs in a
```
AlexanderZ-Band, line 611: Cancelling this await does not cancel `asyncio.to_thread()` or the AgentCore invocation. The remote cycle can keep consuming tokens and running tools, so this does not meet the in-flight halt requirement. Add remote cancellation or exclude this transport from that guarantee.

band-bridge/bridge_core/bridge.py
```diff
@@ -533,6 +567,94 @@ def _remember_processed_message(self, message_id: str) -> None:
...
+            payload.scope,
+            len(room_ids),
+            cid,
+        )
+
+        if payload.mode in ("interrupt", "stop"):
+            for room_id in room_ids:
+                self._cancel_active_forward(room_id)
+        elif payload.mode == "play":
+            await asyncio.gather(*(self._nudge_room(rid) for rid in room_ids))
```
AlexanderZ-Band: `play` awaits the full `/next` forward on the WebSocket callback task. While that forward runs, this receive path cannot process a following stop or interrupt. Schedule the nudge as a tracked background task.

src/band/runtime/execution.py
```diff
@@ -1387,6 +1583,81 @@ async def _execute_message_cycle(self, event: PlatformEvent) -> None:
...
+            if kind == "interrupt" and msg_id:
+                # Consume the message so the idle /next resync does not
+                # re-return it (excludes-only-processed) and re-fire the cycle
+                # the user just interrupted. Mirror the success-path bookkeeping.
+                if await self.link.mark_processed(self.room_id, msg_id):
+                    self._remember_processed_message(msg_id)
+                else:
+                    logger.warning(
+                        "ExecutionContext %s: durable mark_processed failed for "
+                        "interrupted message %s; idle resync may re-deliver it",
```
AlexanderZ-Band: If this ack fails, the interrupted message remains locally replayable and `/next` can run it again. Record `_remember_processed_ack_pending(msg_id)` so the interrupt stays consumed while the durable ack retries.

src/band/runtime/execution.py
```diff
@@ -1522,12 +1816,13 @@ async def _process_event(self, event: PlatformEvent) -> bool:
...
-            if isinstance(event, MessageEvent):
-                await self._execute_message_cycle(event)
-            else:
-                await self._on_execute(self, event)
+            # Call execution handler as a cancellable cycle. A control signal
+            # can abort just this turn; when it does, status is handled inside
+            # _run_cycle and we send nothing further. Only message-driven
+            # cycles report the working signal (see _invoke_handler);
+            # participant add/remove events are housekeeping and skip it.
+            if not await self._run_cycle(event, msg_id):
```
AlexanderZ-Band, line 1872: A control signal during hydration or `mark_processing()` sees no `_active_cycle_task`: interrupt is lost, and stop can still reach this call without another `_stopped` check. Cover the full claimed-message lifecycle or re-check control state immediately before starting the handler.

src/band/runtime/retry_tracker.py
```diff
@@ -55,6 +55,14 @@ def mark_success(self, msg_id: str) -> None:
         """Clear tracking for successfully processed message."""
         self._attempts.pop(msg_id, None)
 
+    def discard_attempt(self, msg_id: str) -> None:
+        """Uncharge an attempt aborted by our own control signal (interrupt/stop).
+
+        A cycle cancelled by interrupt/stop never actually ran the handler, so
+        it must not count against the message's retry budget.
+        """
+        self._attempts.pop(msg_id, None)
```
AlexanderZ-Band: `pop()` removes every prior failed attempt, not just the control-cancelled attempt. Decrement by one so interrupting a retry does not reset the message's entire retry budget.

band-bridge/bridge_core/bridge.py
```diff
@@ -533,6 +567,94 @@ def _remember_processed_message(self, message_id: str) -> None:
         if len(self._processed_message_ids) > _DEDUP_MAX_SIZE:
             self._processed_message_ids.popitem(last=False)
 
+    # --- Control signals (interrupt/stop/play) ---
+
+    async def _handle_control(self, payload: AgentControlPayload) -> None:
```
AlexanderZ-Band: This method does three things: dedup, room resolution, mode dispatch. Split out `_is_duplicate_control(cid)` and `_resolve_control_rooms(payload)` to make it easier to read.

band-bridge/bridge_core/bridge.py
```diff
@@ -533,6 +567,94 @@ def _remember_processed_message(self, message_id: str) -> None:
...
+            "Agent %s: applying control mode=%s scope=%s to %d room(s) "
+            "(correlation_id=%s)",
+            self._config.agent_id,
+            payload.mode,
+            payload.scope,
+            len(room_ids),
+            cid,
+        )
+
+        if payload.mode in ("interrupt", "stop"):
```
AlexanderZ-Band: Use `match`/`case` here (repo convention in CLAUDE.md). Add a `case _:` that logs unknown modes — right now they are silently ignored.

band-bridge/bridge_core/bridge.py
```diff
@@ -382,20 +403,33 @@ def _backlog_event(self, msg: PlatformMessage) -> MessageEvent:
         )
 
     async def _safe_handle_event(self, event: object) -> None:
```
AlexanderZ-Band, line 414: Too many nested levels. Move the inner `try/except` up to the outer `try` — same behavior, flatter code.

AlexanderZ-Band:
`AgentRunner` does forwarding, rehydration, and control signals in one ~800-line file. The control-signal part is self-contained — consider moving it to its own module as a follow-up.

tests/integration/test_control_signals.py
```diff
@@ -0,0 +1,287 @@
...
+
+import asyncio
+import logging
+import time
+from collections.abc import Callable
+
+import httpx
+import pytest
+import pytest_asyncio
+from thenvoi_rest.human_api_chats.types.create_my_chat_room_request_chat import (
```
AlexanderZ-Band: This import fails — the package was renamed to `band_rest` (your conftest already uses it). The file doesn't collect: `ModuleNotFoundError: No module named 'thenvoi_rest'`.

tests/integration/test_control_signals.py
```diff
@@ -0,0 +1,287 @@
...
+    async with httpx.AsyncClient(timeout=30.0) as client:
+        return await client.post(f"{get_base_url()}{path}", headers=_user_headers())
+
+
+async def _get(path: str) -> httpx.Response:
+    async with httpx.AsyncClient(timeout=30.0) as client:
+        return await client.get(f"{get_base_url()}{path}", headers=_user_headers())
+
+
+async def _send_user_mention(chat_id: str, agent_id: str, text: str) -> str:
```
AlexanderZ-Band: No need for raw httpx here — the typed client has `human_api_messages.send_my_chat_message` with mentions, and `user_api_client` is already available. Raw httpx is only justified for the control endpoints.

tests/integration/test_control_signals.py
```diff
@@ -0,0 +1,287 @@
...
+            headers=_user_headers(),
+            json=body,
+        )
+    assert resp.status_code in (200, 201), (
+        f"send failed: {resp.status_code} {resp.text}"
+    )
+    return resp.json()["data"]["id"]
+
+
+async def _wait_until(
```
AlexanderZ-Band: Generic polling helper, nothing control-specific — put it in `tests/conftest_integration.py` so the next integration test doesn't grow its own.

tests/integration/test_control_signals.py
```diff
@@ -0,0 +1,287 @@
...
+) -> bool:
+    deadline = time.monotonic() + timeout
+    while time.monotonic() < deadline:
+        if pred():
+            return True
+        await asyncio.sleep(interval)
+    return pred()
+
+
+class _ControllableHandler:
```
AlexanderZ-Band: Same blocking-handler pattern (started/cancelled events + hang until cancelled) is hand-rolled many times in `tests/runtime/test_execution_interrupt.py`. Extract one shared fake and use it in both places.

tests/integration/test_control_signals.py
```diff
@@ -0,0 +1,287 @@
...
+            raise
+
+
+# --------------------------------------------------------------------------- #
+# Fixtures
+# --------------------------------------------------------------------------- #
+
+
+@pytest_asyncio.fixture
+async def control_room(user_api_client, session_api_client, shared_agent1_info):
```
AlexanderZ-Band: This repeats the reuse-or-create + `is_room_alive` pattern from `shared_room` in `conftest_integration.py`. Make it a shared user-owned room fixture there — future control/permission tests will need it too.

AlexanderZ-Band (changes requested):
See inline comments — main blockers: the in-flight cancellation gaps in the bridge (task registration ordering, play blocking the receive path), the interrupt/stop lifecycle windows in execution.py, and tests/integration/test_control_signals.py failing to import (`thenvoi_rest`).

band-bridge/bridge_core/bridge.py
amit-gazal-band: Going with "exclude this transport from that guarantee." For one-shot bridged agents the remote is invoked per-message over AgentCore with no control channel of its own, and boto3 has no in-flight cancellation — so interrupt/stop is best-effort here: it frees the bridge and never forwards the result onward, but can't halt the remote compute (it runs to completion or its own timeout). Documented at the to_thread call.

band-bridge/bridge_core/bridge.py
amit-gazal-band: Fixed in 44fc7388. `_active_room_tasks` is now `dict[str, set[Task]]`: each forward adds itself to its room's set (discarding in `finally`), and `_cancel_active_forward` cancels *every* live task for the room — the one holding the lock and any same-room waiters queued behind it — so interrupt/stop can't cancel an idle waiter while the real forward keeps running.

band-bridge/bridge_core/bridge.py
amit-gazal-band: Went with excluding this transport from the guarantee rather than adding remote cancellation. For one-shot bridged agents the remote is invoked per-message over AgentCore with no control channel of its own, and boto3 has no in-flight cancellation — so interrupt/stop is best-effort here: it frees the bridge and never forwards the result onward, but can't halt the remote compute (runs to completion or its own timeout). Documented at the `to_thread` call in `forwarder.py` (44fc7388). A hard remote-halt would need an AgentCore-side abort API keyed by session id — follow-up if we want it.

band-bridge/bridge_core/bridge.py
amit-gazal-band: Fixed in 44fc7388. `play` now fires the `/next` nudges as tracked background tasks (`_spawn_play_nudge`, tracked in `_control_tasks`, cancelled on `close()`) instead of awaiting them on the receive task, so a following stop/interrupt isn't queued behind the forward. The spawned forwards still register in `_active_room_tasks`, so a later stop can cancel them.

src/band/runtime/execution.py
amit-gazal-band: Fixed in 44fc7388. The interrupt path now calls `_remember_processed_ack_pending(msg_id)` on durable-ack failure (mirroring the success path), so the message stays consumed locally while the ack retries in the background — the idle `/next` resync won't re-fire the interrupted cycle.

src/band/runtime/execution.py
amit-gazal-band: Fixed in 44fc7388. Added `_cycle_armed`/`_pending_interrupt`: the window is armed right before `mark_processing`, and while armed `interrupt()`/`stop_room()` record the request instead of no-op'ing. `_run_cycle` honors it at entry (before creating the task, no intervening await) via the shared `_abort_cycle` unwind; both `finally` blocks clear the flags so nothing leaks to the next cycle. Covered by new `TestControlSignalInClaimWindow` tests.

src/band/runtime/retry_tracker.py
amit-gazal-band: Fixed in 44fc7388. `discard_attempt` now decrements the count by one (popping only at 0) instead of clearing it, so interrupting a retry uncharges just that attempt and leaves earlier genuine failures on the budget.

band-bridge/bridge_core/bridge.py
amit-gazal-band: Done in 44fc7388 — dovetailed with the module split below. Control handling now lives in `ControlSignalHandler` (`control.py`) with concerns separated: `_is_duplicate(payload)` and `_resolve_rooms(payload)`, with `handle()` orchestrating dispatch.

band-bridge/bridge_core/bridge.py
amit-gazal-band: Done in 44fc7388 — `handle()` uses `match`/`case` with a `case _:` that logs a warning for unknown modes.

band-bridge/bridge_core/bridge.py
amit-gazal-band: Done in 44fc7388 — the inner `try/except` in `_safe_handle_event` was merged into the outer `try`.

tests/integration/test_control_signals.py
amit-gazal-band: Fixed in 44fc7388 — imports are now `band_rest`; the file collects and both tests pass locally against the live platform.

tests/integration/test_control_signals.py
amit-gazal-band: Done in 44fc7388 — `_send_user_mention` uses `user_api_client.human_api_messages.send_my_chat_message` with `ChatMessageRequest`/`ChatMessageRequestMentionsItem`. Raw httpx is kept only for the control endpoints not yet in the client.

tests/integration/test_control_signals.py
amit-gazal-band: Done in 44fc7388 — moved to `tests/conftest_integration.py` as `wait_until` and re-exported.

tests/integration/test_control_signals.py
amit-gazal-band: Done in 44fc7388 — extracted `BlockingHandler` into `tests/conftest.py`; the integration test and the hand-rolled handlers in `tests/runtime/test_execution_interrupt.py` now use it.

tests/integration/test_control_signals.py
amit-gazal-band: Done in 44fc7388 — extracted as `shared_user_owned_room` in `tests/conftest_integration.py`. It reuses only rooms it created (tagged with a marker title) so the returned room is always user-owned and room-scope stop/play doesn't 403.

amit-gazal-band:
@AlexanderZ-Band re: splitting the control-signal code out of `AgentRunner` — done as part of this review pass (44fc7388): the control-signal concern is now its own module, `band-bridge/bridge_core/control.py` (`ControlSignalHandler`), which owns dedup + routing and delegates cancel/nudge/room-list back to `AgentRunner` — trimming the runner and isolating the control logic.

AlexanderZ-Band (dismissed):

PR #443, opened 2026-07-16
feat: band-python-kit base image, isolated SDK venv, CA wiring [INT-977]

AlexanderZ-Band:
## Summary

- Adds a multi-stage, digest-pinned `python:3.12` image for arm64 and x86_64 under `docker/band_python_kit/`, with the Python and uv images exposed as build arguments.
- Builds the Band SDK from the repository lockfile into an isolated, read-only venv at its final runtime path. `$BAND_SDK_PYTHON` exposes the absolute interpreter without putting the venv on `PATH`, keeping customer dependencies separate.
- Supports core-only builds by default and one optional framework extra through `SDK_EXTRA`, avoiding known cross-extra dependency conflicts.
- Installs the per-sandbox `PROXY_CA_CERT_B64` into the system trust store, exports `SSL_CERT_FILE` and `REQUESTS_CA_BUNDLE`, then drops from root to the non-root `agent` user before launching the requested command.
- Includes runtime `git` and `openssh-client` support required by `band.docker.repo_init`.
- Adds reusable, opt-in Docker test infrastructure plus isolation and live-platform scenarios, gated by `DOCKER_TESTS_ENABLED` (and `E2E_TESTS_ENABLED` for the live test).
- Adds the Linux arm64 and Windows AMD64 uv lock environments without changing resolved package versions.
- Documents the image in `docker/band_python_kit/README.md` and surfaces it from the repository's main README.

## Notable reliability fixes

- Builds the SDK venv directly at its final path so installed console-script shebangs remain valid.
- Keeps the OCI entrypoint at `/usr/local/bin/band-entrypoint.sh`, independent of configurable `BAND_SDK_HOME`.
- Gives both Docker scenarios the full cold-build timeout budget from the shared Docker toolkit.
- The live fixture relies on the SDK's startup room reconciliation for messages sent during initial WebSocket subscription; this behavior was verified with an injected pre-connect delay.

## Verification

- [x] Default core-only Docker build and container startup
- [x] `SDK_EXTRA` build and dependency isolation from a customer venv pinned to `httpx==0.13.3`
- [x] `BAND_SDK_HOME` override builds and starts successfully
- [x] `band-acp` and `band-trigger` console scripts use the final venv interpreter
- [x] Multi-architecture build for `linux/amd64` and `linux/arm64`
- [x] Synthetic proxy CA trusted automatically by `httpx`; public TLS remains valid
- [x] Entrypoint drops to uid 1000 (`agent`) with and without proxy CA installation
- [x] Containerized core SDK agent replies over the live Band platform
- [x] Full local unit suite: 3,798 passed
- [x] Main README markdown-doc test: 15 passed
- [x] Required GitHub CI checks pass

## Out of scope

- Customer venv creation, `spec.yaml`, `sbx kit validate`, and `sbx run --kit` are separate launcher and kit-spec deliverables.
- Publishing `ghcr.io/thenvoi/band-python-kit` and catalog integration are separate release deliverables.
- Docker Sandbox tests require nested virtualization and remain developer-local; standard GitHub-hosted runners cannot provide the required environment.


AlexanderZ-Band:
Added `8ce00464`: the runtime image now ships the digest-pinned `uv` at a fixed image-owned path (`$BAND_SDK_UV` = `/opt/band/bin/uv`, read-only, off `PATH`). Surfaced during INT-978 preflight — the sandbox launcher must run `uv sync --locked` for the customer venv without downloading uv at runtime, and the binary was previously only in the builder stage. Covered by a new docker-gated test that execs it as the non-root `agent` user (the launcher's actual uid after the entrypoint's privilege drop).

docker/band_python_kit/Dockerfile
```diff
@@ -0,0 +1,122 @@
...
+WORKDIR /build
+COPY pyproject.toml uv.lock README.md ./
+COPY src ./src
+
+# Prefer HTTPS for GitHub URLs to avoid SSH key requirements in container builds.
+RUN git config --global --add url."https://github.com/".insteadOf "git@github.com:" \
+    && git config --global --add url."https://github.com/".insteadOf "ssh://git@github.com/"
+
+# Strip host-specific [tool.uv.sources] overrides so uv resolves from the registry.
+RUN sed -i '/^\[tool\.uv\.sources\]/,/^\[/{ /^\[tool\.uv\.sources\]/d; /^\[/!d; }' pyproject.toml
```
amit-gazal-band: Sed strip of [tool.uv.sources] is currently a no-op — no such section exists in the committed root pyproject.toml. Legitimate as a guard against a dev's local override leaking into the build, but reads as load-bearing when it isn't. Reword the comment to say "defensive guard; no-op against committed state," or drop it. Failure mode is safe (local source + stale lock → --locked fails loud).

docker/band_python_kit/README.md
```diff
@@ -0,0 +1,113 @@
...
+
+## Privilege drop
+
+The container's default process starts as **root** — required for the CA
+install step above, which only root can perform. The entrypoint
+(`entrypoint.sh`) always ends by dropping to `agent` (uid 1000) via
+`setpriv` before `exec`-ing the real command:
+
+```bash
+exec setpriv --reuid=agent --regid=agent --init-groups -- "$@"
```
amit-gazal-band, line 85: setpriv assumed present, never explicitly installed (only bash/ca-certificates/git/openssh-client are). It ships in util-linux (Essential in Debian trixie-slim), so present today — but container startup hard-depends on it. Add explicit util-linux to the runtime apt-get for a base-bump-proof, legible dependency.

docker/band_python_kit/entrypoint.sh
```diff
@@ -0,0 +1,21 @@
...
+# Deliberately thin: reading workspace config, spec.yaml-driven customer venv
+# sync, and agent launch belong to the sandbox launcher, not this image. This
+# stub only wires CA trust and validates the SDK venv before dropping to the
+# non-root runtime user.
+
+# The proxy CA is session-generated per sandbox, so it can only be installed
+# at container start, not baked into the image. Runs while still root.
+if [[ -n "${PROXY_CA_CERT_B64:-}" ]]; then
+  base64 -d <<<"${PROXY_CA_CERT_B64}" > /usr/local/share/ca-certificates/sandbox-proxy-ca.crt
+  update-ca-certificates >/dev/null
```
amit-gazal-band: CA decode has no validation. Invalid base64 already fails loud under set -e; the silent path is a decode that succeeds but isn't PEM → update-ca-certificates skips it → trust silently not wired. Cheap guard: openssl x509 -noout -in <file> and exit 1 on failure. Silent-untrusted is nasty to debug in a sandbox.

tests/docker/echo_agent.py
```diff
@@ -0,0 +1,53 @@
...
+        await tools.send_message(f"echo: {msg.content}", mentions=[msg.sender_id])
+
+
+async def main() -> None:
+    agent = Agent.create(
+        adapter=EchoAdapter(),
+        agent_id=os.environ["BAND_AGENT_ID"],
+        api_key=os.environ["BAND_API_KEY"],
+        ws_url=os.environ["BAND_WS_URL"],
+        rest_url=os.environ["BAND_REST_URL"],
```
amit-gazal-band: `echo_agent.py` reads config via os.environ[...] — repo standard is pydantic-settings BaseSettings.

docker/band_python_kit/Dockerfile
```diff
@@ -0,0 +1,122 @@
...
+      uv sync --locked --no-dev --no-editable --extra "$SDK_EXTRA"; \
+    else \
+      uv sync --locked --no-dev --no-editable; \
+    fi
+
+# ── Runtime stage ─────────────────────────────────────────────────────────
+FROM ${PYTHON_BASE_IMAGE}
+ARG BAND_SDK_HOME
+
+ENV DEBIAN_FRONTEND=noninteractive
```
amit-gazal-band: `DEBIAN_FRONTEND=noninteractive` set via ENV in the runtime stage leaks build config into the runtime image env; conventionally build-only (ARG or inline on apt-get).

amit-gazal-band (approved):

docker/band_python_kit/Dockerfile
AlexanderZ-Band: Done in 9dba682 — comment reworded to state it's a defensive no-op guard against a local override, not load-bearing.

---
_Generated by [Claude Code](https://claude.ai/code)_

docker/band_python_kit/README.md
AlexanderZ-Band: Done in 9dba682 — util-linux now installed explicitly in the runtime stage, with a comment noting it provides setpriv.

---
_Generated by [Claude Code](https://claude.ai/code)_

docker/band_python_kit/entrypoint.sh
AlexanderZ-Band: Done in 9dba682 — entrypoint now validates the decoded CA with `openssl x509 -noout` and exits 1 with a clear error (openssl added to the runtime image). Verified end-to-end: good cert → trust wired + TLS handshake OK; non-PEM payload → loud exit 1 (old script silently continued).

---
_Generated by [Claude Code](https://claude.ai/code)_

tests/docker/echo_agent.py
AlexanderZ-Band: Done in 9dba682 — echo_agent now uses a pydantic-settings BaseSettings (BAND_ prefix). Safe in-container: pydantic-settings is a core band-sdk dependency, so it's in the image's core-only venv.

---
_Generated by [Claude Code](https://claude.ai/code)_

docker/band_python_kit/Dockerfile
AlexanderZ-Band: Done in 9dba682 — DEBIAN_FRONTEND is now a build-only ARG in both stages, so it no longer persists in the runtime image env.

---
_Generated by [Claude Code](https://claude.ai/code)_

PR #444, opened 2026-07-16
feat: add no-LLM LangGraph example

amit-gazal-band:
## Summary
- Add `examples/langgraph/11_no_llm_graph.py`, a deterministic `graph_factory` graph that replies via `band_send_message` with plain Python logic — no LLM, no tool-calling loop.
- Document it: note in `docs/adapters/langgraph.md` that `graph_factory` doesn't require an LLM, fill in the examples table (which was also missing rows for `09`/`10`), add it to `examples/langgraph/README.md`, and register `no_llm_agent` (plus the already-missing `memory_agent`) in `agent_config.yaml.example`.

## Test plan
- [ ] `uv run --extra langgraph python examples/langgraph/11_no_llm_graph.py` against a live Band room, verify "ping" -> "pong" and echo behavior
- [x] Docs cross-checked against the actual file paths/tables edited

### DIFF
--- a/examples/langgraph/11_no_llm_graph.py
+++ b/examples/langgraph/11_no_llm_graph.py
@@ -42,14 +42,32 @@ _SENDER_PATTERN = re.compile(r"^\[(?P<sender>[^\]]+)\]:\s*(?P<text>.*)$", re.DOT
 _MENTION_TOKEN_PATTERN = re.compile(r"@\[\[[^\]]+\]\]\s*")
 
 
+def _parse_latest_message(message: Any) -> tuple[str, str]:
+    """Split the adapter's ``"[sender]: content"`` format back into parts.
+
+    The adapter formats each platform message this way (see
+    ``PlatformMessage.format_for_llm``) before appending it to
+    ``state["messages"]``.
+    """
+    match = _SENDER_PATTERN.match(str(message.content))
+    sender = match.group("sender") if match else "there"
+    text = match.group("text") if match else str(message.content)
+    # Strip raw "@[[uuid]]" mention tokens before echoing: the echoed text
+    # isn't registered in this reply's `mentions` list, so an unresolved
+    # token would render as "@Unknown" in the chat UI.
+    text = _MENTION_TOKEN_PATTERN.sub("", text).strip()
+    return sender, text
+
+
+def _build_reply(text: str) -> str:
+    return "pong" if "ping" in text.lower() else f"You said: {text}"
+
+
 def build_no_llm_graph_factory() -> Any:
     """Build a graph_factory whose only node is deterministic Python logic.
 
-    The adapter formats each platform message as ``"[sender]: content"``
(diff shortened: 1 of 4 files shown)


examples/langgraph/11_no_llm_graph.py
```diff
@@ -0,0 +1,110 @@
...
+            text = match.group("text") if match else str(last.content)
+            # Strip raw "@[[uuid]]" mention tokens before echoing: the echoed
+            # text isn't registered in this reply's `mentions` list, so an
+            # unresolved token would render as "@Unknown" in the chat UI.
+            text = _MENTION_TOKEN_PATTERN.sub("", text).strip()
+
+            reply_text = "pong" if "ping" in text.lower() else f"You said: {text}"
+            await send_message.ainvoke(
+                {"content": f"{reply_text}", "mentions": [sender]}
+            )
```
AlexanderZ-Band: Use a stable sender ID or handle
display names may be missing or duplicated, causing replies to fail or mention the wrong participant.

examples/langgraph/11_no_llm_graph.py
```diff
@@ -0,0 +1,110 @@
...
+    ``state["messages"]``. The node below parses that prefix back out to
+    know who to @mention, then calls ``band_send_message`` directly — no
+    model, no ``bind_tools``, no ``ToolNode``.
+    """
+    checkpointer = InMemorySaver()
+
+    def graph_factory(band_tools: list[Any]) -> Pregel:
+        send_message = next(t for t in band_tools if t.name == "band_send_message")
+
+        async def reply(state: MessagesState) -> dict[str, list[Any]]:
```
AlexanderZ-Band, line 77: `state["messages"][-1]` is hard to read and assumes the state always contains a message.

examples/langgraph/11_no_llm_graph.py
```diff
@@ -0,0 +1,110 @@
...
+    (see ``PlatformMessage.format_for_llm``) before appending it to
+    ``state["messages"]``. The node below parses that prefix back out to
+    know who to @mention, then calls ``band_send_message`` directly — no
+    model, no ``bind_tools``, no ``ToolNode``.
+    """
+    checkpointer = InMemorySaver()
+
+    def graph_factory(band_tools: list[Any]) -> Pregel:
+        send_message = next(t for t in band_tools if t.name == "band_send_message")
+
```
AlexanderZ-Band, line 76: Could we extract parsing and cleanup from `reply()` so it has a single responsibility and the main flow is easier to read?

```python
async def reply(state: MessagesState) -> dict[str, list[Any]]:
    sender, text = parse_latest_message(state)
    reply_text = create_reply(text)
    await send_reply(send_message, reply_text, sender)
    return {"messages": []}
```

AlexanderZ-Band (changes requested):
Please address the inline comments before merging.

AlexanderZ-Band (approved):
Approved with comments.

PR #447, opened 2026-07-19
fix(examples): make coding_agents Docker stack spin up cleanly [INT-500]

nir-singher-band:
## Summary

Validated the `examples/coding_agents` two-agent Docker Compose stack (Claude SDK **planner** + Codex **reviewer**) end-to-end against `app.band.ai` and fixed everything that stopped it from spinning up. Both containers now boot, connect, load their role prompts, and coordinate in a room.

## Root causes fixed

**1. Phantom `agent_config.yaml` directory → config-load crash.**
Compose bind-mounts `./agent_config.yaml`; when the file is missing Docker auto-creates the *source* as an empty **directory**, so both runners crashed reading config. `create_agents.py` now writes the combined, keyed `agent_config.yaml` **directly** (single source of truth), replacing the old "copy the example, paste creds" step that produced the phantom dir. It refuses to overwrite an existing config without `FORCE=1` (re-running registers new agents and would orphan the old ones).

**2. Platform mismatch / unreachable host → `httpx.ConnectError: [Errno -2] Name or service not known` crash loop.**
Agents registered on one Band platform can't be found by containers pointed at another, and a misspelled/unreachable `BAND_REST_URL` fails DNS before auth. `create_agents.py` now logs the platform it registered against and warns to match `.env`; `.env.example` and the README troubleshooting section document the exact symptom → fix.

## Also in this change

- **Reviewer role prompt now loads** — mount `./prompts` into the reviewer service; it read `/app/prompts/reviewer.md` but nothing was mounted there, so it was booting without its Band collaboration prompt.
- **Repo block is optional throughout** (example, README, script) — with no `repo` block the agents boot chat-only, needing zero GitHub/SSH. Add `REPO_URL` (± `REPO_BRANCH`/`REPO_INDEX`) to wire a shared repo.
- `test_communication.py` reads the combined `agent_config.yaml`; dropped the stale "3 agents" docstring.
- `.gitignore`s the `coding_agents` per-run artifacts (`.agent_ids.txt`, legacy `planner.yaml`/`reviewer.yaml`) so credentials/IDs can't be committed.

## Test plan

Verified live on `app.band.ai`:
- `python create_agents.py` registers planner + reviewer and writes `agent_config.yaml` (no phantom dir).
- `docker compose up -d` → both containers `Up` and stable (no restart loop).
- Logs show both connecting over REST/WS, both loading role prompts (incl. reviewer at `/app/prompts/reviewer.md`), and the planner completing a room turn.
- No secrets in the diff — `.env`, `agent_config.yaml`, `.agent_ids.txt` are gitignored and excluded.


examples/coding_agents/create_agents.py
```diff
@@ -1,95 +1,160 @@
...
+    )
+    return repo
+
 
 async def main() -> None:
     api_key = os.environ.get("BAND_API_KEY")
     if not api_key:
         raise ValueError("BAND_API_KEY environment variable is required")
 
+    if CONFIG_PATH.is_file() and os.environ.get("FORCE", "").strip().lower() not in (
```
AlexanderZ-Band, line 87: is_file() misses the stale agent_config.yaml directory this PR is meant to recover from. The script will register both agents, then fail at write_text() with IsADirectoryError, without writing cleanup IDs. Check exists() or handle directories before registering agents.

examples/coding_agents/create_agents.py
```diff
@@ -1,95 +1,160 @@
...
-        for a in created:
-            f.write(f"{a['id']}\n")
-    logger.info("\nAgent IDs saved to .agent_ids.txt for cleanup")
+        logger.info("%s: id=%s, role=%s", a["name"], a["id"], a["role"])
+    logger.info(
+        "\nRepo: %s", repo_block["url"] if repo_block else "none (chat-only, no clone)"
+    )
+    logger.info("Registered against: %s", base_url)
+    logger.info(
+        "IMPORTANT: set BAND_REST_URL in .env to the SAME platform (%s), or the "
```
AlexanderZ-Band, line 153: The setup guide has users configure BAND_REST_URL in .env, but this script does not load .env. On a non-default platform it registers agents on app.band.ai, while Compose connects to the URL in .env. Please load the same configuration source or make the command pass BAND_REST_URL explicitly.

examples/coding_agents/create_agents.py
```diff
@@ -1,95 +1,160 @@
...
-                "id": agent.id,
-                "api_key": credentials.api_key,
-                "role": agent_def["role"],
-                "file": agent_def["file"],
-            }
-        )
+        config[agent_def["key"]] = section
+        created.append({"name": agent.name, "id": agent.id, "role": agent_def["role"]})
+
+    body = yaml.dump(config, default_flow_style=False, sort_keys=False)
```
AlexanderZ-Band, line 138: Could we move config and cleanup-file writing into small helpers? main() currently registers agents, builds config, writes files, and reports the result. Separating persistence would make the flow easier to follow and test.

examples/coding_agents/create_agents.py
```diff
@@ -1,95 +1,160 @@
...
+    CONFIG_PATH.write_text(f"{CONFIG_HEADER}\n{body}", encoding="utf-8")
+    logger.info("\nWrote combined config: %s", CONFIG_PATH.name)
+
+    CLEANUP_PATH.write_text("".join(f"{a['id']}\n" for a in created), encoding="utf-8")
+    logger.info("Agent IDs saved to %s for cleanup", CLEANUP_PATH.name)
 
     logger.info("\n=== Summary ===")
```
AlexanderZ-Band, line 145: Could we move this summary output into a `log_summary()` helper? It would keep `main()` focused on the registration flow and put the user-facing setup instructions in one place.

examples/coding_agents/create_agents.py
```diff
@@ -1,95 +1,160 @@
...
-            {
-                "name": agent.name,
-                "id": agent.id,
-                "api_key": credentials.api_key,
-                "role": agent_def["role"],
-                "file": agent_def["file"],
-            }
-        )
+        config[agent_def["key"]] = section
+        created.append({"name": agent.name, "id": agent.id, "role": agent_def["role"]})
```
AlexanderZ-Band, line 136: If reviewer registration fails, the planner ID is never written to the cleanup file. `FORCE=1` also overwrites the previous cleanup IDs. Both paths can leave agents orphaned. Could we persist each successful registration immediately or clean up created agents on failure?

examples/coding_agents/test_communication.py
```diff
@@ -26,8 +29,8 @@
 SCRIPT_DIR = Path(__file__).parent
 
 
-def load_agent_config(filename: str) -> dict:
-    """Load agent config from YAML file."""
+def load_config_file(filename: str) -> dict:
```
AlexanderZ-Band, line 32: Could we use `band.config.load_agent_config()` here? It already supports keyed config and validates missing files, empty values, and required credentials, so we do not need another YAML loader.

examples/coding_agents/test_communication.py
```diff
@@ -42,9 +45,10 @@ async def main() -> None:
...
 
-    # Load agent configs
-    planner = load_agent_config("planner.yaml")
-    reviewer = load_agent_config("reviewer.yaml")
+    # Load agent configs from the combined file create_agents.py writes
+    config = load_config_file("agent_config.yaml")
+    planner = config["planner"]
+    reviewer = config["reviewer"]
 
     base_url = os.environ.get("BAND_REST_URL", "https://app.band.ai")
```
AlexanderZ-Band, line 53: This script also ignores the `.env` file used by Compose, so the documented command targets `app.band.ai` on custom deployments. Could we load the same settings here as `create_agents.py`?

examples/coding_agents/README.md
```diff
@@ -62,28 +67,30 @@ When using HTTPS URLs, configure git credentials on host (`credential helper`, P
...
 cp .env.example .env
+# edit .env: set ANTHROPIC_API_KEY and OPENAI_API_KEY
 ```
 
-2. Configure agents and repo:
+2. Register the agents and generate `agent_config.yaml`:
 
 ```bash
-cp agent_config.yaml.example agent_config.yaml
+BAND_API_KEY=band_u_... python create_agents.py
```
AlexanderZ-Band, line 86: Since this script is now the required setup path, can we add PEP 723 metadata and document `uv run create_agents.py`? Plain `python` currently depends on the SDK already being installed.

AlexanderZ-Band:
Closing as stale: this two-month-old Docker-stack fix has not been maintained against the current coding-agent examples and configuration.

PR #449, opened 2026-07-20
feat: kit credential injection — proxy-managed custody [INT-981]

AlexanderZ-Band:
Stacked on #448 (INT-980). Delivers proxy-managed credential custody so the **real Band and LLM keys never enter the sandbox VM** — the VM holds only a sentinel while a host-side proxy injects the real keys on the wire — plus an automated never-in-VM proof.

## What lands here (code-complete, full local gate green)

**Launcher custody**
- `proxy-managed` credential source: no secrets file, no plaintext acknowledgement — the VM carries only the `proxy-managed` sentinel.
- Enforced: a real `BAND_API_KEY` under proxy-managed mode is **rejected** (sentinel-only). Scoped to the Band key on purpose — LLM/provider keys ride sbx's own built-in service-secret placeholder (`SANDBOX_STORED_CREDENTIAL_…`), which the launcher passes through untouched. The less-secure `workspace-env-file` fallback warns at launch.

**SDK — key rides the wire, not the VM**
- REST: proven that the sentinel is emitted as the `X-API-Key` header (pytest-httpx against a real `BandLink`).
- WebSocket: the key is sent as an **`x-api-key` handshake header** (`WebSocketClient` → `additional_headers`), matching the platform's `user_socket.ex`, which reads `x-api-key` from the handshake with precedence over the query param. Upgrade-peer test asserts the header rides the real handshake.

**Kit**
- Injection is provisioned **host-side** via `sbx secret set-custom --host '**.band.ai'` (controllable endpoint, not hardcoded to prod). The kit-declared `credentials[].apiKey.inject[]` form was reverted — it crashes `sbx create` 0.35.0 despite passing `kit validate`. `spec.yaml` bakes **no** credentials (guard test asserts it). LLM providers use built-in `sbx` services.
- Two custody tiers documented (injection preferred, env-file fallback with warning); echo-agent defaults to `credentials.source: proxy-managed`. `sbx` doc/version pins bumped 0.34.0 → 0.35.0.

**Never-in-VM E2E** (`tests/docker/test_kit_proxy_managed_live.py` + `tests/docker/toolkit/sbx_cli.py`)
All three checks are now implemented — the auth round-trip was the last one:
1. **injection path** — the Band host's cert *through the proxy* is the Docker Sandboxes MITM cert (rides the injection path, not a plain tunnel).
2. **never in VM** — the exact key provisioned into `set-custom` is absent from env / files / `/proc`.
3. **round-trip (auth)** — a real agent, provisioned via the baseline toolkit, boots inside the sandbox with only the sentinel in its env and round-trips a room message (REST + WS), proving the proxy injected a *working* key. Its id + endpoints are baked into a throwaway echo-agent workspace (since `sbx create` takes no `--env`); its key is injected host-side and never enters the VM.

Gated `sandbox` + `e2e` (skips on CI). A module-wide `pytest.mark.timeout` now covers `sbx create`'s microVM boot — the previous absence/cert tests had none and would have been killed by the 30s global.

Single source of truth for the `proxy-managed` literal (`src/band/credentials.py`) with a guard test binding the kit docs' `--placeholder` to it.

## Gate
ruff · ruff format · pyrefly (0 errors) · **3966 unit tests** · 15 markdown-doc snippets — all green. The generated round-trip `band.yaml` was additionally validated against the real launcher loader.

## Resolved since opening
- The temporary `[tool.uv.sources]` git-pin for `phoenix-channels-python-client` is **gone** — the handshake-header param shipped, so it's now a plain `>=0.2.2` constraint with `uv.lock` refreshed.

## Not in this PR (needs live infra)
- **The live run of the never-in-VM proof** — on a nested-virtualization host with the `sbx` CLI, against staging. The test is implemented and skips on CI/dev; this live pass is INT-981's headline acceptance criterion. (Platform WS `x-api-key` auth is already implemented; the E2E validates it. A non-prod target also needs `sbx policy allow network`, since the kit only allowlists `app.band.ai`.)
- Kit image ref: swaps to the published OCI artifact from INT-979 when that wiring is finalized.


docker/band_python_kit/README.md
```diff
@@ -85,14 +88,37 @@ deployment, override the endpoints in `band.yaml` (or via `BAND_REST_URL` /
...
+holds the real keys and writes them onto outbound requests; the sandbox only ever
+sees a sentinel. Requires `sbx create --kit`.
+
+- **LLM providers** are built-in `sbx` services — store the one your agent uses:
+  ```bash
+  sbx secret set -g anthropic        # or openai, google, groq, mistral, …
+  ```
+- **Band** is provisioned host-side. The `**.band.ai` wildcard follows whichever
+  Band deployment you point the agent at via `BAND_REST_URL` (prod `app.band.ai`,
+  `platform.dev.band.ai`, staging, …) — the endpoint config is the control knob:
```
amit-gazal-band: Remove `platform.dev.band.ai` please. Only Band employees can use the dev environment.

docker/band_python_kit/README.md
```diff
@@ -85,14 +88,37 @@ deployment, override the endpoints in `band.yaml` (or via `BAND_REST_URL` /
...
+  ```bash
+  export BAND_API_KEY=<your-band-key>   # keep the value out of shell history
+  sbx secret set-custom -g --host '**.band.ai' \
+    --env BAND_API_KEY --placeholder proxy-managed --value "$BAND_API_KEY"
+  ```
+  `--value` is visible in shell history / process args — pass it from an exported
+  variable as above (and clear history if the real value ever appears literally).
```
amit-gazal-band, line 112: Important note.

docker/band_python_kit/RELEASING.md
```diff
@@ -117,9 +117,12 @@ independent job) has already succeeded, but no kit artifacts were produced.
 
 ## Supported `sbx` version
 
-Validated against **`sbx` v0.34.0** at ship time (OCI v2 kit artifacts,
-`kit.allowedSources` default). The kit surface of `sbx` is experimental and has
-moved between releases, so on **every** CLI upgrade, revalidate:
+Targets **`sbx` v0.35.0** — the version whose kit-spec v2 schema the credential
+injection uses (`credentials[].apiKey.inject[]`). `sbx kit validate` passes clean
+on the repo `spec.yaml` under 0.35.0. The `kit push` → `pull` → OCI-consume
+roundtrip below was last run under v0.34.0 and **must be re-run under 0.35.0 at
```
amit-gazal-band: Consider removing the v0.34.0 note.

tests/conftest.py
```diff
@@ -54,6 +54,21 @@
...
 
 
+class CollectionGateSettings(BaseSettings):
+    """Env-var gates for which marked suites collect/run.
+
+    Field name == env var (case-insensitive). Read via a fresh instance in the
+    collection hooks so the current environment is always what decides.
+    """
+
+    model_config = SettingsConfigDict(extra="ignore", case_sensitive=False)
```
amit-gazal-band: CollectionGateSettings raises ValidationError when a gate var is set-but-empty (CI=, E2E_TESTS_ENABLED=), which kills the whole pytest run inside a collection hook. The old os.environ.get("CI") == "true" treated empty as false. Some CI wrappers and env CI= … invocations do produce empty vars. One-line fix: add env_ignore_empty=True to the SettingsConfigDict. (Verified the failure reproduces with pydantic-settings in this venv.)

tests/docker/test_kit_proxy_managed_live.py
```diff
@@ -0,0 +1,273 @@
...
+    assert f"BAND_API_KEY={PROXY_MANAGED_API_KEY}" in environ, "agent env not observed"
+
+    # Absence is now meaningful: the real key is in no process env, the exec
+    # shell's env, or any searched file.
+    assert real_key not in environ
+    assert real_key not in sandbox.shell_env()
+    assert (
+        sandbox.files_containing(
+            real_key, search_paths=["/home/agent", "/etc", "/tmp", "/workspace"]
+        )
```
amit-gazal-band, line 189: files_containing(real_key, …) embeds the key (shell-quoted) in the sbx exec argv, and exec uses check=True — a transport failure (sandbox died mid-test, daemon hiccup) raises CalledProcessError, whose message renders the full command into the pytest output. The module's own _run_redacting_secret exists precisely because "a routine failure would spill the credential into the traceback and captured test logs" — the same argument applies here. Suggest routing secret-bearing exec calls through the same redaction. (Gated local-only test, failure path only — minor, but inconsistent with the PR's own hygiene bar.)

tests/docker/test_kit_spec.py
```diff
@@ -93,18 +73,38 @@ def test_launcher_module_referenced_by_kit_is_importable() -> None:
...
+    # `credentials[].apiKey.inject[]` also crashes `sbx create` on 0.35.0.)
     assert "credentials" not in spec
+
+
+def test_kit_docs_placeholder_matches_the_sdk_sentinel() -> None:
+    """The `sbx secret set-custom --placeholder` value in the shipped kit docs
+    must equal the SDK's PROXY_MANAGED_API_KEY — the single origin of the
+    sentinel. YAML/shell can't import the constant, so this guard fails CI if a
+    doc mirror drifts from it."""
+    docs = [KIT_DIR / "README.md", KIT_DIR / "echo-agent" / "README.md"]
```
amit-gazal-band: test_kit_docs_placeholder_matches_the_sdk_sentinel checks the two READMEs, but spec.yaml's comment block also carries --placeholder proxy-managed and can drift silently. Add it to the scanned files.

amit-gazal-band (changes requested):
Please see my comments

docker/band_python_kit/README.md
AlexanderZ-Band: Done in 85bc6e6 — scrubbed from all four shipped-kit spots (this one plus the two policy-example lines in this README, and the spec.yaml comment); the examples now use a neutral `<band-host>` placeholder. `grep -r platform.dev.band.ai docker/band_python_kit/` is clean.

docker/band_python_kit/README.md
AlexanderZ-Band: Ack — kept that warning as-is.

docker/band_python_kit/RELEASING.md
AlexanderZ-Band: Trimmed in 85bc6e6: dropped the v0.34.0 history from this section, but kept the one-line obligation to run the full push → pull → OCI-consume roundtrip at the next release cut — it genuinely hasn't been run under 0.35.0 yet, so deleting it entirely would hide pending validation. (The v0.34.0 mention further down stays: it's the factual record of how the first-release roundtrip prerequisite was closed.)

tests/conftest.py
AlexanderZ-Band: Fixed in 85bc6e6 — reproduced exactly as you described, and the same failure class existed in the baseline settings (`E2E_TESTS_ENABLED=` kills `BaselineSettings()`, which the sandbox test module constructs at import). Added `env_ignore_empty=True` to CollectionGateSettings and to every class in tests/e2e/baseline/settings.py, updated the canonical pattern in AGENTS.md, and added regression tests (tests/test_env_gates.py) that fail on the old code.

tests/docker/test_kit_proxy_managed_live.py
AlexanderZ-Band: Fixed in 85bc6e6 — agreed it broke the PR's own hygiene bar. Generalized the one redaction helper: `run_redacting_secret` now returns stdout and masks the secret even when it's embedded inside a longer argument (the old exact-arg match wouldn't have caught the grep command). `Sandbox.exec` grew a `redact=` param and `files_containing` routes through it. Unit-tested: a masked failure still keeps the exit code and stderr diagnostics.

tests/docker/test_kit_spec.py
AlexanderZ-Band: Fixed in 85bc6e6 — spec.yaml added to the scanned files, and I verified the guard actually fails on a deliberately drifted spec before restoring it.

amit-gazal-band (approved):

PR #467, opened 2026-07-21
fix(integrations): replay Band room history when a remote ACP session can't be restored [INT-1095]

AlexanderZ-Band:
## Summary

Fixes [INT-1095](https://linear.app/thenvoi/issue/INT-1095) and [INT-1101](https://linear.app/thenvoi/issue/INT-1101). The ACP client adapter relied **solely** on ACP `session/load` to restore conversation context; whenever the remote agent process was fresh (container restart, `,fork`-per-connection spawn, crash) the agent started with **zero context**, even though the Band room transcript is intact on the platform.

The fix adopts the existing `opencode`/`letta` "resume-else-replay" pattern, then hardens it through two review rounds. The governing invariant: **a freshly created ACP session owes the room a transcript replay; a restored one does not.**

## What changed

**Replay fallback** (`converters/acp_client.py`, `integrations/acp/client_adapter.py`, `integrations/acp/client_runtime.py`)
- The converter carries the room's text transcript (`replay_messages`, via the shared `build_replay_messages` helper). Adapter narration events (thought/tool_call/tool_result/task) never replay.
- On bootstrap: `session/load` is validated first; on any miss (no persisted id, unavailable, **or erroring** — protocol errors are now a recoverable load-miss, not a turn-killing exception) the fresh session's first prompt is seeded with the transcript.
- **Off-bootstrap** (INT-1101): a session minted after a mid-run teardown re-fetches the transcript via `tools.fetch_room_context`, so a respawn never starts amnesiac. `_get_or_create_session` reports created-vs-reused; `_load_persisted_session`'s redundant bool contract was deleted.

**Prompt safety** (review-driven)
- Header framed as read-only background ("treat as already handled"), with an escape hatch so explicit recall requests are never refused.
- The live message is attributed (`format_for_llm`) and sits **last**, under a **nonce'd** `[New Message <nonce>]` boundary marker the header names — replayed content predates the turn, so it cannot spoof the boundary.

**Backlog truncation** (all adapters, `runtime/formatters.py` + `preprocessing/default.py`)
- Bootstrap history stops **strictly before** the triggering message (`messages_before`): backlog messages that accumulated while the agent was offline are pending turns of their own; replaying them exposed future requests and duplicated them when their own turn arrived. Oneshot's swallow-if-seen drain is deliberately untouched (its `seen_ids` coupling depends on the LLM having seen the fetched page).

**Codex consolidation** (`adapters/codex.py`)
- Its private history formatting now uses the shared helper (net −10 lines). Drops the dead `"message"` type (not in `MessageType`, no producer anywhere); whitespace-only lines filtered for all four consumers.

**CI / lockfile**
- `uv.lock` caught up to the released 1.3.0 (the release flow bumps pyproject without regenerating the lock — the source of everyone's recurring dirty-tree noise).
- CI now runs `uv sync --locked` everywhere, and the release job refreshes `uv.lock` on the release-please PR (checked out with the App token so the push triggers the PR's CI). First live exercise will be the next release PR — worth watching that run.

## Guarding the known rehydration weaknesses

- **No last-message duplication**: trigger excluded by id at both history producers; post-trigger backlog truncated; a loaded session gets no replay; unit test counts the live message appearing exactly once.
- **Current message not ignored**: prompt shape pinned by test (system context → replay → attributed live message last under the boundary marker).

## Testing

- **Unit**: full suite green (3970 passed). New coverage: converter replay + narration exclusion, load-miss / load-**error** / load-success paths, cold boot, replay-once, exactly-one live message, attribution + ordering + nonce agreement, **mid-run respawn re-seed** (red on old code), backlog truncation (formatters + preprocessor, red on old code).
- **Live e2e** (backends lane, real Copilot CLI + platform), re-run after every behavior-affecting round:
  - `test_acp_recall_via_room_replay_when_session_load_misses` — new, deterministic: fresh `COPILOT_HOME` per phase forces the load-miss; phase-2 recall carries a user-stated marker **and** an agent-produced fact that only the replayed agent side of the transcript can supply.
  - Matrix rehydration + context-recall `copilot_acp` cells: green.

## Follow-ups filed

- [INT-1101](https://linear.app/thenvoi/issue/INT-1101) — **fixed here** (In Review); residual noted there: one room's failure still tears down the runtime momentarily (context now recovers, so no longer silent loss).
- [INT-1102](https://linear.app/thenvoi/issue/INT-1102) — unify the replay-safety framing across codex/opencode/letta (their weaker wording predates this PR; migrating needs each lane's own live validation).

Likely also closes [INT-623](https://linear.app/thenvoi/issue/INT-623) (L4 Rehydration placeholder).


amit-gazal-band (dismissed):
Very clean fix!

amit-gazal-band (approved):

PR #470, opened 2026-07-21
fix(release): survive post-tag failures; isolate and harden PyPI publishing

AlexanderZ-Band:
Fixes INT-1103.

## Problem

The [v1.4.0 release run](https://github.com/band-ai/band-sdk-python/actions/runs/29841953779/job/88672800697) failed after release-please had created and tagged `band-sdk-v1.4.0`, so the PyPI, kit, and add-band publish jobs never ran: the GitHub release exists, PyPI is still at 1.3.0.

Two defects underneath:

1. **Expression crash.** On a release-creating merge `steps.release.outputs.pr` is empty, and GitHub evaluates a step's `env:` templates even when its `if:` is false — the bare `fromJSON('')` in `PR_BRANCH` threw and failed the `release` job. (Only such site in the repo.)
2. **Publishing coupled to the release job's fate.** Downstream jobs gated on `release_created == 'true'` carry an implicit `success()` on `needs`, and release-please tags before it finishes its other work — so any post-tag failure strands a tagged release unpublished, with no re-run path (release-please reports `release_created` only once, and re-runs use the frozen workflow definition).

### `release.yml`
- `PR_BRANCH: ${{ fromJSON(steps.release.outputs.pr || '{}').headBranchName }}` — the expression is total; the `if:` still gates execution.
- `publish-kit` / `bump-add-band` use `!cancelled() && needs.release.outputs.release_created == 'true'` (the bump additionally requires `needs.publish-kit.result == 'success'`) — a created release publishes even if the `release` job fails afterwards.
- No PyPI job here anymore.

### `band-publish.yml` (new) — the PyPI publish
- Triggered by `release: published`, so publishing keys off the durable release rather than the release run's job graph; `workflow_dispatch` (input `tag`) publishes any existing tag, e.g. for recovery. `skip-existing: true` makes re-runs idempotent.
- A top-level file, not a reusable workflow: PyPI trusted publishing binds the publisher to the top-level workflow filename and [does not support reusable workflows](https://docs.pypi.org/trusted-publishers/troubleshooting/) (warehouse#11096).
- Structured per [PyPI's security model](https://docs.pypi.org/trusted-publishers/security-model/): a build job validates the tag (`band-sdk-vX.Y.Z` format, checked out via `refs/tags/` so a branch can't shadow it, ancestor of `main`, version matching `pyproject.toml`) and uploads the distributions; the environment-privileged publish job runs two steps (download, publish) and never executes project code.

### `pr-title.yml`
- `release` added to the allowed scopes.

### Docs
- `docs/ci-cd-workflows.md` and `docker/band_python_kit/RELEASING.md` describe the current topology and security layers.

## Repository settings (already live, set via API)

- `release` **environment deployment policy**: `main` + `band-sdk-v*` tags only — GitHub matches it against the run's `GITHUB_REF` before the job can mint OIDC, so a modified workflow on a side branch can't reach the credential.
- **Tag ruleset 19436545**: creating/moving/deleting `band-sdk-v*` tags is restricted to the release App and repo admins — necessary because a release event runs the workflow file at the tagged commit, so the trusted tag namespace must not be claimable by write collaborators.
- Not configured (open org decision): required reviewers on the `release` environment — a human gate on every release, including the kit jobs.

## ⚠️ Required before the next release: PyPI trusted-publisher migration

1. Add publisher: project `band-sdk` → Publishing → this repo, workflow `band-publish.yml`, environment `release`.
2. Retire the `release.yml` entry (it no longer contains a publish step).

## Recovering 1.4.0 (after merge + migration)

1. Dispatch `band-publish` with tag `band-sdk-v1.4.0` (release events only run workflows present at the tag, so pre-existing tags go via dispatch).
2. Dispatch `kit-publish-manual` with ref `band-sdk-v1.4.0`, version `1.4.0`, major `1`, move-floating `true`.
3. Re-create the add-band pin-bump PR if needed (inert until the catalog entry exists — likely a no-op).

No release/tag deletion needed.


amit-gazal-band (approved):
I had to add a trusted publisher on PyPI cause you changed the file that published. Did it, so we're good to go.

PR #474, opened 2026-07-22
fix(ci): exempt Band-published packages from the quarantine gate

AlexanderZ-Band:
Fixes INT-1106.

## Problem

The kit quarantine gate applies its 7-day minimum public age to **every** locked artifact, including the three PyPI packages Band publishes itself (`band-client-rest`, `band-testing-python`, `phoenix-channels-python-client`). The gate's threat model — from INT-979 — is *upstream* compromise: a freshly-poisoned third-party release that the ecosystem detects and yanks within days. A release Band makes carries no such signal, so gating our own packages only deadlocks the pipeline: the v1.4.1 kit publish tripped on `phoenix-channels-python-client==0.2.2`, published one day earlier by us.

## Fix

`FIRST_PARTY` set in `scripts/check-lock-age.py` (the gate script is the policy's single source of truth — the workflow invocation is unchanged):
- matched by PEP 503 canonical name, so a lock spelling variant can't silently lapse the exemption;
- printed on every gate run, so the active exemptions stay visible in the log;
- third-party artifacts keep the full window.

Verified: 9/9 gate tests pass (2 new — exemption + canonicalization), and the gate now passes the committed lock with the default 7-day window (it fails on `main` today).

## Trade-off, considered

Exempting by name forgoes the age tripwire against Band's own PyPI account being hijacked. Accepted deliberately: a routinely-overridden gate (the alternative, given every first-party bump trips it) protects nothing, and account-hijack is better addressed by trusted publishing/2FA on those packages.

## Scope note

The gate script runs from the checked-out release tag, so tags cut before this change (band-sdk-v1.4.0/v1.4.1) still gate those packages — their kit republish uses the documented `quarantine-max-age-days` override; this fixes all future tags.

### DIFF
--- a/scripts/check-lock-age.py
+++ b/scripts/check-lock-age.py
@@ -14,6 +14,13 @@ gate does NOT work: the cutoff makes uv treat the committed lock itself as
 outdated and every gated build fails — proven live on uv 0.9.13 and 0.11.19;
 see astral-sh/uv#18775 for the underlying option-tracking behavior.)
 
+Packages Band publishes itself (``FIRST_PARTY``) are exempt from the age
+window: the gate models *upstream* compromise — a poisoned third-party
+release that the ecosystem detects and yanks within days — and a release we
+made carries no such signal. Ageing our own artifacts would only deadlock
+our own pipeline for a week after every first-party dependency bump.
+Exemptions are printed on every run so they stay visible in the gate log.
+
 Exit codes: 0 = every artifact is old enough; 1 = violations (each listed on
 stderr); 2 = bad usage / unreadable lock. Artifacts without an upload-time
 are violations too — an undatable source must not slip the gate silently.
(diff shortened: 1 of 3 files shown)


scripts/check-lock-age.py
```diff
@@ -33,13 +40,29 @@
...
 @dataclass(frozen=True)
 class Violation:
     package: str
     version: str
     detail: str
 
 
+def canonical(name: str) -> str:
+    """PEP 503 canonical form, so lock spellings always match FIRST_PARTY."""
+    return name.strip().lower().replace("_", "-").replace(".", "-")
```
amit-gazal-band: canonical() isn't strictly PEP 503 (low severity)

canonical('foo__bar')  -> 'foo--bar'   (PEP 503: 'foo-bar')
canonical('foo.._bar') -> 'foo---bar'  (PEP 503: 'foo-bar')

Harmless for the current FIRST_PARTY names (none contain consecutive/mixed separators, and uv writes already-canonical names), so no live impact today. But it's a latent trap: add a future first-party package whose distribution name has a double separator, spell the constant in true canonical form, and the exemption silently won't match. Since canonicalization is the defense this PR advertises, either make it real —

```
import re
def canonical(name: str) -> str:
    """PEP 503 canonical form, so lock spellings always match FIRST_PARTY."""
    return re.sub(r"[-_.]+", "-", name.strip()).lower()
```

amit-gazal-band (dismissed):

scripts/check-lock-age.py
AlexanderZ-Band: Confirmed and fixed in b6de473 — `canonical()` now uses the PEP 503 spec regex (`re.sub(r"[-_.]+", "-", ...)`), and the test pins the separator-run case (`Our__Package.Name` → `our-package-name`); it fails on the old implementation.

PR #486, opened 2026-07-23
ci: run required checks on merge_group to prepare for merge queue

amit-gazal-band:
## Why

The `main-branch-protection` ruleset combines **dismiss-stale-reviews-on-push** with **strict** ("require branches up to date before merging"). When main moves ahead, strict forces a resync; the resync push trips dismiss-stale and discards the approval. Repo admins can't force-merge (org-admin only), so an approved PR that falls behind gets stuck needing a second re-approval — repeatedly if main keeps moving.

The fix is a **GitHub merge queue**, which resyncs and tests server-side *after* approval and merges without pushing to the PR branch (so dismiss-stale never fires). But the queue only merges once the required checks report on its temporary `merge_group` branch — and today none of them trigger on `merge_group`, so enabling the queue first would deadlock **every** merge.

This PR is the prerequisite: make the required checks run on `merge_group`. The queue itself is enabled separately (ruleset change) after this merges.

## Changes

- **`ci.yml`** — add a `merge_group:` trigger (covers `lint`, `packaging`, `test`, `test-crewai`).
- **`pr-title.yml`** — add `merge_group:` and guard the validation step to `pull_request`. A merge group has no PR title to validate, so the step no-ops there and the job reports success for the `Validate PR Title` required check.

## Rollout

1. Merge this PR (queue still off).
2. Add the `merge_group` rule to the `main-branch-protection` ruleset — queue live.
3. Heads-up to the team: merging becomes "Merge when ready" via the queue, not an instant button.

Closes INT-1125.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -2,10 +2,7 @@ name: CI
 
 on:
   pull_request:
-    # Keep dev here until the single-trunk cutover completes (default branch
-    # flipped to main and dev retired). Removing dev early would stop CI on
-    # dev-targeted PRs — including this migration PR — blocking their merge.
-    branches: [dev, main]
+    branches: [main]
   # Run on the merge queue's temporary branch so these required checks report
   # against the queued merge result. Without this the queue would wait forever.
   merge_group:
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -6,6 +6,9 @@ on:
     # flipped to main and dev retired). Removing dev early would stop CI on
     # dev-targeted PRs — including this migration PR — blocking their merge.
     branches: [dev, main]
+  # Run on the merge queue's temporary branch so these required checks report
+  # against the queued merge result. Without this the queue would wait forever.
+  merge_group:
 
 jobs:
 
--- a/.github/workflows/pr-title.yml
+++ b/.github/workflows/pr-title.yml
@@ -3,6 +3,10 @@ name: PR Title
 on:
   pull_request:
     types: [opened, edited, synchronize, reopened]
+  # Required check must also report on the merge queue's branch, or the queue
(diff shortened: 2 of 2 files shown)


.github/workflows/ci.yml
```diff
@@ -2,10 +2,10 @@ name: CI
...
 on:
   pull_request:
-    # Keep dev here until the single-trunk cutover completes (default branch
-    # flipped to main and dev retired). Removing dev early would stop CI on
-    # dev-targeted PRs — including this migration PR — blocking their merge.
-    branches: [dev, main]
+    branches: [main]
+  # Run on the merge queue's temporary branch so these required checks report
+  # against the queued merge result. Without this the queue would wait forever.
+  merge_group:
```
AlexanderZ-Band, line 8: Did we verify that all required GitHub Actions checks (not just CI and PR title) also run on merge_group? If a required workflow doesn’t support merge_group, the merge queue can get stuck waiting for a check that never starts. Not blocking if this has already been verified.

AlexanderZ-Band (approved):
approved with one question

.github/workflows/ci.yml
amit-gazal-band: Yes, verified. The required checks on main come from exactly two workflows: ci.yml (lint, packaging, the four test matrix cells, and the four test-crewai cells) and pr-title.yml (Validate PR Title). Both now trigger on merge_group, so every required context reports against the queued merge branch. No other workflow produces a required check — e2e, the kit-* publish/rebuild jobs, band-publish, and release aren't gating and won't be waited on.

On pr-title specifically: a merge group has no associated PR, so the validation step is guarded with if: github.event_name == 'pull_request' and no-ops in the queue while the job still reports success — so it satisfies the required check without a title to validate.

PR #487, opened 2026-07-26
feat: upgrade pydantic-ai-slim to 2.x

AlexanderZ-Band:
Closes INT-1121. Supersedes #466.

## Summary

- Upgrade `pydantic-ai-slim` from 1.x to `>=2.18.0` and update the lockfile.
- Migrate the Pydantic AI adapter to 2.x APIs: `ProcessHistory`, async streaming context management, event parts, usage access, and context-aware custom-tool registration.
- Preserve correct behavior for retry exhaustion, usage reporting, history filtering, and custom tools with focused regression coverage.
- Keep CrewAI isolated in its dedicated dependency lane; simplify its CI job to the tests that require the real SDK, verify that coverage cannot drift, and remove test module surgery that caused order-dependent conformance failures.
- Correct the dependency-conflict documentation and parse the missing-framework CI flag as a boolean.

## Validation

- `ruff check .`, `ruff format`, and `pyrefly check`
- Unit suite and markdown-doc tests
- Dedicated `dev-crewai` CI command, CrewAI test files, and standalone CrewAI conformance coverage

## Remaining work

- Run the live core-lane E2E smoke to verify the Pydantic AI 2.x Responses API default.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -137,18 +137,13 @@ jobs:
         # dev-crewai venv lacks langgraph/anthropic/parlant/pydantic-ai/etc.;
         # tolerate missing framework configs instead of failing fast.
         BAND_ALLOW_MISSING_FRAMEWORKS: "1"
+      # Only the tests that *need* this venv. Every other crewai test fakes the
+      # package through sys.modules, so the `test` job above already runs it and
+      # a second pass here would add no signal — just a path to forget.
+      # tests/framework_conformance/test_crewai_job_coverage.py keeps this honest.
       run: |
         uv run pytest \
-          tests/adapters/test_crewai_adapter.py \
-          tests/adapters/test_crewai_adapter_soak.py \
-          tests/adapters/test_crewai_flow_adapter.py \
           tests/adapters/test_crewai_flow_phase3.py \
-          tests/adapters/test_crewai_flow_phase4.py \
-          tests/adapters/test_crewai_flow_phase5.py \
-          tests/adapters/test_crewai_flow_state_source.py \
-          tests/converters/test_crewai.py \
-          tests/converters/test_crewai_flow.py \
-          tests/integrations/test_crewai_tools.py \
           tests/integrations/test_crewai_flow_real_sdk.py \
           tests/test_capability_gating_e2e.py \
           tests/framework_conformance/ -k crewai
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -150,6 +150,7 @@ jobs:
           tests/converters/test_crewai_flow.py \
           tests/integrations/test_crewai_tools.py \
           tests/integrations/test_crewai_flow_real_sdk.py \
+          tests/test_capability_gating_e2e.py \
           tests/framework_conformance/ -k crewai
 
     - name: Pyrefly check (crewai sources)
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -145,6 +145,7 @@ jobs:
         uv run pytest \
           tests/adapters/test_crewai_flow_phase3.py \
           tests/integrations/test_crewai_flow_real_sdk.py \
+          tests/integrations/test_crewai_real_tools.py \
           tests/test_capability_gating_e2e.py \
           tests/framework_conformance/ -k crewai
 
--- a/src/band/adapters/pydantic_ai.py
+++ b/src/band/adapters/pydantic_ai.py
@@ -6,10 +6,12 @@ Extracted from band.integrations.pydantic_ai.agent.BandPydanticAgent.
 
 from __future__ import annotations
 
+import inspect
 import json
 import logging
 import warnings
-from typing import ClassVar, Any, Callable
+from collections.abc import Callable
+from typing import Any, ClassVar, get_origin, get_type_hints
 
 import httpx
 from pydantic_ai import (
@@ -21,6 +23,7 @@ from pydantic_ai import (
     UnexpectedModelBehavior,
     capture_run_messages,
 )
+from pydantic_ai.capabilities import ProcessHistory
 from pydantic_ai.messages import (
     ModelMessage,
     ModelRequest,
@@ -85,7 +88,7 @@ def _is_replayable_history_message(message: Any) -> bool:
     """Drop assistant responses that would replay as content:null.
 
     Keep any response with at least one content-bearing part (text, tool
-    calls, builtin tool calls/returns, files). Drop thinking-only and empty
+    calls, native tool calls/returns, files). Drop thinking-only and empty
     responses, which providers reject when sent back as history.
     """
     if isinstance(message, ModelResponse):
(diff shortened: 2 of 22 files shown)


src/band/adapters/pydantic_ai.py
```diff
@@ -139,6 +147,25 @@ async def native(args: Any) -> Any:
...
+    annotations`` is classified on the real type rather than the string.
+    """
+    first = next(iter(inspect.signature(fn).parameters), None)
+    if first is None:
+        return False
+    try:
+        annotation = get_type_hints(fn).get(first)
+    except (NameError, TypeError):  # unresolvable annotation: treat as plain
+        return False
+    return annotation is RunContext or get_origin(annotation) is RunContext
```
amit-gazal-band, line 170: A tool whose first parameter is unannotated — `async def my_tool(ctx, message: str)` — falls through to `return False` here (`get_type_hints(fn).get("ctx")` is `None`), so it registers via `agent.tool_plain()`. pydantic-ai 2.18 accepts that same callable on `agent.tool()` and injects the context correctly, so this is a behaviour change from the routing rather than a constraint of the upgrade. Verified against a real `Agent`:

```
agent.tool       -> required: ['message']         # ctx injected by the framework
agent.tool_plain -> required: ['ctx', 'message']  # 'ctx': {}  (no type, no description)
```

Under `tool_plain` the model has to invent a value for `ctx`, and the tool body receives that value where it expected the `RunContext` — so `ctx.deps.send_message(...)` raises `AttributeError` at call time in production, not at registration. It's silent in every other respect.

Suggested fix — treat a missing annotation as context-taking, matching pydantic-ai's own behaviour:

```python
try:
    annotation = get_type_hints(fn).get(first)
except (NameError, TypeError):
    return False
if annotation is None:  # unannotated first param: pydantic-ai injects the context
    return True
return annotation is RunContext or get_origin(annotation) is RunContext
```

The `except` branch is fine as-is: I checked both cases it catches (a `TYPE_CHECKING`-only `RunContext` import, an unresolvable return annotation) and pydantic-ai rejects those tools with `NameError` on *either* path, so the choice there can't cause a silent misroute.

One note on coverage: the four `TestCustomTools` tests that used this shape were re-annotated in this PR, but they patch `Agent`, so they never exercised real registration either way — the case is currently untested. A small check against a real `Agent`, like `test_create_agent_registers_context_free_custom_tool` does, would pin it.

src/band/adapters/pydantic_ai.py
```diff
@@ -139,6 +147,25 @@ async def native(args: Any) -> Any:
     return native
 
 
+def _takes_run_context(fn: Callable[..., Any]) -> bool:
+    """Whether ``fn`` takes pydantic-ai's ``RunContext`` as its first parameter.
+
+    Decides the registration path: ``agent.tool`` requires a RunContext-first
+    callable and raises ``UserError`` for anything else, while ``agent.tool_plain``
```
amit-gazal-band: `agent.tool` is more permissive than this. Against 2.18 it accepts a zero-parameter callable and one whose first parameter is unannotated; it raises `UserError` only when the first parameter is annotated as something *other* than `RunContext`:

```
def get_time() -> str                   -> agent.tool: OK
def plain(ctx, message: str) -> str     -> agent.tool: OK (ctx injected)
def weird(first: str, msg: str) -> str  -> UserError: First parameter of tools that take
                                             context must be annotated with RunContext[...]
async def native(args: Any) -> Any      -> UserError (same)
```

That last row is the `_custom_tool_def_to_callable` shape, and it's a real problem this helper correctly solves — the tuple-form custom tools genuinely do need `tool_plain`. It's specifically the "raises `UserError` for anything else" half that overstates it, and that's what leads to the unannotated-`ctx` handling below.

tests/framework_configs/adapters.py
```diff
@@ -396,63 +319,16 @@ def _build_crewai_config() -> AdapterConfig:
...
-    """Import CrewAIFlowAdapter with mocked crewai dependencies.
+    """The CrewAIFlowAdapter class for the conformance config.
 
-    The flow adapter has its own optional crewai.flow.flow import. The
-    integration tools module (used by build_band_crewai_tools when a
-    sub-Crew is constructed at runtime) imports crewai.tools, so the same
-    mock pattern as ``_get_crewai_adapter_cls`` is reused.
+    Plain import, as for ``_get_crewai_adapter_cls``. The module-level
+    ``from crewai.flow.flow import Flow`` is already guarded by try/except, so it
+    binds ``Flow = None`` when crewai is absent.
```
amit-gazal-band: This describes the import that this PR removes from `src/band/adapters/crewai_flow.py`:

```diff
-try:  # pragma: no cover - optional dependency guard
-    from crewai.flow.flow import Flow  # type: ignore[import-not-found]
-except ImportError:  # pragma: no cover - exercised only when crewai missing
-    Flow = None  # type: ignore[assignment,misc]
```

The removal is right — `Flow` has no remaining references in that module (every surviving occurrence is prose in a docstring or comment), so the plain import holds for the reason given in `_get_crewai_adapter_cls` above. Just the second sentence needs to catch up; something like "the adapter no longer imports `Flow` at module scope at all" would match.

.github/workflows/ci.yml
```diff
@@ -137,19 +137,16 @@ jobs:
         # dev-crewai venv lacks langgraph/anthropic/parlant/pydantic-ai/etc.;
         # tolerate missing framework configs instead of failing fast.
         BAND_ALLOW_MISSING_FRAMEWORKS: "1"
+      # Only the tests that *need* this venv. Every other crewai test fakes the
+      # package through sys.modules, so the `test` job above already runs it and
+      # a second pass here would add no signal — just a path to forget.
+      # tests/framework_conformance/test_crewai_job_coverage.py keeps this honest.
```
amit-gazal-band, line 143: Nice catch!

amit-gazal-band (commented):
Reviewed the pydantic-ai 2.x migration and the crewai isolation refactor. I verified the migrated APIs against pydantic-ai-slim 2.18.0 — `AgentRetries` as a `TypedDict{tools, output}`, `capabilities` present and `history_processors` gone, `FunctionToolResultEvent.part` plus the inherited `tool_call_id` property, `AgentRunResult.usage` as a property, `run_stream_events` returning an async CM, and the `Exceeded maximum output retries` wording in `_tool_execution.py`. Full unit suite green locally (4082 passed, 82 skipped), ruff clean, and no stale 1.x API references remain anywhere in the tree.

The `retries={"tools": 3, "output": 0}` change is the standout — the reasoning that an unsatisfiable `output_type=str` turns every output retry into a duplicate room post is correct and non-obvious, and proving it with a `FunctionModel` rather than a mock is the right call. `test_module_isolation.py` and `test_crewai_job_coverage.py` both target genuinely invisible failure modes. I also confirmed the `sys.modules`-surgery removal is safe: every `crewai` import under `src/band/` is either `TYPE_CHECKING`-only or function-local.

One substantive bug below (the custom-tool routing), plus two doc corrections.

Two things for the description rather than the code:

- `uv.lock` also moves **anthropic 0.96.0 -> 0.120.0**, which isn't mentioned. It's a direct dep of two other adapters, so worth naming — a later bisect will otherwise land on "the pydantic-ai PR" and lose time. (`pydantic-graph` and `genai-prices` also move but are pydantic-ai's own siblings, so those are expected.)
- The Responses API default: bare `openai:` now routes there instead of Chat Completions, and all four `examples/pydantic_ai/*.py` plus `examples/pydantic_ai/README.md` use `openai:gpt-5.4-mini` and switch silently. Only the `__init__` docstring records it. Given the description lists the live core-lane smoke as remaining work, I'd treat that run as a merge gate. Good news: I checked the E2E baseline's custom tools separately and they're unaffected — `ToolSpec.as_callable(ctx_annotation=RunContext)` classifies correctly — so the risk is confined to the provider API surface itself.

amit-gazal-band (dismissed):
Approving.

**E2E: ran the `pydantic_ai` lane live — 24/24 green.**

```
E2E_TESTS_ENABLED=true uv run pytest tests/e2e/ -k pydantic_ai
```

That covers the item listed as remaining work: the **Responses API default is verified**. Replies, custom-tool round-trip, burst recall, rehydration (offline / partial / cross-framework), peer delegation, room isolation, concurrent instances, and the usage-reporting cells all pass against the live platform.

Two cells (`test_store_memory_across_memory_adapters`, `test_recall_memory_across_memory_adapters`) failed on the first pass with `403 plan_required — Memory API access requires an Enterprise plan`. That was an entitlement gap on the account I was testing with, not a code issue; both pass on a re-run once the plan was in place.

I also A/B'd the new retry budget against `output: 3` while chasing that failure. On the healthy path the budget is never touched at all, so the change is free — the reasoning in the code comment holds up. The one behaviour worth being aware of, not worth blocking on: when a tool errors and the model's first pass does nothing terminal, `output: 0` removes its only chance to report that in the room and the message is marked permanently failed. Only reachable on an error path, and I only triggered it via the 403 above.

Everything else verified locally: full unit suite (4082 passed, 82 skipped), `ruff check`, `ruff format --check`, and the migrated pydantic-ai APIs checked against 2.18.0.

The custom-tool routing comment above (unannotated first parameter falling through to `tool_plain`) is still open — approving on the assumption you'll pick it up, since it's a small change and the rest is solid.

amit-gazal-band (approved):

PR #488, opened 2026-07-26
refactor(integrations): migrate A2A SDK to 1.1.2

AlexanderZ-Band:
## Summary
- Migrate Band A2A integrations and six examples to `a2a-sdk>=1.1.2,<2` using official route factories, helpers, and task lifecycle APIs.
- Expose per-peer JSON-RPC/REST routes with 1.0 and v0.3 discovery compatibility.
- Preserve Band-specific streaming reduction, room/context correlation, timeout handling, and UUID peer aliases.
- Use the Fern client retry policy for gateway peer discovery: this reduces retry headroom from the removed bespoke loop, but adds `Retry-After`-aware backoff.

## Validation
- `uv run pytest tests/ --ignore=tests/integration/ --ignore=tests/e2e/ -q --no-cov`
- `uv run ruff check .`
- `uv run pyrefly check`
- Isolated wheel installs and imports for `a2a_gateway` and `a2a_gateway_demo` extras.


amit-gazal-band:
## Blocking: per-alias `Mount("/{tenant}")` shadows the REST binding for every peer after the first

Nice migration overall — leaning on the upstream route factories and deleting the hand-rolled JSON-RPC/SSE layer is clearly the right call. One issue in `GatewayServer._build_app` blocks it, though.

### The issue

`create_rest_routes(..., enable_v0_3_compat=True, path_prefix=...)` returns 23 routes, and **the last one is a catch-all `Mount(path="/{tenant}")`** for multi-tenant hosts. `server.py:119-125` calls it once per alias and flattens everything into a single list, so the *first* alias's tenant mount ends up ahead of every later alias's flat routes.

Starlette dispatches on the first `Match.FULL`, and a `Mount` matches on prefix alone — so `/agents/<anything>/...` is captured by alias #1's mount and 404s inside its inner router, which only knows alias #1's paths.

Agent-card and JSON-RPC routes survive purely because `protocol_routes` is concatenated *before* `rest_routes` (`server.py:127`). Everything under the REST binding does not.

Measured against the real `GatewayServer._build_app()` with two peers:

| alias | `/.well-known/agent-card.json` | `POST /message:stream` |
|---|---|---|
| `weather-agent` (first registered) | 200 | **200** |
| `uuid-weather` | 200 | **404** |
| `billing-agent` | 200 | **404** |
| `uuid-billing` | 200 | **404** |

**Impact:** the REST streaming endpoint works for exactly one peer, and never for a UUID alias. A single-peer gateway still loses its UUID alias. This is the endpoint advertised in `README.md:611`, `examples/a2a_gateway/README.md`, `examples/a2a_gateway/01_basic_gateway.py:90,99`, and `examples/run_agent.py:983`.

### Why CI is green

Every test in `test_server.py` uses one peer (`build_server()`, line 42), and the only multi-alias test — `test_uuid_peer_alias_serves_the_same_agent_card` (line 106) — hits the card route, which is in the group that isn't shadowed. Nothing exercises a second peer's REST route.

### Test to reproduce

Add to `tests/integrations/a2a/gateway/test_server.py` (reuses the existing `FakeExecutor` and `make_peer`):

```python
def build_multi_peer_server() -> GatewayServer:
    weather = make_peer("uuid-weather", "Weather Agent", "Gets weather info")
    billing = make_peer("uuid-billing", "Billing Agent", "Answers billing questions")
    return GatewayServer(
        peers={"weather-agent": weather, "billing-agent": billing},
        peers_by_uuid={weather.id: weather, billing.id: billing},
        gateway_url="http://localhost:10000",
        port=10000,
        executor_factory=lambda _slug: FakeExecutor(),
    )


@pytest_asyncio.fixture
async def multi_peer_client() -> AsyncIterator[httpx.AsyncClient]:
    transport = ASGITransport(app=build_multi_peer_server()._build_app())
    async with httpx.AsyncClient(transport=transport, base_url="http://test") as client:
        yield client


async def test_rest_binding_is_reachable_for_every_peer_and_alias(
    multi_peer_client: httpx.AsyncClient,
) -> None:
    """Every peer must serve the REST route the gateway docs advertise.

    A gateway hosting more than one peer is the normal case, and each peer is
    addressable by slug and by UUID, so all four routes have to answer.
    """
    aliases = ("weather-agent", "uuid-weather", "billing-agent", "uuid-billing")

    reached_handler = {}
    for alias in aliases:
        response = await multi_peer_client.post(
            f"/agents/{alias}/message:stream",
            headers={"A2A-Version": "1.0"},
            json={
                "message": {
                    "messageId": "message-1",
                    "role": "ROLE_USER",
                    "parts": [{"text": "Hello"}],
                }
            },
        )
        reached_handler[alias] = response.status_code

    assert reached_handler == dict.fromkeys(aliases, 200)
```

On this branch it fails with:

```
Differing items:
{'uuid-weather':  404} != {'uuid-weather':  200}
{'billing-agent': 404} != {'billing-agent': 200}
{'uuid-billing':  404} != {'uuid-billing':  200}
```

(`weather-agent: 200` is the one matching item.)

### Suggested fix

This gateway namespaces peers by path and isn't multi-tenant, so the tenant mounts can be dropped — in `server.py:119`:

```python
rest_routes.extend(
    route
    for route in create_rest_routes(
        handler,
        enable_v0_3_compat=True,
        path_prefix=f"/agents/{alias}",
    )
    # Each call also returns a catch-all "/{tenant}" mount for multi-tenant
    # hosts. Peers here are namespaced by path, and the first mount would
    # shadow every later peer's REST routes.
    if not isinstance(route, Mount)
)
```

plus `from starlette.routing import BaseRoute, Mount, Route`.

Verified: with that change the new test passes and the rest of the suite is unaffected (36 passed).

---

### Related, separate from the above: `/v1/` is the *legacy* REST path

Worth fixing while in here. Under a2a-sdk 1.1.2 the two REST paths map to different protocol versions:

| Path | Protocol |
|---|---|
| `/agents/<peer>/message:stream` | 1.0 |
| `/agents/<peer>/v1/message:stream` | v0.3 compat |

Posting with `A2A-Version: 1.0` to the `/v1/` path returns:

```
400 — A2A version '1.0' is not supported by this handler. Expected version '0.3'.
```

So the `/v1` prefix reads like a newer route but is the legacy binding. `examples/a2a_gateway/README.md` currently tells 1.0 clients to "use the versioned route" `/v1/message:stream` with a `SendMessageRequest` wrapper — a 1.0 client needs `/agents/<peer>/message:stream` instead. Same path appears in `README.md:611`, `examples/a2a_gateway/01_basic_gateway.py:90,99`, and `examples/run_agent.py:983`.

Also note `test_rest_stream_runs_through_upstream_handler` covers `/message:stream`, so the `/v1` path the docs advertise has no test.

🤖 Review generated with [Claude Code](https://claude.com/claude-code)

amit-gazal-band:
## Follow-up: secondary findings

Separate from the [routing bug above](https://github.com/band-ai/band-sdk-python/pull/488#issuecomment-5105490429). Each of these was reproduced by running it against `a2a-sdk==1.1.2`, not just read off the diff. Roughly in severity order.

---

### 1. `GET /agents/<peer>/tasks` returns conversation content, unauthenticated

Adopting `create_rest_routes` brings in task endpoints the gateway never had before. The task listing is reachable with no credentials and inlines the agent's reply text:

```
GET /agents/weather-agent/tasks   ->  200

{"tasks":[{"id":"3c7e…","contextId":"4242…","status":{"state":"TASK_STATE_COMPLETED",
 "message":{"role":"ROLE_AGENT","parts":[{"text":"The salary review is scheduled Friday"}]…
```

The pre-PR server exposed only the card, `message:stream`, and JSON-RPC `message/send|stream` — no way to enumerate or read past tasks. Combined with the `host="0.0.0.0"` bind (`server.py:166`, pre-existing) and no auth layer, this is a content-disclosure change rather than just a larger surface. `tasks/{id}` and `tasks/{id}:cancel` are live too; `extendedAgentCard` and the push-notification config routes currently fail closed with 400.

If these aren't wanted, they can be filtered out in the same pass as the tenant mounts.

---

### 2. `_handle_event` lost its `try/finally`, stranding a dead `task_id` on the room

`integrations/a2a/adapter.py:185-190`. On `main`, `_emit_task_event` + cleanup ran in a `finally`, so a delivery failure still finalized the task. Now they're sequential, so if `_deliver_task_update` raises the room keeps pointing at a completed task.

Reproduced with `tools.send_message` raising on a completed task carrying artifact text:

```
send_message awaited : 1
exception propagated : RuntimeError('platform down')
_tasks               : {'room-1': 'task-1'}   <- stale
task event emitted   : 0                      <- rehydration record lost
outbound task_id     : 'task-1'               <- next turn continues a completed task
```

`on_message` catches the exception and posts an error event, so nothing crashes — the room just silently keeps sending `task_id` of a finished task on every subsequent turn. Restoring the `finally` fixes it.

---

### 3. Each alias gets its own task store, so tasks are invisible across aliases

`server.py:92-96` constructs a fresh `DefaultRequestHandler` + `InMemoryTaskStore` per alias — 2 handlers and 2 distinct stores for a single peer. A task started on the slug can't be fetched or cancelled on the UUID:

```
task created via POST /agents/weather-agent
  GET /agents/weather-agent/tasks/<id>  -> 200
  GET /agents/uuid-w/tasks/<id>         -> 404
```

Note this is currently *masked* by the tenant-mount bug (the UUID alias 404s for unrelated reasons) — it surfaces the moment that's fixed, so worth handling in the same change. One handler per peer, shared across its aliases, solves this and item 5 together.

---

### 4. The new `httpx.AsyncClient` is never closed — and `stop()` is not the hook

`integrations/a2a/adapter.py:106` creates a client the adapter owns. `main` didn't have this: it used `ClientFactory.connect(...)`, which managed its own transport. So the PR introduces an ownership obligation it doesn't discharge.

The fix isn't a `stop()` method — `Agent.stop()` only ever calls `cleanup_all()`:

```python
# agent.py:304
cleanup_all = getattr(self._adapter, "cleanup_all", None)
```

Neither A2A adapter overrides it, so `async def cleanup_all(self)` on `A2AAdapter` is the right place to `aclose()`.

**Related, pre-existing (not this PR):** `A2AGatewayAdapter.stop()` (`gateway/adapter.py:264`) is dead code for the same reason — nothing in the SDK calls it, so the uvicorn server leaks on agent shutdown. It exists on `main` too, but `SimpleAdapter.cleanup_all`'s own docstring names "a self-hosted server" as exactly what belongs in that hook, and this PR is already in the file.

---

### 5. `_task_cache` entries created by resubscribe are never reclaimed

`on_cleanup` (`integrations/a2a/adapter.py:300`) derives its keys from `_task_senders`, but `_try_resubscribe` → `_reduce_task_event` populates `_task_cache` without ever touching `_task_senders`:

```
after resubscribe-style reduce : _task_cache=[('room-2','task-9')]  _task_senders=[]
after on_cleanup('room-2')     : _task_cache=[('room-2','task-9')]   <- still there
```

Small, but it's an unbounded dict on a long-lived adapter. Iterating `_task_cache`'s own keys fixes it.

---

### 6. Route count grows ~52 per peer

Measured on the real `_build_app()`:

| peers | top-level routes |
|---|---|
| 1 | 53 |
| 2 | 105 |
| 5 | 261 |
| 50 | **2601** |

Two aliases × (1 card + 1 legacy card + 1 JSON-RPC + 23 REST). Starlette matches linearly, so every request on a 50-peer gateway walks a 2601-entry list, and each peer also carries 2 handlers + 2 task stores. A single parameterized `/agents/{peer_id}` route set with peer resolution inside the executor would collapse this and fix item 3 at the same time.

---

### 7. Timeout path logs success

`gateway/adapter.py:326-331`. `_await_response` swallows `TimeoutError`, so the `else:` branch runs anyway:

```
WARNING A2A response timed out: room=room-1 task=b16ba967… timeout=0.01s
INFO    A2A request completed: room=room-1 task=b16ba967…
```

---

### Smaller items

- **`protocol.py:94`** — `is_terminal_state()` is dead code; no caller in `src`, `tests`, or `examples`.
- **`server.py:42`** — `self.peers_by_uuid` is now write-only; aliasing is derived from `peer.id` in `_build_app` instead.
- **`integrations/a2a/adapter.py:174`** — `getattr(event, "HasField", lambda _name: False)("message")`. `StreamResponse` always has `HasField`, and all six `_handle_event` tests pass real protobuf events, so nothing motivates the guard. It's also a trap: `MagicMock().HasField("message")` is **truthy**, so a future mock-based test would silently take the message branch. Plain `event.HasField("message")` reads as what's meant.
- **`gateway/types.py:86`** — `def _message(self, content: str):` is missing its return type.
- **`server.py:147`** — `_handle_list_peers` re-imports `JSONResponse` although it's already imported at line 20; the signature is also `(_request: Any) -> Any` where `(request: Request) -> JSONResponse` is available.
- **`gateway/adapter.py:298`** — `_establish_request` runs before the `try` at line 306, so peer-not-found and room-creation failures skip the structured `logger.exception` the rest of the method provides.
- **`CLAUDE.md:209-211`** — the A2A key-files table still lists only `types.py`; the new `protocol.py` and `config.py` aren't there, and `response_timeout_s` isn't documented outside the README snippet.
- **`test_adapter.py:80`** — `test_timeout_is_adapter_configuration` mostly asserts the constructor stored its argument. `test_timeout_returns_terminal_failure` is the one carrying the weight.

---

### Verified as fine

Worth recording so they don't get re-litigated:

- **Dropping `_list_peers_page_with_retry` is safe.** The Fern client's `_should_retry` already covers `{429, 408, 409}` and 5xx with jittered backoff honoring `Retry-After` / `RateLimit-Reset`, and `DEFAULT_REQUEST_OPTIONS = {"max_retries": 3}` is threaded through. Fewer attempts than the old five, but it respects server-supplied backoff, which the hand-rolled loop did not.
- **Gateway cancellation is correct.** `ActiveTask.cancel` calls `self._producer_task.cancel()` *before* `agent_executor.cancel`, so `_execute_a2a`'s `CancelledError` handler runs and `pending_task`'s `finally` releases the room slot.
- **`raise InternalError()` / `raise UnsupportedOperationError()`** in the examples is valid — in 1.1.2 these derive from `A2AError(Exception)`, so dropping the `ServerError` wrapper is right.
- `PendingA2ATask`'s lock + `done` guard correctly make `fail`/`complete`/`cancel` idempotent and mutually exclusive.

🤖 Review generated with [Claude Code](https://claude.com/claude-code)

src/band/integrations/a2a/gateway/server.py
```diff
@@ -180,283 +76,104 @@ async def _handle_agent_card(self, request: Request) -> JSONResponse:
...
+                        enable_v0_3_compat=True,
+                    )
+                )
+                rest_routes.extend(
+                    create_rest_routes(
+                        handler,
+                        enable_v0_3_compat=True,
+                        path_prefix=f"/agents/{alias}",
+                    )
+                )
```
amit-gazal-band: **Blocking — this shadows the REST binding for every peer after the first.**

`create_rest_routes(...)` returns 23 routes, and the last is a catch-all `Mount(path="/{tenant}")` for multi-tenant hosts. Called once per alias and flattened into one list, the *first* alias's mount lands ahead of every later alias's flat routes. Starlette dispatches on the first `Match.FULL` and a `Mount` matches on prefix alone, so `/agents/<anything>/...` is captured by alias #1's mount and 404s in its inner router.

Measured on the real `_build_app()` with two peers:

| alias | agent-card.json | `POST /message:stream` |
|---|---|---|
| `weather-agent` (first) | 200 | **200** |
| `uuid-weather` | 200 | **404** |
| `billing-agent` | 200 | **404** |
| `uuid-billing` | 200 | **404** |

Cards and JSON-RPC survive only because `protocol_routes` is concatenated before `rest_routes` on line 129. A single-peer gateway still loses its UUID alias. This is the endpoint advertised in `README.md:611`, `examples/a2a_gateway/README.md`, `examples/a2a_gateway/01_basic_gateway.py:90,99` and `examples/run_agent.py:983`.

This gateway namespaces peers by path and isn't multi-tenant, so the mounts can go:

```python
rest_routes.extend(
    route
    for route in create_rest_routes(
        handler,
        enable_v0_3_compat=True,
        path_prefix=f"/agents/{alias}",
    )
    # Each call also returns a catch-all "/{tenant}" mount for multi-tenant
    # hosts. Peers here are namespaced by path, and the first mount would
    # shadow every later peer's REST routes.
    if not isinstance(route, Mount)
)
```

plus `Mount` in the `starlette.routing` import. Verified: with that change the repro test below passes and the suite is unaffected (36 passed).

Repro test for `tests/integrations/a2a/gateway/test_server.py` (reuses the existing `FakeExecutor` / `make_peer`):

```python
def build_multi_peer_server() -> GatewayServer:
    weather = make_peer("uuid-weather", "Weather Agent", "Gets weather info")
    billing = make_peer("uuid-billing", "Billing Agent", "Answers billing questions")
    return GatewayServer(
        peers={"weather-agent": weather, "billing-agent": billing},
        peers_by_uuid={weather.id: weather, billing.id: billing},
        gateway_url="http://localhost:10000",
        port=10000,
        executor_factory=lambda _slug: FakeExecutor(),
    )


@pytest_asyncio.fixture
async def multi_peer_client() -> AsyncIterator[httpx.AsyncClient]:
    transport = ASGITransport(app=build_multi_peer_server()._build_app())
    async with httpx.AsyncClient(transport=transport, base_url="http://test") as client:
        yield client


async def test_rest_binding_is_reachable_for_every_peer_and_alias(
    multi_peer_client: httpx.AsyncClient,
) -> None:
    """Every peer must serve the REST route the gateway docs advertise.

    A gateway hosting more than one peer is the normal case, and each peer is
    addressable by slug and by UUID, so all four routes have to answer.
    """
    aliases = ("weather-agent", "uuid-weather", "billing-agent", "uuid-billing")

    reached_handler = {}
    for alias in aliases:
        response = await multi_peer_client.post(
            f"/agents/{alias}/message:stream",
            headers={"A2A-Version": "1.0"},
            json={
                "message": {
                    "messageId": "message-1",
                    "role": "ROLE_USER",
                    "parts": [{"text": "Hello"}],
                }
            },
        )
        reached_handler[alias] = response.status_code

    assert reached_handler == dict.fromkeys(aliases, 200)
```

Fails on this branch with `uuid-weather`/`billing-agent`/`uuid-billing` at 404.

src/band/integrations/a2a/gateway/server.py
```diff
@@ -180,283 +76,104 @@ async def _handle_agent_card(self, request: Request) -> JSONResponse:
...
-            )
 
-        slug, peer = resolved
-        body = await request.json()
-
-        # Extract JSON-RPC fields
-        method = body.get("method", "")
-        request_id = body.get("id")
-        params = body.get("params", {})
+    def _build_app(self) -> Starlette:
```
amit-gazal-band, line 106: **Unauthenticated task listing returns conversation content.**

`create_rest_routes` brings in task endpoints the gateway never had. Reachable with no credentials, and the listing inlines the agent's reply text:

```
GET /agents/weather-agent/tasks   ->  200

{"tasks":[{"id":"3c7e...","contextId":"4242...","status":{"state":"TASK_STATE_COMPLETED",
 "message":{"role":"ROLE_AGENT","parts":[{"text":"The salary review is scheduled Friday"}]...
```

The pre-PR server exposed only the card, `message:stream`, and JSON-RPC `message/send|stream` — no way to enumerate or read past tasks. With the `host="0.0.0.0"` bind on line 166 (pre-existing) and no auth layer, this is a content-disclosure change, not just a wider surface. `tasks/{id}` and `tasks/{id}:cancel` are live too; `extendedAgentCard` and the push-config routes fail closed with 400.

If that isn't intended, those routes can be filtered in the same pass as the tenant mounts above.

src/band/integrations/a2a/gateway/server.py
```diff
@@ -180,283 +76,104 @@ async def _handle_agent_card(self, request: Request) -> JSONResponse:
...
+        for slug, peer in self.peers.items():
+            aliases = dict.fromkeys((slug, peer.id))
+            for alias in aliases:
+                card = self._agent_card(alias, peer)
+                executor = self.executor_factory(slug)
+                handler = DefaultRequestHandler(
+                    agent_executor=executor,
+                    task_store=InMemoryTaskStore(),
+                    agent_card=card,
+                )
```
amit-gazal-band: **One handler + task store per *alias* means tasks are invisible across aliases**, and it drives the route count.

A single peer gets 2 `DefaultRequestHandler`s and 2 distinct `InMemoryTaskStore`s. A task started on the slug can't be fetched or cancelled on the UUID:

```
task created via POST /agents/weather-agent
  GET /agents/weather-agent/tasks/<id>  -> 200
  GET /agents/uuid-w/tasks/<id>         -> 404
```

Currently masked by the mount bug (the UUID alias 404s for an unrelated reason), so it surfaces the moment that's fixed — worth handling together.

Route growth measured on the real `_build_app()`:

| peers | top-level routes |
|---|---|
| 1 | 53 |
| 2 | 105 |
| 5 | 261 |
| 50 | **2601** |

~52 per peer: 2 aliases x (1 card + 1 legacy card + 1 JSON-RPC + 23 REST). Starlette matches linearly, so a 50-peer gateway walks a 2601-entry list per request. One handler per peer shared across its aliases — or a single parameterized `/agents/{peer_id}` route set resolving the peer inside the executor — fixes both this and the split store.

src/band/integrations/a2a/gateway/server.py
```diff
@@ -1,172 +1,68 @@
...
-
-        Args:
-            peers: Dict of slug → Peer objects (primary lookup).
-            peers_by_uuid: Dict of UUID → Peer objects (fallback lookup).
-            gateway_url: Base URL for AgentCard URLs (e.g., "http://localhost:10000").
-            port: Port to listen on.
-            on_request: Async callback invoked with (peer_id, message) → events.
-        """
         self.peers = peers
         self.peers_by_uuid = peers_by_uuid
```
amit-gazal-band: `self.peers_by_uuid` is now write-only — `_build_app` derives the UUID alias from `peer.id` instead, and nothing else reads this. Either drop the constructor parameter or use it for resolution.

src/band/integrations/a2a/gateway/server.py
```diff
@@ -180,283 +76,104 @@ async def _handle_agent_card(self, request: Request) -> JSONResponse:
...
-            except Exception as e:
-                logger.exception("Error in JSON-RPC stream handler: %s", e)
-                error_response = {
-                    "jsonrpc": "2.0",
-                    "error": {"code": -32000, "message": str(e)},
-                    "id": request_id,
-                }
-                yield f"data: {json.dumps(error_response)}\n\n"
+    async def _handle_list_peers(self, _request: Any) -> Any:
+        from starlette.responses import JSONResponse
```
amit-gazal-band: `JSONResponse` is already imported at line 20; this local re-import is redundant. The signature is also `(_request: Any) -> Any` where `(request: Request) -> JSONResponse` is available.

src/band/integrations/a2a/adapter.py
```diff
@@ -200,91 +164,100 @@ async def on_message(
...
 
-        # Store sender info on first event for this task
-        if key not in self._task_senders:
-            self._task_senders[key] = {"id": sender_id, "name": sender_name or ""}
+        await self._deliver_task_update(task, tools, self._task_senders[key])
+        if task.status.state == TaskState.TASK_STATE_INPUT_REQUIRED:
+            await self._emit_task_event(tools, task, task.status.state)
+        elif task.status.state in TERMINAL_TASK_STATES:
+            await self._emit_task_event(tools, task, task.status.state)
+            self._finalize_task(room_id, task.id)
```
amit-gazal-band: **Regression: the old `try/finally` guaranteed terminal cleanup; this doesn't.**

On `main`, `_emit_task_event` + sender/task cleanup ran in a `finally`, so a delivery failure still finalized the task. Sequential now, so if `_deliver_task_update` raises the room keeps pointing at a completed task.

Reproduced with `tools.send_message` raising on a completed task carrying artifact text:

```
send_message awaited : 1
exception propagated : RuntimeError('platform down')
_tasks               : {'room-1': 'task-1'}   <- stale
task event emitted   : 0                      <- rehydration record lost
outbound task_id     : 'task-1'               <- next turn continues a completed task
```

`on_message` catches the exception and posts an error event, so nothing crashes — the room just silently sends the finished `task_id` on every later turn. Restoring the `finally` around the terminal branch fixes it.

src/band/integrations/a2a/adapter.py
```diff
@@ -136,16 +103,11 @@ async def on_started(self, agent_name: str, agent_description: str) -> None:
...
-        # Build client configuration
-        config = ClientConfig(streaming=self.streaming)
-
-        # Connect to remote A2A agent
-        self._client = await ClientFactory.connect(
-            agent=self.remote_url,
-            client_config=config,
-            interceptors=[_StaticHeadersInterceptor(headers)] if headers else None,
-            resolver_http_kwargs={"headers": headers} if headers else None,
+        self._http_client = httpx.AsyncClient(headers=headers)
```
amit-gazal-band, line 109: **This client is never closed, and `stop()` is not the hook that would close it.**

`main` didn't own a client here — it used `ClientFactory.connect(...)`, which managed its own transport. This introduces an ownership obligation the adapter never discharges.

`Agent.stop()` only ever calls `cleanup_all()`:

```python
# agent.py:304
cleanup_all = getattr(self._adapter, "cleanup_all", None)
```

Neither A2A adapter overrides it, so `async def cleanup_all(self)` on `A2AAdapter` is the right place to `aclose()`.

Related and pre-existing (not this PR): `A2AGatewayAdapter.stop()` is dead for the same reason — nothing calls it, so the uvicorn server leaks on shutdown. It's on `main` too, but `SimpleAdapter.cleanup_all`'s docstring names "a self-hosted server" as exactly what belongs in that hook, and this PR is already in these files.

src/band/integrations/a2a/adapter.py
```diff
@@ -200,91 +164,100 @@ async def on_message(
...
         """Handle A2A event and forward to Band platform."""
-        # Handle direct message reply (rare - most responses come via Task)
-        if isinstance(event, A2AMessage):
-            text = get_message_text(event)
-            if text:
-                await tools.send_message(
-                    content=text,
-                    mentions=[{"id": sender_id, "name": sender_name or ""}],
-                )
+        if getattr(event, "HasField", lambda _name: False)("message"):
```
amit-gazal-band: `StreamResponse` is protobuf and always has `HasField`, and all six `_handle_event` tests pass real protobuf events — nothing motivates the guard.

It's also a trap: `MagicMock().HasField("message")` returns a **truthy** Mock, so a future mock-based test would silently take the message branch. Plain `event.HasField("message")` reads as what's meant.

src/band/integrations/a2a/adapter.py
```diff
@@ -349,6 +300,7 @@ async def on_cleanup(self, room_id: str) -> None:
         keys_to_remove = [key for key in self._task_senders if key[0] == room_id]
         for key in keys_to_remove:
             self._task_senders.pop(key, None)
+            self._task_cache.pop(key, None)
```
amit-gazal-band: `_task_cache` entries created by `_try_resubscribe` are never reclaimed — it goes through `_reduce_task_event`, which populates `_task_cache` without ever touching `_task_senders`, and these keys come from `_task_senders`:

```
after resubscribe-style reduce : _task_cache=[('room-2','task-9')]  _task_senders=[]
after on_cleanup('room-2')     : _task_cache=[('room-2','task-9')]   <- still there
```

Small, but it's an unbounded dict on a long-lived adapter. Iterating `_task_cache`'s own keys for the room fixes it.

src/band/integrations/a2a/gateway/adapter.py
```diff
@@ -295,75 +283,145 @@ def _resolve_peer(self, peer_id: str) -> Peer | None:
...
+                request.context_id,
+                request.pending.task.id,
+            )
+            raise
+        else:
+            logger.info(
+                "A2A request completed: room=%s task=%s",
+                request.room_id,
+                request.pending.task.id,
+            )
```
amit-gazal-band: The timeout path reaches this `else`, so a timed-out request logs success. `_await_response` swallows `TimeoutError`, so no exception propagates:

```
WARNING A2A response timed out: room=room-1 task=b16ba967... timeout=0.01s
INFO    A2A request completed: room=room-1 task=b16ba967...
```

Having `_await_response` signal whether it completed or timed out would let this log the truth.

src/band/integrations/a2a/gateway/adapter.py
```diff
@@ -295,75 +283,145 @@ def _resolve_peer(self, peer_id: str) -> Peer | None:
...
+        )
 
-        Args:
-            peer_id: Target peer slug or UUID.
-            message: A2A message from remote agent.
+    async def _execute_a2a(
+        self, peer_id: str, context: RequestContext, event_queue: EventQueue
+    ) -> None:
+        """Bridge one official A2A execution to a Band room."""
+        request = await self._establish_request(peer_id, context, event_queue)
```
amit-gazal-band: `_establish_request` runs before the `try` below, so peer-not-found and room-creation failures skip the structured `logger.exception` the rest of this method is careful to provide — they surface as a bare traceback from the SDK instead.

src/band/integrations/a2a/protocol.py
```diff
@@ -0,0 +1,101 @@
...
+    for message in reversed(task.history):
+        if message.role == Role.ROLE_AGENT:
+            text = get_message_text(message)
+            if text:
+                return text
+
+    return ""
+
+
+def is_terminal_state(state: int) -> bool:
```
amit-gazal-band: `is_terminal_state()` has no caller anywhere in `src`, `tests`, or `examples` — callers compare against `TERMINAL_TASK_STATES` directly. Dead on arrival; worth dropping.

src/band/integrations/a2a/gateway/types.py
```diff
@@ -34,10 +37,55 @@ class PendingA2ATask:
...
+
+    async def cancel(self) -> None:
+        """Publish a terminal cancellation and release the request."""
+        async with self._lock:
+            if self.done.is_set():
+                return
+            await self._updater.cancel()
+            self.done.set()
+
+    def _message(self, content: str):
```
amit-gazal-band: Missing return type — should be `-> Message`. CLAUDE.md requires annotations on every function.

amit-gazal-band (changes requested):
## Request changes

The A2A 1.1.2 migration is the right direction — leaning on the upstream route factories and deleting ~3200 lines of hand-rolled JSON-RPC/SSE is a clear win, `protocol.py`'s reducer is well factored, and the aliasing test in `test_protocol.py` is the kind that catches real regressions. One blocker and a few correctness issues before it lands.

Everything below was reproduced by running it against `a2a-sdk==1.1.2`, not read off the diff.

**Blocking**

- Per-alias `Mount("/{tenant}")` shadows the REST binding for every peer after the first, so `/agents/<peer>/message:stream` works for exactly one peer and never for a UUID alias. Repro test + one-line fix inline on `server.py`.

**Should fix before merge**

- `GET /agents/<peer>/tasks` returns conversation content unauthenticated — new exposure vs. the pre-PR server.
- `_handle_event` lost its `try/finally`, stranding a completed `task_id` on the room when delivery fails.
- Per-alias task stores make a task invisible across a peer's own aliases (surfaces once the mount bug is fixed).
- The new `httpx.AsyncClient` is never closed; `cleanup_all()` is the hook, not `stop()`.

**Smaller** — `_task_cache` leak, contradictory timeout log, ~52 routes per peer, dead `is_terminal_state()` / `peers_by_uuid`, the `getattr(HasField)` guard, missing return type, redundant import.

**Not inline-able**

- `CLAUDE.md:209-211` — the A2A key-files table still lists only `types.py`; the new `protocol.py` and `config.py` aren't there, and `response_timeout_s` isn't documented outside the README snippet.
- Docs point 1.0 clients at the wrong path: under 1.1.2, `/agents/<peer>/v1/message:stream` is the **v0.3** binding (`A2A-Version: 1.0` there returns `400 - A2A version '1.0' is not supported by this handler. Expected version '0.3'`). 1.0 clients need `/agents/<peer>/message:stream`. Affects `README.md:611`, `examples/a2a_gateway/README.md`, `examples/a2a_gateway/01_basic_gateway.py:90,99`, `examples/run_agent.py:983`. `test_rest_stream_runs_through_upstream_handler` covers `/message:stream`, so the documented `/v1` path has no test.
- `test_adapter.py` — `test_timeout_is_adapter_configuration` mostly asserts the constructor stored its argument; `test_timeout_returns_terminal_failure` is the one carrying the weight.

**Verified as fine**, so they don't get re-litigated: dropping `_list_peers_page_with_retry` is safe (Fern's `_should_retry` already covers `{429, 408, 409}` + 5xx with `Retry-After`-aware backoff, `max_retries=3` threaded through); gateway cancellation is correct (`ActiveTask.cancel` cancels the producer before calling `agent_executor.cancel`, so `pending_task`'s `finally` releases the slot); `raise InternalError()` in the examples is valid since 1.1.2 errors derive from `A2AError(Exception)`; and `PendingA2ATask`'s lock + `done` guard correctly make the terminal transitions idempotent.

Detail on each of these is in the two comments above.

🤖 Review generated with [Claude Code](https://claude.com/claude-code)

src/band/integrations/a2a/gateway/server.py
AlexanderZ-Band: Confirmed and fixed in f4985f3a — reproduced the 404s exactly as measured. Went slightly further than dropping the mounts: `_messaging_rest_routes()` keeps only the message:send/message:stream/card endpoints (allowlist), which also removes the mount. Your repro test is added as `test_rest_binding_is_reachable_for_every_peer_and_alias` and passes.

src/band/integrations/a2a/gateway/server.py
AlexanderZ-Band: Confirmed — reproduced the unauthenticated `GET /tasks` returning conversation content. Fixed in f4985f3a: the REST surface is filtered to the messaging endpoints + compat card (`MESSAGING_REST_SUFFIXES` allowlist), restoring the pre-PR exposure. Guarded by `test_task_rest_routes_are_not_exposed`.

src/band/integrations/a2a/gateway/server.py
AlexanderZ-Band: Confirmed and fixed in f4985f3a: one `DefaultRequestHandler` + `InMemoryTaskStore` per peer, shared across its slug/UUID aliases — a task started on the slug is now visible on the UUID (via JSON-RPC GetTask; the REST task routes are gone per the auth finding). Combined with the REST filtering, route count for 50 peers drops 2601 → 801. Kept the per-alias route sets rather than a parameterized `/agents/{peer_id}` route — the shared handler fixes the store split, and the remaining linear growth is card/JSON-RPC routes that are cheap to match. `test_task_started_on_slug_is_visible_via_uuid_alias` covers it.

src/band/integrations/a2a/gateway/server.py
AlexanderZ-Band: Fixed in f4985f3a — dropped the constructor parameter. The adapter keeps its own `_peers_by_uuid` for `_resolve_peer`, which is the actual resolution path.

src/band/integrations/a2a/gateway/server.py
AlexanderZ-Band: Fixed in f4985f3a — removed the local re-import and typed the handler `(_request: Request) -> JSONResponse`.

src/band/integrations/a2a/adapter.py
AlexanderZ-Band: Confirmed — this was a genuine regression, and the PR's `test_terminal_task_is_retained_when_band_delivery_fails` was asserting the bug (nothing retries the delivery, so retention only leaves a stale task_id). Fixed in f4985f3a: terminal emit + finalize restored under `finally`, and the test inverted to `test_terminal_task_is_finalized_even_when_band_delivery_fails`.

src/band/integrations/a2a/adapter.py
AlexanderZ-Band: Fixed in f4985f3a — `A2AAdapter.cleanup_all()` closes the A2A client and the owned httpx transport. Also took the related fix since we're in these files: `A2AGatewayAdapter.stop()` renamed to `cleanup_all()` so `Agent.stop()` actually stops the hosted uvicorn server. Both covered by new shutdown tests.

src/band/integrations/a2a/adapter.py
AlexanderZ-Band: Fixed in f4985f3a — plain `event.HasField("message")`. Agreed the getattr guard was unmotivated and a Mock trap.

src/band/integrations/a2a/adapter.py
AlexanderZ-Band: Confirmed and fixed in f4985f3a — `on_cleanup` now sweeps the union of `_task_senders` and `_task_cache` keys for the room. Guarded by `test_cleanup_reclaims_tasks_cached_by_resubscribe`.

src/band/integrations/a2a/gateway/adapter.py
AlexanderZ-Band: Fixed in f4985f3a — `_await_response` now returns whether the reply arrived, and the completion log only fires when it did. The existing timeout test now also asserts no 'completed' log via caplog.

src/band/integrations/a2a/gateway/adapter.py
AlexanderZ-Band: Fixed in f4985f3a — `_establish_request` is wrapped so setup failures get the same structured `logger.exception` (peer + context) as execution failures.

src/band/integrations/a2a/protocol.py
AlexanderZ-Band: Fixed in f4985f3a — dropped `is_terminal_state()`; confirmed no callers anywhere.

src/band/integrations/a2a/gateway/types.py
AlexanderZ-Band: Fixed in f4985f3a — `-> Message` added.

AlexanderZ-Band:
## Self-review round (25db472f)

Ran a deep verification pass over the full diff before requesting re-review. Everything below was reproduced empirically before fixing:

**Security — JSON-RPC task exposure (completes the earlier REST fix):** the REST filtering in f4985f3a closed only half the surface. The JSON-RPC dispatcher still served `ListTasks` (returned every caller's tasks with full message history — all unauthenticated callers share one task-store identity), plus push-config and extended-card methods. Now closed via `ALLOWED_JSONRPC_METHODS` (messaging + task-UUID-gated operations, 1.0 and v0.3 spellings), guarded by `test_non_messaging_jsonrpc_methods_are_not_exposed`. Follow-up for the broader bind/auth question: INT-1149.

**Socket leak on stop:** cancelling `serve()` skipped uvicorn's shutdown; a stop/start cycle failed with `EADDRINUSE`. Now sets `should_exit` and awaits. Verified rebind works.

**Docs that didn't work:** all three newly-authored gateway curl commands failed against the shipped server (v0.3 payload shapes fed to the 1.0 parser → `-32602`; `/v1/message:stream` mislabeled as the 1.0 REST binding → silent 200 + empty stream; demo docstring missing the `A2A-Version` header). All rewritten to shapes verified by execution.

**Hardening:** pre-migration (0.x) terminal state names recognized on rehydration; `cleanup_all` fails in-flight gateway requests; concurrent-request rejection no longer leaks the internal room id; resubscribe stream closed deterministically; proto presence checks fixed; `a2a-sdk` capped `<2`; demo `discover_peer` no longer closes the shared connection pool.

**Coverage restored** (dropped by the test rewrite, behaviors still live): auth-plumbing identity into `ClientConfig`, history-converter round trip with the new state vocabulary, remote-error and failed-task paths, context continuity across turns, bootstrap gating, `PendingA2ATask` terminal idempotency. Net +17 tests; full suite 4044 passed, ruff/pyrefly clean.

https://claude.ai/code/session_011Y2bdwA8CxTkD9SXxfU8et

amit-gazal-band (commented):
Migration itself checks out — I verified every changed call against a real `a2a-sdk==1.1.2` install (`ClientFactory.create_from_url`, `send_message(SendMessageRequest)`, `subscribe(SubscribeToTaskRequest(id=...))`, async `Client.close()`, `new_text_message` kwargs, `TaskUpdater.start_work/complete/failed/cancel`, `RequestContext` id generation, and `InternalError`/`UnsupportedOperationError` now being real exceptions). Locally: 4279 passed / 82 skipped, ruff clean, pyrefly 0 errors, README markdown-docs green.

One blocking item below (the `a2a_gateway` extra can't import, one line to fix), one question, one bit of praise.

pyproject.toml
```diff
@@ -88,16 +88,16 @@ gemini = [
     "google-genai>=1.43.0",  # httpx is a transitive dep used for retry exception types
 ]
 a2a = [
-    "a2a-sdk>=0.3.22",  # Official A2A Python SDK; brings protobuf>=5.29.5 transitively
+    "a2a-sdk>=1.1.2,<2",  # Official A2A Python SDK; brings protobuf transitively
 ]
 a2a_gateway = [
-    "a2a-sdk>=0.3.22",  # Official A2A Python SDK; brings protobuf>=5.29.5 transitively
+    "a2a-sdk>=1.1.2,<2",  # Official A2A Python SDK; brings protobuf transitively
```
amit-gazal-band: **This extra can no longer import the gateway.** The 1.1 route factories reach `a2a.compat.v0_3.jsonrpc_adapter`, which imports `sse_starlette` — and a2a-sdk ships that only under its `http-server` / `fastapi` / `all` extras, not in its core deps.

Reproduced in a clean venv with exactly this extra:

```bash
uv venv --python 3.11 repro
uv pip install '.[a2a_gateway]'   # installs a2a-sdk 1.1.2, starlette, uvicorn — no sse-starlette
```

```python
>>> from band.adapters.a2a_gateway import A2AGatewayAdapter, A2AGatewayAdapterConfig
band/integrations/a2a/gateway/server.py:12  -> a2a.server.routes.agent_card_routes
a2a/server/routes/__init__.py:9             -> a2a.server.routes.jsonrpc_routes
a2a/compat/v0_3/jsonrpc_adapter.py:6        -> from sse_starlette.sse import EventSourceResponse
ModuleNotFoundError: No module named 'sse_starlette'
```

It fails on the *first* of the three route-factory imports, at module load, so nothing about the gateway is reachable.

Suggested fix here and at line 100 (and line 181, so the suite stops depending on luck):

```toml
"a2a-sdk[http-server]>=1.1.2,<2",
```

Verified: adding `a2a-sdk[http-server]` pulls `sse-starlette==3.4.6` and the same import succeeds, adapter constructs fine.

**Why CI is green:** `sse-starlette` appears exactly once as a dependency in `uv.lock` — under `mcp` — and `mcp>=1.25.0` is in the `dev` extra. So `uv sync --extra dev` supplies it transitively while the published extra stays broken. Worth a CI step that installs `.[a2a_gateway]` alone and imports the module, since no unit test can catch this class of bug.

The plain `a2a` extra (line 91) is fine — I confirmed `band.adapters.a2a` imports without it, as the client path never touches `a2a.server`.

pyproject.toml
```diff
@@ -88,16 +88,16 @@ gemini = [
...
     "starlette>=0.40.0",  # HTTP server framework
     "uvicorn>=0.32.0",  # ASGI server
     "python-multipart>=0.0.22",
 ]
 a2a_gateway_demo = [
-    "a2a-sdk>=0.3.22",      # A2A client for calling gateway peers
+    "a2a-sdk>=1.1.2,<2",      # A2A client for calling gateway peers
```
amit-gazal-band: Same gap as line 94 — confirmed by clean-venv install: `.[a2a_gateway_demo]` -> `ModuleNotFoundError: No module named 'sse_starlette'`. Both `demo_orchestrator/__main__.py` and `02_with_demo_agent.py` now build their app from `create_jsonrpc_routes` / `create_rest_routes`, so this extra needs `a2a-sdk[http-server]` too.

src/band/integrations/a2a/gateway/adapter.py
```diff
@@ -163,63 +176,32 @@ async def on_started(self, agent_name: str, agent_description: str) -> None:
...
-            on_request=self._handle_a2a_request,
+            executor_factory=partial(BandAgentExecutor, self),
         )
         await self._server.start()
 
         logger.info("Gateway HTTP server started on port %d", self.port)
 
-    async def _fetch_all_peers_with_retry(self) -> list[Peer]:
-        """Fetch all peer pages, retrying if the platform rate-limits startup."""
+    async def _fetch_all_peers(self) -> list[Peer]:
```
amit-gazal-band, line 187: Dropping `_list_peers_page_with_retry` here trades 6 attempts over ~31s for `DEFAULT_REQUEST_OPTIONS`' `max_retries: 3` (~7s). The Fern client does retry 429 (`retryable_400s = [429, 408, 409]`) *and* honours `Retry-After`, which the bespoke loop didn't — so I think this is a net improvement in behaviour. Just flagging that startup gets less headroom, and the code being removed carried a comment saying the platform rate-limits gateway startup specifically.

Also: this is an unrelated behaviour change riding along in an SDK-migration PR. If it's deliberate cleanup, worth a line in the PR description so it isn't discovered in the diff later.

.github/dependabot.yml
```diff
@@ -9,6 +9,11 @@ updates:
       day: "monday"
       time: "09:00"
     open-pull-requests-limit: 10
+    groups:
```
amit-gazal-band, line 12: Very nice!

amit-gazal-band (dismissed):
Approved. Just one blocking issue.

pyproject.toml
AlexanderZ-Band: Fixed in 290df68. Both published gateway extras now require `a2a-sdk[http-server]>=1.1.2,<2`, which supplies `sse-starlette`; the `<2` bound intentionally keeps the migration on the validated A2A 1.x API. Added a packaging CI check that installs the built wheel with `a2a_gateway` alone and imports the adapter. Verified in a clean venv.

pyproject.toml
AlexanderZ-Band: Fixed in 290df68. `a2a_gateway_demo` now also requires `a2a-sdk[http-server]>=1.1.2,<2`; the packaging CI check installs that extra independently and imports the upstream REST route factory. Verified in a clean venv.

src/band/integrations/a2a/gateway/adapter.py
AlexanderZ-Band: Agreed. This is deliberate cleanup: Fern keeps retries and honors `Retry-After`, at the cost of less startup headroom than the bespoke loop. I added that tradeoff to the PR summary in 290df68 so it is explicit.

amit-gazal-band (approved):

PR #489, opened 2026-07-26
feat(sdk): add Strands Agents adapter

AlexanderZ-Band:
## What

Adds a **Strands Agents** adapter (Tier-1 injectable-model family) plus the wiring
that makes it a first-class adapter in this repo.

**Adapter** (`src/band/adapters/strands.py`)
- One `strands.Agent` per turn, so each turn's tools own that turn's
  `AgentToolsProtocol` and no room capability leaks into Strands' untyped
  `invocation_state`.
- Platform tools bridge to native `AgentTool`s from the shared Band registry
  (names, schemas, validation, serialization all stay single-sourced); custom tools
  accept both the portable `(InputModel, handler)` pair and Strands' own `@tool`
  functions.
- `HookProvider` emits `Emit.EXECUTION` pairs and decides the turn's terminal-action
  policy; `Emit.USAGE` maps Strands' accumulated usage.
- Band owns per-room history: the transcript is converted to Converse messages,
  seeded into the fresh `Agent`, and read back after the run.

**Converter** (`src/band/converters/strands.py`) converts Band history to Converse
`toolUse`/`toolResult` messages, batching parallel calls and preserving valid tool
message ordering.

**Baseline/E2E** adds `Adapter.STRANDS` to the shared matrix and offline support
registry, using the core lane and its injectable model seam.

## A2A note (rebase on #488)

This PR originally carried a dual-version A2A compatibility layer (a2a-sdk 0.3 + 1.x
behind a `compat.py` boundary, with a legacy 0.3.26 CI lane). #488 landed on `main`
in the meantime with a full 1.x-only migration of the A2A integration, including
security and lifecycle fixes (JSON-RPC method allowlist, removal of the
unauthenticated task routes, route-shadowing fix, server socket release, task-cache
eviction, pre-migration terminal-state rehydration). The compatibility layer was
superseded and has been dropped from this PR; `main`'s 1.x-only implementation and
its `a2a-sdk>=1.1.2,<2` pin stand as-is.

Retained from that work:
- Import smoke tests for the standalone A2A examples
  (`tests/example_agents/a2a/test_example_imports.py`), so an example drifting from
  the installed SDK's API fails CI at import time. These now guard `main`'s migrated
  1.x examples and pass against them.
- Read-only `GITHUB_TOKEN` permissions on the CI and E2E workflows.

## Also in this PR

- `ToolEventKey` provides one vocabulary for execution-event payload keys.
- `lazy_exports` replaces repeated hand-rolled lazy import maps.
- E2E smokes cover contacts capability, memory rehydration, and per-instance
  concurrency markers (the concurrency smoke folds those markers into the widened
  `slow_turn_budget` deadlines that landed on `main`).

## Docs

Updates the README adapter and emit matrices, configuration example, and
`examples/strands/` provider and usage documentation.

## Verification

Rebased onto `main` @ 03065f5.

Offline:

```text
4371 unit tests passed, 85 skipped
ruff check: passed
ruff format --check: passed
pyrefly check: passed (0 errors)
uv lock --check: passed
```

Live E2E — dispatched `e2e.yml` with `lane=core` / `os=ubuntu` on this branch's head
(run 30638603739, all three jobs green):

```text
258 passed, 135 skipped, 4 rerun in 36m34s  (0 failed)
```

The Strands cells ran live and passed rather than skipping — directly observed in the
job log, including the cells that exercise this PR's load-bearing paths:

```text
test_tool_round_trip.py::test_custom_tool_round_trips[strands]                        PASSED
test_room_isolation.py::test_rooms_keep_isolated_context[strands]                     PASSED
test_rehydration_cross_framework.py::test_rehydrates_foreign_peer_message[strands]    PASSED
test_rehydration_idempotency.py::test_handled_work_not_redrained_on_restart[strands]  PASSED
test_rehydration_idempotency.py::test_restart_usage_splits_replay_and_inference[strands] PASSED
test_rehydration_offline.py::test_recalls_offline_note_on_cold_boot[strands]          PASSED
test_rehydration_partial.py::test_partial_reboot_preserves_context_and_peer[strands]  PASSED
test_peer_delegation.py::test_peer_initiated_delegation_with_self_recall[strands]     PASSED
test_participant_management.py::test_invites_messages_and_removes_a_peer[strands]     PASSED
```

The lane ran with `BAND_ALLOW_MISSING_FRAMEWORKS=0`, so an unwired framework fails
loudly instead of skipping; every skip in the run belongs to an out-of-lane adapter
(crewai, gemini, google_adk, letta, codex, opencode, copilot_acp), each covered by
its own lane.

### DIFF
--- a/src/band/adapters/strands.py
+++ b/src/band/adapters/strands.py
@@ -1,0 +1,746 @@
+"""
+Strands Agents adapter using SimpleAdapter pattern.
+
+Modeled on the pydantic-ai adapter: the framework owns the agent loop, the
+model object is injectable at the public ``Agent(model=...)`` seam, and Band
+platform tools are registered as native framework tools.
+"""
+
+from __future__ import annotations
+
+import json
+import logging
+from typing import Any, Callable, ClassVar
+
+import httpx
+
+try:
+    from strands import Agent, ToolContext, tool
+    from strands.hooks import HookProvider, HookRegistry
+    from strands.hooks.events import AfterToolCallEvent, BeforeToolCallEvent
+    from strands.models import Model
+    from strands.types.tools import (
+        AgentTool,
+        ToolGenerator,
+        ToolResult,
+        ToolSpec,
+        ToolUse,
+    )
+except ImportError as e:
+    raise ImportError(
+        "Strands Agents dependencies not installed. "
+        "Install with: uv add band-sdk[strands]"
+    ) from e
+
+from band_rest.core.api_error import ApiError
+
+from band.core.protocols import AgentToolsProtocol
+from band.core.simple_adapter import SimpleAdapter
+from band.core.types import (
+    AdapterFeatures,
+    Capability,
+    Emit,
+    PlatformMessage,
+    TurnUsage,
+)
+from band.converters.strands import StrandsHistoryConverter, StrandsMessages
+from band.runtime.custom_tools import (
+    CustomToolDef,
+    execute_custom_tool,
+    get_custom_tool_name,
+    is_marked_terminal,
+)
+from band.runtime.prompts import render_system_prompt
+from band.runtime.tools import (
+    band_tool_errored,
+    get_tool_description,
+    is_terminal_success,
+)
+
+logger = logging.getLogger(__name__)
+
+# invocation_state key carrying the current turn's AgentToolsProtocol handle.
+# Strands threads invocation_state from Agent.invoke_async(...) into every tool
+# body (via ToolContext) and hook event, which is what lets the tool functions
+# be built once while each turn binds its own tools handle.
+_BAND_TOOLS_STATE_KEY = "band_tools"
+
+
+def _tools_from_context(tool_context: ToolContext) -> AgentToolsProtocol:
+    """The current turn's platform tools handle, bound via invocation_state."""
+    handle = tool_context.invocation_state.get(_BAND_TOOLS_STATE_KEY)
+    if handle is None:
+        raise RuntimeError(
+            "Band tools handle missing from invocation_state; platform tools "
+            "must be invoked through StrandsAdapter.on_message"
+        )
+    return handle
+
+
+def _result_text(result: ToolResult) -> str:
+    """Flatten a ToolResult's content blocks to text for error detection/reporting."""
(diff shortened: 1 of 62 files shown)


src/band/adapters/strands.py
```diff
@@ -0,0 +1,542 @@
...
+        )
+        await self._run_turn(
+            message=user_message,
+            room_id=room_id,
+            history=room_history,
+            tools=tools,
+            hooks=hooks,
+        )
+        if not hooks.terminal_fired:
+            await self._report_error(tools, missing_reply_error("Strands"))
```
amit-gazal-band, line 507: `missing_reply_error()` shares the wording, but the *policy* around it is duplicated: this same "turn did no terminal work -> post an ERROR event" tail already exists in `src/band/adapters/crewai.py:443` and `src/band/adapters/pydantic_ai.py:888`. I'd make it a shared helper (something like `report_missing_reply(tools, framework, detail=...)` next to `missing_reply_error` in `runtime/tools.py`) so the three adapters can't drift, and so a change to how we treat a reply-less turn is a one-line change instead of three.

src/band/adapters/strands.py
```diff
@@ -0,0 +1,542 @@
...
+            return TurnUsage()
+        return TurnUsage.from_mapping(
+            usage,
+            input="inputTokens",
+            output="outputTokens",
+            cache_read="cacheReadInputTokens",
+            cache_write="cacheWriteInputTokens",
+        )
+
+    async def _report_error(self, tools: AgentToolsProtocol, error: str) -> None:
```
amit-gazal-band, line 529: Same point for `_report_error` — this is the third near-identical copy (`crewai.py:498` catches bare `Exception`, `pydantic_ai.py:964` catches `ApiError`/`httpx.HTTPError` like this one). It has no adapter-specific state, so it could live once as a module-level helper rather than as a private method on each adapter.

amit-gazal-band (commented):
One suggestion on the missing-reply reporting: the message string is shared, but the surrounding policy is now copy-pasted a third time. Worth pulling into a shared helper.

amit-gazal-band (approved):
Well done.

PR #503, opened 2026-07-30
feat: env-driven LogSettings + OTEL-ready logging

AlexanderZ-Band:
## Summary
- Add validated `LogSettings` (`BAND_LOG_*`) as the opt-in entry point for process-wide Band logging, with file rotation, style/stream knobs, and JSON overrides
- Expand `configure_logging` / `build_logging_config` with typed styles and richer sinks; migrate examples, CLIs, Docker runners, and desktop/ACP paths onto the shared API
- Document the env surface in the README and cover settings + logging behavior with unit tests
- **OpenTelemetry (in scope):** host owns providers/exporters; Band keeps root handlers for OTEL `LoggingHandler`, emits JSON trace-correlation fields, passes through pydantic-ai `instrument=`, adds `examples/opentelemetry/` demo + README matrix (no Band TracerProvider / no sbx collector)
- **Also in scope (stacked commit `02c3d8fb`, desktop-app work from this same stack, not scope creep):** degraded (polled REST) start when no role is available, mention-detection read before message truncation, and the transcript attention default flip (`room_first` -> `user_first`)

Closes INT-1153

## Test plan
- [ ] `uv run pytest tests/test_log_settings.py tests/test_logging_config.py tests/test_band_import.py tests/integrations/desktop_app/test_logs.py tests/docker/launcher/test_exec.py -v`
- [ ] Spot-check an example with `BAND_LOG_LEVEL=DEBUG` and optional `BAND_LOG_FILE`
- [ ] OTEL demo / coexistence tests once landed (`examples/opentelemetry/`, keep-handlers + pydantic `instrument=`)
- [ ] Confirm stack: this PR targets the INT-1116 desktop branch, not `main`


src/band/logging_config.py
```diff
@@ -7,121 +7,407 @@
...
+    @model_validator(mode="after")
+    def _fmt_only_with_standard(self) -> LoggingRequest:
+        if self.fmt is not None and self.style != LoggingStyle.STANDARD:
+            raise ValueError("fmt is only supported with style='standard'")
+        if (
+            self.fmt is not None
+            and self.log_file is not None
+            and (self.file_style or LoggingStyle.STANDARD) != LoggingStyle.STANDARD
+        ):
+            raise ValueError("fmt is only supported with file_style='standard'")
```
amit-gazal-band: **H1 (high) - the `fmt=` trap this PR claims to have closed is still open, one axis over.**

Commit `5347444` fixed four provisioning scripts that died under `BAND_LOG_CONSOLE_STYLE=json` by pinning the console style next to `fmt=`. This validator has a **second arm** for the *file* style, and none of those four scripts pin `log_file_style`:

- `examples/coding_agents/create_agents.py:23`, `examples/coding_agents/test_communication.py:26`
- `examples/claude_sdk_docker/create_agents.py:24`, `examples/claude_sdk_docker/test_communication.py:26`

Reproduced: with `BAND_LOG_FILE=/tmp/a.log BAND_LOG_FILE_STYLE=json`, each of them raises `ValueError: fmt is only supported with file_style='standard'` at import and the script dies. Same bug, same four files.

**Please don't pin a fifth kwarg at the call sites.** The asymmetry is what makes this a bug class rather than an oversight: `fmt` comes from the code author, the *style* comes from the environment, so rejecting the combination fires on something the author cannot control. `build_logging_config`'s own docstring already promises the right behavior (line 326-327: "Ignored for `rich` and `json` styles") - make the code honor it and delete `_fmt_only_with_standard`. The rich and json builders already ignore `fmt`, so this is a deletion.

I prototyped exactly that and verified it: no crash; console formatter keeps `{'format': '%(message)s', ...}` while the file formatter is `pythonjsonlogger.json.JsonFormatter`; **full suite byte-identical to baseline (4417 passed)**; and the docker launcher's `fmt=` + `FormatStyle.BRACE` ISO-format test still passes, which was the one live caller worth worrying about. No existing test pins this validator, so nothing needs rewriting - but please **add** the missing test (json file sink + `fmt=` configures cleanly), or the trap grows back.

src/band/logging_config.py
```diff
@@ -142,14 +429,87 @@ def configure_logging(
...
+    return {
+        "version": 1,
+        "disable_existing_loggers": False,
+        "formatters": formatters,
+        "handlers": handlers,
+        "root": {
+            "level": request.root_level,
+            "handlers": root_handlers,
+        },
+        "loggers": _build_logger_configs(
```
amit-gazal-band, line 593: **M1 (medium) - a host handler on a *named* logger is silently destroyed.**

`dictConfig` is non-incremental, and `common_logger_config` removes handlers from every logger it configures - not just root. So the natural embedding pattern (attach my shipper to `band`, leave root alone) breaks silently. Reproduced: the handler is gone from `logging.getLogger('band').handlers` and receives zero records.

`tests/test_log_settings.py:132` pins only the **root** case, and the README documents only root. The narrower, more correct integration is the one that fails without a word.

Agreed fix, after some back-and-forth between the two reviewers - the second half matters:

- **Keep `build_logging_config` complete.** Stripping the `loggers` section here was proposed and **rejected**: this is a documented self-apply API (line 310-311, "Applications can inspect, modify, then apply the returned dict themselves"), and without the section a direct consumer gets `band` at NOTSET inheriting root WARNING - every Band INFO log silently gone. Verified both ways.
- **Have `configure_logging` apply a copy with `loggers` removed**, then `setLevel()` for `band` and each `extra_loggers` entry. Verified this holds every property at once: band INFO still on console, host handler kept *and receiving*, host propagation preserved, both levels correct.

Two consequences to write down rather than discover later: (a) a direct builder consumer applying the complete dict still loses host handlers - irreducible, since handler clearing is unconditional and the builder must name `band` to express its level, so the docstring should say "use `configure_logging` if host handlers matter"; (b) `configure_logging`'s docstring currently promises it "Returns the applied `dictConfig` dictionary so callers can inspect the exact configuration" - after this it returns the *declarative* configuration with named-logger levels applied afterwards. Reword it instead of leaving the claim standing.

src/band/logging_config.py
```diff
@@ -163,51 +523,159 @@ def _build_logger_configs(
...
     for logger_name, logger_level in extra_loggers.items():
-        if not logger_name:
-            raise ValueError("extra_loggers keys must be non-empty logger names")
         loggers[logger_name] = {
-            "level": _normalize_level(
-                logger_level,
-                name=f"extra_loggers[{logger_name!r}]",
-            ),
+            "level": logger_level,
             "propagate": True,
```
amit-gazal-band: **M2 (medium) - `configure()` forces `propagate=True` back on, reversing a decision it never made.**

Every logger named in `extra_loggers` / `BAND_LOG_OVERRIDES` gets `propagate: True` hardcoded. Reproduced: a host sets `logging.getLogger('httpx').propagate = False` to isolate it, then `LogSettings().configure(extra_loggers={'httpx': 'WARNING'})` sets it back to `True` and httpx records start reaching the host's handlers. Band should set a level without undoing the host's isolation.

**Fix: just drop the key** - and note this is fixable *here in the builder*, independently of M1's divergence, because CPython's `DictConfigurator.configure_logger` assigns propagation only when the key is present:

```python
propagate = config.get('propagate', None)
if propagate is not None:
    logger.propagate = propagate
```

Verified: popping `propagate` out of the config leaves a host's `propagate=False` intact, while popping `handlers` does *not* save the handlers (hence M1 needing more). So the dict still expresses exactly the same levels, no contract changes, and every consumer benefits - including direct `dictConfig` users.

**One cost that belongs in the CHANGELOG, not in a bug report six months from now:** this applies to the `band` logger too (line 523). Today Band forces `propagate=True`, so a host that deliberately set `band.propagate = False` gets Band's logs anyway. Omit the key and that host is honored - meaning Band's own logs go where the host put them, possibly nowhere. I think that's correct (it's their logger, and overriding it is the whole complaint here), but it's a real behavior change and it should be a deliberate line rather than a user wondering why their Band logs went quiet.

src/band/logging_config.py
```diff
@@ -142,14 +429,87 @@ def configure_logging(
...
         json_fields=json_fields,
         static_fields=static_fields,
+        log_file=log_file,
+        max_bytes=max_bytes,
+        backup_count=backup_count,
+        file_style=file_style,
+        file_level=file_level,
     )
+    if log_file is not None:
+        Path(log_file).expanduser().resolve().parent.mkdir(parents=True, exist_ok=True)
```
amit-gazal-band: **M3 (medium, security) - `BAND_LOG_FILE` persists conversation content world-readably.**

Reproduced: `LogSettings(log_file=.../nested/dir/agent.log).configure()` creates the directory `drwxr-xr-x` and the file `-rw-r--r--` (umask-derived). And Band logs message content at DEBUG - `integrations/acp/server.py:492` (prompt text), `converters/claude_sdk.py:44` (message content), `adapters/agno.py:764` (full payload). So after this PR, `BAND_LOG_LEVEL=DEBUG BAND_LOG_FILE=/tmp/band/agent.log` is a **two-env-var** way to persist room content at 0644 inside a 0755 directory. That capability is new and README-documented; before it took writing code.

The defense already exists one layer up - `integrations/desktop_app/logs.py:39` deliberately does `STATE_DIR.mkdir(mode=0o700, ...)` - which is evidence the concern is understood. It just lives in the one caller who remembered, instead of in the function that does the dangerous thing. Please push it down to this chokepoint: parent `0o700`, log file `0o600`, plus a README line saying what a DEBUG log file can contain. Same reasoning as any redaction guard - make it structural so a future caller can't get it wrong.

See also N2 on `examples/20-questions-arena/setup_logging.py:39`, which now writes every HTTP request URL into exactly such a file.

src/band/integrations/desktop_app/event_relay.py
```diff
@@ -211,11 +211,26 @@ def __init__(
...
+        would turn a degraded room view into no room view at all, which is the
+        one outcome the user cannot repair without restarting Desktop. The
+        supervisor keeps trying for a role in the background.
+        """
         self._directory.mkdir(mode=0o700, parents=True, exist_ok=True)
         self._task = asyncio.create_task(self._supervise())
         try:
             async with asyncio.timeout(RELAY_TUNING.band_relay_start_timeout_s):
                 await self._ready.wait()
+        except TimeoutError:
```
amit-gazal-band, line 227: **M4 (medium, scope) - is it intentional that this PR carries non-logging desktop changes?**

Commit `02c3d8fb` is in this PR and changes three behaviors that have nothing to do with logging or OpenTelemetry:

1. **Here:** `start()` now swallows a role-acquisition `TimeoutError` and serves degraded (polled REST) instead of failing the launch.
2. `integrations/desktop_app/service.py:365-372` - message addressing is now read *before* truncation, so a mention past `BAND_MAX_MESSAGE_CHARS` still counts.
3. `integrations/desktop_app/tools.py` + `assets/room-view.js` - the attention default flips `room_first` -> `user_first` in the transcript model and the view.

No objection to any of them on the merits: each is individually defensible, each has a test, and the reasoning in the commit message is sound. This is purely a scope question - they read as INT-1116 desktop work rather than INT-1153 logging work, and combined with the CI gap in the review summary (nothing has run against this PR) they'd land without machine verification and without the desktop owner necessarily knowing they're here.

So: **intentional?** If yes, a line in the PR description naming them is enough from my side - no need to split a merge-committed stack.

examples/run_agent.py
```diff
@@ -904,8 +900,10 @@ async def main() -> None:
     parser.add_argument(
         "--log-level",
         "-l",
-        default=os.getenv("LOG_LEVEL", "INFO"),
-        help="Logging level (default: INFO or LOG_LEVEL env var)",
+        default=None,
```
amit-gazal-band, line 903: **N1 (medium) - `LOG_LEVEL` is silently dropped, and the migration is half-done.**

This was `default=os.getenv("LOG_LEVEL", "INFO")`; it's now `None`, so `LOG_LEVEL` no longer does anything. Same removal in `examples/copilot_sdk/setup_logging.py`, which had *explicit* handling for it including the empty and numeric (`LOG_LEVEL=20`) forms, with a comment explaining why they were tolerated.

That would be a clean rename to `BAND_LOG_LEVEL` - except the repo still ships the old name: `band-bridge/bridge_core/bridge.py:756` reads `LOG_LEVEL`, and `band-bridge/.env.example:30` still contains `LOG_LEVEL=INFO`. So after this PR there are two logging vocabularies, and anyone who already has `LOG_LEVEL=DEBUG` exported (or is using that `.env.example`) loses example verbosity with no warning and no CHANGELOG note.

Either option is fine, half-migrated isn't: keep `LOG_LEVEL` as a documented fallback where it existed, **or** migrate band-bridge and its `.env.example` too and record the rename. The numeric/empty compatibility loss deserves an explicit yes/no as well - it was handled on purpose.

examples/20-questions-arena/setup_logging.py
```diff
@@ -2,53 +2,41 @@
...
+    settings = LogSettings.create(
+        log_level=level,
+        log_file=_LOG_DIR / filename,
+        log_file_level="DEBUG",
+        log_max_bytes=5 * 1024 * 1024,
+        log_backups=3,
+    )
+    settings.configure(
+        extra_loggers={
+            **chatty_logger_levels("DEBUG"),
```
amit-gazal-band: **N2 (medium) - this call site gained two loggers it didn't ask for, on the console *and* in the file.**

The old code raised exactly four loggers to DEBUG: `phoenix_channels_python_client`, `langchain`, `langchain_openai`, `langchain_anthropic`. `chatty_logger_levels("DEBUG")` is `httpx` + `httpcore` + `phoenix_channels_python_client`, so **`httpx` and `httpcore` are now at DEBUG for the first time**.

Both sinks are affected, which is easy to miss. Reproduced in clean processes, both shapes:

- **New:** console handler is pinned to level 20 and the file to 10, so httpx *DEBUG* is dropped at the console - but httpx **INFO** passes it. An `httpx` INFO probe (`HTTP Request: POST https://api.example/v1?key=...`) appears **on stderr and in the file**.
- **Old:** `httpx` had no entry, so its effective level was root `WARNING` and the same record appeared in **neither**.

The logger gate is what creates the record; pinning the handler level only filters what's already been created. So this is new console noise *and* new sensitive file content - every HTTP request URL into a 5 MB x 3 rotating file which, per M3, is created 0644 in a 0755 directory.

Spell out the four loggers this call site actually wanted. `chatty_logger_levels()` is a convenience whose membership isn't the membership intended here.

src/band/docker/launcher/run.py
```diff
@@ -146,31 +148,26 @@ def execute(launch: ResolvedLaunch) -> None:
...
-        "{asctime} {name} {levelname} {message}",
+def configure_logging(*, log_file: Path | None = None) -> None:
+    """Apply launcher logging: console now, optional file after path resolution.
+
+    The console stream is left to ``BAND_LOG_STREAM``. Nothing reads this
+    process's stdout — it logs and then execs the customer entrypoint — so
+    unlike band-acp and band-trigger, whose stdout carries JSON-RPC frames and
+    command output, there is no protocol here to protect from a log line.
+    """
+    settings = LogSettings(
```
amit-gazal-band: **L1 (low) - this silently clobbers `BAND_LOG_FILE`, and `LogSettings.create()` exists to prevent exactly that.**

On the first call `log_file` is `None`, and an explicit `None` init kwarg beats the environment in pydantic-settings. Reproduced: with `BAND_LOG_FILE` set, `LogSettings(log_file=None).log_file` is `None` while `LogSettings().log_file` is the env value. So `BAND_LOG_FILE` never works in the launcher - while the new docstring right above advertises env-driven behavior ("The console stream is left to `BAND_LOG_STREAM`").

`LogSettings.create()`, added in commit `81cacd7` for precisely this ("`None` values are omitted so environment variables and defaults still apply"), isn't used here. Switching to it fixes it.

Separately: both styles are hard-pinned to `STANDARD`, so `BAND_LOG_CONSOLE_STYLE` is silently ignored in this process too. That one is forced by the `fmt=` validator - and would resolve itself once H1 lands.

tests/test_log_settings.py
```diff
@@ -0,0 +1,176 @@
...
+        log_stream: LogStream = LogStream.STDOUT
+
+    with band_log_env(monkeypatch, STREAM="STDERR"):
+        assert StdoutLogSettings().log_stream == LogStream.STDERR
+
+    with band_log_env(monkeypatch, STREAM=None):
+        assert StdoutLogSettings().log_stream == LogStream.STDOUT
+
+
+def test_host_telemetry_handler_must_be_attached_after_band_configures(
```
amit-gazal-band, line 136: **L2 (low) - the invariant that makes M1 and M2 tolerable isn't pinned anywhere.**

This test is the right shape and the right instinct - it's the reason M1/M2 are "medium, fix it" rather than "blocker." What's missing is the guard one level up: **nothing pins that no library code path configures logging.** I grepped `src/` and it's true today (only `band-acp`, `band-trigger`, the desktop server, and the docker launcher call it), which is exactly why it's worth a test - given how destructive this `dictConfig` is, a future `configure_logging_from_env()` dropped into `Agent.create` would hijack every host's logging with no failing test anywhere.

Cheap version: `import band`, build an `Agent`, assert root's handler list is identical before and after.

While here - please also extend this test to the **named-logger** case (M1). It currently pins only root, and the `band`-logger variant is the one that silently fails.

src/band/logging_config.py
```diff
@@ -7,121 +7,407 @@
...
+        max_bytes: Max file size before rotation. ``0`` uses a plain
+            :class:`~logging.FileHandler`; a positive value uses
+            :class:`~logging.handlers.RotatingFileHandler`.
+        backup_count: Rotated backup files to keep when ``max_bytes > 0``.
+            Must be at least 1 there: a rotating handler with no backups never
+            rotates. Ignored when ``max_bytes`` is ``0``.
+        file_style: File formatter style: ``"standard"`` or ``"json"``.
+            Defaults to ``"standard"`` when ``log_file`` is set.
+        file_level: Level for the file handler. Defaults to ``level``. When
+            more verbose than ``level``, the ``band`` logger is lowered so
```
amit-gazal-band: **L3 (low, docs) - "the `band` logger is lowered" is the whole story, and the docstring doesn't say what that excludes.**

`effective_band_level` lowers only the `band` logger, so with `BAND_LOG_FILE_LEVEL=DEBUG` the file captures Band's DEBUG records and nothing else's - every other logger is still gated at `root_level`. That's a defensible design (it's Band's file, at Band's level), but "File handler level. Defaults to `level`" reads like a *sink* level, i.e. "everything at DEBUG goes to the file," which is what an operator setting it for a post-mortem would expect.

One clause is enough: a more verbose `file_level` captures extra detail from `band` (and anything named in `extra_loggers`), not from every logger in the process.

examples/slack/setup_logging.py
```diff
@@ -2,14 +2,10 @@
...
 
-
-def setup_logging(level: int = logging.INFO) -> None:
-    """Configure logging to show band + slack-sdk logs, mute noisy deps."""
-    configure_logging(
-        level,
-        extra_loggers={"slack_sdk": level},
-    )
+def setup_logging(level: LogLevel | None = None) -> None:
+    """Configure logging to show only band logs, hiding noisy dependencies."""
```
amit-gazal-band: **N3 (nit) - this docstring now says the opposite of what the function does.**

It was "Configure logging to show band + slack-sdk logs, mute noisy deps," which was accurate: the function raises `slack_sdk`. It's now the generic "show only band logs, hiding noisy dependencies" - while the body still passes `extra_loggers={"slack_sdk": settings.log_level}`.

Worth a moment because of where it landed: commit `44d7941` exists specifically to fix `examples/slack/dev_bridge.py` losing `slack_sdk` diagnostics, and this is the file that documents why they're kept. Same generic string also replaced "Configure concise logging for the mixed example suite" in `examples/mixed/setup_logging.py`. Cosmetic, but it's how the next reader gets it wrong.

amit-gazal-band (changes requested):
Two independent reviews (an architect pass and a Codex reviewer pass) reconciled into one list. Everything below was reproduced against this SHA, not read off the commit messages.

**Verdict: not merge-ready.** One high finding blocks in-code (H1); a second, H2, has no line to attach to: **CI has never run on this PR.** `.github/workflows/ci.yml` is `on: pull_request: branches: [main]` and this targets the INT-1116 branch, so the only checks that ran were *Validate PR Title* and dependabot. 84 files, +3298/-716, zero machine verification. That's an owner call, raised separately.

The gate I ran myself at this SHA: `ruff check` clean, `ruff format --check` 816 files clean, `pyrefly check` 0 errors, markdown-docs 59 passed, unit suite 4417 passed. (1 failure + 9 errors are a **pre-existing** `HOME` leak from `tests/docker/launcher/test_exec.py::test_main_exits_nonzero_on_launch_error`, unrelated to this PR and getting its own issue.)

Credit where due, so nobody re-litigates it: the core design holds up. `LogSettings` is a proper pydantic-settings class, validation lives in one `LoggingRequest`, no library code path configures logging (only the four entrypoints do), stdout is genuinely protected on `band-acp`/`band-trigger`/the desktop stdio server with a real behavior test, and the four "traps" the commits claim to close are actually closed *and* tested. The OTEL half and `PydanticAIAdapter.instrument` were checked and cleared: real in-memory spans, tri-state inherit/opt-out, handler ordering, `ProcessHistory` preserved.

Inline comments follow, ranked: H1 (high), M1-M4 (medium), N1-N2 (medium), L1-L3 (low), N3 (nit).

src/band/integrations/desktop_app/event_relay.py
AlexanderZ-Band: Intentional. This PR is deliberately stacked on the INT-1116 desktop branch, not merged straight to `main` — the desktop-app changes (event relay degraded-start, mention-before-truncation, attention default) are part of that stack, not scope creep into the logging work. Noting it here per your ask; will add a line to the PR description naming them as well.

src/band/logging_config.py
```diff
@@ -142,72 +434,261 @@ def configure_logging(
...
+        Path(log_file).expanduser().resolve() if log_file is not None else None
     )
-    logging.config.dictConfig(config)
+    if resolved_log_file is not None:
+        # Band logs message content at DEBUG in several places (prompt text,
+        # tool payloads), so a log file can hold room content — lock the
+        # directory and file down rather than inherit the umask default
+        # (typically 0o755/0o644, world-readable).
+        resolved_log_file.parent.mkdir(mode=0o700, parents=True, exist_ok=True)
+        resolved_log_file.parent.chmod(0o700)
```
amit-gazal-band: **M3a (blocker) — this `chmod`s a directory Band did not create, and an ordinary log path now crashes at startup.**

`mkdir(..., exist_ok=True)` is a no-op when the directory already exists, and then `chmod(0o700)` is applied to it unconditionally. So the target of the `chmod` is whatever directory the operator happened to name — not something Band owns.

Reproduced, three ways:

1. `BAND_LOG_FILE=/tmp/band.log` → `PermissionError: [Errno 1] Operation not permitted: '/private/tmp'`, raised out of `configure_logging`, so **the process fails to start**. This is not an exotic path: it is the example I used in my own round-1 comment, and it is the obvious thing an operator types.
2. Run as **root** — which is what every `docker/claude_sdk/runner.py`, `docker/codex/runner.py`, `docker/letta/runner.py`, and the launcher do — and the `chmod` *succeeds* instead: `/tmp` goes from `1777` to `0700`, breaking every other process on the box. Silent, and worse than the leak it was meant to close.
3. `BAND_LOG_FILE=$HOME/agent.log` → verified a `0755` directory silently becomes `0700`. On Linux where `$HOME` is commonly `0755`, that breaks anything reading through it.

**Only harden what you create.** Something like: walk up to the first missing ancestor, create the missing segments with `0o700`, and leave any pre-existing directory's mode alone. If the operator points `BAND_LOG_FILE` at a directory they already manage, its permissions are their call — Band's job is to not *widen* them, not to narrow them.

Worth noting `integrations/desktop_app/logs.py` dropped its own `STATE_DIR.mkdir(mode=0o700, ...)` in favour of this. That one is genuinely Band's own directory, so it was correct there — if this moves to create-only hardening, check that STATE_DIR still ends up `0700` on a first run (it should, since Band creates it) and keep a test on it.

src/band/logging_config.py
```diff
@@ -142,72 +434,261 @@ def configure_logging(
...
+    logging.config.dictConfig({**config, "loggers": {}})
+    if resolved_log_file is not None:
+        resolved_log_file.chmod(0o600)
```
amit-gazal-band: **M3b (high) — the file mode does not survive rotation, which this PR is the feature that adds.**

This `chmod` runs once, at configure time, on the file the handler has just opened. `RotatingFileHandler` then creates a new file on every rollover through its own `_open()`, under the process umask. So the lockdown lasts until the log first fills up.

Reproduced with `max_bytes=200, backup_count=3` and 80 records:

```
rot.log:   -rw-r--r--     <- the LIVE file, back to 0644 after rollover
rot.log.1: -rw-r--r--
rot.log.2: -rw-r--r--
rot.log.3: -rw-r--r--
```

Every file including the current one is world-readable again. And the rotating path is exactly the one an operator uses for a long-lived agent — i.e. the case with the most room content in it.

The mode has to be set where the file is *created*, not once afterwards: a small `FileHandler`/`RotatingFileHandler` subclass overriding `_open()` to open with `0o600` (`os.open(..., os.O_APPEND | os.O_CREAT | os.O_WRONLY, 0o600)`) covers the initial file and every rollover with one implementation. Please don't reach for a process-global `os.umask()` — this is a library, and that would change file creation for the whole host application.

See also the test comment on `tests/test_logging_config.py:369`, which passes only because it never rotates.

tests/test_logging_config.py
```diff
@@ -200,6 +222,160 @@ def fake_find_spec(name: str):
...
+    The umask default (typically 0o755/0o644) would otherwise leave both the
+    directory and the file readable by anyone on the box.
+    """
+    log_file = tmp_path / "nested" / "dir" / "agent.log"
+
+    with restored_logging():
+        configure_logging(log_file=log_file)
+
+    assert (log_file.parent.stat().st_mode & 0o777) == 0o700
+    assert (log_file.stat().st_mode & 0o777) == 0o600
```
amit-gazal-band: **M3 test (medium) — this pins the fix only in the one case where the fix works.**

The assertion is right, and against a non-rotating sink it is genuinely green. But no rotation happens here, so it cannot see M3b: with `max_bytes` set, the live file is back to `0644` after the first rollover and so is every backup.

The project already has `test_rotation_rolls_over_and_keeps_the_requested_backups` doing the rollover work. Please assert the mode there too — rotate, then check `0o600` on the live file **and** on each surviving backup. That is the version that would have failed against this commit, which is the only version worth having.

tests/test_log_settings.py
```diff
@@ -0,0 +1,235 @@
...
+    Only band-acp, band-trigger, the desktop server, and the docker launcher
+    call configure_logging today. A future call from inside Agent.create (or
+    similar) would silently hijack every host's logging with no failing test
+    anywhere -- this pins that nothing in the plain Agent construction path
+    does so.
+    """
+    root = logging.getLogger()
+    handlers_before = list(root.handlers)
+
+    Agent(runtime=MagicMock(), adapter=MagicMock())
```
amit-gazal-band: **L2 (medium) — this guard passes with the exact regression it exists to catch.**

The docstring names the threat precisely: "a future call from inside `Agent.create` (or similar) would silently hijack every host's logging." But the test exercises `Agent.__init__`, and `Agent.create` (`src/band/agent.py:108`) is a separate classmethod that does the real work.

I planted the regression to be sure rather than assume: wrapped `Agent.create` so it calls `LogSettings().configure()` first, then ran this test's body. Root's handler list is unchanged and **the test passes**. So it protects the one path a regression is least likely to arrive on.

Two cheap changes make it real:
- Drive `Agent.create(...)` (mocked adapter, fake credentials) rather than `Agent(...)` — that's the documented entry point and where a convenience call would get added.
- Add the import-time half, which nothing covers: in a subprocess, `import band` and assert root has no handlers. `src/band/__init__.py:49` now eagerly imports `band.config.logs`, so "importing the SDK doesn't configure logging" is a live invariant with a live way to break it.

Not a blocker — the production code is correct today. But an invariant guarded by a test that cannot fail is the same as an unguarded invariant, and I asked for this one, so I'd rather it actually hold.

README.md
```diff
@@ -211,23 +211,73 @@ Use [examples/run_agent.py](examples/run_agent.py) when you want one command tha
...
+| `BAND_LOG_FILE` | unset | Optional log file path |
+| `BAND_LOG_FILE_LEVEL` | follows `BAND_LOG_LEVEL` | File handler level (use `DEBUG` for quiet console / verbose file) |
+| `BAND_LOG_MAX_BYTES` | `0` | Rotate when positive; `0` uses a plain file handler |
+| `BAND_LOG_BACKUPS` | `1` | Rotated backups to keep; `0` alongside a size cap is rejected, since a handler with no backups never rotates |
+| `BAND_LOG_CONSOLE_STYLE` | `standard` | `standard`, `rich`, or `json` |
+| `BAND_LOG_FILE_STYLE` | `standard` | `standard` or `json` |
+| `BAND_LOG_STREAM` | `stderr` | `stderr` or `stdout` |
+| `BAND_LOG_OVERRIDES` | `{}` | JSON map of logger name → level |
+
+`BAND_LOG_FILE`'s directory and file are created `0700`/`0600` so the log stays
```
amit-gazal-band: **Docs (low) — this sentence is not true as written, in both halves.**

> `BAND_LOG_FILE`'s directory and file are created `0700`/`0600`

- The directory is not *created* `0700` when it already exists — it is `chmod`ed to `0700`, which is M3a and is a much stronger claim on the operator's filesystem than this sentence implies.
- The file does not *stay* `0600`: after the first rotation the live file and all backups are `0644` again (M3b).

The second paragraph — that Band logs message content at DEBUG, so a file sink can persist room content — is the genuinely valuable part and reads well. Once M3a/M3b are settled, please restate the mechanism to match whatever lands (e.g. "directories Band creates for the log file are `0700`, and the log file and its rotated backups are `0600`"), so the README is not the thing a reader trusts over the code.

amit-gazal-band (changes requested):
**Round 2 review of `ec21c62e`.** Re-verified every round-1 finding against the new code and re-ran the gate myself.

**Nine of eleven are properly fixed, several with better tests than I asked for.** H1 is the deletion we agreed on, with both arms covered (`test_fmt_is_ignored_by_non_standard_console_style`, `test_fmt_configures_cleanly_with_a_json_file_sink`). M1 and M2 are exactly the agreed design — builder stays complete, `configure_logging` applies `{**config, "loggers": {}}` then `setLevel`, `propagate` key dropped — and both have real behavior tests that assert delivery, not bookkeeping. L1 fixed with a test for the pre-resolution call. L3 and N3 docstrings now say what the code does (and N3 went further than asked, across six files). N2 spells out the four loggers the call site actually wanted, restoring `phoenix_channels_python_client` — exactly right. N1 answered by documentation: band-bridge is a separate deployable with its own env surface, so its `LOG_LEVEL` is out of scope. I accept that.

Gate at this SHA, run by me: `ruff check` clean · `ruff format --check` 816 clean · `pyrefly` 0 errors · markdown-docs 59 passed · unit suite **4424 passed** (+7 from the new tests), with the same 1 failure + 9 errors that are the pre-existing `HOME` leak in `tests/docker/launcher/test_exec.py`, unrelated to this PR.

**But M3 has to go back.** The fix does something I did not anticipate when I asked for it, and it is worse than the problem: it `chmod`s a directory Band did not create. `BAND_LOG_FILE=/tmp/band.log` now raises `PermissionError` and the process fails to start; as root — which every `docker/*/runner.py` is — it instead succeeds and sets `/tmp` to `0700` for the whole box. And the protection it does apply is undone by the first log rotation. Both reproduced, details inline. That is my miss as much as the implementation's: I asked for `0700`/`0600` at the chokepoint without saying "only on paths you create, and it has to survive rollover."

One more: the new L2 guard test passes with the exact regression it exists to catch. Inline.

So: **two items open (M3, L2), nine closed.** H2 (no CI on this PR) and M4 (the scope question) are still with @amit.gazal and unchanged by this push.

src/band/logging_config.py
amit-gazal-band: **Agreed remedy for M3a, after an independent second review — both reviewers reproduced this and both hold it at blocker.** Confirmed again separately: a pre-existing `0755` directory becomes `0700`, and via a symlinked path the `chmod` lands on the *resolved* target, so the blast radius includes anything reachable through a link.

Concrete shape we converged on, so this doesn't need another round:

1. Walk up from `log_file.parent` to the **first existing ancestor**.
2. `mkdir` each missing component **without `exist_ok`**, at `mode=0o700`.
3. **`chmod` nothing that already existed** — that is the whole fix.

Why without `exist_ok`: `mkdir` is atomic, so if another actor wins the race it raises `FileExistsError` and aborts rather than adopting a `0755` directory into Band's created set. Fail-closed. (Full ancestor-race containment would need `openat`/dirfd traversal and is out of scope here — the load-bearing rule is just never `chmod` a resolved pre-existing ancestor.) Note umask can only *remove* bits from `0o700`, never widen it, so no umask juggling is needed.

This also fixes a bug the current code has even in the create-everything case: `mkdir(mode=0o700, parents=True)` applies the mode to the **leaf only** — Python creates intermediate parents under the umask default — so `BAND_LOG_FILE=/var/log/band/agents/a.log` leaves `/var/log/band` at `0755` today. The explicit walk covers every segment it creates. Verified: `pre-existing=0o755 (untouched), created=['0o700','0o700','0o700']`.

src/band/logging_config.py
amit-gazal-band: **Agreed remedy for M3b, after an independent second review.** Both reviewers reproduced the same result — after rollover the live file *and* `.1`/`.2`/`.3` are all `0644`, so the current `chmod` protects nothing on a long-lived agent.

The fix is a secure `_open()` override shared by the plain and rotating handlers:

- `os.open(..., O_WRONLY | O_CREAT | O_APPEND, 0o600)`, then `os.fchmod(fd, 0o600)` on the descriptor — the creation mode is masked by umask, `fchmod` is not, so the pair gets you `0600` regardless of the host's umask.
- Use it for **both** `FileHandler` and `RotatingFileHandler`, so every new live file created after a rollover is secured too. The backups then take care of themselves: rollover *renames* the live file, so the mode travels with the inode and each `.N` inherits `0600`.
- Please decide the final-component symlink behaviour explicitly rather than leaving it implicit. `O_NOFOLLOW` where supported is the safer default — the failure mode of following a link and then `fchmod`ing is narrowing permissions on an unrelated or shared target, which is the same class of over-reach as M3a.

Deliberately not: a process-global `os.umask()`. This is a library, and that would change file creation for the entire host application.

tests/test_log_settings.py
```diff
@@ -0,0 +1,235 @@
...
+    """
+    shipper = RecordingHandler()
+
+    with restored_logging(), band_log_env(monkeypatch, LEVEL="INFO", FILE=None):
+        logging.getLogger("band").addHandler(shipper)
+        LogSettings().configure()
+        logging.getLogger("band.runtime").info("shipped record")
+        assert shipper in logging.getLogger("band").handlers
+
+    assert shipper.messages == ["shipped record"]
```
amit-gazal-band, line 175: **Test request (low) — the round-2 fix protects more than this test claims, and the extra protection is unpinned.**

This test is good and it pins the `band` logger. But `dictConfig`'s `_handle_existing_loggers` resets loggers that are *children* of any configured logger — `logger.handlers = []`, `setLevel(NOTSET)`, `propagate = True`. So naming `band` in the `loggers` section was wiping the whole `band.*` subtree, and dropping that section fixed the subtree too, not just `band` itself.

Both reviewers reproduced it independently. Before/after at the two shapes:

```
old (loggers={'band': ...}): handler kept=False, level=NOTSET, propagate=True
new (loggers={})          : handler kept=True,  level=ERROR,  propagate=False
```

So a host that attached a shipper to `band.runtime`, pinned `band.platform` to ERROR, or set `propagate=False` on either was silently reset before this fix and isn't now. That's a genuinely valuable property and nothing currently guards it — if someone later restores the `loggers` section for readability, it regresses invisibly and every test still passes.

Worth a sibling test: attach a handler to `band.runtime`, set `level=ERROR` and `propagate=False`, call `configure()`, and assert all three survive. It exercises a different `DictConfigurator` code path from the `band` case above, so it isn't redundant with it.

amit-gazal-band (commented):
Round-2 addendum, after both reviewers took the open items independently. **No new defects** — this is one extra test request plus the converged implementation detail for M3, which I've added as replies on the two M3 threads so it's next to the code rather than buried here. Nothing changes the round-2 verdict: M3 (blocker) and L2 open, the other nine closed.

Two notes on that verdict after the second pass:

- **M3 stays blocker, independently.** Both reviewers separately reproduced the pre-existing-directory `chmod` (0755 → 0700), the `PermissionError` on `/tmp/band.log`, and the post-rotation `0644` on the live file plus all backups. I explicitly invited the argument for downgrading it to *high* — the `PermissionError` is loud and easy to work around — and it didn't survive: the root-in-Docker case (`/tmp` silently `1777` → `0700`) is process-external, container-wide damage, which is what keeps it at blocker.
- **L2 downgraded medium → low.** Fair correction against me: the false guard is a future risk, not a present defect, since the production code is correct today. The fix asked for is unchanged.

Also confirmed and worth stating: **the M1/M2 implementation is sound beyond what I checked in round 2.** `{**config, "loggers": {}}` is equivalent to omitting the key; the post-apply `setLevel` ordering is correct; repeat calls behave as before (an `extra_loggers` level set by an earlier call is sticky — worth a docs line, not a finding); and the shallow copy leaves the returned declarative config intact for a caller who inspects it.

src/band/logging_config.py
AlexanderZ-Band: **Fixed in `109c10d9` — and thank you for the reproduction, all three cases were real.**

Confirmed before changing anything: a pre-existing `0755` directory became `0700`, and `/tmp` raised `PermissionError: '/private/tmp'`.

Implemented as agreed, with one deliberate difference. `create_log_directory()` walks up to the first existing ancestor and creates each missing segment individually at `0700`, and never touches anything that was already there. That also fixes the bug you spotted in the create-everything case — `mkdir(mode=…, parents=True)` hardens the leaf only. Measured now: pre-existing `0755` untouched, created segments `['0o700','0o700','0o700']`, `/tmp` stays `1777` with no exception.

**The difference: a lost race is not fatal.** You asked for `mkdir` without `exist_ok` so it fails closed; I catch `FileExistsError` and continue. Two agents booting against a shared log volume is ordinary, and killing startup is worse than what it defends — an attacker who can pre-create a directory in your log path can pre-create the log file too. The load-bearing rule you named is untouched: that directory is simply one Band did not create, so nothing chmods it. Happy to switch to fail-closed if you'd rather.

Guarded by `test_band_creates_its_log_directories_and_file_owner_only` and `test_paths_band_did_not_create_keep_the_modes_they_have`. Your `STATE_DIR` note was right too — it lands `0700` on a first run via the same path, now pinned by `test_configure_creates_the_state_directory_owner_only`.

src/band/logging_config.py
AlexanderZ-Band: **Fixed in `109c10d9`.** Reproduced first, exactly as you had it: after rollover the live file *and* `.1`/`.2`/`.3` were all `0644`.

`OwnerOnlyCreate` is a mixin over both `FileHandler` and `RotatingFileHandler`, overriding `_open()` to create the file `0600` (`os.open(O_CREAT|O_EXCL|O_WRONLY, 0o600)` + `os.fchmod`, since the creation mode is umask-masked and `fchmod` is not). Every generation a rollover mints goes through it, and the backups inherit by rename, as you said. Measured now: all four files `0o600`. No `os.umask()` anywhere.

**Two deliberate differences, both from applying your own rule — 'Band's job is to not widen them, not to narrow them' — to files as well as directories:**

1. **A pre-existing log file keeps its mode.** `O_EXCL` means we only set the mode on a file we created. An operator running `logrotate` with `create 0640 band adm` shouldn't have to fight the SDK on every restart, and that file is as much theirs as the directory is. Your reproduction is still fully fixed, because Band creates every rotated generation.
2. **No `O_NOFOLLOW`.** You asked for an explicit decision, so: create-only semantics already close the case you named — if the path exists, symlink or not, we never `fchmod` it, so 'follow a link and narrow an unrelated target' cannot happen. `O_NOFOLLOW` would instead break the legitimate `/var/log/band.log → /mnt/logs/band.log` setup, which felt like the same over-reach in the other direction.

Say the word if you want the unconditional `fchmod` back.

tests/test_logging_config.py
AlexanderZ-Band: **Done in `109c10d9`** — moved into `test_rotation_rolls_over_and_keeps_the_requested_backups` as you asked, asserting `0o600` on the live file and each surviving backup. Verified it fails against the previous commit, which is the point.

The standalone permission test is gone; two replace it: `test_band_creates_its_log_directories_and_file_owner_only` (every created segment, not just the leaf) and `test_paths_band_did_not_create_keep_the_modes_they_have`.

tests/test_log_settings.py
AlexanderZ-Band: **Fair, and fixed in `109c10d9`.** You were right that it protected the one path a regression won't arrive on.

Both changes made. The guard now drives `Agent.create(adapter=…, agent_id=…, api_key=…)`, and a new sibling runs `import band` in a subprocess and asserts root's handler list is `[]`.

I planted both regressions to check the guards actually bite: a `LogSettings().configure()` at the top of `Agent.create` → `1 failed`; a `configure_logging_from_env()` at import in `band/__init__` → `1 failed`. Green on the real code.

README.md
AlexanderZ-Band: **Fixed in `109c10d9`.** You were right on both halves, and the replacement had to wait for M3a/M3b to settle. It now reads:

> Band hardens only what it creates for `BAND_LOG_FILE`: directories it has to create are `0700`, and the log file — plus every file a rotation replaces it with — is created `0600`. A directory or file that already exists keeps the mode you gave it.

The `DEBUG` paragraph you liked is kept, with a closing clause that follows from the new rule: if you point it at a path you manage, the permissions on it are yours to set.

tests/test_log_settings.py
AlexanderZ-Band: **Added in `109c10d9`** — `test_configure_leaves_the_whole_band_subtree_alone`. It attaches a handler to `band.runtime`, sets `level=ERROR` and `propagate=False`, calls `configure()`, and asserts all three survive.

Agreed it isn't redundant with the `band` case: `_handle_existing_loggers` is the code path that resets *children* of a configured logger, which is a different branch from the configured logger itself. Thanks for the before/after — it's the reason the docstring can name the mechanism instead of just the symptom.

AlexanderZ-Band:
## Round 2 addressed — `109c10d9`

Both open items are closed, each reproduced before I touched anything. The nine round-1 items you'd already confirmed are untouched.

**M3 (blocker).** Reproduced exactly as reported: a pre-existing `0755` directory became `0700`; `/tmp` raised `PermissionError`; after rollover the live file and all three backups were `0644`.

The root cause was one thing, twice: both halves hardened *after the fact*. A `chmod` after `mkdir(exist_ok=True)` cannot know whether Band created that directory, and a `chmod` after `dictConfig` cannot survive a handler that opens a new file on every rollover. So both moved to the moment of creation, under a single rule — **Band hardens what it creates and nothing else**:

- `create_log_directory()` walks to the first existing ancestor and creates each missing segment at `0700`, touching nothing pre-existing. This also fixes the bug you spotted in the create-everything case: `mkdir(mode=…, parents=True)` hardens the leaf only.
- `OwnerOnlyCreate` gives `FileHandler` and `RotatingFileHandler` an `_open()` that creates the file `0600`, so every generation a rollover mints is covered and the backups inherit by rename.

Measured after: pre-existing dir `0755` untouched · created segments `['0o700','0o700','0o700']` · `/tmp` still `1777`, no exception · all four rotated files `0o600`.

**Three deliberate departures**, each applying *your* principle — "Band's job is to not widen them, not to narrow them" — to files as well as directories. Details on the threads; happy to reverse any:

1. A pre-existing log **file** keeps its mode too (an operator with `logrotate create 0640 band adm` shouldn't fight the SDK every restart). Your reproduction is still fully fixed, since Band creates every rotated generation.
2. No `O_NOFOLLOW` — create-only semantics already make "follow a link, narrow an unrelated target" impossible, and it would break a symlinked log path.
3. A lost `mkdir` race is not fatal — two agents booting against a shared log volume is ordinary, and the invariant you actually asked for (never chmod what we didn't create) holds either way.

**L2 (low).** You were right that it protected the one path a regression won't arrive on. It now drives `Agent.create`, plus a subprocess sibling asserting `import band` leaves root's handlers `[]`. I planted both regressions to be sure they bite — `configure()` in `Agent.create` → **1 failed**; `configure_logging_from_env()` at import → **1 failed**.

**Also done:** the rotation test now asserts `0600` on the live file and each backup · the `band.*` subtree sibling test (handler + `ERROR` + `propagate=False` on `band.runtime`, all three survive) · the README sentence restated to match what the code does · `STATE_DIR` still lands `0700` on a first run, now pinned by `test_configure_creates_the_state_directory_owner_only`.

**Gate:** `ruff check` clean · `ruff format` clean · `pyrefly` 0 errors · markdown-docs 59 passed · unit suite **4438 passed**, 81 skipped.

The six round-2 threads are answered and resolved. The round-1 threads are still open in the UI — I left those for you, since you're the one who confirmed them fixed.

src/band/logging_config.py
AlexanderZ-Band: Fixed in 7235b0f1 with the deletion we agreed on: `fmt=` is gone rather than patched around, so the trap can't reopen one axis over. Both arms pinned by `test_fmt_is_ignored_by_non_standard_console_style` and `test_fmt_configures_cleanly_with_a_json_file_sink`. You confirmed this closed in your round-2 review.

src/band/logging_config.py
AlexanderZ-Band: Fixed in 7235b0f1: the builder still returns the complete declarative config, but `configure_logging` applies it with `{**config, "loggers": {}}` and then sets levels via `setLevel`, so `dictConfig` never touches an existing named logger and a host's handler survives. The test asserts delivery (the handler still receives records after `configure()`), not bookkeeping. Confirmed closed in round 2.

src/band/logging_config.py
AlexanderZ-Band: Fixed in 7235b0f1: the `propagate` key is dropped from the emitted logger sections, so `configure()` no longer reverses a `propagate=False` decision it never made. Pinned by `test_configure_does_not_reverse_a_hosts_propagate_false`. Confirmed closed in round 2.

src/band/logging_config.py
AlexanderZ-Band: Closed across two rounds. The first cut (7235b0f1, chmod at the chokepoint) had the two defects you caught in round 2 — it narrowed pre-existing directories and didn't survive rotation. The final fix is 109c10d9: hardening moved to creation time under one rule — Band hardens only what it creates. Missing directory segments are created 0700 one by one (nothing pre-existing is touched, so `/tmp` keeps its 1777), and the log file plus every generation a rollover mints is created 0600 via `O_CREAT|O_EXCL` in the handlers' `_open`. Full discussion on the round-2 M3a/M3b threads.

examples/run_agent.py
AlexanderZ-Band: Won't fix in this PR, as agreed: band-bridge is a separate deployable with its own env surface, so its `LOG_LEVEL` is out of scope here. The SDK examples now speak `BAND_LOG_LEVEL` only, and the README documents the boundary (7235b0f1). You accepted this in the round-2 review.

examples/20-questions-arena/setup_logging.py
AlexanderZ-Band: Fixed in 7235b0f1: the call site now spells out the four DEBUG loggers it actually wanted — including the `phoenix_channels_python_client` it had lost — instead of `chatty_logger_levels("DEBUG")`, which silently added httpx/httpcore. Confirmed closed in round 2.

src/band/docker/launcher/run.py
AlexanderZ-Band: Fixed in 7235b0f1: the launcher's `configure_logging()` goes through `LogSettings.create()`, so an explicit `log_file=None` no longer beats `BAND_LOG_FILE` from the environment. Covered by a test for the pre-resolution call in `tests/docker/launcher/test_exec.py`. Confirmed closed in round 2.

tests/test_log_settings.py
AlexanderZ-Band: Closed in two steps. The guard added in 7235b0f1 drove `Agent()` while its docstring named `Agent.create` — your round-2 L2 caught that it passed with the exact regression it exists to catch. 109c10d9 fixes it: the guard now drives `Agent.create`, and a sibling asserts in a subprocess that importing `band` leaves root's handlers empty. Both verified to fail against a planted regression before landing.

src/band/logging_config.py
AlexanderZ-Band: Fixed in 7235b0f1: the docstring now says what actually happens — a verbose `file_level` lowers the `band` logger only, not every logger in the process. Confirmed closed in round 2.

examples/slack/setup_logging.py
AlexanderZ-Band: Fixed in 7235b0f1 for the flagged file, and ec21c62e extended it to four more `setup_logging.py` siblings with the same docstring-vs-behavior mismatch (anthropic, crewai, parlant, claude_sdk). All six now say exactly which extra loggers they raise. You noted in round 2 this went further than asked.

amit-gazal-band (approved):

PR #510, opened 2026-08-02
feat: dual-publish band-python-kit to Docker Hub

AlexanderZ-Band:
## Summary
Dual-publish `band-python-kit` image + OCI kit to Docker Hub (`bandhq/…`) and GHCR in one build. Hub is the customer-facing ref; GHCR stays a mirror. Callers use `secrets: inherit` for `DOCKERHUB_*`.

## Test plan
- [x] CI green
- [x] Rehearsal `kit-publish-manual` (`0.0.0-rc1`, `move-floating=false`) — Hub + GHCR tags OK
- [x] Hub rehearsal tags deleted
- [ ] Delete leftover GHCR `0.0.0-rc1` (needs `delete:packages`)

Fixes INT-1152


.github/workflows/release.yml
```diff
@@ -109,6 +109,7 @@ jobs:
       cancel-in-progress: false
       queue: max
     uses: ./.github/workflows/kit-publish.yml
+    secrets: inherit
```
amit-gazal-band: `secrets: inherit` is a **no-op for the Hub login** — it can be dropped from all three callers.

All three `kit-publish.yml` jobs declare `environment: release`, and per GitHub's docs: *"If you include `environment` in the reusable workflow at the job level, the environment secret will be used, and not the secret passed from the caller workflow."* I confirmed `DOCKERHUB_USERNAME`/`DOCKERHUB_TOKEN` exist on the `release` environment — so your successful rehearsal already authenticated with the environment copies, proving the env-only path works.

Two consequences:

1. The repo-level copies are inert, and they've already drifted: env `DOCKERHUB_TOKEN` was written 06:47:41Z, the repo one 07:05:47Z (18 min later). Whoever next runs `gh secret set DOCKERHUB_TOKEN` will hit the shadowed copy and see no effect.
2. `inherit` hands this workflow all 13 repo secrets plus the org's `APP_PRIVATE_KEY` (App key with `contents: write` on `add-band`), to authenticate to one registry.

Suggest: drop `secrets: inherit` from `release.yml`, `kit-image-rebuild.yml`, `kit-publish-manual.yml`, and delete the two repo-level `DOCKERHUB_*` secrets so the environment is the single source.

docker/band_python_kit/RELEASING.md
```diff
@@ -152,33 +159,53 @@ path does not depend on it.
...
+
+## One-time Docker Hub setup
+
+1. Org `bandhq` on Hub; public repos `band-python-kit` and `band-python-kit-image`.
+2. Repository secrets on `band-ai/band-sdk-python` (also mirrored on the
+   `release` environment): `DOCKERHUB_USERNAME` and `DOCKERHUB_TOKEN`.
+   For an organization access token (OAT), username must be the org name
+   (`bandhq`), not a personal Hub user. Callers pass them with
+   `secrets: inherit` — reusable-workflow caller jobs cannot use
+   `environment:`, so environment-only secrets never reach Hub login.
```
amit-gazal-band: This rationale is inverted, and it's the part that outlives the PR. Caller jobs indeed can't use `environment:` — but the **called** jobs do (`kit-publish.yml` L87/209/309), so environment-only secrets *do* reach Hub login, and in fact take precedence over anything inherited.

As written this instructs future maintainers to keep the duplicated repo+environment copies that have already diverged. Suggest replacing with: secrets live on the `release` environment only; the called jobs declare `environment: release`, so no caller-side passing is needed.

.github/workflows/kit-publish.yml
```diff
@@ -197,55 +222,82 @@ jobs:
...
+          for ref in \
+            "${KIT_NAME}:${{ inputs.version }}" \
+            "${HUB_KIT_NAME}:${{ inputs.version }}"
+          do
+            state="$(python3 scripts/registry-ref-state.py "$ref")"
+            if [ "$state" = "exists" ]; then
+              echo "::error::${ref} already exists and version tags are immutable — re-run only failed jobs, or see RELEASING.md"
+              exit 1
+            fi
+          done
```
amit-gazal-band: Partial-publish recovery regressed. The guard now fails if the tag exists in *either* registry, but the pushes below are two sequential `oras push`es. If the Hub push fails (outage, token, rate limit), re-running this job trips here because the GHCR tag now exists — so "re-run only the FAILED jobs" (the comment at L164) is no longer a complete recovery story; the operator has to delete a just-published immutable tag. Same for `image`, where build-push pushes per-registry.

Either make the guard per-ref and skip an already-published ref (each push targets a distinct registry, so resuming just the missing one is idempotent), or document this recovery path in RELEASING.md.

Related: `kit-image-rebuild.yml`'s `decide` job still derives the next `-rN` from GHCR tags alone (L148) — so a Hub-side failure makes Hub skip that `-rN` permanently, undetected.

.github/workflows/kit-publish.yml
```diff
@@ -197,55 +222,82 @@ jobs:
...
+            --output staging/ghcr/spec.yaml
+          uv run --no-project --with pyyaml scripts/stamp-kit-spec.py \
+            --spec docker/band_python_kit/spec.yaml \
+            --image-ref "${HUB_IMAGE_NAME}:${{ inputs.version }}" \
+            --digest "$digest" \
+            --output staging/hub/spec.yaml
+          echo "── GHCR stamped sandbox.image ──"
+          grep 'image:' staging/ghcr/spec.yaml
+          echo "── Hub stamped sandbox.image ──"
+          grep 'image:' staging/hub/spec.yaml
```
amit-gazal-band, line 283: Nothing fails if these two invocations drift. A copy-paste stamping `${IMAGE_NAME}` into `staging/hub/spec.yaml` would ship a Hub kit pinning a GHCR image — defeating the "a Hub consumer never needs a GHCR pull" property this PR exists for — and the run stays green; the `grep` is only an echo. Cheap fix in the spirit of the guards already here:

```bash
grep -q "${HUB_IMAGE_NAME}@${digest}" staging/hub/spec.yaml
grep -q "${IMAGE_NAME}@${digest}" staging/ghcr/spec.yaml
```

Minor, same file: `registry-ref-state.py` classifies absence from `MANIFEST_UNKNOWN`/`NAME_UNKNOWN`, but Hub answers a **missing repo** with `401 UNAUTHORIZED` (verified). Missing *tag* on an existing repo correctly returns `MANIFEST_UNKNOWN`, so the happy path is fine — but a renamed/typo'd Hub repo dies with an opaque `registry probe failed`. Worth a line in the helper's docstring.

amit-gazal-band (dismissed):
Approving — the shape is right (one build, digest reused, guards extended rather than bypassed, docs updated consistently; I checked every remaining `ghcr.io` mention and none are stale).

I verified the registry state directly: both Hub repos exist and are public, Hub `0.0.0-rc1` is gone, and **GHCR `0.0.0-rc1` is already gone too** on both packages — the last open checkbox is satisfied.

Four comments inline. The `secrets: inherit` one is the only one I'd want resolved before merge; the rest can be follow-ups.

.github/workflows/release.yml
AlexanderZ-Band: Confirmed. The called image, kit, and floating jobs already declare environment: release, so environment secrets take precedence and secrets: inherit was redundant while exposing unrelated caller secrets. Removed secrets: inherit from all three callers, updated the runbook to make the release environment the single source, and deleted the two duplicate repository-level DOCKERHUB_* secrets. The environment copies remain.

docker/band_python_kit/RELEASING.md
AlexanderZ-Band: Confirmed. Updated the runbook: DOCKERHUB_USERNAME and DOCKERHUB_TOKEN live only on the release environment, and the called kit-publish.yml jobs declare that environment. It no longer tells maintainers to duplicate or inherit the credentials.

.github/workflows/kit-publish.yml
AlexanderZ-Band: Confirmed and fixed. Image publishing now builds both refs only when both are absent; after a partial push it copies the completed ref to the missing registry with imagetools create, then resolves the digest and re-runs attestation. Kit publishing checks each ref immediately before pushing and skips completed refs. The rebuild lane now lists both registries, repairs suffix divergence, and marks an incomplete mirror for rebuild instead of deriving the next suffix from GHCR alone.

.github/workflows/kit-publish.yml
AlexanderZ-Band: Confirmed and fixed. The stamping step now fails if either staged spec does not contain its expected registry ref plus the image digest. I also documented that authentication or authorization failures, including a missing or inaccessible repository (Hub 401), intentionally fail closed rather than being classified as absence.

amit-gazal-band (approved):

PR #512, opened 2026-08-02
fix: require shared Band identity in ACP Docker examples

AlexanderZ-Band:
## Summary
- Document and enforce that `copilot_acp_agent` and container `BAND_AGENT_KEY` must be the **same** Band agent in colocated/compose ACP Docker examples (mismatched keys make room tools 404 while text relay still works). The colocated client validates the `.env` **file** value — what `docker run --env-file` actually hands the container — so a stale shell export can't mask a mismatch.
- Extract shared participant helpers (`ensure_participant`, `ensure_in_room`/`ensure_not_in_room`, the `absent_from_room` context manager, `get_participant_role`) and cover the aligned/foreign list-participants boundary in live + mocked tests.
- Add a CI contract test so the example identity invariant cannot drift.

Fixes INT-1131

## Test plan
- [x] `uv run pytest tests/integrations/test_participant_helpers.py tests/integrations/acp/test_copilot_docker_identity_contract.py -v --no-cov`
- [x] `uv run pytest tests/integration/test_participant_permissions.py -k RemoteMcpIdentity -v -s --no-cov`
- [x] Full unit suite: 4380 passed, 85 skipped
- [ ] Spot-check colocated/compose README prerequisites read correctly for a first-time setup

Made with [Cursor](https://cursor.com)


examples/acp/copilot_docker/colocated/.env.example
```diff
@@ -1,8 +1,8 @@
 # Copilot-entitled GitHub token — the Copilot CLI reads this automatically.
 GITHUB_TOKEN=
 
-# Band AGENT API key that band-mcp uses to act on the platform (thnv_a_… / band_a_…).
-# The single container is one Band identity; all tool calls act as this agent.
+# Same api_key as copilot_acp_agent in agent_config.yaml (host client.py).
+# Must be a Band AGENT key (thnv_a_… / band_a_…), not a user key.
```
amit-gazal-band: Please remove `thnv_a_…` (I saw it in multiple places)
We should ignore Thenvoi at this stage

amit-gazal-band (dismissed):
One small comment.

examples/acp/copilot_docker/colocated/.env.example
AlexanderZ-Band: Dropped thnv_a\_ from all 3 spots, kept band_a\_ only.

amit-gazal-band (approved):

PR #514, opened 2026-08-02
fix(integrations): emit canonical ACP tool events

AlexanderZ-Band:
## Summary

- Persist ACP tool calls and results in the SDK canonical JSON event format.
- Preserve failed and denied tool outcomes in the canonical result payload.
- Add an ACP client to Band room to ACP editor regression test with an intent-oriented editor fixture.

## Root cause

The ACP room emitter stored human-readable tool titles and output as plain text. Downstream parsers require structured tool-event payloads, so they treated the events as ordinary text.

## Validation

- uv run pytest tests/integrations/acp -q (266 passed, 8 skipped)
- ruff check on the changed ACP implementation and test files


amit-gazal-band (approved):
LGTM

PR #516, opened 2026-08-03
fix(integrations): stop Codex posting literal reasoning/plan placeholders

AlexanderZ-Band:
## Summary

Codex thought events (`reasoning`/`plan` items) were posting the literal
strings `(reasoning)` / `(plan)` to the room whenever the underlying item
had an empty summary/text — a real, live symptom on the backends E2E lane,
not just a theoretical edge case.

## Changes

- **Fix (INT-1148):** `_extract_thought_text` now returns `None` when a
  reasoning/plan item carries no informative text, and the caller skips
  emission entirely instead of falling back to a placeholder string.
  `_stringify_tool_output` (the shared "protocol payload -> display text"
  helper) now skips a list with nothing extractable rather than dumping it
  as `"[]"`.
- **Test coverage:** unit tests for dict-shaped reasoning summaries, empty
  reasoning/plan skip, and a `dynamicToolCall` result-list edge case; a new
  backends-lane E2E smoke (`test_codex.py`) drives a live Codex turn with
  `Emit.THOUGHTS` on and asserts the room never sees the placeholder
  strings (and, after review, that at least one thought event actually
  landed — otherwise the placeholder assertion would pass vacuously
  without exercising the fix). The smoke's joke-driving prompt asks for a
  harmless pun (not "a joke that might be badly interpreted") to avoid a
  safety-refusal false failure, while still demanding multi-step reasoning
  to reliably populate a real reasoning summary.
- **Follow-up cleanup surfaced by review:** the `item/completed` `"type"`
  vocabulary (`commandExecution`, `reasoning`, `plan`, etc.) was re-typed as
  bare string literals in ~5 places in `codex.py`. Added `CodexItemType`
  (StrEnum, `integrations/codex/types.py`) as the single source of truth,
  reused everywhere the vocabulary appears (gate sets, dispatchers, approval
  labels) — plus a test asserting the enum is fully and disjointly
  partitioned across the adapter's tool-like/thought-like/skipped-message
  buckets, so a new item type can't silently fall through all three.
  `_extract_tool_item` also returned a bare positional
  `tuple[str, dict, str]` and inlined all 7 item types in one ~95-line
  function; split into `CodexToolItem` (NamedTuple) + one small helper per
  item type. Same treatment for the two Codex approval RPC method strings
  (`CodexApprovalMethod` StrEnum backing `CODEX_APPROVAL_METHODS`, replacing
  6 re-typed literal checks).
- **Regression fix in the cleanup itself (caught by review):** three tool
  extractors (`_extract_mcp_tool_call`, `_extract_web_search`,
  `_extract_collab_agent_tool_call`) were switched to reuse
  `_stringify_tool_output` instead of hand-rolled `json.dumps`/`"completed"`
  logic — but that helper's list handling *skips* a non-text list rather
  than dumping it, so a real result like an MCP image/resource content
  block silently collapsed to the generic `"completed"` status instead of
  reporting the actual data. Added an opt-in `raw_fallback` mode to
  `_stringify_tool_output`: thought extraction and `dynamicToolCall` keep
  the skip behavior (avoids reintroducing placeholder noise), the three
  tool-output extractors now dump the raw list as JSON when it has nothing
  extractable as text. Covered by a new regression test per extractor.
- **Test-file cleanup:** extracted `events_of_type(tools, message_type)` and
  replaced ~30 duplicated inline
  `[e for e in tools.events_sent if e["message_type"] == ...]`
  comprehensions across `test_codex_adapter.py` with it.
- **Docs:** strengthened `AGENTS.md`'s Pre-Commit Checklist with an explicit
  self-review step for single-source-of-truth and intent-oriented-code
  violations — previously only stated as prose elsewhere in the file, never
  re-triggered at the point code actually gets checked before commit.
- **Docs (`.cursor` cleanup):** the branch carried an untracked
  `.cursor/rules/use-env-test.mdc` — a second, competing Cursor-rule
  mechanism alongside this repo's existing `.cursorrules`/`CLAUDE.md`
  symlinks to `AGENTS.md`. Folded its content (prefer `.env.test` over
  ad-hoc/.env/invented credentials, never print secrets) into `AGENTS.md`'s
  Environment Variables section and deleted the file, so the guardrail is
  visible from Claude, Cursor, and anything else reading any of the three
  symlinked names, with one copy instead of two.

A full `codex.py` -> package split (mirroring
`adapters/opencode/{adapter,approvals,config}.py`) was considered and
explicitly deferred as a separate, much larger-blast-radius effort (3200+
lines, 28 dependent files).

## Testing

- [x] `uv run pytest tests/adapters/test_codex_adapter.py -v` (191 passed)
- [x] `uv run pytest tests/ --ignore=tests/integration/ --ignore=tests/e2e/` (4381 passed, 85 skipped)
- [x] `uv run pyrefly check`

## Related Issues

Fixes INT-1148


amit-gazal-band (approved):
Ran Tom and Jerry. Looks great.

PR #518, opened 2026-08-04
fix: OneShotInvoker history pagination + participant roster surfacing

AlexanderZ-Band:
## Summary
Two fixes found while validating the AgentCore integration against the coding-agent adapter conformance ladder (INT-1119):

- **History pagination**: `get_agent_chat_context`'s `page`/`page_size` params are deprecated on the real API and scheduled for removal (2026-10-01). `OneShotInvoker` used them for a single fetch and never read `has_more`/`next_cursor`, so any room's history past one page was silently dropped on every invocation. Now follows the cursor while `has_more`, bounded by a `history_page_cap` (mirrors the existing `drain_cap` pattern), surfacing `history_truncated` in the result if the cap is hit. Code review then caught two more edge cases in that same loop: a page-0 fetch failure reported `history_truncated=False` (the worst case looked identical to full success), and `has_more=True` with a `null` `next_cursor` re-requested `cursor=None` forever, duplicating page-0 items until the cap. Both fixed with regression tests confirmed red on the old code.
- **Participant roster surfacing**: the long-running path sends the roster to the model on a room's first message too (`ExecutionContext.participants_changed` treats "never sent before" as changed). `OneShotInvoker` has no cross-call state to diff against — every invocation is that same "first time" case — but hardcoded `participants_msg=None` regardless, so the model got zero roster signal unless it proactively called a tool. Now builds it from the current fetch every call.

**Custom tool dispatch coverage**: nothing exercised the seam between OneShotInvoker's real AgentTools (REST-backed) and an adapter configured with `additional_tools` — added a test driving a real `AnthropicAdapter` through `handle_event`, proving a custom-tool call resolves to the custom handler while the same turn's platform tool call still reaches the real REST client. Also adds `examples/agentcore/custom_tools_llm_server.py`, a runnable container variant wiring `additional_tools`, and fixes BUILDING.md's custom-tools snippet (it showed a constructor call that doesn't match the real `CustomToolDef` type, a `(InputModel, handler)` tuple).

**Tool-call replay across a crash (L4#4)**: characterized rather than fixed — added a test proving a side-effecting custom tool fires twice across a simulated container crash+restart (nothing survives between a tool's side effect and `mark_processed`, and the platform re-serves the same still-open message). Confirmed this is an accepted architectural boundary, not an oversight — `single_instance.py` and `claims.py` already document the same cross-process boundary as out of the SDK's reach for the long-running path too. Added a note to BUILDING.md: side-effecting custom tools need their own idempotency.

**Co-resident instances (L3)**: added an integration test (`tests/integration/test_agentcore_concurrency.py`, `requires_api`-gated) spawning real container processes — three concurrent instances each complete a real Band REST startup and answer `/ping` independently, and a deliberate port clash fails at the OS bind step, not a shared lock or singleton. Also verified live against the dev platform for both this and the custom-tool dispatch above.

Also extracted the BandLink/REST test doubles from `test_oneshot.py` into a shared `tests/runtime/conftest.py`, built from real `band_rest` model instances instead of bare `MagicMock`s (catches real schema drift instead of silently tolerating it), and replaced the fake adapter's `call_args`-based inspection with a plain `received_input` attribute. Also extracted a shared `tests/ports.py::reserve_port()` ephemeral-port helper and refactored the two real duplicates onto it (`tests/e2e/vscode/server.py`, `tests/integration/test_agentcore_concurrency.py`).

## Test plan
- [x] `uv run pytest tests/runtime/test_oneshot.py -v --no-cov` — 40 passed
- [x] `uv run pytest tests/ --ignore=tests/integration/ --ignore=tests/e2e/ --no-cov` — 4311 passed, 20 pre-existing unrelated failures (strands adapter, config-drift — confirmed failing identically on clean `main`, tracked as INT-1165), 0 from this change
- [x] `uv run pytest tests/integration/test_agentcore_concurrency.py -v --no-cov` (real Band creds) — 2 passed
- [x] `uv run ruff check` / `uv run ruff format --check` — clean

https://claude.ai/code/session_01EoBzp9vWrsrPhxcgLw9yWT

### DIFF
--- a/src/band/runtime/oneshot.py
+++ b/src/band/runtime/oneshot.py
@@ -474,13 +474,22 @@ class OneShotInvoker:
                     exc_info=True,
                 )
                 if page_num == 0:
-                    return [], set(), False
+                    return [], set(), True
                 truncated = True
                 break
             items.extend(response.data or [])
             if not response.metadata.has_more:
                 break
             cursor = response.metadata.next_cursor
+            if cursor is None:
+                logger.warning(
+                    "has_more=True but no next_cursor for room %s (page %d) — "
+                    "stopping pagination",
+                    room_id,
+                    page_num,
+                )
+                truncated = True
+                break
         else:
             truncated = True
             logger.warning(
--- a/src/band/runtime/oneshot.py
+++ b/src/band/runtime/oneshot.py
@@ -40,6 +40,7 @@ Example usage::
 from __future__ import annotations
 
 import logging
+from collections import deque
 from datetime import datetime, timezone
 from typing import Any
 
@@ -67,11 +68,18 @@ DEFAULT_DRAIN_CAP = 50
 # Items per page when following the context endpoint's cursor pagination.
 DEFAULT_HISTORY_LIMIT = 50
 
-# Defensive cap on how many pages of history to follow per invocation. A room
-# shouldn't backlog thousands of messages for a single agent in normal
-# operation; if it does, surface it via ``history_truncated`` rather than
-# paginating indefinitely on every stateless invocation.
+# How many of the most recent pages of history to retain per invocation. The
+# context endpoint is oldest-first with no way to request newest-first (see
+# ``_fetch_history``), so pagination always walks to the true end regardless
+# of this value — it only bounds memory/what reaches the LLM, never causes
+# the newest messages to be dropped in favor of older ones.
 DEFAULT_HISTORY_PAGE_CAP = 20
+
+# Absolute ceiling on pagination round-trips per invocation, independent of
+# how many pages are retained (``DEFAULT_HISTORY_PAGE_CAP``). Guards against
+# a backend that never reports ``has_more=False`` — hitting this is a
+# backend contract violation, not a real conversation length.
+DEFAULT_HISTORY_FETCH_CEILING = 500
 
 
 class OneShotEnvelopeError(ValueError):
@@ -93,8 +101,8 @@ class OneShotInvoker:
         agent_id: This container's Band agent identity (used for
             self-message filtering and the adapter's runtime identity).
         drain_cap: Defensive ceiling on the drain loop. Default 50.
-        history_page_cap: Defensive ceiling on how many pages of history to
-            follow per invocation. Default 20.
+        history_page_cap: How many of the most recent pages of history to
+            retain per invocation. Default 20.
     """
 
     def __init__(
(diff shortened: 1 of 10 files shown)


src/band/runtime/oneshot.py
```diff
@@ -416,24 +440,64 @@ async def _fetch_history(
...
-                chat_id=room_id,
-                page=1,
-                page_size=50,
-                request_options=DEFAULT_REQUEST_OPTIONS,
-            )
-        except Exception:
+        items: list[Any] = []
+        cursor: str | None = None
+        truncated = False
+        for page_num in range(self._history_page_cap):
```
amit-gazal-band: **Truncation keeps the oldest pages and drops the newest.**

Confirmed against the platform: the keyset query is ascending (`chat.ex:3035` defaults `sort_order` to `:asc` and the controller never overrides it; `chat.ex:3053` orders `asc: inserted_at, asc: id`; `chat.ex:3098` filters `m.inserted_at > ^t`). So `cursor=None` is the oldest page and this loop walks forward in time.

That means when the cap is hit, the surviving history is the room's *first* `cap × limit` messages and the recent turns — the ones the triggering message is actually a reply to — are the ones dropped. The LLM answers today's question with the conversation's opening.

The endpoint offers no way to page from the newest end (its `parameters:` list is only `cursor`, `limit`, `page`, `page_size`), so the options are: keep paginating to the end and retain the **last** N pages (bounded memory, same worst-case round-trips), or make the cap high enough that hitting it is genuinely exceptional and treat `history_truncated` as an alarm rather than a routine outcome.

amit-gazal-band (changes requested):
Checked the pagination against the real platform implementation (`thenvoi-platform`), not just the generated client's docstring. One blocker: the page cap truncates the wrong end of history.

The keyset path is **ascending**, so `cursor=None` is the *oldest* page and pagination walks oldest → newest:

- `context_controller.ex:show` → `Chat.get_agent_context_cursor/3`, never passing `sort_order`
- `chat.ex:3035` — `sort_order = if Keyword.get(opts, :sort_order, :asc) == :desc, do: :desc, else: :asc`
- `chat.ex:3053` — `order_by([m], [{:asc, m.inserted_at}, {:asc, m.id}])`
- `chat.ex:3098` — the `:asc` cursor filter is `m.inserted_at > ^t`

So `items.extend(...)` accumulates in correct chronological order (no reversed transcript) — but when `history_page_cap` is hit, `_fetch_history` returns the room's **first** 1000 messages and never reaches the recent ones, including the turns the triggering message is replying to. The endpoint exposes no `sort_order`/`before` parameter (only `cursor`, `limit`, `page`, `page_size`), so the SDK can't ask for newest-first — the fix has to be "paginate to the end, keep the last N pages", or a cap high enough that hitting it is genuinely exceptional.

For context on the stakes: the offset path (`chat.ex:3011`, `order_by asc: inserted_at` + `offset`) means the old `page=1, page_size=50` returned the **oldest 50** relevant messages, so the container has been running on ancient history for any room past 50 relevant messages. This PR raises that ceiling 20× — real improvement — but keeps the same wrong-end semantics beyond it.

Details inline. The rest of the API migration checks out: `cursor_or_limit_present?/1` still routes a `limit`-only first call to the keyset path (Fern drops `None` query params), `mixed_pagination_families?/1` isn't tripped, and `maybe_add_deprecation_headers/2` no longer fires.

src/band/runtime/oneshot.py
AlexanderZ-Band: Confirmed and fixed in 4dc0374c. Verified the ascending-order claim directly against `thenvoi-platform` (`context_controller.ex` show/2 never passes `sort_order`; `chat.ex:2865` defaults to `:asc`; `chat.ex:2883` orders `asc: inserted_at, asc: id`).

Went with your first option: `_fetch_history` now always paginates to the true end (`has_more=False`), bounded by a new `DEFAULT_HISTORY_FETCH_CEILING` (500) that only guards against a backend that never terminates. `history_page_cap` now governs a `deque(maxlen=...)` retention window over the fetched pages, so a low cap only bounds memory — it never causes the newest messages to be dropped in favor of older ones.

Updated `test_room_active_for_months_gets_a_prompt_reply_not_a_full_replay` to pin this: asserts the retained content is the newest days (2/3/4, not 0/1/2) and that all 5 pages get fetched (`get_context.await_count == 5`), not just that `history_truncated` is set. Full suite green.

amit-gazal-band (approved):

PR #520, opened 2026-08-04
feat!: cohesive adapter construction surface (INT-1166)

AlexanderZ-Band:
## Summary

INT-1166 — one cohesive construction shape across all 20 adapters, keeping each adapter's parameter vocabulary in parity with its underlying framework (major-version change; old construction paths break where the result is cleaner):

```python
adapter = XAdapter(<framework-native config only>)      # sync, cheap: no resources, no Band credentials
agent = Agent.from_config("my_agent", adapter=adapter)  # Band credentials exactly once
await agent.run()                                       # resources boot on start, die on stop
```

### Commit 1 — ParlantAdapter owns the Parlant server lifecycle (P2)

Parlant was the only adapter forcing customers to hold a server open and wire ports/tools/agent by hand in every example. Now:

- `ParlantAdapter(name=, description=, nlp_service=, ...)` — plain sync constructor; ports reserved and `p.Server` booted in `on_started`, torn down in `cleanup_all` (like codex's subprocess or the gateway's HTTP server).
- Guidelines declared up front via `adapter.add_guideline(condition=, action=, **kwargs)` and created on the live agent at startup, Band platform tools attached by default (`tools=[]` opts out; extra kwargs forward verbatim to Parlant's `create_guideline`).
- Escape hatches: `configure=` async callback receiving the live `(server, parlant_agent)` for full native API access; `server=` / `parlant_agent=` bring-your-own (borrowed — never torn down by the adapter).
- The hang-free in-process server helper moved from the E2E toolkit into `band.integrations.parlant.server` as the single source of truth (Parlant's `__aexit__(None, …)` is the serve-forever path; the exception path leaks the plugin-server port — the promoted helper cancels the serve loop only after Parlant's own cleanup ran).
- Dead `additional_tools` parameter removed — stored but never bridged into Parlant tool-calling (INT-1167 tracks the real bridge).
- All 5 `examples/parlant/` rewritten; `run_agent.py`'s parlant branch fixed (passed `model=`/`guidelines=` kwargs that never existed on any version — pre-existing rot).

### Commit 2 — platform connection injection; bridges stop re-asking for Band creds (P4)

- The ad-hoc `_band_agent_id` setattr is formalized: the runtime injects `adapter.platform` (`PlatformConnection`: agent_id, api_key, rest_url, ws_url) before `on_started`, in both `Agent.start` and the one-shot invoker; `SimpleAdapter` declares the attribute and offers `require_platform()`.
- The four bridge adapters build their REST client at startup from the injection instead of duplicating credentials already given to `Agent.create()`:
  - `A2AGatewayAdapter` — `rest_url`/`api_key` params dropped (`rest_client=` seam added)
  - `SlackAdapter` — `rest_url`/`api_key` params dropped (`rest_client=` seam kept)
  - `BandACPServerAdapter` — `rest_url`/`api_key` params dropped (`rest_client=` seam added)
  - `ACPClientAdapter` / `CopilotACPAdapterConfig` — dead `rest_url` param dropped (assigned + validated, never consumed)
- `band.testing.platform_connection_stub()` for tests driving adapters without the Band runtime.

### Commit 3 — feature-kwarg flatten: `features=AdapterFeatures(...)` → direct kwargs

Every adapter constructor now takes `emit=`, `capabilities=`, `include_tools=`, `exclude_tools=`, `include_categories=` directly (`**features: Unpack[FeatureKwargs]`), never a wrapping `AdapterFeatures(...)` object:

```python
adapter = ClaudeSDKAdapter(model="...", emit=Emit.TOOL_CALLS | Emit.THOUGHTS)
adapter = AgnoAdapter(agent, capabilities=Capability.MEMORY)
```

- `Emit`/`Capability` become a `StrEnum` mixin (`_FlagEnum`): members combine with `|` into a `frozenset`, a lone member is also accepted directly. Not `enum.Flag` — that loses string values (int bitmask) and breaks the frozenset-style set algebra (`-`, `in`) already used throughout.
- **Default flip**: `emit` omitted now defaults to everything the adapter's `SUPPORTED_EMIT` declares (was silent-unless-asked); pass `emit=()` for silence. `capabilities` stays opt-in — omitted, it's empty (unchanged: it puts extra tool schemas in front of the model on every turn).
- A value outside `SUPPORTED_EMIT`/`SUPPORTED_CAPABILITIES` raises `BandConfigError` immediately at construction — was a lazy `UserWarning` in `on_started`.
- `Emit.EXECUTION` renamed `Emit.TOOL_CALLS` (clearer; no wire-format ambiguity).
- Removed outright, not deprecated (pre-1.0, breaking is acceptable): per-adapter `enable_execution_reporting`/`enable_memory_tools` booleans, and the Codex/Letta/Opencode config-object equivalents (`emit_thought_events`, `enable_task_events`).
- Found and documented in the same pass: `Emit.TASK_EVENTS` is load-bearing (not just narration) on Codex/Letta/Opencode — each persists its session/thread/agent-resume mapping in task-event metadata gated by that flag. Narrowing `emit` to exclude it also stops resumption across restarts; now called out in each adapter's class docstring and in AGENTS.md.
- `AdapterFeatures` remains valid as an *internal* container type (and as a parameter to internal helpers like `render_system_prompt`) — only the public adapter-constructor kwarg surface changed.
- `SlackAdapter` gained a proper `_resolve_features()` override (the extension point `SimpleAdapter` already documented for exactly this) instead of constructing-then-discarding an `AdapterFeatures` before adopting the inner adapter's.
- Full sweep off the retired pattern: all 16 brain + 3 bridge adapters in `src/band`, ~50 examples, ~40 test files, `docs/adapters/*.md`, `MIGRATING-v0.3.0.md` deleted (both its "before" and "after" states were retired, so it described nothing current).

### Commit 4 — Codex/Letta/Opencode config classes self-source from the environment

`CodexAdapterConfig`, `LettaAdapterConfig`, `OpencodeAdapterConfig` are now `pydantic_settings.BaseSettings`: any field left unset falls back to a `<PREFIX>_`-prefixed environment variable (e.g. `CODEX_MODEL`, `LETTA_BASE_URL`, `OPENCODE_PROVIDER_ID`) before its code default — an explicit constructor kwarg always wins. Removes the hand-rolled `os.getenv(...)` per field that every example previously needed.

- `LettaAdapterConfig`'s deprecated-field shims (`api_key`, `mcp_server_url`/`mcp_server_name`) moved from dataclass `__post_init__` to a pydantic `model_validator`; `provider_key` reads the unprefixed `LETTA_API_KEY` (Letta Cloud's own name), while the deprecated `api_key` field deliberately stays environment-blind so a real `LETTA_API_KEY` never spuriously triggers its deprecation warning.
- `CopilotSDKAdapterConfig`/`CopilotACPAdapterConfig` evaluated and **not** converted: frozen dataclass, holds an external SDK type (`ProviderConfig`) and a callable field (`ask_user`), only one field (`github_token`) has real env precedent — poor fit.
- Surfaced and fixed a real test-hermeticity gap this exposed: `tests/e2e/baseline/settings.py` eagerly calls `load_dotenv(ENV_TEST_FILE)` at import time, so a contributor's local `.env.test` could leak `CODEX_*`/`LETTA_*`/`OPENCODE_*` values into unrelated unit tests' "no override" defaults, non-deterministically depending on which test happened to import that module first. Fixed via a `pytest_configure` hook in `tests/conftest.py` that neutralizes those keys before any subdirectory `conftest.py` can trigger the leak (skipped when `E2E_TESTS_ENABLED=true`, where the real values are the point), plus a defense-in-depth autouse fixture.

### Commit 5 — review fixes

- `SlackAdapter`: an explicit `emit=`/`capabilities=` override now propagates to `inner.features` too — a turn dispatches straight to `inner.on_message()`, which reads the inner instance's own `self.features`, so the override was previously a no-op.
- `docker/{codex,letta,claude_sdk}/runner.py`: dropped retired constructor/config fields (`enable_task_events`, `enable_execution_reporting`, `emit_thought_events`) that now raise at construction; migrated to `emit=`.
- `docs/adapters/codex.md`: removed a stale row documenting the removed `enable_task_events` field.
- `test_surface_guards.py`'s file scan widened to also cover `docker/*/runner.py` and `docs/adapters/*.md` (previously `examples/` only) — same root cause behind the last two.

## Breaking changes

- `ParlantAdapter(server=, parlant_agent=)` are now optional keyword-only escape hatches; `additional_tools` removed.
- Bridge adapters no longer accept `api_key`/`rest_url` constructor parameters.
- Adapters reading the injected agent id use `adapter.platform.agent_id` instead of `_band_agent_id`.
- `emit` omitted now defaults to everything the adapter's `SUPPORTED_EMIT` declares (was silent-unless-asked): previously-silent adapters start posting tool-call/thought/task-event narration to rooms after upgrade — more room messages and platform traffic. Pass `emit=()` to restore the old silence.
- `features=AdapterFeatures(...)` no longer accepted by any adapter constructor — use `emit=`/`capabilities=`/etc. directly. `Emit.EXECUTION` renamed `Emit.TOOL_CALLS`.
- Per-adapter `enable_execution_reporting`/`enable_memory_tools` booleans (and the Codex/Letta/Opencode config-object equivalents) removed, not deprecated.

## Deferred

- Flat-kwarg sugar vs `config=XConfig(...)` for codex/letta/copilot_sdk/copilot_acp/opencode — parked (INT-1166 description); those configs are legitimately larger (10–18 adapter-specific fields) than the shared 5-field feature vocabulary, so a config object stays the right shape.
- INT-1167 — real custom-tools bridge for Parlant.

## Validation

- 4554 unit tests pass (84 skipped), 0 failed; pyrefly 0 errors; ruff check + format clean; 45/45 markdown-docs snippets pass.
- Live E2E baseline lanes: `google` 92 passed/0 failed, `crewai` 71/0, `letta` 57/0, `backends` 73 passed/8 failed, `core` 252 passed/6 failed. All failures are model-nondeterminism/infra-transient (`@flaky_model`/`@flaky_infra`, e.g. `test_memory_survives_adapter_rehydration`) except one: `core`'s `test_handled_work_not_redrained_on_restart[strands]` hits a pre-existing bug in `src/band/converters/strands.py`'s Converse-to-OpenAI role merge, newly exercised because this PR's `emit` default flip makes Strands narrate tool calls (and therefore replay them) for the first time. Filed as INT-1172, left red intentionally — not in this diff.

Linear: INT-1166

https://claude.ai/code/session_017uq15BnoT1Bdw9jjaZYRPB


src/band/adapters/parlant.py
```diff
@@ -129,37 +215,144 @@ def __init__(
...
+                self._server_cm = server_cm
+                assert prepared_agent is not None and prepared_app is not None
+                self._server = server
+                self._parlant_agent = prepared_agent
+                self._app = prepared_app
+            else:
+                self._parlant_agent, self._app = await self._prepare_server(
+                    self._server, agent_name, agent_description
+                )
+        except BaseException:
```
amit-gazal-band, line 307: On the **borrowed-server** path (`server=` given, adapter-created agent → `_owns_server=False`, `_created_agent=True`), a failure inside `_prepare_server` leaves `_guidelines_applied_count` advanced while `_parlant_agent` is still `None`, and neither reset that would clear it can run:

- `_release_server()` does reset the count (lines 876-878), but that code sits *after* its `if server_cm is None: return` guard on line 873 — and `_server_cm` is only ever assigned in the owned-server branch (line 298), so it early-returns here.
- The `if self._owns_server:` block just below is false.

Meanwhile the agent itself was built into a local in `_prepare_server` (line 331) and only published by the successful tuple-unpack on line 304, so it's dropped.

Concretely: five `add_guideline` specs, the third raises inside `agent.create_guideline`. State left behind is `_guidelines_applied_count == 2`, `_parlant_agent is None`. The caller retries `agent.start()`; `_prepare_server` sees `agent is None`, creates a brand-new Parlant agent, and applies `self._guideline_specs[2:]` — specs 1 and 2 never land on it. The agent then runs looking healthy, missing two guidelines, with nothing in the logs.

The count's comment ("it resets alongside that agent") holds only for the owned-server branch. Suggest tying the reset to the thing it actually tracks — whichever agent `_parlant_agent` currently points at — rather than to server ownership: either reset it in this `except` whenever `self._parlant_agent is None` after `_release_server()`, or have `_prepare_server` count against the local `agent` it built and publish the count alongside the successful agent assignment.

Not blocking — it only bites when something retries `start()`.

amit-gazal-band (dismissed):
Approving. Read every `src/` hunk and ran the full gate at this head — unit suite, `ruff check`, `pyrefly check`, and the markdown-docs suite are all clean.

The flattened `**features: Unpack[FeatureKwargs]` surface reads much better than the old `features=AdapterFeatures(...)` wrapper, and moving feature validation to construction time (`BandConfigError` instead of a `UserWarning` in `on_started`) is the right call. Same for dropping the repeated `rest_url=`/`api_key=` from the bridge adapters in favor of the injected `PlatformConnection`.

One non-blocking correctness note inline on the Parlant borrowed-server path. Two smaller things worth a follow-up rather than holding the PR:

- `src/band/adapters/opencode/adapter.py:296` — `_preflight_server` catches `httpx.HTTPError`, which includes the `HTTPStatusError` from `health()`'s `raise_for_status()`. Any non-2xx from `/global/health` (404 on a build without the route, 401/403/405 through a reverse proxy) hard-fails `Agent.start()` for a server that served turns fine before this PR, and there's no config opt-out (`_preflight_enabled` is `client_factory is None`). Narrowing to `httpx.TransportError` would keep the fail-fast intent — any HTTP response at all is proof of life.
- Please make sure the release notes lead with the `emit` default flip. Omitted `emit` now resolves to `SUPPORTED_EMIT` rather than empty, so an existing deployment that upgrades without touching code starts POSTing tool_call/tool_result pairs, thoughts, and a per-turn usage event into every room. It's documented in the docstrings, but it's the change most likely to surprise someone.

src/band/adapters/parlant.py
AlexanderZ-Band: Confirmed — fixed in 791d8ed4: the reset now fires whenever `self._parlant_agent is None` after `_release_server()`, not only on the owned-server path, so a retried `start()` always redoes every guideline against the fresh agent it creates rather than skipping ahead by the stale count.

amit-gazal-band (approved):

PR #532, opened 2026-08-13
feat: nightly baseline E2E schedule + release-please gate

AlexanderZ-Band:
## Summary

Runs the baseline E2E suite nightly over the full lane × OS matrix and turns the result into a signal that gates the release.

- **Nightly `schedule:` cron** on `e2e.yml`. A `workflow_dispatch` with `lane: all`/`os: all` reproduces it exactly.
- **`gate()`/`GateResult` in `scorecard.py`**: any `fail` cell, or a `skip` cell whose home lane was expected this run, reddens the run; `N/A`+reason passes. The `scorecard` job writes the grid to the step summary and backstops on a whole leg crashing.
- **Lane-level retry**: failed nodeids only, once, folded back via `overlay()` (last-write-wins, not rank-based).
- **`mark-baseline`**: posts the `baseline-green` commit status and a digest on one persistent tracking issue.
- **`release-gate.yml`**: a no-op for every ordinary PR; for the release-please PR it blocks until the baseline is green. Not a blanket "E2E is a required PR check".

## How the release gate decides

A nightly marks one commit — main's tip at 03:17 UTC — so gating on the live tip would go red the moment anything merged after it.

`check-release-baseline.sh` walks the base branch newest-first (`COMMIT_SCAN_LIMIT`, default 40) for the most recent commit carrying a `baseline-green` status and blocks unless it is `success`. A green one older than `MAX_BASELINE_AGE_DAYS` (default 7) stops vouching. The context name lives in `baseline-status-context.sh`, sourced by both producer and consumer, so a typo cannot silently desync them.

## Per-leg timeout: 240 minutes

`timeout-minutes` in `e2e.yml` is the single definition. GitHub reports a timeout as `cancelled`, which the gate treats as a failure, so a too-tight cap manufactures red baselines — a healthy `core/windows` leg took 110 minutes.

## Also in this diff

Hardening of the scripts the workflow calls: exit-code propagation from the gate, a loud failure for a lane×OS selection with no runnable cell, `!cancelled()` and normalized dispatch inputs on both reporting jobs, `--lfnf=none` on the retry, a 3-way overlay branch, a reachable empty-roster guard, and a checkout before `release-gate.yml` invokes a repo script. `test_e2e_ci_scripts.py` and `test_e2e_lane_drift.py` cover these by driving the real scripts.

## Test plan

- [x] Unit suite — 4577 passed, 111 skipped
- [x] `ruff` / `pyrefly` / `actionlint` / `shellcheck -x` — clean
- [x] Full-matrix run [31799268514](https://github.com/band-ai/band-sdk-python/actions/runs/31799268514): 11 legs, `letta` ubuntu-only; retry ran only the failed nodeids; `overlay` folded 345 cells preserving attempt 1's passes
- [x] `release-baseline-gate` reports success here — the no-op path
- [ ] After INT-1206: confirm the gate reflects real `baseline-green` state on the release PR

Not yet exercised live: the 240 cap, `check-release-baseline.sh` against the real API, `!cancelled()`, the `merge_group` path.

## Known failures in that run — pre-existing, tracked in INT-1212

This PR changes no `src/` code. 3 of 11 legs failed:

- **Anthropic credits exhausted** — `anthropic` and `copilot_sdk` return HTTP 400; four further cells likely collateral. Also gates the E2E judge.
- **`test_acp_band_tool_result_is_a_single_clean_payload`** — test unsatisfiable, adapter correct: #514 nests the payload in a `str` field and escapes its quotes, so `.containing('"success"')` cannot match.
- **`test_codex_thoughts_are_not_placeholders`** — asserts ≥1 thought while `reasoning_summary` defaults to `None`, so none is requested.
- **Copilot CLI unpinned** in `setup-copilot.sh`.

## Not included

Adding `release-baseline-gate` to the `main-branch-protection` required checks — that is what makes the gate block merges. Tracked as INT-1206.

Closes INT-1195.


.github/workflows/e2e.yml
```diff
@@ -293,17 +287,23 @@ jobs:
...
+    needs: [lanes, e2e]
     if: always()
     runs-on: ubuntu-latest
     defaults:
       run:
         shell: bash
+    outputs:
+      # Consumed by mark-baseline to decide the baseline-green commit status and the
+      # nightly digest's header/warning.
+      passed: ${{ steps.verdict.outputs.ok }}
```
amit-gazal-band, line 349: **[BLOCKING]** A nightly that breaks *before* the merge step posts no status and no digest at all, and the release gate silently keeps vouching with an earlier green.

`passed` is `steps.verdict.outputs.ok`, and every consumer hard-guards with `:?` (`compute-gate-verdict.sh:14`, `mark-commit-status.sh:9`, `post-baseline-digest.sh:11`), so an **empty** output aborts instead of defaulting to red. Concrete path:

1. The `lanes` job fails (registry drift — exactly what its asserts exist to catch).
2. `needs.lanes.outputs.lanes` is empty, so `Resolve expected lanes` runs `echo "" | jq ...` → empty output, **exit 0** (verified locally).
3. `EXPECTED_LANES=""` trips `merge-scorecard.sh:7`'s `:?`; the step fails without writing `passed`.
4. `compute-gate-verdict.sh` aborts on `MERGE_OK:?` without writing `ok`.
5. `mark-baseline` still runs (`!cancelled()` overrides the implicit `success()`) but dies on its **first** step's `PASSED:?` → **no `baseline-green` status, and every later step (digest, issue sync) skipped.**

`check-release-baseline.sh` then walks back to an earlier `success` and lets the release PR merge — for up to `MAX_BASELINE_AGE_DAYS` (7) of silently-missing nightlies. The same chain likely fires if `Download lane scorecards` errors on zero matching artifacts (every leg died before upload).

Fix: default these to false (`MERGE_OK="${MERGE_OK:-false}"`, `PASSED="${PASSED:-false}"`, and likewise for `EXPECTED_LANES`) rather than aborting. A verdict that can't be computed is red, not absent.

.github/workflows/e2e.yml
```diff
@@ -340,4 +356,164 @@ jobs:
...
+  # the separate job below, so their limited coverage can never affect release gating.
+  mark-baseline:
+    name: mark baseline-green
+    needs: scorecard
+    # `!cancelled()` rather than `always()`: this job writes durable, externally
+    # visible state (a commit status, plus reopening the tracking issue), and a
+    # cancelled run is not evidence of anything — under `always()` a nightly whose
+    # legs were cancelled mid-flight (the per-(lane,OS) concurrency group, or a
+    # human cancelling the run) posts `baseline-green: failure` and reopens the
+    # issue with nothing actually broken. Declining to post is the safe direction:
```
amit-gazal-band: **[BLOCKING]** This safety argument is inverted, and it's what makes the finding above dangerous.

"release-gate.yml reads a missing status as `pending` and blocks anyway" is not what `check-release-baseline.sh` does: it **skips** commits with no status (`:39-44`) and gates on the *most recent commit that has one*. So declining to post is **not** the safe direction — it silently falls through to an older verdict.

`!cancelled()` may well still be the right condition here (see the separate note below on what it actually covers), but please fix the stated rationale — as written it invites the next person to reintroduce a fail-open "just don't post" path. The same claim is repeated in `test_e2e_lane_drift.py:106`.

tests/framework_conformance/test_e2e_lane_drift.py
```diff
@@ -36,3 +54,91 @@ def test_workflow_lane_extraction_is_not_vacuous() -> None:
...
+
+@pytest.mark.parametrize("job_name", _REPORTING_JOBS)
+def test_reporting_jobs_do_not_certify_a_cancelled_run(job_name: str) -> None:
+    """A cancelled run is not evidence, so it must not post a status or comment.
+
+    Under ``always()`` a nightly whose legs were cancelled mid-flight (the
+    per-(lane,OS) concurrency group, or a human cancelling) reports the baseline as
+    *red* and reopens the tracking issue with nothing actually broken. Declining to
+    report is the safe direction: ``release-gate.yml`` reads a missing status as
+    ``pending`` and blocks anyway. A ``timeout-minutes`` leg kill does not cancel the
```
amit-gazal-band: **[BLOCKING]** (same fix as `e2e.yml:388`) This docstring repeats the inverted premise: `release-gate.yml` does **not** read a missing status as `pending` — `check-release-baseline.sh` skips statusless commits and gates on the newest commit that has a status, so a missing status inherits an older verdict.

The assertions below are fine; it's the stated reasoning that would mislead the next reader of this guard.

.github/scripts/check-release-baseline.sh
```diff
@@ -0,0 +1,65 @@
...
+jq_latest="$jq_latest | select(. != null) | \"\\(.state) \\(.created_at)\""
+
+mapfile -t shas < <(
+  gh api "repos/$REPO/commits?sha=$BASE_REF&per_page=$COMMIT_SCAN_LIMIT" --jq '.[].sha'
+)
+
+entry=""
+tested_sha=""
+for sha in "${shas[@]}"; do
+  entry=$(gh api "repos/$REPO/commits/$sha/statuses" --jq "$jq_latest" || true)
```
amit-gazal-band: **[BLOCKING]** `|| true` makes an API failure indistinguishable from "no status on this commit", which can walk *past* a red baseline onto an older green one.

This loop makes up to `COMMIT_SCAN_LIMIT` (40) sequential `gh api` calls. If the call for the newest tested commit — the one carrying `failure` — hits a transient 5xx or a secondary rate limit, the error is swallowed, the loop continues, finds an older `success`, and the gate passes on a baseline that is actually red.

Suggest capturing `gh`'s exit status separately from its output and aborting on a non-zero rc that isn't "no status of this context" (that case is already an empty result thanks to the `select(. != null)` in `jq_latest`). An unreadable baseline should block, not scroll past.

.github/workflows/e2e.yml
```diff
@@ -340,4 +356,164 @@ jobs:
...
+    # `timeout-minutes` leg kill does NOT cancel the run, so it still reddens here.
+    #
+    # Both dispatch inputs are normalized with `|| 'all'`, matching every other
+    # consumer (the `lanes`/`e2e` jobs): compared bare, an input that resolved
+    # empty would read as "scoped", so a full-matrix run would silently never
+    # certify the commit while `lanes` had in fact fanned out the whole matrix.
+    if: >-
+      !cancelled() &&
+      (github.event_name == 'schedule' ||
+       ((github.event.inputs.lane || 'all') == 'all' &&
```
amit-gazal-band, line 455: **[BLOCKING]** A full-matrix `workflow_dispatch` certifies `github.sha` on **any** branch — the condition checks `lane`/`os`, never the ref.

So a developer running the full matrix from a feature branch posts `baseline-green` on a non-main commit *and* closes/reopens the canonical nightly issue based on branch state. I checked the repo's merge settings: `allow_merge_commit` and `allow_rebase_merge` are both **true**, so that branch's SHAs can land in main's history verbatim — after which `check-release-baseline.sh` scanning `sha=main` will find that status and let a feature-branch baseline vouch for main. Not hypothetical.

Fix: add a default-branch ref check to this job's `if`, e.g. `github.ref == format('refs/heads/{0}', github.event.repository.default_branch)`.

.github/workflows/e2e.yml
```diff
@@ -340,4 +356,164 @@ jobs:
...
+    # issue with nothing actually broken. Declining to post is the safe direction:
+    # release-gate.yml reads a missing status as `pending` and blocks anyway. A
+    # `timeout-minutes` leg kill does NOT cancel the run, so it still reddens here.
+    #
+    # Both dispatch inputs are normalized with `|| 'all'`, matching every other
+    # consumer (the `lanes`/`e2e` jobs): compared bare, an input that resolved
+    # empty would read as "scoped", so a full-matrix run would silently never
+    # certify the commit while `lanes` had in fact fanned out the whole matrix.
+    if: >-
+      !cancelled() &&
```
amit-gazal-band, line 452: **[NON-BLOCKING]** (fails closed — false red, not false green) `!cancelled()` does not cover the scenario this comment cites, so a concurrency-cancelled leg still posts a false `baseline-red`.

`cancelled()` reflects the **workflow run**, not a `needs` job — which this PR already knows: `compute-gate-verdict.sh:6-8` and the line just below here both note that a `timeout-minutes` leg kill lands as job conclusion `cancelled` *without* cancelling the run. The same is true of a job-level concurrency cancel, and the group at `:154` (`e2e-baseline-<lane>-<os>`) is neither run- nor ref-scoped.

Scenario: the nightly starts at 03:17; someone dispatches a scoped `core`/`ubuntu` run (from any branch) → the nightly's `core/ubuntu` leg is cancelled → `needs.e2e.result == 'cancelled'` → `matrix_ok=false` → this job posts `baseline-green: failure` on main's tip and reopens the tracking issue "with nothing actually broken" — precisely what the comment claims is prevented. As written, `!cancelled()` only guards against a human cancelling the whole run.

Guard on `needs.e2e.result != 'cancelled'` instead, or scope the concurrency group by run/ref. Worth confirming against a live run before changing the condition, since this rests on GitHub's `cancelled()` scoping.

.github/workflows/release-gate.yml
```diff
@@ -0,0 +1,92 @@
...
+# feedback while the PR is open, `merge_group` so the check also reports against the
+# merge queue's temporary merge commit — without it the queue would wait forever once
+# this job's name is added as a required status check.
+#
+# Registered as a required status check on main-branch-protection under the context
+# "release-baseline-gate" (this job's name) — a live ruleset setting, not a file in
+# this repo.
+
+on:
+  pull_request:
```
amit-gazal-band, line 35: **[NON-BLOCKING]** (fails closed) The gate only re-evaluates on PR events, so a green nightly cannot clear an already-red gate.

Triggers are `pull_request: [opened, synchronize, reopened, ready_for_review]` and `merge_group`. If the release-please PR's gate failed against a red baseline and then **no** further commits land on main, release-please never pushes to its branch, no `synchronize` fires, and the required check stays red after the next nightly turns green — until someone re-runs the job by hand.

A `schedule` or `workflow_run`-on-e2e re-trigger would close it; a documented "re-run this check after a green nightly" note in the header comment would be an acceptable stopgap.

.github/scripts/run-baseline-e2e.sh
```diff
@@ -0,0 +1,64 @@
...
+# the retry's rows over the original's (last attempt always wins per cell,
+# regardless of pass/fail rank) into the one file the rest of the job expects.
+#
+# Overlaying needs BOTH files, so each case is handled on its own: keyed on ATTEMPT2
+# alone, a first attempt that died before session end would hand overlay a missing
+# base, and the FileNotFoundError would cost the lane its whole fragment -- surfacing
+# at merge time as "missing" cells instead of the retry's real results. Whichever
+# attempts exist are used; only both-absent contributes nothing.
+if [ -f "$ATTEMPT1" ] && [ -f "$ATTEMPT2" ]; then
+  uv run python -m tests.e2e.baseline.scorecard overlay "$ATTEMPT1" "$ATTEMPT2" --out "$FINAL"
```
amit-gazal-band, line 65: **[NON-BLOCKING]** A failing `scorecard overlay` is invisible in the leg's exit code.

`set -e` is deliberately off (good reason, well documented), so if this command — or either `cp` below — errors, `$FINAL` is never written yet the script still `exit "$code"` from pytest: **0** when attempt 1 passed. The leg reports green while contributing no scorecard fragment.

In a full matrix the gate still catches it via sibling lanes' `skip` rows, so this is narrow — but a single-lane scoped dispatch would report a green matrix over an empty grid (only `merge-scorecard.sh`'s zero-files branch saves it, and only if no other fragment landed). Suggest folding the overlay/cp rc into `code`.

tests/e2e/baseline/scorecard.py
```diff
@@ -178,6 +182,113 @@ def merge(scorecards: Iterable[list[ScorecardRow]]) -> list[ScorecardRow]:
...
+    registry for a nightly/full-matrix run, or just the chosen lane for a scoped
+    dispatch) — never inferred from the rows themselves, so an intentionally
+    out-of-scope lane's cells can never be mistaken for a silent failure.
+    """
+    home_lane = adapter_home_lanes()
+    failing = tuple(r for r in rows if r.status == "fail")
+    missing = tuple(
+        r
+        for r in rows
+        if r.status == "skip" and home_lane.get(r.adapter) in expected_lanes
```
amit-gazal-band: **[NON-BLOCKING]** (latent) The `missing` test uses the adapter's **home** lane and ignores `@lane()` pins, so a legitimately-skipped cell can be reported as a silent failure.

Scheduling resolves the pin *first* (`lane_selection.py::_lane_skip_reason` checks `_override_lane(item)` before `_home_lanes`), and `ADDING_AN_ADAPTER.md` prescribes exactly that pin for a cross-lane cell. Add an `@lane(Lane.GOOGLE)` pin to a `@per_adapter` test whose adapter's home lane is `core`: in the `core` leg the cell is skip-marked with a correct reason, and a `--expected-lanes core` run then lists it under "Missing (lane ran, no result)" and reddens the digest.

No such matrix cell exists today, so nothing is broken — but the docs actively recommend the pattern that would trip it, so it's a trap for whoever adds the first cross-lane cell. Reusing the same override-aware lane resolution here (rather than `adapter_home_lanes()` alone) would keep the gate and the scheduler on one rule.

amit-gazal-band (changes requested):
Review of the nightly baseline schedule + release gate. Nice, well-commented design — the per-lane concurrency, the retry/overlay split, and the shared status-context file are all the right calls.

One theme runs through the blocking findings: **every path that fails to write a verdict fails _open_.** The `:?` guards abort instead of defaulting to red, and `check-release-baseline.sh` treats a missing or unreadable status as "look further back" rather than "block" — so a nightly that dies early, or one flaky `gh api` call, lets the gate keep vouching with an older green (bounded only by `MAX_BASELINE_AGE_DAYS`, default 7).

Findings are tagged **[BLOCKING]** / **[NON-BLOCKING]** inline. The split is one criterion: does it let a bad baseline vouch for a release?

- **[BLOCKING]** — fail open: `e2e.yml:305`, `e2e.yml:388` (+ the matching test docstring), `check-release-baseline.sh:39`, `e2e.yml:399`. All four are small, localized edits, and together they close every path where the gate certifies something it hasn't verified.
- **[NON-BLOCKING]** — fail closed, latent, or cosmetic: `e2e.yml:397`, `release-gate.yml:28`, `run-baseline-e2e.sh:58`, `scorecard.py:234`. Happy to see these as follow-ups.

One non-finding: `e2e.yml:374` (`Fail the job on a red verdict`) carries an implicit `success()`, so it's skipped on the most common red path where the merge step already failed. The job is red either way — the explicit gate step is just dead code there.

.github/workflows/e2e.yml
AlexanderZ-Band: Confirmed — traced the exact chain you described. Fixed by defaulting the verdict outputs to false/empty instead of aborting via `:?`: `MERGE_OK` in compute-gate-verdict.sh, `PASSED` in mark-commit-status.sh and post-baseline-digest.sh, and `EXPECTED_LANES` in merge-scorecard.sh (which now reaches the python module with an empty value, and `known_lane_ids()` validation there still fails the merge — reddening it — instead of the shell aborting before `passed` is ever written). Also updated the stale docstring in test_e2e_ci_scripts.py that described the old `:?` guard. See aea61d2f.

.github/workflows/e2e.yml
AlexanderZ-Band: You're right, the rationale was backwards. Rewrote the comment to describe what check-release-baseline.sh actually does: a missing status is skipped over (not read as "pending"), so declining to post falls back to the last commit that was actually tested (bounded by MAX_BASELINE_AGE_DAYS), not a blocking pending state. See aea61d2f.

tests/framework_conformance/test_e2e_lane_drift.py
AlexanderZ-Band: Fixed the same inverted premise in this docstring to match the e2e.yml comment fix. See aea61d2f.

.github/scripts/check-release-baseline.sh
AlexanderZ-Band: Confirmed. Now captures `gh`'s exit code separately from `entry`'s output: a non-zero rc aborts the scan loudly instead of being swallowed by `|| true`; the legitimate "no status of this context" case stays a clean empty string via `jq_latest`'s existing `select(. != null)`. See aea61d2f.

.github/workflows/e2e.yml
AlexanderZ-Band: Confirmed, not hypothetical — added your suggested fix: `github.ref == format('refs/heads/{0}', github.event.repository.default_branch)` in mark-baseline's `if`, plus a guard test asserting the condition is present. See aea61d2f.

.github/workflows/e2e.yml
AlexanderZ-Band: Confirmed the gap, but went a different route than the suggested `needs.e2e.result != 'cancelled'` guard: that would also swallow a genuine timeout-minutes kill, which this same PR documents as landing with the identical "cancelled" conclusion and deliberately needs to still redden — so the two causes aren't distinguishable from `needs.e2e.result` alone. Fixed it at the root instead: scoped the e2e job's concurrency group by `github.event_name` (`e2e-baseline-${{ github.event_name }}-${{ matrix.lane }}-${{ matrix.os }}`), so a scoped manual dispatch can no longer collide with — and cancel — a nightly leg for the same lane+OS in the first place. The false cancellation can't happen rather than needing to be detected after the fact. Added a guard test asserting the group is scoped by event kind. Open to revisiting if you see a reason the event-kind scoping isn't sufficient. See aea61d2f.

.github/workflows/release-gate.yml
AlexanderZ-Band: Added the documented stopgap you suggested: a comment on release-gate.yml's trigger block naming the limitation and the manual remediation (re-run the job, or push a commit that resyncs release-please's PR). Left the auto re-trigger mechanism out of scope for this non-blocking finding — happy to open a follow-up if you'd rather have it built now instead. See aea61d2f.

.github/scripts/run-baseline-e2e.sh
AlexanderZ-Band: Confirmed. Folded the overlay/cp exit code into the script's own: `merge_code` now captures whichever of overlay/cp actually runs, and a non-zero value overrides `code` before the final `exit`. See aea61d2f.

tests/e2e/baseline/scorecard.py
AlexanderZ-Band: Confirmed and fixed at the root: pulled the override-aware resolution out into `lane_selection.expected_lane()` (an `@lane` pin wins, else the shared home lane) — `_lane_skip_reason` now delegates to it too, so the scheduler and the gate can't drift onto two different rules. `ScorecardCollector.scorecard()` stamps each `skip` row with its resolved `expected_lane` while the live item is still available, and `gate()` consults that field (falling back to the adapter's home lane only for older data that predates it). Added a test reproducing your exact scenario — a cell pinned cross-lane via `@lane`, gated against its home lane vs. its pin. See aea61d2f.

amit-gazal-band (approved):

PR #538, opened 2026-08-13
feat(delegation): typed msg.delegation, kept out of the model [INT-992]

eric-descourtis-thenvoi:
M1-S1 of Credential & Identity Propagation (INT-992). The platform now mints a secret-free identity envelope on `metadata["delegation"]` when a message crosses an ownership boundary (thenvoi-platform #2358/#2359). This PR gives agent code a typed, read-only view of it and proves the model can never see it.

## Invariants delivered
- **I1 — never in the model context**: the `delegation` key is stripped at `format_message_for_llm`, the single choke point every history path routes through (DefaultPreprocessor hydration, `get_history_for_llm`, oneshot). Live turns were already structurally safe: both `PlatformMessage.format_for_llm` implementations render `[name]: content` only — now pinned by a regression test per class.
- **I2 — typed read-only access**: `ctx.delegation` on `ExecutionContext` (set per message by the preprocessor) and `msg.delegation` on the runtime `PlatformMessage` (populated by `BandLink.get_next_message` / `get_stale_processing_messages`). Envelope models are `frozen=True`.
- **I3 — malformed never raises**: `parse_delegation` accepts `MessageMetadata`, plain dict, or `None`, and returns `None` on malformed input (logged); the REST re-wrap boundary (`MessageMetadata(**metadata)`) is guarded by a tolerant `mode="before"` validator. Six tolerance tests.
- **I4 — adapter contract untouched**: only the `delegation` key is stripped, only at the LLM-formatting seam — A2A/ACP/claude_sdk converters keep reading their session keys off metadata (`test_preserves_metadata` amended consciously; all other keys still pass through).

## Evidence
- CI-equivalent suite: 4586 passed, +44 new tests, **0 new failures** (the 1 failure + 11 errors in `tests/integrations/desktop_app/` reproduce identically at base `091a942` and pass in isolation — pre-existing ordering interaction, not this change).
- `ruff check` / `ruff format --check` / `pyrefly check` all clean.
- TDD throughout; each commit body records the injected mutation that proves its test bites (e.g. disabling the strip fails 4 formatter tests with the envelope present).

## Notes for review
- The preprocessor builds the **core** `PlatformMessage`, so the typed field lives on the **runtime** class it is actually constructed on; adapter-path code reads `ctx.delegation`. Core types stay untouched.
- Frozen platform shape this codes against: `{"version": 1, "originator": {"uuid", "handle", "display_name"}, "message_id", "minted_at", "hop": null}` — `extra="allow"` absorbs M2.5's future `hop` payload.

Linear: INT-992

### DIFF
--- a/src/band/client/streaming/client.py
+++ b/src/band/client/streaming/client.py
@@ -15,9 +15,16 @@ from phoenix_channels_python_client.client import (
 from phoenix_channels_python_client.client_types import ReconnectPolicy
 from phoenix_channels_python_client.exceptions import PHXConnectionError
 from phoenix_channels_python_client.phx_messages import PHXMessage
-from pydantic import BaseModel, ConfigDict, ValidationError, model_validator
+from pydantic import (
+    BaseModel,
+    ConfigDict,
+    ValidationError,
+    field_validator,
+    model_validator,
+)
 
 from band.client.streaming.errors import classify_initial_upgrade_error
+from band.core.delegation import DelegationEnvelope, parse_delegation_value
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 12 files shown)


AlexanderZ-Band:
Closing as part of the abandoned delegation PR stack (#538–#541): no activity since August and the branch conflicts with the current main.

PR #539, opened 2026-08-14
feat(delegation): CredentialResolver — token_for(audience) with cache and injection [INT-993]

eric-descourtis-thenvoi:
M2 SDK 1/3. **Stacked on #538 (INT-992)** — until #538 merges, this diff shows both changes; the INT-993 increment is the last 3 commits.

Adds the credential-exchange client for M2 ("your tokens, their agent"):

- `band.platform.CredentialResolver` — `await resolver.token_for(audience)` calls `POST /api/v1/agent/delegation/token` (hand-rolled httpx; `band_rest` 0.0.26 has no delegation client) with `X-API-Key` from the link, returning a frozen `DelegatedToken` (`access_token`, `token_type`, `expires_at: datetime | None`, `obo`).
- **Typed error taxonomy** mirroring the platform's frozen codes 1:1 — the first-run funnel pair `ConsentMissing` / `ProviderNotConnected` carries the platform's remediation strings; `WindowExpired`/`Revoked`/`AudienceNotAllowed`/`CrossOrgBlocked`/`DelegationDenied` are terminal; `NoDelegation` raised client-side when the turn has no envelope.
- **Cache discipline per the token-endpoint contract**: memory only (response is `cache-control: no-store`), expiry-aware with 30s skew, **no reuse when `expires_at` is null** (null = unknown, never fabricate a TTL), single-flight per audience so tool fan-out can't stampede the per-message exchange cap (default 20).
- `ctx.credentials` — per-turn resolver on `ExecutionContext`, rebuilt when the envelope identity changes; reads `ctx.delegation` from INT-992.

Evidence: TDD throughout, mutation recorded per commit (e.g. disabling the expiry check or the lock fails the matching test); suite 4620 passed with the pre-existing desktop_app reds byte-identical to base; ruff + pyrefly clean. Testing note: pytest-httpx resolves without yielding, so the single-flight test uses a suspending async callback to make concurrency real.

Linear: INT-993

### DIFF
--- a/src/band/client/streaming/client.py
+++ b/src/band/client/streaming/client.py
@@ -15,9 +15,16 @@ from phoenix_channels_python_client.client import (
 from phoenix_channels_python_client.client_types import ReconnectPolicy
 from phoenix_channels_python_client.exceptions import PHXConnectionError
 from phoenix_channels_python_client.phx_messages import PHXMessage
-from pydantic import BaseModel, ConfigDict, ValidationError, model_validator
+from pydantic import (
+    BaseModel,
+    ConfigDict,
+    ValidationError,
+    field_validator,
+    model_validator,
+)
 
 from band.client.streaming.errors import classify_initial_upgrade_error
+from band.core.delegation import DelegationEnvelope, parse_delegation_value
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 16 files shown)


AlexanderZ-Band:
Closing as part of the abandoned delegation PR stack (#538–#541): no activity since August and the branch conflicts with the current main.

PR #542, opened 2026-08-14
fix: bump agent-client-protocol and bind ACP handlers by name

AlexanderZ-Band:
## Summary

`agent-client-protocol` 0.11.0 removed `session/set_model`, `SetSessionModelRequest` and `SetSessionModelResponse`. `src/band/integrations/acp/server.py` imports `SetSessionModelResponse`, so any bump past the `>=0.9.0` pin fails at import.

## Changes

**Dependency**

- `agent-client-protocol` `>=0.9.0` -> `>=0.11.0`. Lock resolves to 0.12.0.
- Left uncapped. `test_server_routing.py` dispatches every method through the SDK router, so a future bump that removes a method or changes a handler signature fails CI.

**Removed session-model support**

- Deleted `ACPServer.set_session_model` and the `SetSessionModelResponse` import.
- Deleted `BandACPServerAdapter.set_session_model` and `_session_models`. Nothing read the dict; it was written by the handler and popped in `on_cleanup`.

**Handler binding**

The 0.11.0 schema bump also reordered `Agent` protocol parameters and added `additional_directories`. `ACPServer` subclassed `acp.Agent`, so it had to mirror those positional signatures, and `pyrefly` flagged seven `bad-override` errors after the bump.

The mirroring was unnecessary. The SDK router resolves handlers with `getattr` and invokes them as `func(**request_model_fields)` (`acp/router.py:94-107`) — it binds by name, never by position.

- Dropped the `acp.Agent` base. Handlers declare only the fields they read, keyword-only, with `**kwargs` absorbing the rest, so an upstream reorder or added field cannot break dispatch.
- Removed the `additional_directories` placeholder parameters and the `del` statements that went with them.
- This also fixes a defect the base class introduced: an unimplemented method resolved to the protocol's inherited stub, so the router answered the request with a null result instead of `method_not_found`. Without the base, `getattr` returns `None` and the route reports `method_not_found`.
- `run_agent`'s parameter is typed `Agent`, so there is one `cast` at that call site.

**`run_acp_server`**

`ACPServer` implements `fork_session`, `resume_session` and `close_session`, and returns `SessionCapabilities(list, resume, fork)` from `initialize`. The SDK registers `session/fork`, `session/resume` and `session/close` as unstable routes: `build_agent_router` returns `method_not_found` for them unless `use_unstable_protocol=True`. None of the eight `run_agent(server)` call sites set it.

- Added `run_acp_server()` in `integrations/acp/server.py`, which calls `acp.run_agent` with the flag set.
- Migrated the CLI, the `__init__`/`server_adapter` docstrings and the four `examples/acp/servers/` scripts.

**Test coverage**

Existing server tests call handlers directly with hand-written keywords, so they cannot exercise the router's name-based binding. The client side has router-level coverage via `acp_toolkit/harness.py`; the server side had none.

- Added `tests/integrations/acp/test_server_routing.py` (22 tests). Dispatches by JSON-RPC name through `build_agent_router` with camelCase wire payloads. Only the REST boundary is faked (`mock_rest_client`); the adapter runs its real session bookkeeping and sessions come from a real `create_session()`.
- Assertions are on observable outcomes: room creation and session registration, session removal on close, fork producing a distinct session, mode recorded, prompt text reaching the room message, pending prompt cleared on cancel. Expected dispatch results are the router's real output, including the `normalize_result` serialization applied to five routes.
- Also asserts the unstable routes return `method_not_found` without the flag, that `run_acp_server` sets it, and that the payload table covers every handler `ACPServer` defines.
- Repointed the `test_cli.py` patch target from `acp.run_agent` to `band.integrations.acp.server.run_acp_server`.

## Related Issues

Relates to [INT-1207](https://linear.app/thenvoi/issue/INT-1207/bump-agent-client-protocol-breaks-import-setsessionmodelresponse)

## Testing

- [x] `uv run pytest tests/ --ignore=tests/integration/ --ignore=tests/e2e/` — 4574 passed, 111 skipped (+22)
- [x] `uv run pytest --markdown-docs $(git ls-files '*.md' ':!:examples/*')` — 59 passed
- [ ] Integration tests — not run, no live platform in this environment

## Checklist

- [x] Documentation updated as needed — no user-facing doc referenced the old serving pattern; `AGENTS.md` key-files row names `run_acp_server`


src/band/adapters/__init__.py
```diff
@@ -56,6 +56,7 @@
         ACPClientAdapter as ACPClientAdapter,
         ACPServer as ACPServer,
         BandACPServerAdapter as BandACPServerAdapter,
+        run_acp_server as run_acp_server,
```
AlexanderZ-Band: Why? Not needed

src/band/integrations/acp/server.py
```diff
@@ -160,12 +163,15 @@ async def new_session(
...
             cwd: Working directory from the editor.
+            additional_directories: Extra workspace directories the editor
+                grants access to. Not used by the Band adapter.
             mcp_servers: Optional MCP server configs from the editor.
             **kwargs: Additional keyword arguments.
 
         Returns:
             NewSessionResponse with the session identifier.
         """
+        del additional_directories
```
AlexanderZ-Band: Bad patch let's do it right, no bolt on fix root cause

src/band/integrations/acp/server.py
```diff
@@ -189,11 +196,14 @@ async def load_session(
...
             session_id: The ACP session to load.
             mcp_servers: Optional list of MCP servers from the editor.
+            additional_directories: Extra workspace directories the editor
+                grants access to. Not used by the Band adapter.
             **kwargs: Additional keyword arguments.
 
         Returns:
             LoadSessionResponse if session exists, None otherwise.
         """
+        del additional_directories
```
AlexanderZ-Band: Bad design

src/band/integrations/acp/server.py
```diff
@@ -208,8 +218,8 @@ async def load_session(
 
     async def list_sessions(
         self,
-        cursor: str | None = None,
         cwd: str | None = None,
+        cursor: str | None = None,
```
AlexanderZ-Band: Consider Newark's to avoid such bugs

src/band/integrations/acp/server.py
```diff
@@ -576,3 +565,34 @@ def _extract_text(prompt: list[Any]) -> str:
...
+
+async def run_acp_server(
+    server: ACPServer,
+    input_stream: Any = None,
+    output_stream: Any = None,
+    **connection_kwargs: Any,
+) -> None:
+    """Serve an :class:`ACPServer`, with every route it advertises reachable.
+
+    Drop-in replacement for :func:`acp.run_agent`, and the only supported way
```
AlexanderZ-Band: Keep comments factual and to the point

tests/integrations/acp/test_server_routing.py
```diff
@@ -0,0 +1,154 @@
...
+
+from band.integrations.acp.server import ACPServer, run_acp_server
+from band.integrations.acp.server_adapter import BandACPServerAdapter
+
+# Payloads keyed by ACP method, in the camelCase wire form an editor sends.
+# Every method ACPServer implements is represented.
+WIRE_REQUESTS: dict[str, dict[str, Any]] = {
+    "initialize": {"protocolVersion": 1},
+    "authenticate": {"methodId": "api_key"},
+    "session/new": {"cwd": "/workspace", "mcpServers": []},
```
AlexanderZ-Band: Don't over mock, make sure all tests are meaningful not asserting obvious 
Not asserting assumed results

src/band/integrations/acp/server.py
AlexanderZ-Band: Agreed, `del additional_directories` was a bolt-on. Fixed at the source in 80ef312.

Root cause: `ACPServer` subclassed `acp.Agent`, which forced it to mirror the protocol's positional signatures. The router never calls positionally — `_resolve_handler` looks handlers up with `getattr` and `_make_func` invokes `func(**request_model_fields)` (`acp/router.py:94-107`). So the positional mirroring bought nothing and was the reason an added upstream field had to be absorbed by a placeholder parameter.

Dropped the base class. Handlers now declare only the fields they read, keyword-only, with `**kwargs` taking the rest. The placeholders and `del` statements are gone, and an upstream reorder or added field can no longer break dispatch.

Second defect the base class was hiding: an unimplemented method resolved to the protocol's inherited stub, so the router would answer the request with a null result instead of `method_not_found`. Verified:

```
subclass inherits callable stub: True
stub returns: None
no-base getattr: None
```

Note I tried keeping the base and only making the handlers keyword-only first — pyrefly rejected it with 12 `bad-override` errors, correctly, since keyword-only is not substitutable for positional-or-keyword. Removing the base is what actually resolves it rather than suppressing it. `run_agent`'s parameter is still typed `Agent`, so there is one `cast` at that call with a comment explaining that we satisfy the router's contract and not the nominal protocol.

---
_Generated by [Claude Code](https://claude.ai/code)_

src/band/integrations/acp/server.py
AlexanderZ-Band: Reading this as "consider kwargs" — that is what 80ef312 does, and it removes this bug class.

`list_sessions` had `cwd`/`cursor` declared purely to hold upstream's positional order; it reads neither. It is now `async def list_sessions(self, **kwargs: Any)`. Across the class, handlers declare only what they read, keyword-only, with `**kwargs` absorbing the rest — so a reorder or an added field in the request model is a no-op for us.

If you meant something else, say so and I will redo it.

---
_Generated by [Claude Code](https://claude.ai/code)_

src/band/integrations/acp/server.py
AlexanderZ-Band: Same root cause as the `new_session` thread — fixed in 80ef312. `load_session` no longer declares `additional_directories` at all, so there is nothing to `del`. Details in [that thread](https://github.com/band-ai/band-sdk-python/pull/542#discussion_r3785383200).

---
_Generated by [Claude Code](https://claude.ai/code)_

src/band/adapters/__init__.py
AlexanderZ-Band: Removed in 80ef312. `run_acp_server` is not an adapter, so it does not belong in `band.adapters`. `src/band/adapters/` is now byte-identical to `main`; the examples import it from `band.integrations.acp`.

---
_Generated by [Claude Code](https://claude.ai/code)_

src/band/integrations/acp/server.py
AlexanderZ-Band: Done in c2fdb5c. Rewrote the `run_acp_server` docstring, the `session_capabilities` comment, the dependency comment and the test docstrings to state what the code does and what the SDK requires, dropping the narrative framing. Also removed the AGENTS.md section I had added — the docstring already carried it and tests pin the behavior, so it was a fourth copy of one fact and the only one nothing verifies.

---
_Generated by [Claude Code](https://claude.ai/code)_

tests/integrations/acp/test_server_routing.py
AlexanderZ-Band: Rewritten in 80ef312. You were right on both counts — the old version stubbed the adapter's own methods (`create_session`, `handle_prompt`, `on_cleanup`, `verify_credentials`), so several tests only proved a mock got called, and the parametrized test asserted nothing at all beyond "did not raise".

Now only the REST boundary is faked, via the existing `mock_rest_client` fixture. The adapter runs its real session bookkeeping, and sessions come from a real `create_session()` call rather than hand-seeded dicts.

Assertions are on observable outcomes:

| Test | Asserts |
|---|---|
| `test_new_session_creates_a_room_and_registers_the_session` | room created, session registered, cwd stored |
| `test_close_session_removes_the_session` | session gone afterwards |
| `test_load_session_reports_a_miss_for_unknown_session` | empty result, no session invented |
| `test_fork_session_creates_a_second_session` | new id, original still live |
| `test_set_mode_is_applied_to_the_session` | mode recorded |
| `test_prompt_posts_the_text_to_the_room` | prompt text reaches the room message |
| `test_cancel_notification_releases_the_pending_prompt` | pending prompt cleared |

On "not asserting assumed results": the expected dispatch results are the router's real output, captured by running it rather than by guessing. That caught a wrong assumption of mine — five routes (`session/load`, `session/close`, `session/set_mode`, `session/set_config_option`, `authenticate`) go through the SDK's `normalize_result`, so they return a serialized `{}`, not the response model I had asserted. `session/prompt` is no longer mocked away either; it blocks until the peer replies, so it dispatches as a task and is released via `cancel_prompt`.

22 tests, all passing.

---
_Generated by [Claude Code](https://claude.ai/code)_

amit-gazal-band (approved):

PR #543, opened 2026-08-15
fix: bring Copilot CLI (ACP) to the L4 bar

AlexanderZ-Band:
## Why

Copilot CLI (ACP) integration to the L4 common bar ([INT-1111](https://linear.app/thenvoi/issue/INT-1111/intel-github-copilot-cli-mvp-integration-l0-l4-validation)). Gaps identified by the 2026-08-14 baseline validation run (e2e.yml #72).

## Changes

- **Roster/contacts context**: `ACPClientAdapter` now injects the platform's participants/contacts updates into the ACP prompt as `[System]` lines (previously discarded).
- **Tool-name canonicalization**: Copilot's `<server>-<tool>` MCP spelling (e.g. `band-band_send_message`) is rewritten to the canonical name at tool-call construction, so narration and reply-suppression use one vocabulary (`canonicalize_mcp_tool_name` in `runtime/tools.py`); the legacy band-mcp ≤1.3.1 spelling is included too.
- **Memory capability support**: `ACPClientAdapter` declares `Capability.MEMORY`; declaring it gates the memory tool group on the loopback MCP server. Contact tools stay unconditionally registered, matching the adapter's pre-existing default that every caller without `features=` relies on.
- **Usage N-A**: `test_restart_usage_splits_replay_and_inference` now excludes `copilot_acp` (ACP has no per-turn usage data), consistent with the existing usage-smoke exclusion.
- **CI auth**: the `backends` lane runs `copilot_acp` matrix cells on Anthropic BYOK instead of the Copilot-hosted quota (which was exhausted mid-run in the validation run). One smoke (`test_copilot_hosted_auth_replies`) keeps the Copilot-hosted auth path covered; skips without `GITHUB_TOKEN`.
- **Concurrency fix**: the shared loopback MCP server could be started twice by two rooms' concurrent first turns (leaking the loser); now guarded by a dedicated lock.
- Registry: `copilot_acp` declares `supports=` for the memory/contacts capability matrix; `runs_tool_loop` stays `False` (unchanged from `main`) until the custom-tool round trip has live proof, matching codex/opencode/letta.

## Test plan

- [x] Unit suite green (4573 passed), ruff/pyrefly/markdown-docs clean
- [x] `tests/e2e/baseline/` collection includes the memory/contacts matrix cells for `copilot_acp`
- [x] Full two-OS `backends` lane run against the live platform ([run](https://github.com/band-ai/band-sdk-python/actions/runs/31872432238)): matrix scorecard 14 pass / 4 N-A / 0 fail on both OS, closing the memory/contacts gap
- [x] Scorecard evidence + L0–L4 recorded on [INT-1111](https://linear.app/thenvoi/issue/INT-1111/intel-github-copilot-cli-mvp-integration-l0-l4-validation)

Known residual (tracked on INT-1111, not blocking): `test_copilot_hosted_auth_replies` hits the Copilot-hosted monthly quota (expected — it deliberately tests the non-BYOK path); `test_acp_band_tool_result_is_a_single_clean_payload` has an unrelated content-shape gap, root cause not yet isolated.


src/band/integrations/acp/client_adapter.py
```diff
@@ -225,7 +224,12 @@ async def on_message(
             )
 
         prompt_text = await self._build_prompt_text(
-            room_id, session_id, msg, replay=replay
+            room_id,
+            session_id,
+            msg,
```
AlexanderZ-Band: alwayes prefer kwargs

tests/integrations/acp/test_client_adapter_behavior.py
```diff
@@ -543,6 +543,51 @@ async def test_bootstrap_with_empty_history_has_no_replay_block(fake_agent) -> N
...
+
+    async with acp_adapter(fake_agent) as session:
+        await session.send(
+            "hello",
+            bootstrap=True,
+            participants_msg="## Current Participants\n- @rev — Rev (Agent)",
+            contacts_msg="Contact update: Rev is now a contact",
+        )
+
+    prompt = fake_agent.prompt_texts()[0]
```
AlexanderZ-Band: intent oriented code, hide plumbing use fixtures.
look at our now test style fires then adapte the way we write tests

tests/integrations/acp/test_client_adapter_behavior.py
```diff
@@ -543,6 +543,51 @@ async def test_bootstrap_with_empty_history_has_no_replay_block(fake_agent) -> N
...
+    roster_at = prompt.index("[System]: ## Current Participants")
+    contacts_at = prompt.index("[System]: Contact update")
+    live_at = prompt.index("[Peer]: hello")
```
AlexanderZ-Band: magic strings , we should allow access to the asserted content wrapp it intent oriented style hide plumbing and not assert the obviuse

src/band/integrations/acp/client_adapter.py
```diff
@@ -38,15 +39,23 @@
...
+    iter_tool_definitions,
+)
 
 logger = logging.getLogger(__name__)
 
 LocalMcpServerConfig = HttpMcpServer | SseMcpServer
 
+# The name the loopback Band MCP server registers under; ACP runtimes that
+# prefix MCP tool names key off it (e.g. Copilot's ``band-<tool>``).
+BAND_MCP_SERVER_NAME = "band"
```
AlexanderZ-Band: single source of truth

src/band/integrations/acp/client_adapter.py
```diff
@@ -274,6 +290,7 @@ async def handler(
         ) -> dict[str, object]:
             del kwargs
             call = ACPToolCall.from_acp(tool_call)
+            call = replace(call, name=self._canonical_tool_name(call.name))
```
AlexanderZ-Band: feels hacky, can we fix at root?

src/band/integrations/acp/client_adapter.py
```diff
@@ -274,6 +290,7 @@ async def handler(
         ) -> dict[str, object]:
             del kwargs
```
AlexanderZ-Band: this is a bad patch, we need to adress by solving root cause

src/band/integrations/acp/client_adapter.py
AlexanderZ-Band: single consern functions

src/band/runtime/tools.py
```diff
@@ -772,6 +772,24 @@ def is_room_posting_tool(tool_name: str) -> bool:
     return _matches_tool_name(tool_name, ROOM_POSTING_TOOL_NAMES)
 
 
+def canonicalize_mcp_tool_name(
```
AlexanderZ-Band: feels like a bolton patch, cant we solve at root?

tests/e2e/baseline/toolkit/builders.py
```diff
@@ -377,7 +377,7 @@ def _build_opencode(
     )
 
 
-@adapter(Adapter.COPILOT_ACP, requires=[Dep.COPILOT_CLI], runs_tool_loop=False)
+@adapter(Adapter.COPILOT_ACP, requires=[Dep.COPILOT_CLI], supports=_LLM_TOOL_LOOP)
```
AlexanderZ-Band: why _ private? where is it defined?

tests/integrations/acp/test_client_adapter_behavior.py
```diff
@@ -126,6 +128,63 @@ async def test_text_relayed_alongside_non_posting_tool(fake_agent) -> None:
...
+        raw_input={"title": "war room"},
+        result='{"id": "room-9"}',
+    ).will_say("Created the room.")
+    async with acp_adapter(fake_agent) as session:
+        reply = await session.send("make a room")
+
+    (call,) = reply.tool_calls
+    (result,) = reply.tool_results
+    assert json.loads(call["content"])["name"] == "band_create_chatroom"
+    assert json.loads(result["content"])["name"] == "band_create_chatroom"
```
AlexanderZ-Band: should we create helper fauntion to asset such casses? or is it over engeneering

tests/integrations/acp/acp_toolkit/harness.py
```diff
@@ -171,6 +172,17 @@ def tool_calls(self) -> list[dict[str, Any]]:
     def tool_results(self) -> list[dict[str, Any]]:
         return self._events_of("tool_result")
 
+    @property
+    def tool_call_names(self) -> list[str]:
+        """Narrated tool_call names in order — the canonical vocabulary
+        consumers match on (see ToolCallRoomEvent)."""
+        return [json.loads(e["content"])["name"] for e in self.tool_calls]
```
AlexanderZ-Band: not feeling robust, any chance to leverage pydantic and parse at root? carful not to over engeneer unless rely gain is acheved

tests/integrations/acp/acp_toolkit/harness.py
```diff
@@ -285,14 +300,26 @@ async def _spawn(
...
+
+    The adapter's contract puts it last in every prompt, so tests assert
+    position against this projection instead of hand-building the line.
+    """
+    return f"[{LIVE_SENDER_NAME}]: {content}"
+
+
 def _message(content: str, room_id: str) -> PlatformMessage:
     return PlatformMessage(
         id=str(uuid4()),
```
AlexanderZ-Band: consider uuid7

src/band/integrations/acp/client_adapter.py
AlexanderZ-Band: Fixed in 3e3a9898 — call site is all-kwargs now.

tests/integrations/acp/test_client_adapter_behavior.py
AlexanderZ-Band: Reworked in 3e3a9898: roster fixture comes from the real producer (build_participants_message), assertions run on projections (system_updates / final_line / live_line from the toolkit) instead of index arithmetic.

tests/integrations/acp/test_client_adapter_behavior.py
AlexanderZ-Band: Fixed in 3e3a9898 — the [System]: prefix is now SYSTEM_UPDATE_PREFIX (one spelling, adapter + tests), and Reply grew tool_call_names/tool_result_names so tests stop hand-parsing event JSON.

src/band/integrations/acp/client_adapter.py
AlexanderZ-Band: Moved in 16b03b98 — the constant now lives in integrations/mcp/backends.py, which was already naming the LocalMCPServer "band"; the adapter imports it.

src/band/integrations/acp/client_adapter.py
AlexanderZ-Band: Root-fixed in 16b03b98: ACPToolCall.from_acp takes a canonicalize seam, so the name is canonical at construction — no post-hoc replace at either site.

src/band/integrations/acp/client_adapter.py
AlexanderZ-Band: Same fix as the sibling thread (16b03b98) — canonicalization happens at object construction via from_acp, both here and in the chunk parser.

src/band/integrations/acp/client_adapter.py
AlexanderZ-Band: Split in 16b03b98: _claim_session_bootstrap (once-per-session gate), _system_update_sections (roster/contacts blocks), _framed_replay (nonce-framed replay). _build_prompt_text is assembly only.

src/band/runtime/tools.py
AlexanderZ-Band: Agreed — unified in 16b03b98. is_room_posting_tool and canonicalize_mcp_tool_name now share one resolver (_resolve_mcp_tool_name), so the <server>- convention has exactly one implementation; canonicalize also accepts any server name, same rule suppression always used, and no longer takes a server_name param.

tests/e2e/baseline/toolkit/builders.py
AlexanderZ-Band: _LLM_TOOL_LOOP is defined at the top of this same module (builders.py:37) — (Capability.MEMORY, Capability.CONTACTS), the supports= vocabulary every adapter entry in the file uses. Private because it is module-internal registry shorthand (this line only switched the copilot entry to use it, from the C27-era runs_tool_loop=False). Happy to rename/export if you want it visible outside the registry.

tests/integrations/acp/test_client_adapter_behavior.py
```diff
@@ -44,6 +47,38 @@ def replay_boundary(prompt: str) -> int:
...
+        for line in prompt.splitlines()
+        if line.startswith(SYSTEM_UPDATE_PREFIX)
+    ]
+
+
+def first_line(text: str) -> str:
+    return text.splitlines()[0]
+
+
+def final_line(prompt: str) -> str:
```
AlexanderZ-Band: is it safe?

tests/integrations/acp/test_client_adapter_behavior.py
AlexanderZ-Band: Not over-engineering — done in 3e3a9898: Reply grew tool_call_names/tool_result_names, and in 3f2a41ac they parse through the pydantic room-payload models, so these assertions are now one-liners against a typed projection.

tests/integrations/acp/acp_toolkit/harness.py
AlexanderZ-Band: Done in 3f2a41ac — the projections parse via ToolCallRoomEvent/ToolResultRoomEvent.model_validate_json (the payload's declared contract), so a malformed event fails loudly instead of a KeyError. No new machinery — the models already existed.

tests/integrations/acp/acp_toolkit/harness.py
AlexanderZ-Band: Pushing back on this one: stdlib uuid7 lands in Python 3.14 and this repo's floor is 3.11, so it would mean a third-party dep — and this id is a fake test-message id with no ordering semantics anywhere in the harness (only equality checks). Leaving uuid4.

tests/integrations/acp/test_client_adapter_behavior.py
AlexanderZ-Band: Fair catch — splitlines()[-1] was only safe for single-line fixtures (and raises on an empty prompt). Replaced in 3f2a41ac with closes_the_prompt(prompt, live_line(...)), a suffix check that matches the actual contract: the live message comes last, not "is one line".

src/band/integrations/acp/client_adapter.py
```diff
@@ -108,7 +118,9 @@ class ACPClientAdapter(SimpleAdapter[ACPClientSessionState]):
     """
 
     SUPPORTED_EMIT: ClassVar[frozenset[Emit]] = frozenset()
-    SUPPORTED_CAPABILITIES: ClassVar[frozenset[Capability]] = frozenset()
+    SUPPORTED_CAPABILITIES: ClassVar[frozenset[Capability]] = frozenset(
+        {Capability.MEMORY, Capability.CONTACTS}
+    )
 
     def __init__(
```
AlexanderZ-Band: init is long and holde logic time to refactor
single concern be be carful

.github/scripts/setup-copilot.sh
```diff
@@ -1,15 +1,17 @@
...
+# gate is the CLI plus ANTHROPIC_API_KEY.
 set -euo pipefail
 
-# Fail with a clear message if the token is missing rather than letting a later
+# Fail with a clear message if the key is missing rather than letting a later
 # ACP session fail opaquely. printenv takes the name as an argument, so `set -u`
-# alone wouldn't catch an unset token.
-: "${GITHUB_TOKEN:?GITHUB_TOKEN (from a Copilot-entitled account) is required for the Copilot CLI}"
+# alone wouldn't catch an unset key.
+: "${ANTHROPIC_API_KEY:?ANTHROPIC_API_KEY is required for the Copilot CLI's Anthropic BYOK auth}"
```
AlexanderZ-Band: dont assume we will add it look for what we have we should v=have it inspect secret

src/band/integrations/acp/client_adapter.py
AlexanderZ-Band: Refactored in cfd0d19c: __init__ is assignments only — command shaping, tool-vocabulary computation, transport selection, and runtime construction each moved to a single-concern helper (_shape_command / _registered_tools / _select_transport / _build_runtime). Behavior-preserving; the 67 adapter unit tests pass unchanged.

.github/scripts/setup-copilot.sh
AlexanderZ-Band: Verified against the repo's actual secrets (gh secret list): E2E_ANTHROPIC_API_KEY exists (set 2026-06-30, already mapped at job level for the core lane's Anthropic adapters) and E2E_GITHUB_TOKEN exists (set 2026-07-05 — the mapping this PR restores for the hosted smoke is the same one that existed pre-BYOK). Nothing new to provision.

src/band/runtime/tools.py
```diff
@@ -752,24 +752,42 @@ class ListMyPeersInput(BaseModel):
...
+    legacy spelling arrives as ``band-create_agent_chat_message``). Other
+    spellings (``mcp__server__tool``, ``server.tool``) are not matched — no
+    wired backend uses them. Extend here when such a backend is added.
+    """
     if tool_name in names:
-        return True
-    return tool_name.endswith(tuple(f"-{name}" for name in names))
+        return tool_name
+    for name in names:
+        if tool_name.endswith(f"-{name}"):
```
amit-gazal-band: Non-blocking, but the suffix match is broader than the naming convention it documents: `endswith(f"-{name}")` accepts *any* prefix before the hyphen, not just an MCP server name.

That's harmless for the platform tools (all `band_*`-prefixed, so nothing realistic collides), but `_own_tool_names` also carries custom-tool names derived from the input model via `get_custom_tool_name` — `SearchInput` -> `search`, `EchoInput` -> `echo`. So for a caller that registers a custom tool named `search`, a remote agent's own unrelated `web-search` canonicalizes to `search`: it gets narrated under the Band name and its `tool_result` correlates to the Band tool, misattributing a foreign call in the room record.

Narrow (needs a short generic custom-tool name plus a remote tool that happens to end in `-<that name>`), and the failure is a wrong narration label rather than a wrong call. Anchoring the prefix to the actually-advertised server name — `BAND_MCP_SERVER_NAME` / the configured `mcp_servers` names — instead of accepting any prefix would close it, and would match what the docstring already says the convention is.

src/band/integrations/acp/client_adapter.py
```diff
@@ -507,14 +653,23 @@ async def cleanup_all(self) -> None:
             self._room_to_session.clear()
             self._room_tools.clear()
             self._bootstrapped_sessions.clear()
+        async with self._mcp_backend_lock:
```
amit-gazal-band, line 682: Holding the lock across `stop()` is the right move, but I don't think the residual is quite what the comment describes — it reads as "only the unlocked fast-path read can still race," and the *locked* slow path can too.

There's no shutdown flag, so once `cleanup_all` has fully returned (backend nulled, stopped, lock released, `_runtime.stop()` done), a later `_ensure_band_mcp_backend` takes the lock, sees `None`, and starts a **fresh** `LocalMCPServer` that nothing will ever stop — `cleanup_all` is idempotent but won't be called again. Reachable because `_load_persisted_session` awaits `_session_mcp_servers()` outside `_session_lock`, so room B's bootstrap can be parked there while room A's failing turn drives `on_message` -> `stop()` -> `cleanup_all`.

So it's the same "started, never stopped" leak `_mcp_backend_lock` was added to prevent, just moved from concurrent-creation to post-teardown-creation. A `self._stopped` flag set under `_mcp_backend_lock` and checked in `_ensure_band_mcp_backend`'s locked branch (raise instead of create) would close it properly and make the turn fail loudly, which is what the comment says it wants. Fine to leave for a follow-up given it takes a shutdown racing a first turn to hit — but worth either fixing or describing accurately.

amit-gazal-band (dismissed):
Reviewed the full diff against `main` (src + tests + CI). Ran `tests/integrations/acp` + `tests/runtime/test_tools.py`: 395 passed, 8 skipped.

Approving — the substantive issues from my pass are already handled in `8922052d`, and what's left is narrow. Two inline notes below, neither blocking.

**Verified fixed in `8922052d`:**
- `render_system_prompt`'s `include_base_instructions=False` path dropping the capability-gated `MEMORY_SECTION`/`CONTACT_SECTION`. Fixed at the root in `prompts.py` rather than worked around in the adapter, which is the right call — `ACPClientAdapter` is just the first caller to combine the two unconditionally — and the two tests that had codified the old behavior were correctly flipped. This also makes the new AGENTS.md sentence ("`render_system_prompt` carries the matching capability sections") true, which it wasn't when I first read it.
- The `copilot.cmd`-without-`shell=True` bug in the backends-lane metadata step. Good catch; that would have reddened backends/windows independent of test outcomes.

**Withdrawing** one sub-point from my earlier pass: I'd noted that `_resolve_mcp_tool_name` iterating a `frozenset` makes the resolved name order-dependent when two members both suffix-match. That needs one vocabulary member to be a hyphen-suffix of another, which can't happen with the current tool names. Not worth acting on.

I also agree with the two findings you declined. Re-gating contact tools behind `Capability.CONTACTS` was already tried and reverted in `9f57c958` for a good reason, and `_claim_session_bootstrap` is genuinely atomic by construction — no `await` between the check and the add — which its docstring already states precisely.

Rest of the diff checked out: lock ordering is consistent (`_session_lock` -> `_mcp_backend_lock`, no inversion), the `[System]:` prefix matches codex/opencode/anthropic parity, `features` is forwarded correctly by `_build_copilot_acp`, and `Dep.ANTHROPIC` being a default-lane dep keeps `adapter_lane` resolving `copilot_acp` to `backends`.

src/band/integrations/acp/client_adapter.py
AlexanderZ-Band: Confirmed and fixed in 37b819f8. Added a self._stopped flag exactly as suggested, checked under _mcp_backend_lock in _ensure_band_mcp_backend -- raises loudly instead of creating.

One wrinkle caught while implementing the naive version of this: cleanup_all()/stop() isn't only the final-shutdown hook -- on_message's except-handler also calls it to recover a wedged turn, and expects a *later* turn on any room to self-heal by respawning both the ACP connection and the MCP backend from scratch. A flag set unconditionally would have turned every transient turn failure into a permanently bricked adapter, which is worse than the leak being fixed.

So cleanup_all(self, *, final: bool = True) now distinguishes the two callers: Agent.stop() and the test harness get the default (blocks recreation, matches your ask), while the on_message error path explicitly passes final=False and keeps self-healing. Two new regression tests cover both directions -- one fails on the pre-fix code (DID NOT RAISE), one guards against re-breaking the recovery path.

src/band/runtime/tools.py
AlexanderZ-Band: Agreed this is real, and agreed non-blocking -- leaving as documented rather than anchoring to BAND_MCP_SERVER_NAME. Anchoring would mean threading the actual server name through _resolve_mcp_tool_name's callers (is_room_posting_tool has none available at most call sites today), for a collision that needs a short generic custom-tool name plus a remote tool ending in -<that name>, not reachable with any custom tool registered today. Cost of the fix outweighs the benefit at current scope; the docstring already states the exact boundary (miss-only, no wrong-post) so a future caller adding a collision-prone custom tool name has something to trip over. Leaving open as a follow-up rather than re-litigating here.

AlexanderZ-Band: Tracked as a follow-up: https://linear.app/thenvoi/issue/INT-1222/acp-tool-name-canonicalization-matches-any-hyphen-prefix-not-just-the

Includes the concrete custom-tool-name collision scenario and why anchoring to the real server name is the right fix but not worth doing at current scope (no colliding tool name exists today). Resolving this thread since it's now tracked rather than silently dropped.

AlexanderZ-Band: Update: this is now actually fixed, superseding my earlier "leaving as documented" comment above.

A later review pass on the same code surfaced the sharper consequence: `is_room_posting_tool` uses the identical unanchored resolver, so an unrelated MCP server's own tool ending in `-band_send_message` (not just a custom-tool suffix collision) would get treated as a Band room-posting tool and silently suppress the assistant's real text reply -- reachable with zero custom tools, independent of `inject_band_tools`. That raised this past "non-blocking, narrow" into a real fix in `b5f37fe1`.

`_resolve_mcp_tool_name` now anchors specifically to `BAND_MCP_SERVER_NAME` (`"band"`) instead of accepting any hyphen-delimited prefix -- one change in the shared resolver, no per-call-site plumbing needed after all (confirmed via codex/copilot_sdk/opencode's full test suites: none of them exercise a non-`band` prefix, so no behavior change there). This also closes the original custom-tool-collision scenario from this thread (`web-search` no longer resolves to a custom `search` tool) as a side effect.

INT-1222 marked Done rather than left open for a fix that already landed.

amit-gazal-band (approved):

PR #544, opened 2026-08-16
fix: surface errored and silent Claude SDK turns as room errors

AlexanderZ-Band:
## Summary
- `ClaudeSDKAdapter._process_response` never checked `ResultMessage.is_error`, so a Claude CLI auth failure (`is_error=True`, `subtype="success"`) logged "Complete" and returned normally — no room error, no exception, delivery recorded as successful.
- A clean turn that never called a Band tool also went silent (no reply, no error).
- Fix: gate on `is_error` (not `subtype`, which is misleading — reports "success" on the reproduced auth failures) and surface `result`/`errors`/`api_error_status` as a room `error` event. Wire the existing `missing_reply_error()` helper for the residual "completed but posted nothing" case (already used by strands/pydantic_ai/crewai; claude_sdk was never wired up).
- The missing-reply guard's "did a tool call succeed" tracking keyed off `ToolResultBlock` inside `AssistantMessage`, but the protocol delivers tool results in **user-type envelopes** — that branch was dead in production. Tool results from both envelopes now route through one shared handler, which also revives the never-firing tool_result narration.
- A background code review then found three more real bugs in this diff itself, each verified independently (confirmed to reproduce, then confirmed the regression test fails against the bug and passes against the fix):
  - Custom terminal-tool opt-in (`band_terminal=True`) was keyed by the Python handler's `__name__`, but the tool is actually registered under `get_custom_tool_name(input_model)` — the two routinely differ, so it never matched.
  - A declined tool call (manual `/decline`, `auto_decline`, or an approval timeout) already posts its own room-visible decline notice, but the new missing-reply guard didn't know that and piled a second, contradictory "no reply" error on top. A per-room "declined this turn" flag now suppresses the guard for that turn.
  - `_on_user_message` only handled `ToolResultBlock`; the vendored SDK's own message parser also allows a `tool_use` block inside a `"user"` envelope (e.g. a subagent's nested call), which was silently dropped. Both envelope handlers now route `ToolUseBlock` through one shared `_on_tool_use`.
- Also cleaned up: `tool_executed` renamed to `replied_this_turn` (it tracked "the turn already has an answer," not "a tool ran"); the three `Emit`-gated `send_event` blocks collapsed into one helper; the `"claude_sdk_session_id"` metadata key (written by the adapter, read back by the converter) pulled into one `SESSION_ID_METADATA_KEY` constant instead of being retyped in both places.

## Verification
- Live, before/after against the real Claude CLI: isolated `HOME` + no API key forces a genuine "not logged in" auth failure. Pre-fix code posts only the session task event (silent). Post-fix code additionally posts `Error: Claude SDK turn failed: Not logged in · Please run /login` — matching the ticket's own production reproduction (`is_error=True`, `subtype="success"`).
- Every one of the 4 new regression tests (from both the initial fix and the code-review fallout) was checked to fail when its corresponding fix is reverted, and pass with it restored — not just added and left green.

## Test plan
- [x] `uv run pytest tests/adapters/test_claude_sdk_adapter.py tests/converters/test_claude_sdk.py -q --no-cov` — 125 passed
- [x] `uv run pytest tests/ --ignore=tests/integration/ --ignore=tests/e2e/ -q --no-cov` — 4565 passed, 111 skipped
- [x] `uv run ruff check .` / `uv run ruff format --check .` / `uv run pyrefly check` (only pre-existing, unrelated errors remain — confirmed via diff against main)

Fixes INT-1210.

https://claude.ai/code/session_01VF7wM1rU36vm217o9zkyWD


src/band/adapters/claude_sdk.py
```diff
@@ -683,127 +720,320 @@ async def on_message(
...
+        # The ResultMessage carries the turn's total usage (the SDK runs
+        # its own tool loop internally, so this is already aggregated).
+        await self.emit_usage(tools, self._usage_from_result(sdk_message))
+        # A denied tool call already posted its own room-visible decline
+        # notice (see _resolve_tool_permission) — that is the turn's answer
+        # for why nothing else happened, so it must not also trigger the
+        # missing-reply guard below. Read-only: _process_response's finally
+        # owns clearing this marker, since it must fire on every exit path,
+        # not just this one.
+        declined = self._declined_tool_this_turn.get(room_id, False)
```
amit-gazal-band: **The decline suppression is turn-global, not scoped to the declined tool.**

None of the three set-sites (`:1161`, `:1270`, `:1293`) checks *which* tool was denied, so any decline suppresses the guard.

Scenario: the user asks a question, the model tries `Bash`, the user `/decline`s it, and the model then answers in **plain text** without `band_send_message`. `replied_this_turn` is False and `declined` is True, so the guard is skipped at :947 and the answer is silently dropped — the silent turn this PR exists to fix.

The room isn't literally silent (`_handle_approval_command` posts "Approval `t` resolved as **decline**"), but that notice explains the declined `Bash` call, not the unanswered question — so the premise in the comment above ("that is the turn's answer for why nothing else happened") only holds when the declined tool was the one that would have posted the reply.

Suggest scoping suppression to declines of a room-posting/terminal tool. `ResultMessage.permission_denials` is available at the pinned floor if you'd rather drive it off the result message than off the callback.

src/band/adapters/claude_sdk.py
```diff
@@ -1006,23 +1267,30 @@ async def _resolve_manual_approval(
...
                     logger.debug(
                         "Room %s: Failed to send timeout notification", room_id
                     )
 
             if decision == "accept":
                 return PermissionResultAllow()
+            # Suppressing the missing-reply guard requires a delivered notice:
+            # a timeout nobody heard about must still surface as an error.
+            if notified:
+                self._declined_tool_this_turn[room_id] = True
```
amit-gazal-band: **This can set the flag *after* `_process_response`'s `finally` already cleared it, leaking a decline into the next turn** (same applies to `:1270`).

`can_use_tool` doesn't run inside the receive loop: the SDK dispatches it from its own task group (`claude_agent_sdk/_internal/query.py`: `self._tg.start_soon(self._handle_control_request, request)`), independent of the consumer's `receive_response()` iteration. The client is cached across turns, so that task outlives the turn that spawned it.

So if a turn ends while an approval is pending (stream dies, an exception mid-receive, or an abandoned call), `_process_response` pops the marker at :779, and up to `approval_wait_timeout_s` (default 300s) later this line sets it again. The room's next, unrelated turn then has its missing-reply guard wrongly suppressed — precisely what that `finally` was written to prevent.

Making the marker turn-scoped would close it: capture a turn token at turn start and only honor a set from the matching turn (or hang the marker off a per-turn object the approval captures by reference).

src/band/adapters/claude_sdk.py
```diff
@@ -683,127 +720,320 @@ async def on_message(
...
+            "Room %s: Tool result: %s... error=%s",
+            room_id,
+            block.tool_use_id[:20],
+            block.is_error,
+        )
+        result_tool_name = pending_tool_names.pop(block.tool_use_id, None)
+        await self._send_narration_event(
+            tools,
+            gate=Emit.EXECUTION,
+            content=lambda: json.dumps(
```
amit-gazal-band, line 1035: **This payload omits `ToolEventKey.NAME` and `IS_ERROR`, which `parse_tool_result` requires.**

`src/band/converters/parsing.py` drops any tool_result payload without a `name` (logs "Skipping tool_result with missing name") and defaults `is_error` to False. Every sibling adapter sets `NAME` on both events — anthropic, codex, langgraph, gemini, copilot_sdk, letta, agno, google_adk — and `_narrate_tool_call` sets it here too at :908; only this result payload leaves it out.

The omission predates the PR, but it was unreachable before: results arrive in user envelopes, and the old code only handled `ToolResultBlock` under `AssistantMessage`. Now that `_on_user_message` routes them here, these events actually post for the first time, so the consequence is new.

Impact with `Emit.EXECUTION` on: in a mixed-framework room, every result event this adapter posts is discarded on replay by parsing.py-based converters, leaving the `tool_call` records unpaired (anthropic then has to patch them via `_patch_orphaned_tool_uses`). The Claude SDK's own converter passes the JSON through raw, so it survives there — but without `IS_ERROR` it can't tell a failed call from a successful one either way.

`_on_tool_result` already has `result_tool_name` in hand one line below; passing it plus `block.is_error` into the payload fixes both.

amit-gazal-band (changes requested):
The core fix is sound — gating on `ResultMessage.is_error` rather than `subtype` is the right signal, routing results through the user-envelope branch fixes a genuinely dead path, and the `>=0.1.81` floor checks out (I verified `errors` and `api_error_status` are both present in that release).

Three issues below. Two of them leave the exact failure mode this PR exists to close still reachable, and the third means the tool_result events this PR newly makes live are discarded downstream. All look like small fixes.

src/band/adapters/claude_sdk.py
AlexanderZ-Band: Fixed in 710428e9. The decline record is now keyed by the tool that was declined, not a room-wide bool: `_on_turn_complete` only suppresses the missing-reply guard when the denied tool is one whose success would have counted as terminal work (`is_terminal_success`). Declining Bash (or another side tool) while the model then answers in plain text still fires the guard — covered by the new `test_declined_side_tool_still_reports_missing_reply`.

src/band/adapters/claude_sdk.py
AlexanderZ-Band: Fixed in 710428e9, but not by scoping the marker to a turn token — went with your suggested alternative instead, driving this off `ResultMessage.permission_denials`. Verified live against the real CLI (bundled `@anthropic-ai/claude-agent-sdk` `.d.ts` confirms the shape: `{tool_name, tool_use_id, tool_input}`) that it's populated per-turn and never carries a denial from a different turn — a two-turn probe with a decline in turn 1 showed `permission_denials=[]` on turn 2's ResultMessage. Since it's the CLI's own authoritative record, there's no local turn-scoped state left to race: the only thing still tracked in the adapter is whether a decline notice actually reached the room, keyed by `tool_use_id` (globally unique, so a late write from a decision that resolves after its turn ended can only ever match that same turn's `permission_denials`, never a later turn's).

src/band/adapters/claude_sdk.py
AlexanderZ-Band: Fixed in 710428e9 — `_on_tool_result`'s payload now sets `ToolEventKey.NAME` (the already-resolved `result_tool_name`) and `ToolEventKey.IS_ERROR` (`block.is_error`) alongside `OUTPUT`/`TOOL_CALL_ID`. Added `test_tool_result_payload_includes_name_and_is_error` asserting the JSON body against a real declined-tool result.

amit-gazal-band (approved):
All three addressed — thanks.

- **Decline scoping**: `_declined_the_reply` now requires the denial to be both room-notified *and* on a tool that would have delivered the reply (`is_terminal_success`, MCP prefix stripped). The declined-side-tool case reports missing-reply again.
- **Cross-turn leak**: replacing the room-level bool with a set of `tool_use_id`s cross-referenced against `ResultMessage.permission_denials` is a better fix than the turn token I suggested — a stale id from a turn that died mid-approval simply can't match a later turn's denials, which is what makes dropping the `finally` safe. Nice that the test reproduces the dead-stream sequence rather than poking the marker directly.
- **tool_result payload**: `NAME` + `IS_ERROR` present, with the `parse_tool_result` dependency named in a comment.

Also picked up the approval-eviction gap (`_FORCED_DECLINE`), which I hadn't flagged.

Verified locally at 482701c0: full unit suite passes (the only failure is a local asdf `node` shim returning 126 in `desktop_app/test_protocol.py`, unrelated and green in isolation), ruff clean, pyrefly clean on the adapter. I also confirmed `ToolPermissionContext.tool_use_id` exists at the pinned `0.1.81` floor, so the new dependency doesn't need a floor bump.

Two non-blocking notes for whenever you're next in here:

1. `ToolEventKey.NAME: result_tool_name` is `None` when the `tool_use` wasn't observed — a result replayed by a resumed session that misses `pending_tool_names`. `parse_tool_result` still drops those, so the replay case that room-scoped map exists for isn't fully covered by the fix.
2. `_declined_the_reply` reads `tool_name`/`tool_use_id` off entries the SDK types `list[Any]` and passes through as raw CLI JSON (true at both 0.1.81 and 0.2.139), so those key names rest on your live check and `_denial` encodes the same assumption. Non-silent if it ever drifts (a spurious missing-reply error beside the decline notice), and the malformed-entry test proves it won't crash — just worth a comment pointing at what verified it.

PR #545, opened 2026-08-16
fix: Parlant tool descriptions drift from SDK master models

AlexanderZ-Band:
## Summary
- Parlant hand-wrote its 12 tool docstrings instead of reading the master Pydantic models in `runtime/tools.py`, so a master reword never reached Parlant agents — it had already drifted (stale unprefixed `add_participant(...)`/`remove_participant(...)` example calls for tools actually named `band_add_participant`/`band_remove_participant`).
- Adds a `band_tool` decorator that sets each function's `__doc__` from `get_tool_description(func.__name__)` before registering with `p.tool` — the tool name is never retyped as a string (it's just `func.__name__`, always written to match its `TOOL_MODELS` entry), and decorator syntax is preserved.
- Only genuine Parlant-specific addition kept: a mentions-format note on `band_send_message` (Parlant takes a comma-separated string; the master model describes a list).
- Adds a regression test (`test_descriptions_match_master_source`) asserting every Parlant tool description starts with its master `get_tool_description()` text.

Fixes [INT-1170](https://linear.app/thenvoi/issue/INT-1170/parlant-tool-descriptions-are-hand-written-and-drift-from-the-sdk).

## Test plan
- [x] `uv run --extra dev-parlant pytest tests/integrations/parlant/test_tools.py -v --no-cov` — 39 passed (real Parlant SDK, not mocked)
- [x] `uv run ruff check` / `uv run ruff format --check`
- [x] `uv run pyrefly check`

https://claude.ai/code/session_0114NCg97KJ6mNNK5RmvsbpH

### DIFF
--- a/src/band/integrations/parlant/tools.py
+++ b/src/band/integrations/parlant/tools.py
@@ -22,14 +22,16 @@ NOTE: We intentionally do NOT use `from __future__ import annotations` here
 because Parlant's @p.tool decorator checks annotation types at runtime.
 """
 
+import inspect
 import json
 import logging
 import warnings
-from typing import Any, Callable, Optional
+from typing import Annotated, Any, Callable, Optional
 
 from band.core.exceptions import BandToolError
 from band.core.types import AdapterFeatures, Capability
 from band.runtime.tools import (
+    TOOL_MODELS,
     append_available_mention_handles,
     get_tool_description,
     serialize_tool_result,
@@ -53,6 +55,14 @@ _session_message_sent: dict[str, bool] = {}
 SEND_MESSAGE_MENTIONS_NOTE = (
     "\n\nThis tool's mentions argument is a single string: separate multiple "
     'handles with commas, e.g. "@alice, @bob/agent".'
+)
+
+# Same divergence as SEND_MESSAGE_MENTIONS_NOTE, appended to the per-argument
+# description instead of the tool-level one: the master field's list-oriented
+# text would otherwise reach the LLM unqualified for this comma-separated param.
+SEND_MESSAGE_MENTIONS_PARAM_NOTE = (
+    " This tool takes it as a single comma-separated string, not a list, "
+    'e.g. "@alice, @bob/agent".'
 )
 
 # The master model describes lookup_peers' raw return shape (a 'data'/'metadata'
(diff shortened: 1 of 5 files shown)


src/band/integrations/parlant/tools.py
```diff
@@ -592,28 +551,13 @@ async def band_remove_contact(
             logger.error("[Parlant Tool] Error removing contact: %s", e, exc_info=True)
             return ToolResult(data=f"Error removing contact: {e}")
 
-    @p.tool
+    @band_tool()
     async def band_list_contact_requests(
         context: ToolContext,
         page: int = 1,
         page_size: int = 50,
         sent_status: str = "pending",
```
amit-gazal-band, line 585: `sent_status`'s allowed values still never reach the LLM. The master field is `Literal["pending", "approved", "rejected", "cancelled", "all"]` (`src/band/runtime/tools.py:336`), but the Parlant parameter is a bare `str`, so `band_tool` propagates only the `Field` description — which doesn't enumerate them. The constructed schema on this head:

```
{'type': 'string', 'description': 'Filter sent requests by status'}
```

No enum, no values in the prose. A model that passes `"open"` or `"PENDING"` sends it straight through to `tools.list_contact_requests` and gets a backend error, with nothing in the tool surface to have prevented it. This is the same class of bug the PR set out to fix — the master model knows the vocabulary and the LLM doesn't — so it's worth closing here rather than leaving the one typed field as the exception.

Annotating the parameter with the master's `Literal` (rather than `str`) would let Parlant build the enum into the schema and keep the single source of truth; failing that, folding the literal values into the description via `param_overrides` would at least get them in front of the model.

amit-gazal-band (changes requested):
Reviewed the diff and ran it live: booted `examples/parlant/01_basic_agent.py` against a real Band platform with parlant 3.3.2 and drove it as a user through three turns (greet + `band_get_participants`; `band_lookup_peers` + `band_add_participant`; membership re-check). It works — the agent replies, completes multi-tool turns, and the add-participant really landed. `tests/integrations/parlant/`: 44 passed on this head.

Most of what I found on the previous head (`4e32a5b2`) — every parameter reaching the LLM with `description=None`, the lost `@john/agent-name` handle format on `band_add_contact`, the lost `identifier` guidance on `band_add_participant`, the lost either/or hints on the contact tools — is fixed by `4a0b5794`. I re-dumped every constructed `ToolEntry` on this head and confirmed each one now carries real per-argument text. Nice fix, and the `Annotated[...]`-wrapping approach keeps the single-source-of-truth property intact.

One gap survives it, inline below. Everything else from my earlier pass is resolved.

One thing outside this diff, for a follow-up: `examples/parlant/01_basic_agent.py:89` and `:110` instruct the agent to call `band_add_participant` / `band_remove_participant` "with the **name** parameter" — both take `identifier`. In my live run the model dropped the argument once (`[ToolCaller] Inference::Completion: Argument 'identifier' is missing`) before recovering, so it costs a wasted inference round rather than breaking the turn. The restored `identifier` description makes it much less likely; the guidelines are still worth correcting.

src/band/integrations/parlant/tools.py
AlexanderZ-Band: Fixed in 048ac466. You're right that a bare `Literal[...]` annotation is the more direct read of "give Parlant the master's real type" — but I checked `parlant.core.services.tools.plugins._describe_parameters` directly: it only turns a real `enum.Enum` subclass into a JSON Schema `enum` (`issubclass(param_type, enum.Enum)`); a bare `Literal[...]` falls into the "best-effort string" branch, which does `resolved_type.__name__ != "list"` unconditionally — and `Literal[...].__name__` is the string `'Literal'`, so it raises `Exception("Only \`list\` is supported as a generic container in parameters")`. Confirmed by actually trying it against a live `dev-parlant` venv (parlant 3.3.2) before ruling it out.

So instead: the master field's `Literal` choices are now folded into the parameter description as prose (`_literal_choices` in `src/band/integrations/parlant/tools.py`), same mechanism used for every other divergence in this file. Live re-dump on this head:

```
sent_status {'type': 'string', 'description': 'Filter sent requests by status One of: pending, approved, rejected, cancelled, all.'}
```

Also closed the single-source-of-truth gap one level up: `sent_status`/`status` was independently hand-typed as the identical inline `Literal["pending", "approved", "rejected", "cancelled", "all"]` in *two* master models (`ListContactRequestsInput` and `ListSentContactRequestsInput`, `src/band/runtime/tools.py`). Factored both into one `ContactRequestSentStatus` alias in `band.core.types` (same pattern as the existing `EventMessageType`), so there's one definition instead of two copies that could silently drift apart.

Added `test_list_contact_requests_sent_status_description_lists_literal_choices` (`tests/integrations/parlant/test_tools.py`) asserting the master's literal choices reach the built parameter description — derives the expected choices from `ListContactRequestsInput.model_fields`, not hardcoded, so it can't pass on stale text. Full `tests/integrations/parlant/` (46 now) + `tests/runtime/` + `tests/core/` green on `dev-parlant`; `ruff check/format` and `pyrefly check` clean on the touched files.

amit-gazal-band (approved):

PR #550, opened 2026-08-17
fix: tool text drifts from the SDK master across adapters

AlexanderZ-Band:
## Summary

- `runtime/tools.py` is meant to be the single source of truth for platform tool text (docstring = tool description, `Field(description=...)` = arg description). CrewAI hand-duplicated all 17 input models with independently written descriptions — measured drift on 14 of 42 fields, not the single instance INT-1171 named. pydantic_ai duplicated nothing but silently dropped every per-argument description, since pydantic-ai's griffe parser only reads `Args:` sections and the adapter supplied bare type hints.
- Adds two factories in `runtime/tools.py` that make the mistake structurally hard to make: `platform_args_schema(name, validators=...)` returns the master input model (optionally with adapter-only parsing leniency layered on via `create_model`), and `platform_tool(name)` decorates a function with the master docstring plus a rendered `Args:` section. Neither accepts description text.
- CrewAI's 17 duplicated model classes are deleted; both adapters rewired onto the factories.
- Restores INT-336's lenient `mentions` normalizer (`list[str]` + JSON-string/bracketed-handle coercion), silently reverted to a JSON-string field by a later refactor.
- Adds a registry-driven guardrail (`test_tool_text_drift.py`) that fails if an adapter's advertised schema diverges from the master — wired for pydantic_ai and CrewAI, the two adapters that build their own schema objects.
- Fixes a real gap found in review: `band_lookup_peers.page`/`page_size` lost their `ge=1` lower bound when CrewAI's locally-bounded copy was deleted in favor of the (unbounded) master — fixed on the master model, closing it for every adapter.
- Parlant already had this bug and is already fixed by #545 (INT-1170, separate PR, currently open) — not touched here.

## Behavior change

`band_send_event.message_type` is now **required** for CrewAI instead of defaulting to `"thought"`, matching every other adapter and the master model. A CrewAI agent that omits it now gets a validation error instead of a silently mislabeled event.

## Test plan

- [x] `uv run pytest tests/ --ignore=tests/integration/ --ignore=tests/e2e/` — 4559 passed
- [x] `uv run pytest tests/integrations/test_crewai_real_tools.py ...` against the real `crewai` package (`--extra dev-crewai`) — 145 passed
- [x] `uv run ruff check . && uv run ruff format --check . && uv run pyrefly check` — clean
- [x] `uv run pytest --markdown-docs $(git ls-files '*.md' ':!:examples/*')` — 61 passed
- [x] Verified all 17 pydantic_ai tool schemas now match the master exactly (0/42 mismatches; was 42/42 before the fix)
- [x] Verified the new CrewAI probe catches injected drift (re-declaring a field description on the local subclass fails the guardrail)

https://claude.ai/code/session_019mZ9meinC1E4WPLzNjY2LE

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -12,6 +12,37 @@ This is a Python SDK that connects AI agents to the Band collaborative platform.
 
 ## Platform Tools
 
+### Tool text is never written in an adapter
+
+`src/band/runtime/tools.py` owns every word an LLM reads about a platform tool:
+the input model's class docstring is the tool description, and each
+`Field(description=...)` is an argument description. An adapter must reach for
+whichever of these fits its framework instead of retyping the text:
+
+| Framework wants | Use | Result |
+|---|---|---|
+| A Pydantic `args_schema` class | `platform_args_schema(name)` | the master model itself |
+| The same, but its tool layer emits a value the master won't parse | `platform_args_schema(name, validators={...})` | a subclass with the master's text plus the extra validators |
+| Schema derived from a function docstring | `@platform_tool(name)` | docstring = master description + a rendered `Args:` section |
+| A raw JSON/dict schema | `iter_tool_definitions()`, `get_openai_tool_schemas()`, `get_anthropic_tool_schemas()` | built live from the master |
+
+None of these accept description text — an adapter that needs different wording
+has a modeling problem to fix on the master, not a local string to write. If a
+framework's leniency really is adapter-specific (CrewAI's `mentions` coercion),
+express it as a `validators=` entry, never as a re-declared field.
+
+```python
+from band.runtime.tools import SendMessageInput, platform_args_schema
+
+assert platform_args_schema("band_send_message") is SendMessageInput
+```
+
+Guardrail: `tests/framework_conformance/test_tool_text_drift.py` runs each
+`AdapterConfig.advertised_arg_text` probe and fails if what the adapter
+advertises differs from the master. Wire the probe up for a new adapter that
+builds its own schema objects; leave it `None` when the master schema is passed
+through untouched.
+
 ### Chat Tools
 - `band_send_message`: Send message to chat room (requires mentions)
 - `band_send_event`: Send non-message event (thought, error, task)
@@ -601,7 +632,7 @@ When adding a new framework adapter and converter, follow this TDD workflow. Use
 
 1. Add an output adapter in `tests/framework_configs/output_adapters.py` — choose base class matching output format (`BaseDictListOutputAdapter`, `StringOutputAdapter`, `SenderDictListAdapter`, or `LangChainOutputAdapter`).
 2. Register converter config in `tests/framework_configs/converters.py` — factory function, builder function returning `ConverterConfig` with behavioral flags, append to `_CONVERTER_CONFIG_BUILDERS`.
-3. Register adapter config in `tests/framework_configs/adapters.py` — factory function with mocked constructor args, builder function returning `AdapterConfig`, append to `_ADAPTER_CONFIG_BUILDERS`.
+3. Register adapter config in `tests/framework_configs/adapters.py` — factory function with mocked constructor args, builder function returning `AdapterConfig`, append to `_ADAPTER_CONFIG_BUILDERS`. If the adapter builds its own tool schema objects, also set `advertised_arg_text` (see [Tool text is never written in an adapter](#tool-text-is-never-written-in-an-adapter)).
 
 ### Phase 3: Run Conformance Tests (Expect Failures)
 
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -23,7 +23,7 @@ whichever of these fits its framework instead of retyping the text:
 |---|---|---|
 | A Pydantic `args_schema` class | `platform_args_schema(name)` | the master model itself |
 | The same, but its tool layer emits a value the master won't parse | `platform_args_schema(name, validators={...})` | a subclass with the master's text plus the extra validators |
-| Schema derived from a function docstring | `@platform_tool(name)` | docstring = master description + a rendered `Args:` section |
+| Schema derived from a function docstring | `@platform_tool` (bare — reads `fn.__name__`, takes no name argument) | docstring = master description + a rendered `Args:` section |
 | A raw JSON/dict schema | `iter_tool_definitions()`, `get_openai_tool_schemas()`, `get_anthropic_tool_schemas()` | built live from the master |
 
 None of these accept description text — an adapter that needs different wording
@@ -32,9 +32,15 @@ framework's leniency really is adapter-specific (CrewAI's `mentions` coercion),
 express it as a `validators=` entry, never as a re-declared field.
 
 ```python
-from band.runtime.tools import SendMessageInput, platform_args_schema
+from band.runtime.tools import SendMessageInput, platform_args_schema, platform_tool
+
+
+@platform_tool
+async def band_send_message(content: str, mentions: list[str]) -> None: ...
+
 
 assert platform_args_schema("band_send_message") is SendMessageInput
+assert "Args:" in (band_send_message.__doc__ or "")
 ```
 
 Guardrail: `tests/framework_conformance/test_tool_text_drift.py` runs each
--- a/src/band/integrations/crewai/tools.py
+++ b/src/band/integrations/crewai/tools.py
@@ -58,6 +58,7 @@ from band.runtime.tools import (
     is_terminal_success,
     platform_args_schema,
     serialize_tool_result,
+    validate_tool_arguments,
 )
 
 logger = logging.getLogger(__name__)
@@ -415,18 +416,16 @@ def _make_platform_tools(
         cache_function: Any = _no_cache
 
         def _run(self, *_args: Any, **kwargs: Any) -> Any:
-            content: str = kwargs.get("content", "")
-            metadata: dict[str, Any] | None = kwargs.get("metadata")
-
             async def execute(tools: AgentToolsProtocol) -> str:
-                if "message_type" not in kwargs:
-                    raise ValueError(
-                        "Invalid arguments for band_send_event: "
-                        "message_type: Field required"
-                    )
+                # Validated here, not in _run: _run is also called directly,
+                # bypassing args_schema validation, and _exec turns the
+                # ValueError into the error result the caller expects.
+                args = validate_tool_arguments(
+                    "band_send_event", self.args_schema, kwargs
+                )
                 # No execution reporting for send_event to avoid meta-events.
                 await tools.send_event(
-                    content, kwargs["message_type"], metadata=metadata
+                    args["content"], args["message_type"], metadata=args.get("metadata")
                 )
                 return json.dumps({"status": "success", "message": "Event sent"})
 
--- a/src/band/integrations/crewai/tools.py
+++ b/src/band/integrations/crewai/tools.py
@@ -26,29 +26,16 @@ from typing import (
     Any,
     Awaitable,
     Callable,
-    Literal,
     Protocol,
-    Type,
     cast,
     runtime_checkable,
 )
 
-from pydantic import BaseModel, Field, field_validator, model_validator
+from pydantic import BaseModel, field_validator
 
 if TYPE_CHECKING:
     from crewai.tools import BaseTool
 
-from band.core.memory_types import (
-    MemoryListScope,
-    MemorySegment,
-    MemoryStatus,
-    MemoryStoreScope,
-    MemorySystem,
-    MemoryType,
-    memory_type_field_description,
-    validate_memory_type_for_system,
-    validate_subject_scope,
-)
 from band.core.exceptions import BandToolError
 from band.core.protocols import AgentToolsProtocol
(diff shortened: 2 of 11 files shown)


AGENTS.md
```diff
@@ -12,6 +12,37 @@ This is a Python SDK that connects AI agents to the Band collaborative platform.
...
+`src/band/runtime/tools.py` owns every word an LLM reads about a platform tool:
+the input model's class docstring is the tool description, and each
+`Field(description=...)` is an argument description. An adapter must reach for
+whichever of these fits its framework instead of retyping the text:
+
+| Framework wants | Use | Result |
+|---|---|---|
+| A Pydantic `args_schema` class | `platform_args_schema(name)` | the master model itself |
+| The same, but its tool layer emits a value the master won't parse | `platform_args_schema(name, validators={...})` | a subclass with the master's text plus the extra validators |
+| Schema derived from a function docstring | `@platform_tool(name)` | docstring = master description + a rendered `Args:` section |
```
amit-gazal-band: This row is stale and will actively mislead: `platform_tool` takes the **function**, not a name — `c86b8e1e` removed the name argument, and the decorator's own docstring says it "reads `fn.__name__` rather than taking a tool name argument."

A developer following this table writes `@platform_tool("band_foo")` and gets an import-time crash:

```
AttributeError: 'str' object has no attribute '__name__'
```

The prose right below the table is correct, and the executed `python` fence in this section only covers `platform_args_schema` — so markdown-docs can't catch this one. Suggest `@platform_tool` (bare).

Worth considering: adding the decorator to that executed fence would make the table self-policing, e.g.

```python
@platform_tool
async def band_send_message(content: str, mentions: list[str]) -> None: ...

assert "Args:" in (band_send_message.__doc__ or "")
```

src/band/runtime/tools.py
```diff
@@ -1286,24 +1303,98 @@ def get_tool_description(name: str) -> str:
...
+
+
+def format_arg_doc(name: str, description: str) -> str:
+    """Render one Google-style ``Args:`` entry.
+
+    Continuation lines are indented past the argument name so a multi-line
+    ``Field(description=...)`` stays part of that argument — flush-left
+    continuations end the entry as far as a docstring parser is concerned.
+    """
+    head, *rest = description.strip().splitlines()
```
amit-gazal-band, line 1324: This unpack has no floor. A whitespace-only `Field(description=...)` is truthy, so it survives the `if field.description` filter in `get_tool_docstring_with_args` below, and then:

```python
>>> format_arg_doc("x", "   ")
ValueError: not enough values to unpack (expected at least 1, got 0)
```

Because `platform_args_schema` / `platform_tool` are evaluated at **import** time (`SEND_MESSAGE_ARGS_SCHEMA` at `src/band/integrations/crewai/tools.py:333`, the `@platform_tool` decorators in the pydantic-ai adapter), one stray-whitespace description on a master model takes down adapter import rather than degrading to a missing arg description.

Cheapest fix is to filter on the stripped value instead, so the empty case never reaches here:

```python
if field.description and field.description.strip()
```

or guard locally with `if not lines: return f"    {name}:"`.

src/band/integrations/crewai/tools.py
```diff
@@ -559,7 +411,7 @@ async def execute(tools: AgentToolsProtocol) -> str:
     class SendEventTool(BaseTool):
         name: str = "band_send_event"
         description: str = get_tool_description("band_send_event")
-        args_schema: Type[BaseModel] = _SendEventInput
+        args_schema: type[BaseModel] = platform_args_schema("band_send_event")
```
amit-gazal-band, line 415: Routing through the master here makes `message_type` **required** (`SendEventInput.message_type: EventMessageType = Field(...)`), which is the behavior change the PR description claims: "a CrewAI agent that omits `message_type` now gets a validation error instead of a silently mislabeled event."

But `_run` five lines down (line 419) still has the old fallback:

```python
message_type: str = kwargs.get("message_type", "thought")
```

So the schema and the executor now disagree about whether the field is required. That gap matters because this PR itself treats direct-`_run` as a live path — `SendMessageTool._run` just above carries `normalize_mentions_lenient(kwargs.get("mentions"))` with the comment *"Normalized here too, not just in the schema: `_run` is also called directly, bypassing args_schema validation."* Confirmed the bypass is real: crewai's `BaseTool.run()` validates via `_validate_kwargs`, but a Flow calling `create_crewai_tools()` and then `._run(...)` does not (as the fake Flows in `tests/adapters/test_crewai_flow_phase3.py:791,851` do).

On that path an omitted `message_type` is still silently posted as `thought` — exactly the mislabeling the PR says it fixed. Either drop the default so `_run` fails the same way the schema does, or keep it and soften the claim in the description.

amit-gazal-band (changes requested):
Reviewed the full diff (11 files, merge-base `be68fa01`) and verified the risky claims empirically rather than from the PR body. The central design is right and I'd like it in: one master, adapters reach for `platform_args_schema` / `platform_tool` / the schema builders, and the drift guardrail is a real tripwire (confirmed `framework_id="crewai"` carries the probe and the crewai job selects it via `-k crewai`).

What I checked that isn't obvious from the diff:

- **3.11/3.12 docstring indentation** — `get_tool_description` returns `model.__doc__` *raw* (no `cleandoc`), so pre-3.13 indentation really does reach the rendered docstring. I simulated indented class docstrings across all 17 tools and re-ran both invariants through pydantic-ai's real schema builder: 0 blurb mismatches, 0 arg-text mismatches. The flush-left `Args:` header sets `cleandoc`'s margin to 0, so nothing shifts. CI's matrix is safe.
- **No constraint regression** — every bound on the four deleted CrewAI list models (`ge=1`, `le=100`, `le=50`) is present on the masters. The `ge=1` added to `LookupPeersInput` is porting CrewAI's constraint, not new scope.
- **No lost validation** — master `StoreMemoryInput` still enforces `validate_memory_type_for_system` and `validate_subject_scope`; the field is `type`, matching `kwargs.get("type")` in `_run`.
- **Mentions leniency** — exercised 10 input shapes (list / JSON string / `"[@john/agent]"` / bare / `None` / `""` / single-quoted) plus idempotency; the `create_model` subclass preserves the master docstring and every field description.
- `ruff check` and `ruff format --check` clean; `pytest --markdown-docs AGENTS.md` passes. Unit suite: 4627 passed; the 1 failure + 11 errors are all in `tests/integrations/desktop_app/`, which this PR touches in zero commits.

Three things to fix before merge — one is a docs bug that will bite the next person who follows the new guidance, and one is an inconsistency the PR's own description contradicts. Details inline.

AGENTS.md
AlexanderZ-Band: Fixed — table now says `@platform_tool` (bare) with a note that it reads `fn.__name__` and takes no name argument. Also added the decorator to the executed fence per your suggestion, so this stays self-policing: `@platform_tool` on a stub function + `assert "Args:" in (band_send_message.__doc__ or "")`. Verified with `pytest --markdown-docs AGENTS.md`.

src/band/runtime/tools.py
AlexanderZ-Band: Fixed — filtered on `field.description.strip()` upstream in `get_tool_docstring_with_args` (the cheaper option you suggested), so a whitespace-only description never reaches `format_arg_doc`. Verified directly: a probe model with `Field(..., description="   ")` now degrades to no `Args:` line instead of raising `ValueError` at import.

src/band/integrations/crewai/tools.py
AlexanderZ-Band: Fixed — dropped the `kwargs.get("message_type", "thought")` fallback in `SendEventTool._run`. It now raises inside the `execute()` coroutine (caught by `_execute_tool`'s existing error envelope) when `message_type` is missing, so the direct-`_run` bypass path fails the same way the schema does instead of silently mislabeling. Added `test_send_event_run_rejects_missing_message_type`, which I confirmed fails without the fix (asserted status: success was returned) and passes with it.

src/band/runtime/tools.py
AlexanderZ-Band: Follow-up: added a regression test (`test_whitespace_only_description_is_omitted` in `tests/runtime/test_tool_definitions.py`) so this stays caught in CI. Confirmed it fails against the pre-fix code with the exact reported `ValueError: not enough values to unpack` and passes against the fix.

src/band/integrations/crewai/tools.py
```diff
@@ -559,25 +411,31 @@ async def execute(tools: AgentToolsProtocol) -> str:
...
 
         def _run(self, *_args: Any, **kwargs: Any) -> Any:
             content: str = kwargs.get("content", "")
-            message_type: str = kwargs.get("message_type", "thought")
             metadata: dict[str, Any] | None = kwargs.get("metadata")
 
             async def execute(tools: AgentToolsProtocol) -> str:
+                if "message_type" not in kwargs:
+                    raise ValueError(
+                        "Invalid arguments for band_send_event: "
```
amit-gazal-band: Non-blocking, and please merge without it if you'd rather — but this string is generated 40 lines below in `src/band/runtime/tools.py`, which this file already imports from. `validate_tool_arguments` validates, raises `ValueError(format_tool_validation_error(...))` — the same exception type `_execute_tool` already converts — and returns normalized kwargs:

```python
async def execute(tools: AgentToolsProtocol) -> str:
    # Validated here, not in _run: _run is also called directly, bypassing
    # args_schema validation, and _exec turns the ValueError into the
    # error result the caller expects.
    args = validate_tool_arguments("band_send_event", self.args_schema, kwargs)
    await tools.send_event(
        args["content"], args["message_type"], metadata=args.get("metadata")
    )
```

I checked rather than assumed — applied exactly this and re-ran the suite: 4629 passed, including `test_send_event_run_rejects_missing_message_type`. Verified details:

- The message is **byte-identical** to the hand-written one: `validate_tool_arguments("band_send_event", TOOL_MODELS["band_send_event"], {"content": "c"})` raises exactly `Invalid arguments for band_send_event: message_type: Field required`.
- `args["message_type"]` comes back as `MessageType.THOUGHT` rather than a raw `str`. Safe **only** because `MessageType` is a `StrEnum` — worth knowing if that base ever changes.
- `SendEventInput.model_config` is `{}`, so extras are ignored: a direct caller passing `room_id` or stray kwargs isn't newly rejected.

The pull is single-source-of-truth: the master already knows which fields are required and pydantic already words the error, so this can't drift when the model changes — the same argument the PR makes for descriptions. Fine as a follow-up.

tests/adapters/test_crewai_adapter.py
```diff
@@ -1296,8 +1296,34 @@ async def test_send_event_message_type_validation(
...
-        message_type_field = schema_fields["message_type"]
-        assert message_type_field.default == "thought"
+        # Required, not defaulted to "thought" — the master model is authoritative
+        # and every other adapter already makes the agent state the event type.
+        assert schema_fields["message_type"].is_required()
+
+    def test_send_event_run_rejects_missing_message_type(
+        self, CrewAIAdapter, crewai_mocks, mock_tools, room_context
+    ):
+        """A direct `_run` call bypasses args_schema validation (crewai Flows do
```
amit-gazal-band: Tiny thing: the sentence is split across the summary/blank-line boundary, so the docstring's first line reads as truncated — "A direct `_run` call bypasses args_schema validation (crewai Flows do" ... "this), so ...". Reads like a formatter artifact rather than intent.

A real one-line summary with the why below scans better, e.g. `"""A direct _run call must enforce the requiredness the schema enforces."""` — or drop the docstring and use an inline comment, which is what the sibling `test_send_event_message_type_validation` does. Same in `test_whitespace_only_description_is_omitted`.

amit-gazal-band (dismissed):
All three findings addressed — approving. Re-ran the gate on `1450001b`: `ruff check` + `ruff format --check` clean, `pytest --markdown-docs AGENTS.md` passes, unit suite 4629 passed (+2 from the new tests), with the 1 failure + 11 errors still confined to `tests/integrations/desktop_app/`, untouched by this PR.

On each fix:

- **Docs table** — fixed, and extending the executed fence to actually apply `@platform_tool` is better than what I suggested: the table is now self-policing, so a future signature change breaks the doc instead of rotting silently.
- **Whitespace-only description** — `field.description.strip()` closes the reachable path, and the regression test is a real tripwire (`format_arg_doc("x", "   ")` still raises on its own, so dropping the filter fails the test).
- **`send_event` requiredness** — behavior is right now, and the test genuinely fails without the fix. One non-blocking note inline.

Two things I want to correct from my previous review, since both were unfair to this diff:

1. I flagged that `content` stays permissive while `message_type` got hardened. That's **pre-existing**, not a gap in this fix — `kwargs.get(field, "")` on master-required fields is how every one of these `_run`s already worked at the merge base (`content` ×3, `identifier` ×2, `handle` ×3). This PR hardened exactly the field its description made a claim about; broader `_run` hardening is a separate follow-up and shouldn't hold this up.
2. I read the check's placement inside `execute` as odd. It's correct — `_execute_tool` is what turns the `ValueError` into the `{"status": "error"}` result. I confirmed that hoisting it into `_run` makes the error escape uncaught and fails the new test.

amit-gazal-band (approved):

PR #552, opened 2026-08-18
feat: consolidate band-mcp into one SDK-owned MCP engine (INT-1096)

AlexanderZ-Band:
## Summary

Migrates `band-mcp` into this repo as a sibling workspace package (`packages/band-mcp`), and builds one MCP-framework-neutral tool-registration engine consumed by two front doors, per the plan attached to INT-1096 ("Migration plan: one MCP engine (v5)").

**All 13 execution steps are now complete, all 20 review threads addressed and resolved.** One user-owned action item remains before a `band-mcp` release can actually ship (see below); the repo-archiving item is done whenever you get to it post-merge.

## Progress (one commit per step)

- [x] 1. Spike: proved the FastMCP embedding mount recipe end-to-end (real client round trip, start/stop/start cycle, DNS-rebinding protection).
- [x] 2. Workspace scaffold: `packages/band-mcp` as a uv workspace member. Found and resolved a real crewai/mcp dependency conflict (band-mcp's own lock).
- [x] 3. Mechanical copy of band-mcp's source into the workspace, verified against the real SDK.
- [x] 4-5. Test migration (124 unit tests + integration suite) + get-it-green. Found and fixed 2 pre-existing bugs via a live run (`band_lookup_peers` room-scoping, `band_create_chatroom`'s string return).
- [x] 6. Dependency refresh (band-client-rest 0.0.27, mcp floor). Found and resolved a second crewai/mcp conflict — this one on the *published* band-sdk extras.
- [x] 7. Testing toolkit: `FakeHumanTools`, wire-schema snapshot test locking in band-mcp 1.3.2's published contract before the engine touches anything.
- [x] 8. Built `src/band/integrations/mcp/engine.py` — the one engine, verified via real MCP protocol round trips (SDK in-memory transport).
- [x] 9. `src/band/integrations/mcp/local_server.py` (`LocalMCPServer`) — the embedded front door, rebuilt on the engine. `src/band/runtime/mcp_server.py` is now a pure re-export shim for backward compat.
- [x] 10. Rippled the `chat_id` rename through prompts, `claude_sdk`, and custom tools — the Python-side variable is still `room_id` everywhere; only model-facing text (schemas, prompts) says `chat_id` now, uniformly across both doors.
- [x] 11. Shrunk band-mcp's CLI package to a thin front door: `standalone_spec()` (per-tool room classification, `SendEventWideInput` widening, chat_id pinning) + `StandaloneResolver` (human singleton, LRU-cached/lock-striped per-room `AgentTools`). Legacy `thnv_*` key prefixes dropped per your decision — no compatibility shim.
- [x] 12. Publishing & CI: `--all-packages` everywhere in `ci.yml`, fixed the packaging job's now-broken `LocalMCPServer` smoke (real start/stop cycle instead), added band-mcp wheel/import/CLI verification steps, new `band-mcp-publish.yml` (mirrors `band-publish.yml`'s trusted-publishing layout, waits for the declared band-sdk floor to land on PyPI before its install-check), tag-prefix guard on `band-publish.yml` so a band-mcp release doesn't also fire (and fail) the band-sdk publish job, `packages/band-mcp` registered as its own release-please component with `exclude-paths` on the root component, `mcp` added to `pr-title.yml`'s scopes.
- [x] 13. Docs: new "MCP Engine" section in `AGENTS.md`/`CLAUDE.md`, fixed two now-stale `room_id`-vs-`chat_id` mentions.
- [x] Follow-up cleanup: single-sourced three vocabularies that were still duplicated as bare string literals — `CHAT_ID_FIELD_NAME` (the model-facing room-id argument name, now referenced by the engine's schema builders and every adapter prompt that mentions it), `Surface` (replaces `ToolDefinition.surface`'s bare `Literal["agent","human"]`), and `Scope`/`ToolGroup`/`Transport` (replace band-mcp's own `Literal` types plus their hand-maintained parallel `VALID_SCOPES`/`VALID_TOOLS` lists). Also removed the `# type: ignore[arg-type]`/`cast()` escape hatches `resolve_config`'s `Mapping[str, object]` boundary needed, by giving it a concrete `CliArgs` `TypedDict` instead.
- [x] Follow-up hardening: replaced every multi-line `python3 -c "..."` snippet in `ci.yml`/`band-mcp-publish.yml` with a quoted heredoc (`<<'PYEOF'`) — the old form only worked because YAML strips the block's indentation, and silently allowed shell `$`/backtick/backslash expansion inside what's meant to be inert Python source.
- [x] **Real bug found and fixed via the Letta lane smoke**: `build_engine()` never told FastMCP the caller's real bind host, so FastMCP always assumed loopback and locked DNS-rebinding protection to `127.0.0.1`/`localhost` only — rejecting a `host.docker.internal` caller with a 421 even when `LocalMCPServer` was explicitly bound to `0.0.0.0` for exactly that documented Docker-callback case. Fixed by threading the real host through; added regression tests in `test_engine.py`/`test_local_server.py`.
- [x] **Your call, applied**: dropped legacy single-key `BAND_API_KEY` support entirely (breaking change) — `_legacy_key_capabilities`, `Config.legacy_key`, the pure-legacy escape hatch in `server.py`, and the dead `Settings.band_api_key` field are all gone. Set `BAND_USER_KEY`/`BAND_AGENT_KEY` explicitly now.
- [x] **Pre-existing bug found and fixed while touching the above**: `tests/integration/mcp/` (the live-API integration suite) has been silently uncollectable since step 11 — its `conftest.py` still imported `band_mcp.tools.registrar`/`band_mcp.shared.build_app_context`, both deleted that step. Rewrote the fixtures onto the current engine API and deleted `test_forwarding.py` (superseded by `test_standalone_spec.py`/`test_engine.py`). Verified live against the dev platform: 4 passed, 4 skipped, 2 xfailed (pre-existing, unrelated), 0 failures.
- [x] **Test-suite strengthening pass**: swapped mock-overuse (`MagicMock`/`AsyncMock` standing in for `AgentTools`) for the real `FakeAgentTools` fixture across `tests/mcp/test_standalone_spec.py` and `tests/mcp/test_shared.py`, so the participant-refresh/discard-on-failure/mention-error paths exercise real dispatch logic instead of a hand-typed assumption about it. Added a multi-step room-lifecycle test, a custom-tool-alongside-builtin-tools test, and a concurrent-dispatch-across-rooms test to `test_engine.py`; added a live retry test (`test_agent_send_message_retry_after_lookup_peers`) to `tests/integration/mcp/test_full_workflow.py`. Deleted `test_engine_mount_spike.py` (superseded feasibility spike, docstring said as much).
- [x] Replaced the wire-schema JSON snapshot (`tests/fixtures/wire_schemas/*.json`) with `tests/mcp/test_wire_contract.py` — declarative, hand-written per-tool contracts checked against a real `tools/list` round trip, instead of an opaque diff against a checked-in blob a reviewer can't meaningfully read.
- [x] Automated the one manual step the plan called out: added a `bump-mcp-sdk-floor` job to `release.yml` (mirrors the existing `bump-add-band` job's shape) that opens a PR bumping `packages/band-mcp/pyproject.toml`'s `band-sdk>=` floor to match every future band-sdk release, so this never needs a human to remember it again.

## Decisions made along the way

- **Legacy `thnv_*` key prefixes**: the plan's one open decision (keep vs. drop). User decided: **drop** — no compatibility retained. Applied in step 11.
- Two real dependency conflicts between crewai's transitive `mcp` pin and this migration's `mcp` floors, resolved by capping at `mcp>=1.28.1,<2` instead of `>=1.29.0,<2` everywhere — confirmed with the user both times.
- `SendEventWideInput` (the CLI door's widened event-type input) is an independent model, not a subclass of the SDK's `SendEventInput` — pyrefly correctly flagged that widening a mutable Pydantic field's type via subclassing is unsound.
- The embedded door's OpenCode packaging smoke now does a real `async with LocalMCPServer(...)` start/stop cycle (the old `_build_server()` it called no longer exists on the new class) — this exercises the actual FastMCP-mount + uvicorn serve/shutdown recipe against a freshly resolved `mcp`, which is the failure mode that step originally existed to catch.

## User-owned action items

1. ~~PyPI trusted-publisher binding for `band-mcp`~~ — **done.** `band-mcp-publish.yml` is registered as a trusted publisher on the `band-mcp` PyPI project (environment `release`), same as `band-publish.yml` is for `band-sdk`. The project's old trusted-publisher entry (from the standalone `band-mcp` repo) is intentionally left in place until the new one is confirmed working, then should be removed.
2. **Archiving the old `band-mcp` repo** — still open. Once this PR merges and a `band-mcp-v*` release ships from here, the standalone `band-mcp` repo should be archived (or at minimum READMEd to point here) so it doesn't keep accepting PRs against dead code.
3. **Bump `packages/band-mcp/pyproject.toml`'s `band-sdk` floor before tagging the first `band-mcp-v*` release — now automated end to end, but not yet correctable.** It's currently `>=1.6.0` — a real, already-published version, but one that predates this PR and lacks `src/band/integrations/mcp/engine.py` (confirmed live: `band-sdk>=1.6.0` + `band-client-rest==0.0.27` is unsatisfiable today, since published `band-sdk==1.6.0` pins `band-client-rest==0.0.10`). This is a deliberate placeholder — the real floor is the band-sdk version this PR's merge actually ships as, which isn't known until release-please cuts that release. Full automated sequence: band-sdk releases → the new `bump-mcp-sdk-floor` job (`release.yml`) opens a `fix:`-typed PR correcting the floor (typed `fix:`, not `chore:`, specifically so merging it also triggers release-please's own version bump for the `packages/band-mcp` component — a `chore:` would've merged silently and left `pyproject.toml`'s version stuck, which `band-mcp-publish.yml`'s tag-must-match-version check would then reject) → merge that PR → release-please opens band-mcp's own version-bump release PR → merge that → tag `band-mcp-v*`. `band-mcp-publish.yml`'s build job also fails loudly if any of this is skipped (its install-check step hits the exact resolver conflict above before anything reaches PyPI), so there's no risk of a silently-broken publish either way.

## Test plan

- [x] Full unit suite green at every step (4801 passed, 116 skipped pre-existing, at HEAD — up from 4578 at the start).
- [x] `ruff check`, `ruff format --check`, `pyrefly check` clean at every step.
- [x] Real MCP protocol round trips (SDK in-memory transport) for the engine and the CLI's `standalone_spec`, no mocking of engine internals.
- [x] Subprocess-level contract tests (`tests/mcp/test_cli_contract.py`) against the real `band-mcp` CLI — stdio stdout purity, `--version`/`--help`, published tool names, the `health_check` title-leak regression.
- [x] Live integration run against a dev Band environment for the migrated band-mcp integration suite.
- [x] Live smoke: standalone `band-mcp` over real stdio against a real dev-environment Band room — `initialize` → `tools/list` (all 8 tools incl. `health_check`) → `health_check` OK → `band_create_chatroom` → `band_get_participants` → `band_lookup_peers`, all real HTTP round trips, all succeeded.
- [x] Live smoke: `claude_sdk` core lane — 27 passed, 1 skipped, 1 rerun (transient timing flake, not a regression), 0 failures, against real Anthropic API + real Band rooms. Covers the exact code this migration touched (`chat_id`/`room_id` handling in `claude_sdk/tools.py`/`prompts.py`, `extend_with_chat_id` in the rebuilt engine).
- [x] Live smoke: Letta lane — ran, found and fixed the DNS-rebinding bind-host bug noted above.
- [x] All GitHub checks green on this PR (lint, packaging, CodeQL, full test matrix incl. crewai/parlant on ubuntu+windows × 3.11/3.12).
- [ ] e2e backends lane (codex/opencode/copilot_acp) — reported complete in an earlier pass of this work; worth one final confirmation run before merge rather than relying on that memory.


src/band/core/types.py
```diff
@@ -40,6 +40,19 @@ class ToolEventKey(StrEnum):
 # event kinds. Derived from MessageType so the taxonomy stays single-sourced.
 EventMessageType = Literal[MessageType.THOUGHT, MessageType.ERROR, MessageType.TASK]
 
+# The MCP engine's CLI-door widening of EventMessageType: a standalone MCP
+# agent has no adapter narrating tool_call/tool_result events on its
+# behalf, so band_send_event needs a self-narration channel there that the
+# embedded SDK door doesn't (adapters author tool_call/tool_result
+# programmatically for embedded agents).
+WideEventMessageType = Literal[
```
AlexanderZ-Band: Why this indirection?

packages/band-mcp/src/band_mcp/server.py
```diff
@@ -0,0 +1,372 @@
...
+        "-p",
+        type=int,
+        default=None,
+        help="Port to bind for SSE mode (default: 8000)",
+    )
+
+    return parser.parse_args(argv)
+
+
+def _cli_mapping(args: argparse.Namespace) -> CliArgs:
```
AlexanderZ-Band: Let's look for modern frameworks with handle the arguments parse and validation it feels messy plumbing

src/band/adapters/claude_sdk.py
```diff
@@ -636,8 +636,8 @@ async def on_message(
             else:
                 raise
 
-        # Add room_id context (Claude needs this for tool calls)
-        room_context = f"[room_id: {room_id}]"
+        # Add chat_id context (Claude needs this for tool calls)
+        room_context = f"[chat_id: {room_id}]"
```
AlexanderZ-Band: Single source of truth

tests/integration/mcp/conftest.py
```diff
@@ -0,0 +1,255 @@
...
+    return Config(
+        user_key=user_key,
+        agent_key=agent_key,
+        scope=[Scope.AGENT, Scope.HUMAN],
+        tools=[ToolGroup.CONTACTS, ToolGroup.MEMORY],
+    )
+
+
+@pytest.fixture(scope="session")
+def live_config_2() -> Config:
```
AlexanderZ-Band: Bad name hides bad design

tests/integration/mcp/conftest.py
```diff
@@ -0,0 +1,255 @@
...
+
+
+@pytest.fixture
+def harness(live_config: Config, monkeypatch: pytest.MonkeyPatch) -> LiveHarness:
+    """Primary agent's driver (agent + human scope)."""
+    return _build_harness(live_config, monkeypatch)
+
+
+@pytest.fixture
+def harness_2(live_config_2: Config, monkeypatch: pytest.MonkeyPatch) -> LiveHarness:
```
AlexanderZ-Band: Bad name bad design

tests/integration/mcp/conftest.py
```diff
@@ -0,0 +1,255 @@
...
+    (same known platform limitation noted in ``tests/integration/test_agent_contacts.py``).
+    These tests require real API access and are skipped in CI.
+    """
+    created = await harness.call("band_create_chatroom")
+    room_id = _extract_id(created)
+    assert room_id, f"band_create_chatroom returned no id: {created!r}"
+    yield room_id
+
+
+async def ensure_mentionable_participant(
```
AlexanderZ-Band: Bad name review design decision

tests/integration/mcp/conftest.py
```diff
@@ -0,0 +1,255 @@
...
+
+    A freshly created agent room has no other participant, and self-mention is
+    disallowed by design. Pass `identifier` for a known peer (e.g. a second
+    test agent); omit it to add the room-owning human, discovered via
+    ``band_lookup_peers`` (the ``type: "User"`` entry).
+    """
+    if identifier is None:
+        peers = _unwrap(
+            await harness.call(
+                "band_lookup_peers", chat_id=room_id, page=1, page_size=100
```
AlexanderZ-Band: Single source of truth

tests/integration/mcp/test_full_workflow.py
```diff
@@ -0,0 +1,112 @@
...
+async def test_two_agents_collaborate_in_shared_room(
+    harness: LiveHarness, harness_2: LiveHarness, agent_room: str
+) -> None:
+    """Agent 1 adds a second, genuinely distinct agent identity and @mentions
+    them; agent 2 independently confirms membership through its own session,
+    not agent 1's participant cache."""
+    second_agent_id = get_test_agent_id_2()
+    assert second_agent_id, "TEST_AGENT_ID_2 must be set in .env.test"
+
+    await ensure_mentionable_participant(
```
AlexanderZ-Band: Repeats it self and very hacky just create a scenario with more then o e participant human or agent

src/band/runtime/tools.py
```diff
@@ -904,176 +973,176 @@ def canonicalize_mcp_tool_name(tool_name: str, own_names: Collection[str]) -> st
...
+        surface=Surface.HUMAN,
     ),
     "band_list_my_contacts": ToolDefinition(
         name="band_list_my_contacts",
         input_model=ListMyContactsInput,
         method_name="list_my_contacts",
-        surface="human",
+        surface=Surface.HUMAN,
     ),
     "band_create_contact_request": ToolDefinition(
```
AlexanderZ-Band: Don't we already have a strenum for house keys?

src/band/core/types.py
AlexanderZ-Band: Fixed in ada84b2a: WideEventMessageType had exactly one consumer (this module's own SendEventWideInput), so it now lives there instead of core/types.py, and derives from EventMessageType instead of retyping its members.

packages/band-mcp/src/band_mcp/server.py
AlexanderZ-Band: Swapped to Typer in ada84b2a. It's already a real transitive dependency (mcp[cli] requires it), so no new install footprint -- just promoted to an explicit direct dependency. parse_args/_cli_mapping are gone; Typer hands the command function typed parameters directly, native Enum support replaces the manual choices= wiring, and resolve_config/validate in config.py are untouched (deliberately pure/framework-agnostic). Verified live: --version, --help, the missing-credential exit(2) path, and a real stdio tools/list round trip all match the pre-swap contract.

src/band/adapters/claude_sdk.py
AlexanderZ-Band: Fixed in ada84b2a: both claude_sdk.py and copilot_sdk.py now derive this from CHAT_ID_FIELD_NAME (already used correctly by opencode.py and acp/client_adapter.py). copilot_sdk.py's version actually said "room_id" -- a real violation of the documented invariant that model-facing text always says "chat_id", not just a style gap. Fixed the test that had encoded the wrong string too.

tests/integration/mcp/conftest.py
AlexanderZ-Band: Fixed in ada84b2a: renamed to second_agent_config (and harness_2 to second_agent_harness below) -- names the fixture's role instead of just numbering it.

tests/integration/mcp/conftest.py
AlexanderZ-Band: Same fix as above (ada84b2a): renamed to second_agent_harness.

tests/integration/mcp/conftest.py
AlexanderZ-Band: Fixed in ada84b2a: replaced the dual-mode ensure_mentionable_participant (identifier=None secretly meaning "discover the owner instead") with single-purpose add_room_owner. Adding a *known* second agent is just a one-line band_add_participant call at the test call site -- no helper needed for that case.

tests/integration/mcp/conftest.py
AlexanderZ-Band: Same fix (ada84b2a) -- add_room_owner replaces the dual-mode helper, so there's one lookup path instead of an optional-identifier branch.

tests/integration/mcp/test_full_workflow.py
AlexanderZ-Band: Fixed in ada84b2a: merged this into test_agent_create_room_send_and_read_back as one test_agent_room_with_human_and_second_agent -- a single room with both a human and a second real agent identity as participants, mentioned together in one send_message.

src/band/runtime/tools.py
AlexanderZ-Band: Fixed in 2f5d6bfd: not a StrEnum (this is 45 entries, one per registered tool, not a small closed vocabulary), but every entry did type its name twice -- once as the dict key, once as ToolDefinition's own name= field, with nothing structural stopping the two from drifting apart. Now built from a tuple of ToolDefinition entries (name typed once each) with TOOL_DEFINITIONS derived by keying on .name.

src/band/runtime/tools.py
```diff
@@ -809,10 +817,71 @@ def canonicalize_mcp_tool_name(tool_name: str, own_names: Collection[str]) -> st
...
+            return (definition.name in AGENT_ROOM_BOUND_TOOL_NAMES, False)
+        case Surface.HUMAN:
+            return (False, CHAT_ID_FIELD_NAME in definition.input_model.model_fields)
+        case _:
+            return (False, False)
+
+
+# The one tool name referenced by name outside its own ToolDefinition entry
+# (engine.py's mention-handle error enrichment, band-mcp's SEND_MESSAGE_METHOD_NAME).
+SEND_MESSAGE_TOOL_NAME = "band_send_message"
```
AlexanderZ-Band: Don't we have it somewhere else? Is so then Let's find all of those and consolidate to one str enum

packages/band-mcp/src/band_mcp/shared.py
```diff
@@ -0,0 +1,218 @@
...
+# own startup/warning logs would be silently suppressed below BAND_LOG_LEVEL.
+LogSettings(log_stream=LogStream.STDERR).for_application().configure()
+logger = logging.getLogger(__name__)
+
+AGENT_TOOLS_CACHE_MAX_SIZE = 128
+# Matches the cache size: a coarser stripe count lets two unrelated chat_ids
+# share a lock, so one room's in-flight REST call (the send_message
+# participant refresh below) can block an unrelated room's call for no
+# reason. One stripe per possible cache entry removes that false contention.
+AGENT_TOOLS_LOCK_STRIPES = AGENT_TOOLS_CACHE_MAX_SIZE
```
AlexanderZ-Band: Why we need this indirection smells bad

packages/band-mcp/src/band_mcp/shared.py
```diff
@@ -0,0 +1,218 @@
...
+        ``AgentTools.from_context(ctx)``.
+        """
+        if self._agent_id_resolved:
+            return self._agent_id
+        async with self._agent_id_lock:
+            # Re-check: a concurrent caller (different chat_id stripe) may
+            # have already resolved it while this one waited for the lock.
+            if self._agent_id_resolved:
+                return self._agent_id
+            assert self._agent_rest is not None
```
AlexanderZ-Band: Assert in code is bad practice keep asserts to tests only

src/band/integrations/mcp/engine.py
```diff
@@ -0,0 +1,683 @@
...
+    ToolDefinition,
+    append_available_mention_handles,
+    iter_tool_definitions,
+    serialize_tool_result,
+    validate_tool_arguments,
+)
+
+logger = logging.getLogger(__name__)
+
+CHAT_ID_MAX_LENGTH = 255
```
AlexanderZ-Band: Seems important enough to keep in central place also I think we have it elsewhere already find and consolidate

packages/band-mcp/src/band_mcp/shared.py
AlexanderZ-Band: Fixed in dbbc7c81: AGENT_TOOLS_LOCK_STRIPES was a second name for exactly AGENT_TOOLS_CACHE_MAX_SIZE's value with nothing distinguishing the two -- removed it, reference AGENT_TOOLS_CACHE_MAX_SIZE directly at its one use site.

packages/band-mcp/src/band_mcp/shared.py
AlexanderZ-Band: Fixed in dbbc7c81: replaced with an explicit if self._agent_rest is None: raise RuntimeError(...) -- matches the sibling guard three lines below it in _get_or_create_agent_tools.

src/band/integrations/mcp/engine.py
AlexanderZ-Band: Fixed in c08fe629: searched this repo and the band-client-rest dependency for another 255/max_length constraint tied to chat_id/room_id -- found none, so it wasn't actually duplicated elsewhere. Moved it to sit next to CHAT_ID_FIELD_NAME in runtime/tools.py though (same field, two properties, shouldn't be split across files), and engine.py now imports it from there.

src/band/runtime/tools.py
AlexanderZ-Band: Consolidated all the duplicated 'band_send_message' literals in dbbc7c81 (crewai/claude_sdk/letta/agno + this file's own two frozensets), but kept SEND_MESSAGE_TOOL_NAME as a plain str constant rather than a StrEnum -- checked, and it's the only standalone tool-identity constant of its kind in this file; every other tool name lives as a key in TOOL_DEFINITIONS (the real registry), not as an individually-named constant. A StrEnum with exactly one member doesn't add anything a plain constant doesn't already give -- no second member to justify the abstraction. Open to it if you see a grouping I'm missing, otherwise leaning toward leaving it a constant.

tests/adapters/copilot_sdk/test_reply.py
```diff
@@ -44,7 +44,7 @@ async def test_prompt_contains_room_context_and_message(self):
         await run_message(adapter, tools, content="What's up?")
 
         prompt = client.sessions[0].prompts[0]
-        assert "[room_id: room-1]" in prompt
+        assert "[chat_id: room-1]" in prompt
```
AlexanderZ-Band: Dont we have a const for it?

tests/websocket/test_client.py
```diff
@@ -53,6 +55,21 @@
 }
 
 
+async def dispatch(client: WebSocketClient, event: str, payload: dict) -> Any:
+    """Feed one event through _handle_events via a single registered
+    callback; return what that callback received (None if never called)."""
+    received = None
+
+    async def callback(p: Any) -> None:
+        nonlocal received
```
AlexanderZ-Band: nonlocal is a clear sign of bad design lets think harder on how to avoid it plan search onlin look at other similar places in our code, suggest better sollutions

tests/mcp/test_wire_contract.py
```diff
@@ -0,0 +1,329 @@
...
+    checked generically via ``AGENT_ROOM_BOUND_TOOL_NAMES`` instead."""
+
+    required: frozenset[str] = frozenset()
+    fields: Mapping[str, FieldContract] = field(default_factory=dict)
+
+
+# Only tools with a real, non-obvious wire invariant. A plain required
+# string/int/bool field has nothing here worth pinning.
+CONTRACTS: dict[str, ToolContract] = {
+    "band_send_message": ToolContract(
```
AlexanderZ-Band, line 163: can we avoid hardcoded strings? how are the y used? violates single source of truth?

tests/mcp/test_transport_security.py
```diff
@@ -0,0 +1,266 @@
...
+        """When allowed_hosts is empty, all hosts are blocked."""
+        middleware = TransportSecurityMiddleware(
+            TransportSecuritySettings(
+                enable_dns_rebinding_protection=True,
+                allowed_hosts=[],
+            )
+        )
+
+        # All hosts should be blocked
+        assert middleware._validate_host("localhost:8000") is False
```
AlexanderZ-Band: why do we test it here? is it really our consern to test?

tests/mcp/test_transport_security.py
```diff
@@ -0,0 +1,266 @@
...
+
+    ``server.py`` builds a fresh engine per ``run()`` call -- no
+    module-level FastMCP singleton to import. Build one here the same way
+    (``standalone_spec`` + ``build_engine`` with ``_build_transport_security()``).
+    """
+
+    def _build_mcp(self) -> object:
+        from band.integrations.mcp.engine import build_engine
+        from band_mcp.config import Config
+        from band_mcp.config import settings
```
AlexanderZ-Band: only top level imports, everywhere!

tests/adapters/copilot_sdk/test_reply.py
AlexanderZ-Band: Fixed in e7ff7468: was hardcoding "[chat_id: room-1]" as a literal, duplicating the CHAT_ID_FIELD_NAME constant copilot_sdk.py's own _compose_prompt builds that label from. Now imports CHAT_ID_FIELD_NAME and asserts f"[{CHAT_ID_FIELD_NAME}: room-1]" -- same constant on both sides.

tests/websocket/test_client.py
AlexanderZ-Band: Fixed in e7ff7468: dispatch() now uses a plain AsyncMock() as the callback instead of a hand-rolled closure with nonlocal -- reads the received payload off callback.await_args.args[0] (None if never awaited). Checked the rest of the codebase: this nonlocal-closure-capture idiom is actually pervasive (20+ pre-existing occurrences across tests/), but this particular one is the only instance this PR introduced, so that's the scope of this fix -- the pre-existing ones are a separate, much larger cleanup.

tests/mcp/test_transport_security.py
AlexanderZ-Band: Fixed in e7ff7468: hoisted all of this file's function-local band_mcp/band.integrations.mcp imports (12 of them, across every test method plus _build_mcp) to the top of the file. Every other file in tests/mcp/ already imports band_mcp at module level -- this file was the one outlier.

tests/mcp/test_transport_security.py
AlexanderZ-Band: Fixed in e7ff7468: deleted TestDnsRebindingProtectionBehavior. It's not our concern -- it drove mcp SDK's TransportSecurityMiddleware directly (including its private _validate_host), constructed by hand, disconnected from anything band-mcp builds. Its stated goal ("ensure our configuration is applied correctly") is already covered by TestMcpTransportSecurityIntegration right above it, which builds through band-mcp's real standalone_spec/build_engine/_build_transport_security factories and checks the resulting settings match ours. That's the actual integration boundary we own; characterizing the vendor middleware's own host-matching logic in isolation is the mcp SDK's job, not ours.

tests/mcp/test_wire_contract.py
AlexanderZ-Band: Checked: these tool-name strings (band_send_message, band_send_event, etc.) aren't duplicating a canonical declaration elsewhere -- they're dict keys in CONTRACTS, an editorial curation of which 8 of the 45 registered tools have a genuinely non-obvious wire invariant worth pinning (enum, array item type, a required-set beyond boilerplate). There's no second place these same literals are already typed: per the earlier thread on SEND_MESSAGE_TOOL_NAME, only band_send_message has a named constant -- every other tool name lives solely as a key in TOOL_DEFINITIONS, by design (a StrEnum/constant per tool wasn't judged worth it there either). So there's nothing to import here instead. Drift protection already exists and is stronger than a shared constant would give: test_full_profile_matches_contract does live[name] (not .get) for every CONTRACTS key, so a renamed/deleted tool fails the test loudly on the next run, not silently. Leaving as-is -- no SSOT violation, since nothing else declares these same strings.

src/band/runtime/tools.py
AlexanderZ-Band: No further grouping surfaced, so finalizing: SEND_MESSAGE_TOOL_NAME stays a plain str constant. Resolving.

amit-gazal-band (commented):
test

src/band/integrations/acp/client_adapter.py
```diff
@@ -531,7 +524,6 @@ async def _ensure_band_mcp_backend(self) -> BandMCPBackend:
                     additional_tools=self._custom_tools,
                 )
                 self._band_mcp_backend = backend
```
amit-gazal-band, line 533: The memoized band-mcp backend (`_ensure_band_mcp_backend`) is never health-checked before reuse. `LocalMCPServer.start()` launches uvicorn's `serve()` as a fire-and-forget `asyncio.create_task` with no `add_done_callback`; if that task later crashes, `self._band_mcp_backend` stays non-None forever and every later room in this process keeps getting handed the same dead host/port, with no restart path and no error surfaced until a tool call times out. Worth at least a liveness check (or a done-callback that clears the cached backend) so a crash gets detected instead of silently poisoning future rooms.

packages/band-mcp/src/band_mcp/shared.py
```diff
@@ -0,0 +1,221 @@
...
+        # SDK constructor -- "" is the sentinel, matching the None cache key.
+        instance = AgentTools(
+            room_id=chat_id if chat_id is not None else "",
+            rest=self._agent_rest,
+            agent_id=agent_id,
+        )
+        self._agent_tools_cache[chat_id] = instance
+        self._agent_tools_cache.move_to_end(chat_id)
+        while len(self._agent_tools_cache) > AGENT_TOOLS_CACHE_MAX_SIZE:
+            self._agent_tools_cache.popitem(last=False)
```
amit-gazal-band, line 185: This LRU eviction (`popitem(last=False)`) doesn't hold the evicted room's own stripe lock, so a room's cached `AgentTools` can be evicted here while a call for that same room is still in flight elsewhere. Impact is bounded today — `send_message`/`add_participant`/`remove_participant` all do a pre-flight `get_participants()` refresh regardless, so a race just costs one extra REST round trip — but it's a real, currently-untested race (`tests/mcp/test_shared.py` only covers sequential eviction). Consider a regression test for eviction-during-in-flight-call, even if no fix is needed given the bounded impact.

src/band/integrations/mcp/engine.py
```diff
@@ -0,0 +1,682 @@
...
+            field_kwargs["description"] = field_info.description
+
+        annotation = (
+            Annotated[base_annotation, Field(**field_kwargs)]
+            if field_kwargs
+            else base_annotation
+        )
+
+        default = (
+            inspect.Parameter.empty if field_info.is_required() else field_info.default
```
amit-gazal-band: For a non-required field, this reads `field_info.default` directly, which is Pydantic's `PydanticUndefined` sentinel (not the resolved value) for any field declared with `Field(default_factory=...)` rather than a literal `default=`. Dormant today since no current tool input model uses `default_factory`, but the first one that does (including a third-party custom tool) will get `PydanticUndefined` synthesized into its advertised JSON schema default for every MCP client, with nothing currently guarding against it. Worth handling `default_factory` here (or asserting against it) before it's needed.

.github/workflows/band-mcp-publish.yml
```diff
@@ -0,0 +1,139 @@
...
+# layout (see that file's header for the full security-model rationale:
+# separate build/publish jobs, environment `release`, tag-anchored version
+# validation) with one addition: band-mcp declares a runtime floor on
+# band-sdk (packages/band-mcp/pyproject.toml's `band-sdk>=X.Y.Z`) that a
+# two-phase release bumps to the band-sdk version that first ships the code
+# band-mcp needs. Publishing can lag PyPI's index by a couple minutes, so the
+# build job waits for that exact version to actually resolve before
+# installing the freshly-built wheel against it.
+#
+# The PyPI trusted publisher must reference this filename
```
amit-gazal-band, line 11: Flagging inline since this is the line documenting it: can we confirm Alex is going to register the PyPI trusted publisher for `band-mcp` (owner/repo `band-ai/band-sdk-python`, workflow `band-mcp-publish.yml`, environment `release`) before the first tagged release, since it's a brand-new PyPI project and this can only be configured on pypi.org itself.

amit-gazal-band (dismissed):
Reviewed against ../band-mcp as well — the MCP-engine consolidation and the new band-mcp-publish workflow both look right. A few non-blocking findings inline, plus two notes below that aren't tied to a diff line:

1. **`src/band/adapters/claude_sdk.py` (`_create_mcp_backend`, ~line 500, pre-existing/untouched by this diff)**: the Claude Agent SDK adapter still routes tool dispatch/error-enrichment/result-formatting through the hand-rolled path in `src/band/integrations/claude_sdk/tools.py` rather than the shared `engine.py` this PR consolidates band-mcp CLI and the embedded server onto. A future fix to `engine.py`'s `enrich_send_message_error` or result-serialization shape won't reach Claude SDK users, defeating 'one engine, two front doors' for probably the most-used adapter. Worth a follow-up ticket to fold this third path into the engine — not a blocker here since this PR didn't touch that function.

2. **PyPI trusted publisher for `band-mcp`**: can we confirm Alex is going to configure it before we cut the first release? Trusted publishers are registered per-project on pypi.org itself and can't be created by `band-mcp-publish.yml` — since `band-mcp` is a brand-new PyPI project, someone with PyPI owner access needs to add a trusted publisher pointing at `band-ai/band-sdk-python`, workflow filename `band-mcp-publish.yml`, environment `release` (matching what that workflow's header documents), or the first tagged `band-mcp-v*` publish will fail.

src/band/integrations/acp/client_adapter.py
AlexanderZ-Band: Fixed in 8d03c1bd: added \`LocalMCPServer.is_running\` (reflects the serve task's \`done()\` state) and \`BandMCPBackend.is_running\` (delegates to it; always \`True\` for the in-process \`sdk\` kind, which has no server task to crash). \`\_ensure_band_mcp_backend\` now checks this under the existing \`\_mcp_backend_lock\` before returning the cached backend -- a crashed backend is stopped and cleared so the next call starts a fresh one instead of handing out the dead host/port. Added a crash-simulation regression test in \`tests/integrations/mcp/test_local_server.py\` and a cache-restart test in \`tests/integrations/acp/test_client_adapter.py\`.

src/band/integrations/mcp/engine.py
AlexanderZ-Band: Confirmed live, and it's worse than a leaked sentinel value: passing PydanticUndefined through as the synthesized parameter's literal default makes FastMCP's create_model() (mcp/server/fastmcp/utilities/func_metadata.py) treat it as "no default provided" -- so the field gets advertised as \*required\* in the JSON schema, the opposite of what a default_factory field should show. Repro'd with a \`tags: list\[str\] = Field(default_factory=list)\` field: schema came back with \`"required": \["tags"\]\` and no \`default\` key.

Fixed in 261b151c: when \`field_info.default_factory is not None\`, the synthesized parameter's default is now \`Field(default_factory=field_info.default_factory)\` instead of the raw \`field_info.default\`. Verified the resulting schema matches \`input_model.model_json_schema()\`'s own behavior for the same field (optional, no advertised default), and that omitting the field at call time still resolves through the real factory. Added a regression test in \`tests/mcp/test_engine.py\` that fails without the fix (confirmed by reverting it locally) and passes with it.

packages/band-mcp/src/band_mcp/shared.py
AlexanderZ-Band: Validated the race and traced it end to end: since \`\_invoke_agent\` holds a room's own stripe lock for its entire call, the only way this fires is a \*different\* room's concurrent cache-miss insert overflowing the cache and evicting this room's entry via the un-locked \`\_get_or_create_agent_tools\` path while this room's own call is still deep in \`dispatch_tool\`. Confirmed the impact stays bounded exactly as described: the in-flight call already holds a direct Python reference to its own \`AgentTools\` instance, so the dict eviction doesn't touch it -- it completes normally. The only cost is the next call for that room rebuilding a fresh instance (\`send_message\`/\`add_participant\`/\`remove_participant\` already re-fetch participants regardless).

Agreed no fix is warranted given the bounded impact. Added \`test_invoke_agent_survives_its_own_cache_entry_evicted_mid_flight\` in \`tests/mcp/test_shared.py\` (261b151c): seeds one room mid-dispatch (blocked on a gated \`send_message\`), overflows the cache with distinct rooms until that room's entry gets evicted, then confirms its in-flight call still completes and delivers correctly afterward.

.github/workflows/band-mcp-publish.yml
AlexanderZ-Band: Confirmed -- I'll register the trusted publisher on pypi.org for \`band-mcp\` (owner/repo \`band-ai/band-sdk-python\`, workflow \`band-mcp-publish.yml\`, environment \`release\`) before the first tagged \`band-mcp-v\*\` release, matching what the workflow header documents.

amit-gazal-band (approved):

PR #568, opened 2026-08-23
docs: add the shared-core boundary recommendation and deck

gavrie:
**The recommendation**: ship `core-protocol` across the FFI boundary and let every consumer own its own I/O. The shared core is the sans-io protocol logic — session state machine, reconnect planner, disconnect taxonomy, frame codec, rejoin policy. Each SDK drives it with its own transport, on its own runtime, using its own sockets.

Landing here rather than in the Rust core repo, which has since been archived.

## Files

| | |
|---|---|
| `docs/rust-core-recommendations.md` | The full recommendation — a glossary for readers without a Rust background, twelve sections, and an appendix on what a Go SDK would change |
| `docs/rust-core-recommendations-deck.html` | A 16-page presentation of the same argument. Self-contained: no external assets, works offline, light/dark aware |

Authored by Gavrie Philipson, based on input from and discussions with Amit Gazal, Nir Singher and Alexander Zaikman.

### DIFF
--- a/docs/rust-core-recommendations-deck.html
+++ b/docs/rust-core-recommendations-deck.html
@@ -1,0 +1,1565 @@
+<meta charset="utf-8">
+<title>Band SDK: A Shared Rust Core</title>
+
+<!-- Open in light regardless of the viewer's theme: this is presented off a laptop
+     in a room. The switch below hands control back. Runs before body paint. -->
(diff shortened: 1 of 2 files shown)


AlexanderZ-Band:
Closing as stale: this documentation-only proposal has had no activity or review discussion since August.

PR #573, opened 2026-08-24
feat: add file transfer capability to the Python SDK (INT-1261)

AlexanderZ-Band:
## Summary

Adds file transfer capability (INT-1261): agents can list, read, and send room files, gated by a new `Capability.FILES` negotiated against the platform's `ff_file_transfer` deployment flag (on-prem only, off on SaaS today).

## Changes

- `Capability.FILES` + `runtime/capabilities.py::prune_unsupported`, negotiated via a new `SimpleAdapter.apply_effective_features()` hook wired into `Agent.start()`/`OneShotInvoker.startup()`
- Three new tools on `AgentTools` (+ `FakeAgentTools`): `band_list_room_files`, `band_read_room_file`, `band_send_room_file`
- **Breaking:** `include_memory`/`include_contacts` booleans replaced by a single `capabilities: frozenset[Capability] | None` param across `iter_tool_definitions`, `AgentTools` schema methods, and the MCP engine — migrated through every adapter
- `claude_sdk`: vision-passthrough fix so an image result reaches the model as real vision input, plus narration redaction for image bytes; only adapter wired to `Capability.FILES` this phase (`band-mcp` CLI unaffected — no `files` tool group yet)
- `AGENTS.md`: new File Tools list + Capability Negotiation section

## Related Issues

Fixes INT-1261

## Testing

- [x] `uv run pytest tests/ --ignore=tests/integration/ --ignore=tests/e2e/` (5059 passed)
- [ ] Manual round-trip against a live platform with `ff_file_transfer` on — not CI-wired, per plan

## Checklist

- [x] Tests added/updated
- [x] Documentation updated


src/band/runtime/tools.py
```diff
@@ -2423,6 +2585,251 @@ async def archive_memory(self, memory_id: str) -> Any:
...
+                "content_type": attachment.content_type,
+                "bytes": attachment.bytes,
+            }
+            try:
+                result["text"] = body.decode("utf-8")
+            except UnicodeDecodeError:
+                # content_type has no charset (derived from magic bytes
+                # alone), so a non-UTF-8 file can't be decoded correctly.
+                # Say so rather than silently handing back replaced bytes.
+                result["text"] = body.decode("utf-8", errors="replace")
```
amit-gazal-band: Minor efficiency nit: on a UnicodeDecodeError, this re-decodes the same bytes from scratch with errors="replace" after the strict decode above already failed -- doubling the decode work on every non-UTF-8 text read. Functionally correct (body is capped at MAX_INLINE_TEXT_BYTES, so the cost is negligible), but could be a single decode: decode once with errors="replace", then detect corruption via `U+FFFD in text` instead of a second full decode.

src/band/runtime/tools.py
```diff
@@ -1497,31 +1658,29 @@ def iter_tool_definitions(
...
-        include_memory: Include memory tools (enterprise). Default False.
-        include_contacts: Include contact-management tools. Default True for
-            backward compatibility. Pass False to gate contact tools behind
-            ``Capability.CONTACTS``. The hub-room execution path always
-            forces this to True regardless of adapter preference (see
-            ``AgentTools.get_tool_schemas`` HUB_ROOM auto-enable rule).
+        capabilities: Optional tool categories to include. ``None`` (default)
+            means contacts only, for backward compatibility.
     """
+    resolved = DEFAULT_CAPABILITIES if capabilities is None else capabilities
```
amit-gazal-band: DEFAULT_CAPABILITIES if capabilities is None else capabilities is duplicated verbatim here and in AgentTools.get_tool_schemas (line 2995). Per the repo's single-source-of-truth standard for closed-vocabulary resolution, worth factoring into one small helper (e.g. resolve_capabilities(capabilities)) so a future change to the default-capability rule can't be applied to one call site and silently missed on the other.

src/band/runtime/tools.py
```diff
@@ -2423,6 +2585,251 @@ async def archive_memory(self, memory_id: str) -> Any:
...
+            cursor: Pagination cursor from a previous call's response.
+
+        Returns:
+            Dict with "data" (attachment dicts, deduplicated by id -- a file
+            can be attached to more than one message) and "next_cursor".
+        """
+        response = await self._list_message_page(cursor)
+        seen: set[str] = set()
+        attachments: list[dict[str, Any]] = []
+        for message in response.data:
```
amit-gazal-band: This nested for message in ...: for attachment in message.attachments or [] traversal is duplicated in _find_attachment below (lines 2660-2663). Consider sharing one attachment-iteration helper so a future change to the traversal rule (e.g. skipping soft-deleted attachments, or a dedup key change) can't be fixed in one copy and left stale in the other.

src/band/runtime/tools.py
```diff
@@ -2423,6 +2585,251 @@ async def archive_memory(self, memory_id: str) -> Any:
...
+        """
+        cursor: str | None = None
+        more_pages = True
+        while more_pages:
+            response = await self._list_message_page(cursor)
+            yield response
+            cursor = response.metadata.next_cursor
+            more_pages = bool(response.metadata.has_more and cursor)
+
+    async def _find_attachment(self, file_id: str) -> "Attachment":
```
amit-gazal-band, line 2670: _find_attachment restarts pagination from the first page on every call, with no cache of previously-discovered attachment locations. In a long-lived, high-traffic room, reading an older file (or re-reading the same file id) re-walks the full message history page-by-page each time -- O(pages) sequential REST round trips per read_room_file call, scaling with room age rather than file position. Note: a per-instance cache was already deliberately removed earlier in this branch since AgentTools is recreated per turn and it didn't help across turns -- a fix here would need a differently-scoped cache (e.g. keyed by room, living longer than one AgentTools instance), not a revert of that removal.

amit-gazal-band (dismissed):
Reviewed the file-transfer capability changes end to end (including the capabilities: frozenset[Capability] | None migration off include_memory/include_contacts). No correctness bugs found -- the branch has already been through several self-hardening passes. Left four minor reuse/efficiency notes inline; none are blocking.

src/band/runtime/tools.py
AlexanderZ-Band: Fixed in e164080c: decode once with errors="replace" and detect corruption by checking for the U+FFFD replacement character in the result, instead of a second strict-then-lenient decode pass. (Sorry for the garbled first reply -- shell escaping mangled it.)

src/band/runtime/tools.py
AlexanderZ-Band: Fixed in e164080c: extracted resolve_capabilities() right next to DEFAULT_CAPABILITIES, both call sites now use it.

src/band/runtime/tools.py
AlexanderZ-Band: Fixed in e164080c: added a shared _attachments_in() generator, used by both list_room_files() and _find_attachment().

src/band/runtime/tools.py
AlexanderZ-Band: Agreed this needs a differently-scoped cache and is bigger than a quick fix here -- leaving as-is per your note, opening a low-priority followup ticket to track it rather than fixing inline.

AlexanderZ-Band: Tracked as a followup: https://linear.app/thenvoi/issue/INT-1306 (low priority).

amit-gazal-band (approved):

PR #575, opened 2026-08-26
feat: add agent-scoped memory support

AlexanderZ-Band:
## Summary

- Platform ([PLT-1396](https://linear.app/thenvoi/issue/PLT-1396/fix-memory-read-scoping-when-organization-is-missing)) closed a cross-tenant data leak in organization-scoped memory reads and added a third real scope value — `agent` (private to the storing agent, no organization required). The SDK only exposed `subject`/`organization`, and the live system-prompt guidance every real agent reads steered it toward `organization` scope with no `agent` alternative — which now 422s for any agent whose owner has no organization (verified live: true of every E2E test identity, and there's no self-service org-creation API to fix that).
- Add `AGENT` to `MemoryStoreScope`/`MemoryListScope`; fix `validate_subject_scope`'s fallback suggestion (was pointing back at the now-broken `organization` scope); update master tool text and the live `runtime/prompts.py` guidance every agent gets every turn.
- Migrate the baseline E2E memory suite off `organization` scope onto `agent` scope; invert the cross-agent org-sharing test into a real agent-scope isolation regression test instead of skipping it.
- Add two live (unmocked) tests locking in the platform's fail-closed 422 contract on both read and write.
- Fix a stale customer-facing example that hardcoded `organization` scope.
- Derive the store/list `scope` field descriptions and `validate_subject_scope`'s error text from the enum instead of hand-retyping the org-membership caveat in four places.
- `AgentTools.list_memories`/`store_memory` now catch the platform's org-scope-rejected 422 specifically and raise actionable `scope="agent"` retry guidance instead of letting it fall through to the generic handler, which dumped the raw REST error body (`headers: ..., status_code: 422, body: ...`) at the LLM.

Root-caused via live reproduction against the E2E platform + verification against actual `thenvoi-platform` source (not just the Linear writeup). Full context in [INT-1307](https://linear.app/thenvoi/issue/INT-1307/add-agent-scoped-memory-support-migrate-baseline-e2e-off-organization). A genuine happy-path organization-scope test needs a persistent test org + admin key that doesn't exist yet — tracked separately in [INT-1308](https://linear.app/thenvoi/issue/INT-1308/provision-a-persistent-e2e-test-organization-admin-key-add-real).

## Test plan

- [x] `uv run pyrefly check`
- [x] `uv run pytest tests/ --ignore=tests/integration/ --ignore=tests/e2e/` — 5084 passed
- [x] Live: `E2E_TESTS_ENABLED=true uv run pytest tests/e2e/baseline/smoke/inspection/test_memory.py` — 11/11 passed
- [x] Live: `test_capability_matrix.py` memory tests, `anthropic` cell — 2/2 passed

https://claude.ai/code/session_01Nyz1uWgNb2yHExFP6yzfk4

### DIFF
--- a/examples/band_mcp/02_claude_agent_sdk_external.py
+++ b/examples/band_mcp/02_claude_agent_sdk_external.py
@@ -96,7 +96,7 @@ async def main() -> None:
     store_reply = await run_turn(
         options,
         "Store this fact as a durable memory using band_store_memory with "
-        f'system="long_term", type="semantic", segment="agent", scope="organization": '
+        f'system="long_term", type="semantic", segment="agent", scope="agent": '
         f'"{fact}" Then tell me the memory id you got back.',
     )
     logger.info("Claude (session 1): %s", store_reply)
@@ -106,7 +106,7 @@ async def main() -> None:
     )
     recall_reply = await run_turn(
         options,
-        "Use band_list_memories to search your organization-scoped semantic memories "
+        "Use band_list_memories to search your agent-scoped semantic memories "
         f'for content containing "stdio-{nonce}", then tell me the passphrase you found.',
     )
     logger.info("Claude (session 2): %s", recall_reply)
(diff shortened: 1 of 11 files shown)


amit-gazal-band (approved):
LGTM

PR #583, opened 2026-08-26
refactor(sdk): delegate delivery state to band-sdk-core

AlexanderZ-Band:
## Summary

- Delegate `ExecutionContext` claim, retry, and participant-roster state to `band-sdk-core==0.7.2`; Python retains I/O, asyncio orchestration, handlers, and callbacks.
- Preserve `MessageRetryTracker` as a public compatibility facade.
- Fix pending-ACK ordering during in-process cancellation and keep participant context synchronized with the core roster.

## Validation

- Runtime/session: 768 passed
- Non-live suite: 5127 passed, 122 skipped
- Full GitHub CI passes on Linux/Windows, Python 3.11/3.12, packaging, lint, and CodeQL.

Requires the published [band-sdk-core#37](https://github.com/band-ai/band-sdk-core/pull/37) validation fix.


tests/runtime/test_execution_interrupt.py
```diff
@@ -487,3 +488,120 @@ async def test_interrupt_between_cycles_arms_nothing(self, mock_link):
...
+
+        return AsyncMock(side_effect=_gate)
+
+    async def test_websocket_path_cancellation_during_mark_processed(self, mock_link):
+        entered, release = asyncio.Event(), asyncio.Event()
+        mock_link.mark_processed = self._gated_mark_processed(entered, release)
+        handler = BlockingHandler(block=False)
+        ctx = ExecutionContext("room-123", mock_link, handler, agent_id="agent-123")
+
+        proc = asyncio.create_task(
```
AlexanderZ-Band: isnt there a context manager for tion with auto close?

tests/runtime/test_execution.py
```diff
@@ -2141,3 +2344,27 @@ def test_falls_back_to_class_name_when_whitespace_only(self):
 
     def test_strips_surrounding_whitespace(self):
         assert _error_label(ValueError("  trimmed  ")) == "trimmed"
+
+
+class TestBandSdkCoreConstructorValidation:
```
AlexanderZ-Band, line 2372: is this check really meaningful? what it really test? keep only if you think it is real test

src/band/runtime/types.py
```diff
@@ -93,7 +93,12 @@ class SessionConfig:
     enable_context_cache: bool = True
     context_cache_ttl_seconds: int = 300
     max_context_messages: int = 100
-    max_message_retries: int = 1  # Max attempts per message before permanently failing
+    # Max attempts per message before permanently failing. Passed straight into
```
AlexanderZ-Band: feels like a really long coment

src/band/runtime/tools.py
```diff
@@ -2135,9 +2135,20 @@ async def add_participant(
...
+        # Sync back to ExecutionContext so future turns see the update. The
+        # REST add above already succeeded server-side, so a local roster
+        # failure (a new failure surface as of this migration) must not fail
+        # this tool call out from under a real, already-applied change.
         if self._ctx is not None:
-            self._ctx.add_participant(new_participant)
+            try:
+                self._ctx.add_participant(new_participant)
+            except (ValueError, TypeError) as err:
+                log_roster_error(
```
AlexanderZ-Band: this error handling code is repeated a lot, can we clean the code to be DRY? consider modern pythonic options

src/band/runtime/retry_tracker.py
```diff
@@ -1,34 +1,36 @@
...
 from __future__ import annotations
 
 import logging
 
+from band_sdk_core import RetryTracker
+
 logger = logging.getLogger(__name__)
 
 
 class MessageRetryTracker:
```
AlexanderZ-Band, line 12: this class can benefit from some pythonic sugar like context manager or other techniques .
we need it to show more intent smells of too much plumbing

src/band/runtime/participants.py
```diff
@@ -14,25 +18,32 @@ def participant_snapshot(participant: Mapping[str, Any]) -> dict[str, Any]:
...
-    the key (plausible for the participants-list endpoint, precisely the
-    sparse source this merge exists to defend against).
+def log_roster_error(
+    logger_: logging.Logger, *, room_id: str, action: str, err: Exception
+) -> None:
+    """Log a ``band_sdk_core.ParticipantRoster`` failure with its structured fields.
+
+    ``ParticipantRoster.add``/``set_all`` are real, new failure surfaces at
+    call sites that never raised before this migration (a plain ``TypeError``
+    for a malformed field, or a ``ValueError`` with ``.issues``/
```
AlexanderZ-Band: more comment then code why?

src/band/runtime/execution.py
```diff
@@ -1875,12 +1923,30 @@ async def _process_event_body(
...
             elif isinstance(event, ParticipantRemovedEvent) and event.payload:
-                self.remove_participant(event.payload.id)
+                try:
+                    self.remove_participant(event.payload.id)
+                except (ValueError, TypeError) as err:
+                    log_roster_error(
+                        logger,
+                        room_id=self.room_id,
+                        action="remove participant",
+                        err=err,
```
AlexanderZ-Band: functions and files should be single consern, consider where it worth a refactor

src/band/runtime/execution.py
```diff
@@ -1875,12 +1923,30 @@ async def _process_event_body(
...
+            # a malformed field, where the old Python merge never raised.
+            # Never let it crash the delivery cycle -- log and keep the
+            # previous roster state; the callback still fires either way
+            # since it reports the platform event, not local roster state.
             if isinstance(event, ParticipantAddedEvent) and event.payload:
-                self.add_participant(event.payload.model_dump())
+                try:
+                    self.add_participant(event.payload.model_dump())
+                except (ValueError, TypeError) as err:
+                    log_roster_error(
```
AlexanderZ-Band: maybe a contextmanager to handle this repeated error handler? or anything else to keep DRY

src/band/runtime/execution.py
```diff
@@ -1875,12 +1923,30 @@ async def _process_event_body(
             # history only if enable_context_hydration is True)
             await self._ensure_fresh_context()
 
-            # Handle participant events internally
+            # Handle participant events internally. A new failure surface as
+            # of this migration: ParticipantRoster.add can raise TypeError for
+            # a malformed field, where the old Python merge never raised.
```
AlexanderZ-Band: comment too long

src/band/runtime/execution.py
```diff
@@ -457,6 +463,24 @@ async def _retry_processed_ack(self, message_id: str) -> bool:
 
         return False
 
+    @contextlib.contextmanager
+    def _claim_message(self, message_id: str) -> Iterator[bool]:
+        """Yield whether an in-flight claim was acquired; release iff acquired.
+
+        The sole SDK ergonomic adapter over ``ClaimRegistry.try_claim``/
+        ``.release`` (band_sdk_core has no context-manager equivalent by
```
AlexanderZ-Band: comment only what is not obvious from code

tests/runtime/test_execution_interrupt.py
AlexanderZ-Band: Good catch -- the gated coroutine parks on an Event nothing else sets, so an assertion failing before the explicit `proc.cancel()` would leak a task that never finishes on its own. Added a small `_running()` asynccontextmanager that cancels+drains the task on any exit, wrapping both tests (239386f).

tests/runtime/test_execution.py
AlexanderZ-Band: Kept it -- this is the actual regression guard for INT-1315 (band-sdk-core's `RetryTracker`/`ClaimRegistry` raised a bare `OverflowError` instead of `ValueError` for out-of-range constructor args, which I found and fixed upstream: band-ai/band-sdk-core#37). It runs against the real installed `band_sdk_core` artifact, not a mock, so it's the only thing that would catch a future band-sdk-core downgrade or regression silently reintroducing that bug. Did tighten it in the /simplify pass just before this -- was 4 near-identical one-line tests, now one parametrized test.

src/band/runtime/types.py
AlexanderZ-Band: Trimmed to one line (239386f).

src/band/runtime/tools.py
AlexanderZ-Band: Already fixed just before you posted this -- I ran /simplify on the diff and one of its findings was exactly this. Extracted `apply_roster_change()` in `participants.py`; this site and 3 others (the two `_process_event_body` handlers, and `get_participants`' sync-back) now each collapse to one call (aed8e1b).

src/band/runtime/retry_tracker.py
AlexanderZ-Band: Pushed back on this one, but considered it seriously: `RetryTracker`'s operations (record_attempt/mark_success/discard_attempt/mark_permanently_failed) aren't acquire/release pairs -- they're independent calls scattered across a cycle's lifecycle (attempt at cycle start, success or failure at different points afterward, discard on an aborted control signal). There's no single `with` block that would wrap them meaningfully. This class is also deliberately kept as a thin, honestly-labeled compatibility facade (CLAUDE.md: `ExecutionContext` must not even use it) -- adding sugar here would dress up plumbing that's supposed to read as plumbing, not hide it. Happy to revisit with a concrete shape if you have one in mind.

src/band/runtime/participants.py
AlexanderZ-Band: Trimmed to 3 lines (239386f).

src/band/runtime/execution.py
AlexanderZ-Band: Same fix as the other two comments on this hunk -- /simplify's apply_roster_change() extraction addresses this: each branch is now one call instead of a try/except block inline in _process_event_body (aed8e1b).

src/band/runtime/execution.py
AlexanderZ-Band: Went with a plain function (apply_roster_change) rather than a context manager -- it's a single call wrapped in try/except with no acquire/release pairing to manage, so a cm would just be a fancier way to write the same try/except. Fixed via /simplify just before this comment landed (aed8e1b).

src/band/runtime/execution.py
AlexanderZ-Band: Trimmed -- the try/except this comment was explaining is gone now too (apply_roster_change, aed8e1b), so the down-to-one-line comment doesn't need to carry the old rationale either (239386f).

src/band/runtime/execution.py
AlexanderZ-Band: Trimmed to 2 lines, kept only the non-obvious part (why no cm equivalent in band_sdk_core) -- the rest was restating what the 6-line body already shows (239386f).

src/band/runtime/participants.py
```diff
@@ -14,25 +18,40 @@ def participant_snapshot(participant: Mapping[str, Any]) -> dict[str, Any]:
...
+    the previous roster state rather than propagating -- see
+    :func:`log_roster_error` for why -- so they share this one try/except
+    shape.
     """
-    return {
-        name: snapshot.get(name) or existing.get(name) for name in _PARTICIPANT_FIELDS
-    }
+    try:
+        fn()
+    except (ValueError, TypeError) as err:
```
AlexanderZ-Band, line 54: this is not elegant solution, with log_roster_error: would look much cleaner

src/band/runtime/execution.py
AlexanderZ-Band: Follow-up: after the later review, I adopted the context-manager shape. `apply_roster_change(fn=...)` is now `with log_roster_errors(...):` at all four mutation sites (da56f966), which reads more directly and removes the lambda plumbing.

src/band/runtime/participants.py
AlexanderZ-Band: Agreed. Replaced the callback wrapper with a scoped `with log_roster_errors(...):` context manager at all four roster mutation boundaries, and trimmed the nearby comments (da56f966). This keeps the error policy centralized while leaving each mutation visible at its call site.

amit-gazal-band (approved):
LGTM

PR #585, opened 2026-08-27
feat: integrate band-sdk-core SubscriptionTracker into BandLink

AlexanderZ-Band:
## Summary
- Replace `BandLink`'s hand-rolled `_subscribed_rooms` set with `band_sdk_core.SubscriptionTracker` (already pinned, no core bump needed): ticket-based claims close a TOCTOU race on concurrent `subscribe_room()` calls, and a reconnect-boundary reconciliation set blocks resubscribing a room/agent-topic left ambiguous by cancellation or a failed rollback, until the transport is forced back to a known-clean state.
- Also extends dedup to the three agent-level topics (`agent_rooms`, `agent_contacts`), which previously had zero tracking at all.
- Fixes a pre-existing `RoomPresence` bug (independent of the tracker): a room that failed to subscribe stayed marked as joined in `self.rooms` forever, since `subscribe_room()` is non-raising by design. The new `is_room_subscribed()` query makes the real outcome checkable instead of assuming "no exception" means "subscribed."

## Follow-up (not in this PR)
`band-sdk-typescript`'s `BandLink.ts` has the identical unpatched bug (same bare `Set<string>`, same zero-dedup gap on agent topics) and no TS-side ticket exists yet. This design isn't Python-specific — the wasm bindings expose the same `SubscriptionTracker` API with 1:1 method parity — so a companion TS ticket mirroring this one is recommended.

## Test plan
- [x] `uv run pytest tests/platform/test_link.py tests/platform/test_link_contacts.py tests/runtime/test_presence.py -v` — all green, including new coverage for concurrent-subscribe races, reconciliation-boundary blocking, gated mid-await cancellation, and the reconnect-boundary best-effort leave
- [x] `uv run --all-packages pytest --ignore=tests/integration/ --ignore=tests/e2e/` — 5123 passed, 122 skipped, 0 failed

Claude-Session: https://claude.ai/code/session_01LccTh6HdVnp4JdQQPN6Vhq

### DIFF
--- a/src/band/platform/link.py
+++ b/src/band/platform/link.py
@@ -56,12 +56,20 @@ if TYPE_CHECKING:
 logger = logging.getLogger(__name__)
 
 
+# Single source of truth for the two single-topic agent-channel kinds: every
+# site that builds a topic string (_agent_*_topic) or parses one back apart
+# (_drain_reconciliation) reads these, so a typo in one can't silently
+# diverge from the other.
+_AGENT_ROOMS_KIND = "agent_rooms"
+_AGENT_CONTACTS_KIND = "agent_contacts"
+
+
 def _agent_rooms_topic(agent_id: str) -> str:
-    return f"agent_rooms:{agent_id}"
+    return f"{_AGENT_ROOMS_KIND}:{agent_id}"
 
 
 def _agent_contacts_topic(agent_id: str) -> str:
-    return f"agent_contacts:{agent_id}"
+    return f"{_AGENT_CONTACTS_KIND}:{agent_id}"
 
 
 class BandLink:
@@ -232,6 +240,40 @@ class BandLink:
 
     # --- Subscription management (from BandAgent) ---
 
+    def _blocked_by_reconciliation(
+        self, key: str, pending: set[str], *, noun: str
+    ) -> bool:
+        """Whether ``key`` is blocked from a fresh subscribe until the next
+        reconnect drains ``pending`` — the single check both subscribe_room
+        and _subscribe_agent_topic gate on (see design doc for why this, not
+        core's own status, is the authoritative block condition)."""
+        if key not in pending:
+            return False
+        logger.warning(
+            "%s %s needs reconciliation, blocking subscribe until next reconnect",
+            noun,
+            key,
+        )
+        return True
+
+    async def _leave_channel(
+        self,
+        leave: Callable[[], Awaitable[None]],
+        *,
+        description: str,
+        level: int = logging.WARNING,
+    ) -> bool:
+        """Attempt one best-effort channel leave: log and swallow any
+        failure, report whether it actually succeeded. Shared by every leave
+        attempt in this module (rollback, unsubscribe, reconciliation drain)
+        so "did we manage to leave this?" has one implementation."""
+        try:
+            await leave()
+            return True
+        except Exception as e:
+            logger.log(level, "Error %s: %s", description, e)
+            return False
+
     async def subscribe_agent_rooms(self, agent_id: str) -> None:
         """
         Subscribe to agent room events (room_added/removed).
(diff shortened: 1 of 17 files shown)


tests/platform/test_link_contacts.py
```diff
@@ -83,38 +87,117 @@ async def test_subscribe_agent_contacts_requires_connection(
...
+        here, proven with a gated coroutine so the cancel lands truly
+        mid-flight."""
+        mock_ws_class.return_value = mock_ws_client
+        side_effect, started, _release = gated_coroutine()
+        mock_ws_client.join_agent_contacts_channel.side_effect = side_effect
+
+        link = BandLink(agent_id="agent-123", api_key="test-key")
+        await link.connect()
+
+        task = asyncio.create_task(link.subscribe_agent_contacts("agent-123"))
```
AlexanderZ-Band: why not context manager?

tests/platform/test_link_contacts.py
```diff
@@ -83,38 +87,117 @@ async def test_subscribe_agent_contacts_requires_connection(
...
+    async def test_cancelled_join_blocks_agent_contacts_until_reconnect(
+        self, mock_ws_class, mock_ws_client
+    ):
+        """record_agent_topic_join(joined=False) never reaches core's own
+        NeedsReconciliation (unlike a room's second-phase rollback failure)
+        — the local reconciliation set is what actually blocks the retry
+        here, proven with a gated coroutine so the cancel lands truly
+        mid-flight."""
+        mock_ws_class.return_value = mock_ws_client
+        side_effect, started, _release = gated_coroutine()
```
AlexanderZ-Band: this can be a fixture with all operations wired in context manager co handle close and cancel and wait assert functions inspire from the toolkit.

show intent hide plumbing we need declarative code in tests

tests/platform/test_link_contacts.py
AlexanderZ-Band: Good catch — pulled it into an async context manager: `cancelled_mid_await(gate, call)` in `tests/platform/conftest.py` (09109762). Wires the gate, starts the task, waits for it to be genuinely in-flight, hands control back so a caller can inject something else while it's still suspended (needed for the disconnect-during-cancelled-subscribe test), then cancels + asserts `CancelledError` + clears the gate on exit. Every call site is now `async with cancelled_mid_await(mock, link.subscribe_room(...)): ...` — intent, not plumbing.

tests/platform/test_link_contacts.py
AlexanderZ-Band: Same fix — added `cancelled_mid_await(gate, call)` as an async context manager in `tests/platform/conftest.py` (09109762), inspired by the toolkit's declarative style. It owns the whole lifecycle (gate wiring, task start, wait-until-in-flight, cancel, assert CancelledError, cleanup) and hands control back mid-flight for the one test that needs to inject something (a concurrent `disconnect()`). Every cancellation test site across both files now reads as `async with cancelled_mid_await(mock, call): ...` instead of the hand-rolled task/wait/cancel/assert-raises sequence.

amit-gazal-band (approved):
LGTM

src/band/platform/link.py
```diff
@@ -55,6 +55,22 @@
...
+_AGENT_ROOMS_KIND = "agent_rooms"
+_AGENT_CONTACTS_KIND = "agent_contacts"
+
+
+def _agent_rooms_topic(agent_id: str) -> str:
+    return f"{_AGENT_ROOMS_KIND}:{agent_id}"
+
+
+def _agent_contacts_topic(agent_id: str) -> str:
+    return f"{_AGENT_CONTACTS_KIND}:{agent_id}"
```
amit-gazal-band, line 71: These should live in the rust core in my opinion.

PR #586, opened 2026-08-27
feat: self-registration boot flow for sandboxed agents

AlexanderZ-Band:
## Summary
- Host-side `band-kit provision` CLI (INT-982): registers a Band agent with a user key, writes the non-secret `agent.id` into `band.yaml`, and injects the minted agent key via `sbx secret set-custom` (stdin only, never argv/history) under the `proxy-managed` sentinel — the real key never enters the sandbox VM, matching INT-981's custody model.
- Host-side idempotent, checked independently per resource: registration is skipped when the workspace already has both a non-placeholder `agent.id` and the injected secret; `sbx create` is skipped separately only when `sbx ls --json` already reports a sandbox by that name — so a `create` that failed transiently on a prior run still gets retried on the next `provision --create` instead of silently no-opping. 403 -> clear plan-cap message, 422 -> defers re-registration/adopt-vs-new semantics to PLT-1066 (still Backlog) rather than hard-coding them.
- Registration uses zero HTTP retries: `register_my_agent` has no idempotency key and returns the agent's key exactly once, so a transport-level retry after the platform already committed the write could hit the duplicate-name path with that key already gone, orphaning an agent against the plan cap with no way to recover its credentials.
- A registration that times out client-side is disambiguated rather than left ambiguous: `provision` looks the agent up by name before failing, so the error says whether the platform actually committed the write (naming the orphaned agent id, whose key is already unrecoverable) or confirms nothing landed (safe to retry).
- A `--name` whose secret scope is already occupied but has no matching local `agent.id` is refused outright rather than silently overwritten: `sbx secret set-custom` is create-or-update, so proceeding would clobber an unrelated sandbox's live Band credential.
- Kit README + echo-agent README document the new self-registration path, including the per-resource idempotency split, the collision guard, and the timeout-disambiguation behavior, alongside the existing pre-provisioned flow.
- `examples/sandbox/self-registration/demo.py`: a dev/validation script (mirrors `examples/sandbox/staging-smoke`'s shape) that runs the whole flow live — `band-kit provision` -> `sbx create` -> room round-trip -> idempotency recheck -> full teardown — with a `--cleanup <name>` recovery path for an interrupted run. Not customer-facing, not run in CI (same `sbx`/nested-virt constraint as its sibling).
- Also fixes a pre-existing, unrelated bug found while live-testing this: `docker/band_python_kit/Dockerfile` never copied any `[tool.uv.workspace]` member's manifest into the build context, so `uv sync --locked` has failed since `band-mcp` joined the workspace (#552) — never caught because `DOCKER_TESTS_ENABLED` defaults off everywhere, including CI. Folded in here rather than a separate PR.
- Also fixes a kit README doc bug caught during review: the cleanup section claimed `sbx rm` removes the scoped Band secret alongside the sandbox — the opposite of what the live test had to work around (a removed sandbox's secret persists in `sbx secret ls`). Now documents the required `sbx secret rm <name> --host <host> -f`.
- `/simplify` pass (4 parallel reviews: reuse, simplification, efficiency, altitude) — no behavior change: collapsed the four sbx-subprocess run+raise blocks and the two `AsyncRestClient` construct/close blocks in `provision.py` into shared helpers; dropped a redundant boolean recheck; imported `DEFAULT_REST_URL` instead of a second, drifted copy; split the timeout-disambiguation helper into a pure classifier + a pure formatter (mirroring `_describe_register_error`); decomposed `run()` into `_ensure_agent_registered`/`_ensure_sandbox_created`; parametrized near-duplicate tests and replaced three sleep-racing timeout tests with fast direct unit tests + one wiring test; `examples/sandbox/self-registration/demo.py` now reuses `test_kit_proxy_managed_live`'s workspace/host helpers and a new shared `allow_network_for_hosts` (`tests/docker/toolkit/sbx_cli.py`) instead of redefining them, and its echo-reply wait is now event-driven (`reply_capture`/`user_ws_observer`, matching `staging-smoke/probe.py`) instead of fixed-cadence REST polling.
  - Deliberately deferred (out of scope for this branch): one review pass flagged that `src/band/runtime/tools.py`'s pre-existing `register_my_agent` platform tool call site still retries (`DEFAULT_REQUEST_OPTIONS`, `max_retries: 3`) the same non-idempotent, key-orphaning endpoint this PR made retry-safe in `provision.py`. That call site predates this branch and isn't touched by it — worth a follow-up, not fixed here.

- Mid-provision-crash gap: if `register_agent` succeeds and `write_agent_id` persists the real id but `inject_agent_key` then fails, a retry previously registered a brand-new agent instead of reconciling — now fails loudly with a recovery pointer instead of silently orphaning the first agent (`_ensure_agent_registered`).
- `_describe_register_error` now reads the platform's real 422 `details` map (verified against `thenvoi-platform`'s actual Ecto changeset vs. schema-cast error shapes) and surfaces the real field error instead of always claiming a name collision.
- Extracted `band.docker.sbx_process.run_sbx_subprocess`, shared by `provision.py` and the test toolkit's `run_redacting_secret`, instead of two independently-drifting copies.
- Extracted `band.client.rest.aclose_rest_client`, shared by `provision.py` and `cli/trigger.py`.
- Added a tripwire test pinning `PLACEHOLDER_AGENT_ID` against the shipped echo-agent `band.yaml`'s literal.
- `demo.py`'s final status line uses `logger.info` instead of `print()`.

All 6 review threads resolved. Verified live against the dev deployment post-fix: real agent registration, secret injection, and sandbox creation all succeeded end to end.

## Test plan
- [x] `uv run pyrefly check`
- [x] `uv run pytest tests/ --ignore=tests/integration/ --ignore=tests/e2e/` (5228 passed, 0 failed)
- [x] `tests/docker/test_provision.py` (50 unit tests: orchestration, per-resource idempotency, name-collision refusal, timeout disambiguation — confirmed-absent / confirmed-orphaned / check-itself-failed, tested directly against the classifier plus one end-to-end wiring test — 403/422 mapping, zero-retry registration, key never on argv, key delivered via stdin, key never in logs, recovery banner on mid-flight persist failure, missing-sandbox recovery on re-provision)
- [x] `docker build -f docker/band_python_kit/Dockerfile -t band-python-kit:local .` — now succeeds (previously failed: "lockfile needs to be updated")
- [x] `uv run pytest tests/docker/` (excluding the three live-gated modules) — 212 passed, 2 skipped
- [x] `uv run pytest --markdown-docs docker/band_python_kit/README.md examples/sandbox/self-registration/README.md`
- [x] **Live E2E, run twice against dev** (`tests/docker/test_kit_self_registration_live.py`, `SANDBOX_TESTS_ENABLED=true E2E_TESTS_ENABLED=true`): a fresh agent with no pre-existing identity is registered by `provision.run()`, a real sandbox boots from it, a room message round-trips through the injection path, and a repeat `provision.run()` call proves no duplicate registration.
  - 1st run passed but leaked the injected `sbx` secret on teardown (sandbox removal alone doesn't remove a scoped custom secret) — fixed the test fixture to reap it explicitly, matching the sibling live test's own pattern.
  - 2nd run passed with zero leftovers, verified by hand: no sandbox, no secret, no network policy rule, no orphaned agent on the platform.
- [x] **`examples/sandbox/self-registration/demo.py` run live multiple times against dev**, including the `--cleanup` recovery path against a deliberately orphaned agent (registered, then recovered and deleted by name) — confirmed clean every time.
- [x] Regression check: the two pre-existing live tests this branch's changes touch (`tests/docker/test_band_python_kit_live.py`, `tests/docker/test_kit_proxy_managed_live.py`) still pass unchanged, against the same rebuilt image.
- [x] **Post-`/simplify` live re-verification against dev**: `examples/sandbox/self-registration/demo.py` (full flow, event-driven reply wait), `tests/docker/test_kit_self_registration_live.py`, and `tests/docker/test_kit_proxy_managed_live.py` all pass against the refactored `provision.py`/`sbx_cli.py`/`demo.py`, with zero leftover sandboxes/secrets confirmed via `sbx ls`/`sbx secret ls` after each run.
- [x] **Post-review-fix live re-run against dev** (`examples/sandbox/self-registration/demo.py`): real agent registration, secret injection via `run_sbx_subprocess`, and sandbox creation all succeeded; run hit an unrelated account-level Human-API plan-tier restriction at the room-creation step (`plan_required`), cleanup ran cleanly including a force-delete of the newly-registered agent.
- Not part of CI: every live/sandbox test here is gated behind `sandbox`/`e2e`/`docker_build` marks by design — `sbx` needs a nested-virtualization-capable host, which GitHub-hosted runners don't have.

https://claude.ai/code/session_01L95pyAWjRJoEEPNwbjxA23


examples/sandbox/self-registration/demo.py
```diff
@@ -0,0 +1,294 @@
...
+    logger.info("Run name: %s", name)
+
+    room_id: str | None = None
+    agent_id: str | None = None
+    try:
+        with tempfile.TemporaryDirectory(prefix="band-selfreg-demo-") as tmp_dir:
+            workspace = _prepare_workspace(Path(tmp_dir), endpoints=settings.endpoints)
+            hosts = _deployment_hosts(settings.endpoints)
+
+            with allow_network_for_hosts(hosts, kit=KIT_DIR):
```
amit-gazal-band: `allow_network_for_hosts()` computes the kit's baseline network allowlist from the hardcoded local `KIT_DIR` checkout instead of the `--kit` value actually passed to `sbx create` a few lines below.

**Failure scenario:** run this demo with `--kit docker.io/bandhq/band-python-kit:X.Y.Z` where the published tag's `spec.yaml` has a narrower `caps.network.allow` than the local repo's current `spec.yaml`. `kit_baseline_hosts(KIT_DIR)` reports the deployment host as already covered, so `allow_network_for_hosts` skips granting it — but the actually-running published image's proxy never had that host allowed, so the sandbox can't reach the Band deployment. Step 3 (send message / await echo) then hangs or times out, misattributing a demo network-config bug to the agent/platform.

Should compute the baseline from the actual `kit` argument, not the hardcoded `KIT_DIR`.

src/band/docker/provision.py
```diff
@@ -0,0 +1,588 @@
...
+        raise ValueError(f"no band.yaml found in workspace: {path}")
+    data = _YAML.load(path.read_text(encoding="utf-8")) or {}
+    return str((data.get("agent") or {}).get("id") or "")
+
+
+def write_agent_id(workspace: Path, agent_id: str) -> None:
+    """Write `agent.id` into `band.yaml`, preserving every comment and the rest
+    of the file (a round-trip load/dump, not a regenerate-from-scratch)."""
+    path = _band_yaml_path(workspace)
+    data = _YAML.load(path.read_text(encoding="utf-8"))
```
amit-gazal-band: `write_agent_id` crashes with an unhandled `AttributeError` instead of a clear error when `band.yaml` is empty/whitespace-only, unlike `read_agent_id` above which guards the same load with `or {}`.

`_YAML.load(path.read_text(...))` returns `None` for an empty or whitespace-only `band.yaml` (verified: `YAML().load('')` -> `None`). `data.setdefault("agent", {})["id"] = agent_id` then raises `AttributeError: 'NoneType' object has no attribute 'setdefault'` instead of the module's usual clear `RuntimeError`/`ValueError`, right after a successful registration — leaving the operator with a raw traceback and (per the surrounding except-Exception handler) the printed key-recovery banner as the only clue anything happened.

src/band/docker/provision.py
```diff
@@ -0,0 +1,588 @@
...
+
+def _describe_register_error(err: ApiError) -> str:
+    """A human-readable message for a failed registration call."""
+    if err.status_code == 403:
+        return "Plan agent cap reached — cannot register another external agent."
+    if err.status_code == 422:
+        return (
+            "An agent with this name already exists. Re-registration semantics "
+            "for a re-created sandbox (new identity vs. adopt-existing) are not "
+            "decided yet — see PLT-1066 (stale-agent lifecycle). Pick a new "
```
amit-gazal-band: Runtime string literals embed Linear-style issue-tracker IDs (`PLT-1066` here and at line 430; `INT-981` in the module docstring at lines 5 and 10), violating this repo's CLAUDE.md rule: "Never put issue-tracker references in code — no Linear issue IDs (e.g. INT-123), Linear URLs, or ticket numbers in comments, docstrings, or strings."

This message is shown directly to CLI users on a 422 registration error. The same pattern (`INT-982`) also appears in `examples/sandbox/self-registration/demo.py`, its README, `test_kit_self_registration_live.py`, and `test_provision.py` (which even asserts on the literal string `"PLT-1066"`). Anyone updating or resolving the referenced ticket leaves permanently stale ticket IDs baked into user-facing output and tests.

amit-gazal-band (changes requested):
Automated code review found 3 issues worth addressing before merge.

examples/sandbox/self-registration/demo.py
AlexanderZ-Band: Fixed: demo.py now passes `kit=kit` (the actual --kit argument) instead of the hardcoded `KIT_DIR`, so the network allowlist is computed from the kit that's actually running.

src/band/docker/provision.py
AlexanderZ-Band: Fixed: `write_agent_id` now guards the load with `or {}`, matching `read_agent_id`. An empty/whitespace band.yaml now fails loudly with the module's existing clear `LaunchError` (via the round-trip reload) instead of a raw AttributeError. Added a regression test.

src/band/docker/provision.py
AlexanderZ-Band: Fixed: dropped the Linear issue IDs (INT-981, INT-982, PLT-1066) from all runtime strings, docstrings, and tests in this diff, per CLAUDE.md's rule. The user-facing 422 message and timeout message now explain the undecided semantics in plain terms with no ticket reference.

src/band/docker/provision.py
```diff
@@ -0,0 +1,630 @@
...
+    if has_local_identity and already_has_secret:
+        logger.info(
+            "Sandbox %s already registered (agent.id=%s, secret present) — "
+            "skipping registration",
+            args.name,
+            existing_id,
+        )
+        return existing_id
+
+    if has_local_identity:
```
amit-gazal-band, line 534: This handles restart (secret + id both persisted — skip) and same-name re-registration (422, per PLT-1066), but there's a third case neither of those two cover: a **mid-provision crash**. If `register_agent` succeeds, `write_agent_id` persists the real id, but `inject_agent_key` then fails transiently, a retry sees `has_local_identity=True` and `already_has_secret=False` and falls into this branch — which registers a **brand-new agent** instead of reconciling the existing one. The first agent's key was shown exactly once and is now permanently unrecoverable, and it still counts against the plan's agent cap. Worth reconciling (e.g. detect this state and fail loudly with a recovery pointer, rather than silently minting a second agent) rather than treating it as the same case as a normal restart.

src/band/docker/provision.py
```diff
@@ -0,0 +1,630 @@
...
+        help="Enable verbose logging",
+    )
+    return parser
+
+
+def _describe_register_error(err: ApiError) -> str:
+    """A human-readable message for a failed registration call."""
+    if err.status_code == 403:
+        return "Plan agent cap reached — cannot register another external agent."
+    if err.status_code == 422:
```
amit-gazal-band, line 214: This maps every 422 to a fixed "agent already exists, pick a new name" message. That's right for the by-design same-name collision case, but the platform also returns 422 for other registration validation failures (e.g. a `--description` shorter than the platform's minimum — there's no client-side length check in `build_parser`). In that case the user is told to change `--agent-name`/`--name`, which does nothing, while the real, actionable API message (available via `err.body.error.message`, per the fallback branch a few lines below) is discarded.

src/band/docker/provision.py
```diff
@@ -0,0 +1,630 @@
...
+SBX_CREATE_TIMEOUT_S: Final[int] = 600
+# Bound on the best-effort lookup that disambiguates a registration timeout
+# (see _describe_registration_timeout); independent of --timeout since this
+# is a plain read, not the write whose outcome is in question.
+TIMEOUT_RECOVERY_CHECK_S: Final[int] = 30
+
+# The echo-agent starter's shipped-but-unset marker (docker/band_python_kit/
+# echo-agent/band.yaml); a workspace still carrying it has never been
+# provisioned even though `agent.id` is non-empty.
+PLACEHOLDER_AGENT_ID = "replace-with-agent-id"
```
amit-gazal-band: `PLACEHOLDER_AGENT_ID` re-types the exact literal that lives in `docker/band_python_kit/echo-agent/band.yaml`'s `agent.id` field, with no shared constant tying producer and consumer together. Per the repo's single-source-of-truth standard: a future edit to the starter template's placeholder string would silently break `has_local_identity` detection with no test or type error to catch it.

src/band/docker/provision.py
```diff
@@ -0,0 +1,630 @@
...
+    targets_idx = header_cols.index("TARGETS")
+    targets = set()
+    for line in lines[header_idx + 1 :]:
+        cols = re.split(r"\s{2,}", line.strip())
+        if len(cols) > targets_idx:
+            targets.add(cols[targets_idx])
+    return targets
+
+
+def _run_sbx(
```
amit-gazal-band: `_run_sbx` reimplements `tests/docker/toolkit/sbx_cli.py`'s `run_redacting_secret` (subprocess.run + non-zero-exit RuntimeError + secret redaction) as an independent second copy, since `src/` can't import `tests/`. A future fix to one (e.g. also redacting from a raised `TimeoutExpired`'s output) won't propagate to the other, so the two can silently drift on what they consider safe to expose in an error message. Worth pulling the shared logic somewhere both can import from.

src/band/docker/provision.py
```diff
@@ -0,0 +1,630 @@
...
+async def _rest_client(
+    *, api_key: str, rest_url: str
+) -> AsyncIterator[AsyncRestClient]:
+    """A short-lived `AsyncRestClient`, closed on exit -- this CLI never keeps
+    one alive across calls."""
+    client = AsyncRestClient(api_key=api_key, base_url=rest_url.rstrip("/"))
+    try:
+        yield client
+    finally:
+        await client._client_wrapper.httpx_client.httpx_client.aclose()
```
amit-gazal-band: This closes the client via three levels of private attributes (`client._client_wrapper.httpx_client.httpx_client.aclose()`), duplicating the identical chain already in `src/band/cli/trigger.py:282`. If `band_rest`'s internal wrapper structure changes in a future dependency bump, both call sites break independently. A shared close-helper would only need updating once.

examples/sandbox/self-registration/demo.py
```diff
@@ -0,0 +1,294 @@
...
+                rerun_id = _band_kit_provision(
+                    name=name, workspace=workspace, create=False
+                )
+                if rerun_id != agent_id:
+                    raise RuntimeError(
+                        f"expected the idempotent no-op to return {agent_id!r}, got {rerun_id!r}"
+                    )
+                logger.info("Confirmed idempotent: no duplicate agent registered")
+
+        print(f"\nSuccess: agent {agent_id} self-registered and round-tripped.\n")
```
amit-gazal-band: This uses `print()` while every other line in the file uses `logger.info` (e.g. lines 237, 253, 263), which violates the repo's "never use print(), use logging" convention. Unlike `provision.py`'s CLI stdout output (which has an explicit carve-out as user-facing CLI output), this demo script has no such exception — should be `logger.info`.

amit-gazal-band (changes requested):
Reviewed against INT-982 and Alex's decision comment there (option (b), host-side/never-in-VM; restart dedup via host secret; adopt-vs-new deferred to PLT-1066; `sbx rm` leak deferred to PLT-1066 and documented). Those design tradeoffs are sound and intentional. The inline comments below are code-level issues independent of that decision — one is a real correctness gap in the idempotency path, the rest are smaller cleanups.

One more, not left inline because it's in a file this PR doesn't touch (`src/band/docker/launcher/config.py`, so GitHub won't anchor a diff comment there), but directly relevant to this feature: `resolve_agent_id` (line 250) only checks that `agent.id` is non-empty — it never rejects `provision.py`'s new `PLACEHOLDER_AGENT_ID` sentinel (`"replace-with-agent-id"`, the starter kit's shipped-but-unprovisioned value in `docker/band_python_kit/echo-agent/band.yaml`). A workspace that was never run through `band-kit provision` boots anyway with the literal placeholder as its agent id, producing a confusing downstream REST 401/404 instead of the clear pre-flight `LaunchError` this new sentinel seems intended to enable. Worth a follow-up (in this PR or a fast-follow) to have `resolve_agent_id` reject the placeholder explicitly.

src/band/docker/provision.py
AlexanderZ-Band: Fixed in 27de47b — this state now fails loudly instead of silently registering a second agent: `_ensure_agent_registered` raises with a recovery pointer (manually re-inject the key if you still have it, or clear `agent.id` and pick a new `--agent-name`) whenever a local `agent.id` exists with no injected secret, since that's indistinguishable from a mid-provision crash. Added `test_refuses_when_secret_missing_despite_saved_id` covering it.

src/band/docker/provision.py
AlexanderZ-Band: Fixed in 27de47b. Verified against the real platform (band-sdk-python and the thenvoi-platform backend source): a 422 body is an untyped dict, and the 'details' map is keyed differently depending on where the failure happens -- a schema-level cast failure (e.g. --description too short) uses a JSON-pointer key like '/agent/description', while the field's own Ecto changeset validation (the name-uniqueness collision) uses the bare field name 'name'. `_describe_register_error` now surfaces the real per-field message for anything other than a name-collision, and only falls back to the fixed 'already exists' text when the only failing field is the name (or the body is unclassifiable, matching the historical default). Added tests for both shapes.

src/band/docker/provision.py
AlexanderZ-Band: Fixed in 27de47b -- added a tripwire test (`test_starter_ships_the_placeholder_id_provision_detects` in test_echo_agent.py) asserting `PLACEHOLDER_AGENT_ID` matches the shipped echo-agent band.yaml's literal. provision.py can't import that template at runtime (it isn't packaged with the installed SDK), so a test comparing the two is the single source of truth this can have.

src/band/docker/provision.py
AlexanderZ-Band: Fixed in 27de47b -- extracted `band.docker.sbx_process.run_sbx_subprocess` as the one place both `provision.py` and `tests/docker/toolkit/sbx_cli.py`'s `run_redacting_secret` run+redact an sbx subprocess (src/ can't import tests/, so the shared code lives in src/ and the test toolkit imports it). `run_redacting_secret` is now a thin adapter over it.

src/band/docker/provision.py
AlexanderZ-Band: Fixed in 27de47b -- extracted `band.client.rest.aclose_rest_client`; both `provision.py` and `cli/trigger.py` call it now instead of each reaching through the private attribute chain.

examples/sandbox/self-registration/demo.py
AlexanderZ-Band: Fixed in 27de47b -- swapped to `logger.info`.

amit-gazal-band (dismissed):
All six inline findings from the previous round are fixed in 27de47b — thanks. Approving.

One outstanding, non-blocking item that couldn't be left as an inline comment because `src/band/docker/launcher/config.py` isn't part of this PR's diff: `resolve_agent_id` (line 250) only checks that `agent.id` is non-empty — it never rejects `provision.py`'s `PLACEHOLDER_AGENT_ID` sentinel (`"replace-with-agent-id"`, the starter kit's shipped-but-unprovisioned value in `docker/band_python_kit/echo-agent/band.yaml`). A workspace that was never run through `band-kit provision` currently boots anyway with the literal placeholder as its agent id, producing a confusing downstream REST 401/404 instead of the clear pre-flight `LaunchError` this new sentinel seems intended to enable. Worth a fast-follow to have `resolve_agent_id` reject the placeholder explicitly (sharing `PLACEHOLDER_AGENT_ID`, or an equivalent check, from `provision.py`).

amit-gazal-band (dismissed):

amit-gazal-band (approved):

PR #589, opened 2026-08-27
feat: integrate band-sdk-core RoomRoster into RoomPresence

AlexanderZ-Band:
## Summary

Replaces `RoomPresence`'s hand-rolled room-lifecycle tracking (`self.rooms: Set[str]`, the early-claim dedup in `_join_room`, the reconnect set-diffing in `_handle_reconnect`) with `band_sdk_core.RoomRoster` — an explicit `Unadmitted`/`Admitting`/`Admitted` state machine with ticket-based admission claims and a `reconcile()` call that replaces the manual `old_rooms & current_room_ids` set diffing.

Not a pure lift-and-shift — three deliberate corrections land alongside it:

- **`_handle_room_left` no longer fires `on_room_left` for a room that was never actually admitted.** `record_room_removed`'s `was_tracked` return gates the notification.
- **`auto_subscribe_existing=False` no longer leaks an admission ticket on reconnect.** `reconcile()` atomically claims a ticket for every `Unadmitted` room it's given, with no way to undo that claim afterward — so the reconnect snapshot is now pre-filtered by `auto_subscribe_existing` before it ever reaches `reconcile()`, instead of gating only what happens with the result.
- **`start()` now rejects a concurrent/repeated call** instead of silently overwriting `self._event_task` and orphaning the previous one (a second call previously leaked a background task that `stop()` could never cancel).

Also fixes a latent bug: `_join_room` is split into a synchronous claim (`_join_room`) and an async resolve (`_complete_room_admission`), so a cancelled or failed `subscribe_room()` call now always rolls the ticket back to `Unadmitted` via a bare `finally` — previously, cancellation during `subscribe_room()` left the room stuck "joined" in `self.rooms` forever.

Six near-duplicate "invoke an optional callback, log-and-swallow on failure" call sites (inconsistent log levels between copies) are consolidated into one `_notify` helper with an explicit level/`exc_info` per call site.

Refs INT-1302.

- `_complete_room_admission` now catches `CancelledError` specifically around the `subscribe_room` await and attempts a best-effort unsubscribe before re-raising — a join cancelled after reaching the server previously left a live transport subscription with nothing in the roster to ever clean it up. Strengthened the existing cancellation regression test to assert the unsubscribe actually happens, not just the roster state.
- `_subscribe_rooms` pre-filters rooms a concurrent `room_added` event already admitted, instead of counting `_join_room`'s benign no-op as a failed admission in the results log.
- Pushed back on collapsing `stop()`'s `tracked_room_ids()` + `clear()` into just `clear()`'s return value: `clear()` includes every tracked room regardless of membership (an in-flight `Admitting` one too), which broke `test_stop_does_not_notify_for_a_room_still_admitting` — left as two calls with a comment explaining why.

All 3 review threads resolved.

## Test plan

- [x] `uv sync --extra dev --all-packages`
- [x] `uv run pytest tests/runtime/test_presence.py tests/runtime/test_presence_real_wire.py tests/runtime/test_presence_contacts.py tests/runtime/test_resync.py -v` — 63 passed
- [x] `uv run pytest tests/ --ignore=tests/integration/ --ignore=tests/e2e/ -v` — 5181 passed, 122 skipped
- [x] New `tests/runtime/test_presence_real_wire.py`: one test proves the one seam a fully-mocked suite can't — a real severed WebSocket connection triggering `BandLink`'s actual reconnect supervisor, firing a real `ReconnectedEvent` through the real event queue, dispatched by `RoomPresence`'s real event-consumer task (every prior reconnect test called `_handle_reconnect()` directly)
- [x] New regression coverage for: cancellation-mid-subscribe rollback (now also asserting the best-effort unsubscribe fires), concurrent double-claim on one room, never-admitted room_removed no-op, `stop()` mid-admission no-op, failed reconnect admission not tracked, stale `room_added` after reconnect admission is a no-op, `auto_subscribe_existing=False` ticket-leak fix, `start()` concurrent-call guard, `_subscribe_rooms` skipping an already-admitted room

https://claude.ai/code/session_017QPgnbSx6F19CiHZ1XsaE5

### DIFF
--- a/src/band/platform/link.py
+++ b/src/band/platform/link.py
@@ -117,6 +117,9 @@ class BandLink:
         # WebSocket client (from BandAgent._ws_client)
         self._ws: WebSocketClient | None = None
         self._is_connected = False
+        # True for the span of connect() before _ws is assigned, so a second
+        # concurrent connect() sees it even mid-handshake (see connect()).
+        self._connecting = False
 
         # Subscription tracking (band_sdk_core.SubscriptionTracker) plus local
         # bookkeeping for claims whose real-world outcome is ambiguous (a
@@ -179,36 +182,45 @@ class BandLink:
 
         Extracted from BandAgent.start() lines 158-164.
         """
-        if self._ws is not None:
+        if self._connecting or self._ws is not None:
             logger.warning("Already connected or connecting")
             return
 
-        self._last_disconnect_reason = None
-        ws = WebSocketClient(
-            self.ws_url,
-            self.api_key,
-            self.agent_id,
-            on_reconnect=self._on_reconnected,
-            on_disconnect=self._on_disconnected,
-        )
-        self._ws = ws
+        # Set synchronously, before the first await below, so a second
+        # concurrent connect() sees it immediately -- _ws alone can't do
+        # that job, since it's only assigned once fully connected below
+        # (never a half-connected client a caller could act on).
+        self._connecting = True
         try:
-            await ws.__aenter__()
-            await ws.join_agent_control_channel(
+            self._last_disconnect_reason = None
+            ws = WebSocketClient(
+                self.ws_url,
+                self.api_key,
                 self.agent_id,
-                on_supersede=self._on_supersede,
-                on_control=self._on_control,
+                on_reconnect=self._on_reconnected,
+                on_disconnect=self._on_disconnected,
             )
-        except BaseException:
-            # BaseException, not Exception: a cancellation reaching this
-            # await (e.g. the caller's own task being cancelled) must roll
-            # _ws back too, or every later connect() sees it as non-None and
-            # silently no-ops forever with _is_connected still False.
-            await ws.__aexit__(None, None, None)
-            self._ws = None
-            raise
-        self._is_connected = True
-        logger.info("Connected to platform")
+            try:
+                await ws.__aenter__()
+                await ws.join_agent_control_channel(
+                    self.agent_id,
+                    on_supersede=self._on_supersede,
+                    on_control=self._on_control,
+                )
+            except BaseException:
+                # BaseException, not Exception: a cancellation reaching one
+                # of these awaits (e.g. the caller's own task being
+                # cancelled) must still close the half-opened client, or it
+                # leaks -- _ws itself was never assigned, so there's nothing
+                # to roll back on that side.
+                await ws.__aexit__(None, None, None)
+                raise
+
+            self._ws = ws
+            self._is_connected = True
+            logger.info("Connected to platform")
+        finally:
+            self._connecting = False
 
     async def disconnect(self) -> None:
         """
--- a/src/band/platform/link.py
+++ b/src/band/platform/link.py
@@ -184,22 +184,27 @@ class BandLink:
             return
 
         self._last_disconnect_reason = None
-        self._ws = WebSocketClient(
+        ws = WebSocketClient(
             self.ws_url,
             self.api_key,
             self.agent_id,
             on_reconnect=self._on_reconnected,
             on_disconnect=self._on_disconnected,
         )
-        await self._ws.__aenter__()
+        self._ws = ws
         try:
-            await self._ws.join_agent_control_channel(
+            await ws.__aenter__()
+            await ws.join_agent_control_channel(
                 self.agent_id,
                 on_supersede=self._on_supersede,
                 on_control=self._on_control,
             )
-        except Exception:
-            await self._ws.__aexit__(None, None, None)
+        except BaseException:
+            # BaseException, not Exception: a cancellation reaching this
+            # await (e.g. the caller's own task being cancelled) must roll
+            # _ws back too, or every later connect() sees it as non-None and
+            # silently no-ops forever with _is_connected still False.
+            await ws.__aexit__(None, None, None)
             self._ws = None
             raise
         self._is_connected = True
--- a/src/band/runtime/presence.py
+++ b/src/band/runtime/presence.py
@@ -9,7 +9,9 @@ from __future__ import annotations
 
 import asyncio
 import logging
-from typing import Any, Awaitable, Callable, Set
+from typing import Any, Awaitable, Callable
+
+from band_sdk_core import RoomMembership, RoomRoster
 
 from band.client.rest import DEFAULT_REQUEST_OPTIONS
 from band.platform.event import (
@@ -87,7 +89,7 @@ class RoomPresence:
         self.auto_subscribe_existing = auto_subscribe_existing
 
         # Track rooms we're present in
-        self.rooms: Set[str] = set()
+        self.roster = RoomRoster()
 
         # Callbacks (set by user or AgentRuntime)
         self.on_room_joined: Callable[[str, dict], Awaitable[None]] | None = None
@@ -102,6 +104,23 @@ class RoomPresence:
         # Internal task for consuming events from link
         self._event_task: asyncio.Task | None = None
 
+    async def _notify(
+        self,
+        callback: Callable[..., Awaitable[None]] | None,
+        *args: Any,
+        label: str,
+        level: int = logging.WARNING,
+        exc_info: bool = False,
+    ) -> None:
+        """Invoke an optional user callback; a raise there is the caller's
+        problem, never ours to propagate."""
+        if callback is None:
+            return
+        try:
+            await callback(*args)
+        except Exception as e:
+            logger.log(level, "%s callback error: %s", label, e, exc_info=exc_info)
+
     async def start(self) -> None:
         """
         Start presence management.
@@ -111,6 +130,12 @@ class RoomPresence:
         3. Subscribe to existing rooms (if configured)
         4. Spawn task to consume events from link
         """
+        if self._event_task is not None and not self._event_task.done():
+            raise RuntimeError(
+                f"RoomPresence for agent {self.link.agent_id} is already running; "
+                "call stop() before starting again"
+            )
+
         # Connect if needed
         if not self.link.is_connected:
             await self.link.connect()
@@ -153,15 +178,16 @@ class RoomPresence:
                 pass
             self._event_task = None
 
-        # Notify left for all rooms
-        for room_id in list(self.rooms):
-            if self.on_room_left:
-                try:
-                    await self.on_room_left(room_id)
-                except Exception as e:
-                    logger.warning("on_room_left error for %s: %s", room_id, e)
-
-        self.rooms.clear()
+        admitted_room_ids = self.roster.tracked_room_ids()
+        self.roster.clear()
+        for room_id in admitted_room_ids:
+            try:
+                await self.link.unsubscribe_room(room_id)
+            except Exception as e:
+                logger.warning(
+                    "Failed to unsubscribe room %s during stop: %s", room_id, e
+                )
+            await self._notify(self.on_room_left, room_id, label="on_room_left")
         logger.info("RoomPresence stopped")
 
     async def _on_platform_event(self, event: PlatformEvent) -> None:
@@ -235,19 +261,21 @@ class RoomPresence:
             logger.warning("%s event without room_id", event.type)
             return
 
-        # Unsubscribe from room channels
+        # Unconditional, harmless no-op if never subscribed.
         await self.link.unsubscribe_room(room_id)
+        if not self.roster.record_room_removed(room_id):
+            logger.debug(
+                "%s event for untracked room %s, ignoring", event.type, room_id
+            )
+            return
 
-        # Untrack room
-        self.rooms.discard(room_id)
-
-        # Notify callback
-        if self.on_room_left:
-            try:
-                await self.on_room_left(room_id)
-            except Exception as e:
-                logger.error("on_room_left error for %s: %s", room_id, e, exc_info=True)
-
+        await self._notify(
+            self.on_room_left,
+            room_id,
+            label="on_room_left",
+            level=logging.ERROR,
+            exc_info=True,
+        )
         logger.info("Agent left room via %s: %s", event.type, room_id)
 
     async def _handle_reconnect(self) -> None:
@@ -264,8 +292,6 @@ class RoomPresence:
         transient API failure.
         """
         logger.info("Handling reconnection — syncing rooms from API")
-        old_rooms = self.rooms.copy()
-
         try:
             try:
                 rooms_from_api = await self._list_existing_rooms()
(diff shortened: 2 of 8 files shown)


src/band/runtime/presence.py
```diff
@@ -102,6 +104,23 @@ def __init__(
         # Internal task for consuming events from link
         self._event_task: asyncio.Task | None = None
 
+    async def _notify(
```
AlexanderZ-Band, line 115: feels like it should be context manager

src/band/runtime/presence.py
```diff
@@ -153,15 +178,16 @@ async def stop(self) -> None:
...
-                    await self.on_room_left(room_id)
-                except Exception as e:
-                    logger.warning("on_room_left error for %s: %s", room_id, e)
-
-        self.rooms.clear()
+        admitted_room_ids = self.roster.tracked_room_ids()
+        self.roster.clear()
+        for room_id in admitted_room_ids:
+            try:
+                await self.link.unsubscribe_room(room_id)
```
AlexanderZ-Band: cant we asyncio gater here?

tests/runtime/test_presence.py
```diff
@@ -122,29 +127,47 @@ async def test_start_subscribes_existing_rooms(self, mock_link, presences):
...
+        presence = presences(auto_subscribe_existing=False)
+        await presence.start()
+        first_task = presence._event_task
+
+        with pytest.raises(RuntimeError):
+            await presence.start()
+
+        assert presence._event_task is first_task
+
+        await presence.stop()
```
AlexanderZ-Band, line 149: smells lice a case for contextmanager  or a fixture

tests/runtime/test_presence.py
```diff
@@ -389,63 +465,64 @@ async def on_left(room_id):
...
+    async def test_room_removed_for_a_never_admitted_room_does_not_notify(
+        self, mock_link, presences
+    ):
+        """A room_removed/room_deleted for a room we never actually admitted
+        must not fire on_room_left — record_room_removed's was_tracked
+        return exists specifically to gate this."""
+        presence = presences(auto_subscribe_existing=False)
+        await presence.start()
+        on_left = AsyncMock()
+        presence.on_room_left = on_left
```
AlexanderZ-Band: this smeels like plumbing can we sugger coate it?

tests/runtime/test_presence.py
```diff
@@ -495,6 +566,87 @@ async def on_event(room_id, event):
...
+
+        assert presence.roster.tracked_room_ids() == []
+        assert received == []
+
+    async def test_a_stale_room_added_after_reconnect_admission_is_a_noop(
+        self, mock_link, presences
+    ):
+        """A room_added delivered after reconcile() already admitted the
+        same room via its own admitting list must be a harmless no-op — not
+        a second subscribe or announce. Proves the TOCTOU self-healing the
```
AlexanderZ-Band: show intent in code not in comments

tests/runtime/test_presence.py
```diff
@@ -495,6 +566,87 @@ async def on_event(room_id, event):
...
+        )
+        await presence._handle_reconnect()
+
+        assert mock_link.subscribe_room.call_count == 1
+        assert len(joined) == 1
+
+        await presence._handle_room_added(make_room_added_event(room_id="room-new"))
+
+        assert mock_link.subscribe_room.call_count == 1
+        assert len(joined) == 1
```
AlexanderZ-Band, line 702: are we testing real life casses not only syntetic mocked flows?

tests/runtime/test_presence.py
```diff
@@ -549,24 +701,24 @@ async def test_routes_room_added(self, mock_link, presences):
...
+        assert presence.roster.room_membership("room-123") is RoomMembership.Unadmitted
 
     async def test_routes_message_to_room_event(self, mock_link, presences):
         """Should route message events to on_room_event."""
         presence = presences(auto_subscribe_existing=False)
         await presence.start()
-        presence.rooms.add("room-123")
+        admit_room(presence, "room-123")
 
         received = []
```
AlexanderZ-Band, line 807: huge test file consider single conser class per file split

tests/runtime/test_presence_real_wire.py
```diff
@@ -0,0 +1,114 @@
...
+    )
+
+
+@asynccontextmanager
+async def running_presence(link: BandLink):
+    """Tear down every real resource (event-consumer task, room
+    subscriptions, WS connection) even if an assertion in between raises.
+    ``presence.stop()`` only reaches room-level state -- the connection is
+    ``link``'s own, so it needs its own explicit close."""
+    presence = RoomPresence(link)
```
AlexanderZ-Band, line 52: this should be in conftest and reused

src/band/runtime/presence.py
AlexanderZ-Band: Verified — but pushing back: converting `_notify` to a context manager would reintroduce the `if callback is None:` check at all 8 call sites, which is exactly the duplication `_notify` exists to remove. A CM here would only centralize the try/except-log half, not the null-check half, since every call site is one already-known coroutine call, not an arbitrary multi-statement block a CM is meant to wrap. Keeping the function-call form.

src/band/runtime/presence.py
AlexanderZ-Band: Confirmed and fixed in ec64066a. Verified `BandLink.unsubscribe_room` is keyed entirely by room_id (per-room ticket + per-room topic leaves, only shared-state touch is a set .add() which is race-free under asyncio's single-threaded scheduling) — safe to gather. Extracted the shared unsubscribe+notify body into `_leave_and_notify`/`_leave_one_room` and parallelized via `asyncio.gather`, matching the admit side's existing fan-out (this also fixed a real duplication between `stop()` and `_leave_removed_rooms` I'd left as copy-paste).

tests/runtime/test_presence.py
AlexanderZ-Band: Checked — the `presences` fixture is already the reuse point (its teardown calls `presence.stop()` for every built presence). The explicit `await presence.stop()` in `test_start_while_running_raises_without_orphaning_the_task` isn't redundant cleanup though — it's the test's actual assertion mechanism: the test needs to call stop() mid-test and inspect `first_task.done()` right after, which fixture teardown (running after the test body returns) can't give it. Keeping as-is.

tests/runtime/test_presence.py
AlexanderZ-Band: That line is just `presence.on_room_left = on_left` in a new test of mine — confirmed real but minor: the local `on_left` var was unneeded plumbing since nothing else references it. Fixed in ec64066a: assign `AsyncMock()` directly to `presence.on_room_left` and assert off the attribute.

tests/runtime/test_presence.py
AlexanderZ-Band: Fixed in ec64066a — trimmed the 6-line docstring on `test_a_stale_room_added_after_reconnect_admission_is_a_noop` down to the essential why (the TOCTOU/single-consumer-task reasoning is architectural, lives in `presence.py`'s docstrings and the plan's "Race conditions considered" section, not this test).

tests/runtime/test_presence.py
AlexanderZ-Band: That's on `test_a_stale_room_added_after_reconnect_admission_is_a_noop`, which proves the ticket-dedup logic via direct sequential method calls rather than a real race. Checked whether that's actually a gap: this specific interleaving isn't timing-dependent — `_consume_events` awaits each `_on_platform_event` to completion before dequeuing the next (single consumer task, no concurrency), so there's no real race window to reproduce over a live wire; the sequential-call unit test is a faithful simulation of the only reachable ordering. `test_presence_real_wire.py` already covers the one seam that IS genuinely timing-dependent (the actual reconnect-triggered wiring between `RoomRoster.reconcile()` and BandLink's real reconnect supervisor) — that's deliberately the one live test per the plan; adding more real-wire coverage for pure state-machine logic that's already structurally proven would be redundant, not more correct.

tests/runtime/test_presence.py
AlexanderZ-Band: Fair — the file's at ~750 lines now with 9 test classes. Checked scope though: this multi-class-per-file structure predates this diff (test_presence.py already had 8 classes/~600 lines before this PR, and it's this codebase's established convention — tests/runtime/test_resync.py is the same shape, 5 classes/~480 lines). I added new tests within the existing per-topic-class pattern rather than growing a monolith from scratch. Splitting it into one-class-per-file is a real, separate structural refactor (touches every import of these fixtures/helpers across the repo) that the implementation plan didn't scope — out of scope for this PR. Happy to open a followup if you want it done, just say the word rather than me unilaterally deciding to file one.

tests/runtime/test_presence_real_wire.py
AlexanderZ-Band: Checked both helpers separately. `chat_row` was a real, exact duplicate of test_presence.py's own helper (the plan's own Testing section literally says "reuse, don't reinvent" for this exact helper, and I'd violated that) — fixed in ec64066a, moved to tests/runtime/conftest.py, both files now import it. `running_presence` (the async context manager) I left local: it has exactly one caller today, so extracting it now would be reuse-before-there's-a-second-use — this codebase's own precedent (admit_room) only got promoted to conftest once it had 6 concrete call sites across 3 files. Will move it if/when a second real-wire test needs it.

src/band/platform/link.py
```diff
@@ -184,22 +184,27 @@ async def connect(self) -> None:
...
             self.api_key,
             self.agent_id,
             on_reconnect=self._on_reconnected,
             on_disconnect=self._on_disconnected,
         )
-        await self._ws.__aenter__()
+        self._ws = ws
         try:
-            await self._ws.join_agent_control_channel(
+            await ws.__aenter__()
```
AlexanderZ-Band: Smells of hacky plumbing let's reconsider this design

src/band/runtime/presence.py
```diff
@@ -111,19 +138,36 @@ async def start(self) -> None:
...
+                "call stop() before starting again"
+            )
 
-        # Subscribe to room added/removed events
-        await self.link.subscribe_agent_rooms(self.link.agent_id)
+        # Set synchronously, before the first await below, so a second
+        # concurrent start() call sees it even while this one is still
+        # mid-connect/mid-subscribe -- _event_task alone can't do that job,
+        # since it isn't assigned until the very end of this method.
+        self._starting = True
```
AlexanderZ-Band, line 153: Is this safe? Concurrency

src/band/platform/link.py
AlexanderZ-Band: Agreed, and you were right — reconsidered. Fixed in fb86a97e: separated the reentrancy guard from `_ws` itself, mirroring the `_starting`/`_event_task` split already used in `RoomPresence.start()`. New `_connecting` flag guards re-entrancy (set synchronously before the first await); `_ws` is now assigned exactly once, only after the client is fully connected — never a half-joined client another caller could read via the `if not self._ws: raise "Not connected"` checks used everywhere else in this file. A cancelled/failed connect() closes the half-opened client but never touches `_ws`, since it was never set.

src/band/runtime/presence.py
AlexanderZ-Band: Verified safe. The check (line ~141) and the set (line 153) run with no `await` between them, so under asyncio's single-threaded cooperative scheduling nothing can interleave there — it's the same atomic check-then-set pattern as the pre-existing `_ws`-based guard in `BandLink.connect()`. Concretely:
- `test_concurrent_start_calls_only_one_wins` (tests/runtime/test_presence.py) drives two genuinely concurrent `start()` calls via a gated mid-await connect() and asserts only one proceeds — it fails on the pre-`_starting` code (deadlocks) and passes now.
- `_starting` always resets via the enclosing `try/finally`, including if cancelled while waiting on `_lifecycle_lock`, so it can't get stuck `True`.
- This class isn't meant to be called from multiple OS threads against the same event loop (no `RoomPresence` code is thread-safe by design), so that's the only kind of concurrency out of scope here — consistent with the rest of the class.

No code change; this one was already correct.

src/band/runtime/presence.py
```diff
@@ -399,48 +464,85 @@ async def _join_room(
...
+        succeeded = False
         try:
-            await self.link.subscribe_room(room_id)
-        except Exception as e:
-            logger.warning(
-                "Failed to subscribe to room %s during %s: %s", room_id, context, e
-            )
-            self.rooms.discard(room_id)
-            return False
+            try:
```
amit-gazal-band, line 520: The docstring above claims the bare `finally` fully rolls back "a cancelled or failed subscribe," but a genuine `CancelledError` here skips that rollback. If `subscribe_room` has already sent the channel-join frame server-side when the cancellation lands, the `finally` at line 523 correctly resets the roster ticket to `Unadmitted`, but the `CancelledError` propagates straight past the `if not admitted: await self._unsubscribe_room(...)` block (lines 526-532) — that block never runs on this path. Result: a live transport subscription for `room_id` with no roster entry, so no later `stop()`/`reconcile()` ever targets it for cleanup, and `_handle_room_event`'s membership check silently drops all future events for that room.

The new `test_cancellation_mid_subscribe_rolls_back_to_unadmitted` regression test only asserts roster state, not that `unsubscribe_room` was actually called, so this gap isn't covered. Suggest wrapping the cancellation path so `_unsubscribe_room` still runs when `succeeded` is `False`, regardless of whether the exit is a normal return or a cancellation.

src/band/runtime/presence.py
```diff
@@ -451,21 +553,22 @@ async def _subscribe_rooms(
...
-            if room_id not in self.rooms
-        }
-        if not pending:
+        """Join every room in parallel."""
+        if not rooms_to_join:
             return
 
         results = await asyncio.gather(
             *[
                 self._join_room(room_id, payload, context=context)
```
amit-gazal-band, line 594: `_subscribe_rooms` dropped the old pre-filter for rooms already admitted (previously `if room_id not in self.rooms`). A room admitted concurrently via a live `room_added` event during startup — the exact race `_join_room`'s docstring describes — now gets miscounted as a failed subscription.

Concretely: the REST snapshot in `_subscribe_to_existing_rooms` includes room X, but a live `room_added` event for X already admitted it via `_handle_room_added` first. `_subscribe_rooms` still calls `_join_room("X", ...)`, which returns `False` because `begin_room_admission` returns `None` (already admitted). `_log_admission_results` folds that `False` into the failure count, so the log ends up saying "Subscribed to N rooms during startup (1 failed)" for a room that in fact joined successfully. Worth re-adding a pre-filter (or otherwise distinguishing "already admitted" from "actually failed" in the result) before this gather.

amit-gazal-band (commented):
A few findings from review, most severe first.

src/band/runtime/presence.py
```diff
@@ -143,25 +187,24 @@ async def stop(self) -> None:
...
-                    await self.on_room_left(room_id)
-                except Exception as e:
-                    logger.warning("on_room_left error for %s: %s", room_id, e)
+                    await self._event_task
+                except asyncio.CancelledError:
+                    pass
+                self._event_task = None
 
-        self.rooms.clear()
+            admitted_room_ids = self.roster.tracked_room_ids()
```
amit-gazal-band, line 210: `roster.clear()` already returns the list of removed room ids per `band_sdk_core`'s `def clear(self) -> list[str]` (confirmed against the installed stub, `band_sdk_core/__init__.pyi:305`). Calling `tracked_room_ids()` first and then `clear()` is a redundant extra roster call for a value `clear()` already hands back — `admitted_room_ids = self.roster.clear()` is equivalent, cheaper, and keeps a single source of truth for "what was cleared" instead of a separate snapshot call before the mutating one.

src/band/runtime/presence.py
AlexanderZ-Band: Fixed in 3a80759 -- `_complete_room_admission` now wraps the `await self.link.subscribe_room(room_id)` call itself in `except asyncio.CancelledError: await self._unsubscribe_room(room_id, context=context); raise`, so a cancellation that lands after the join reached the server still gets a best-effort leave before the CancelledError propagates. Strengthened `test_cancellation_mid_subscribe_rolls_back_to_unadmitted` to assert `mock_link.unsubscribe_room` was actually called, per your note that the existing test didn't cover it.

src/band/runtime/presence.py
AlexanderZ-Band: Fixed in 3a80759 -- `_subscribe_rooms` now filters out rooms already `RoomMembership.Admitted` before the gather, so a room a concurrent `room_added` admitted first is skipped instead of counted as a failure. Added `test_subscribe_rooms_skips_a_room_already_admitted` covering it.

src/band/runtime/presence.py
AlexanderZ-Band: Tried this, but it's not actually equivalent -- `clear()`'s return value includes every tracked room regardless of membership state, not just `Admitted` ones. Seeded a room via `begin_room_admission` only (left in `Admitting`, never resolved) and called `stop()`: with `admitted_room_ids = self.roster.clear()`, `on_room_left` fired for it, breaking `test_stop_does_not_notify_for_a_room_still_admitting` (a room never announced as joined must not be announced as left). `tracked_room_ids()` apparently filters to `Admitted` only; `clear()` doesn't. Left the two-call form as is and added a comment explaining why -- happy to revisit if band_sdk_core's actual contract here is different from what the test demonstrates, but wanted to flag it rather than take the redundant-call framing at face value.

amit-gazal-band (approved):
All three review findings addressed in 3a80759 — verified the cancellation-mid-subscribe fix, the already-admitted-room skip, and confirmed the roster.clear()/tracked_room_ids() pushback against the real band_sdk_core runtime. Full test_presence.py suite (42 tests) + ruff pass.

PR #591, opened 2026-08-28
fix: cache attachment lookups in AgentTools file reads

AlexanderZ-Band:
## Summary
- `_find_attachment` restarted pagination from page one on every call — reading an older file in a busy room re-walked the whole message history each time, and `AgentTools` is rebuilt fresh every turn, so a per-instance cache would never survive to help.
- The lookup logic (`AgentTools._fetch_attachment_uncached`, a `@staticmethod`) is now wrapped by `AgentTools._attachment_cache()` — a lazy singleton (`@staticmethod @functools.lru_cache(maxsize=None)`) that builds the `@alru_cache`-wrapped lookup on first real use, keyed on `(room_id, rest, file_id)`. `async-lru` has no way to exclude an argument (e.g. `self`) from its cache key, so keeping the lookup a `@staticmethod` is what makes it cache correctly instead of keying on the `AgentTools` instance, which is rebuilt every turn by design.
- Building the cache lazily (not decorating the function directly at class-body/import time) matters: `RuntimeSettings()` — and thus `BAND_ATTACHMENT_CACHE_MAXSIZE` — would otherwise get read before an app's `load_dotenv()` ever runs (every example does `from band import Agent` first), silently ignoring a `.env`-set override. Verified live before landing.
- Real access-order LRU eviction and free single-flight dedup on concurrent cache misses (verified from `async-lru`'s source, not just docs), bounded by `BAND_ATTACHMENT_CACHE_MAXSIZE` (new `RuntimeSettings`, default 1000) — a process-wide budget shared across every room/agent in the process, not per-room. The key includes the REST client instance too, so two agents/pools never share cached metadata.
- A cached attachment past its own `expires_at` is evicted and re-fetched once before being rejected — the cached copy may simply predate the platform extending that deadline, so a single stale-looking read shouldn't cost more than one extra lookup. If the refresh is still expired, that entry is evicted too, so the cache never keeps serving metadata it's already given up on. `expires_at` comparison treats a naive (offset-less) timestamp as UTC instead of raising `TypeError` on a naive/aware comparison (the Fern model doesn't enforce a timezone).
- The pagination walk (`_iter_message_pages`) has no depth cap — a room's message history has no realistic ceiling to bound against, unlike `iter_chat_pages`'s room-count cap. The only guard is a malformed response repeating a cursor it already returned, which would otherwise loop forever making no progress; safe because `get_agent_chat_context` documents `cursor` as keyset pagination (derived from the boundary row), so two distinct pages can't collide.
- `_attachment_cache()` returns a small structural `AttachmentCache` Protocol (the 4 methods actually used) instead of `Any`, without coupling to `async-lru`'s private, unexported wrapper class.
- Also folded in: `## External Research & Code Reuse` guidance added to `AGENTS.md`, prompted by this session's own detour — `aiocache`'s docs describe a `maxsize` option that only exists on its unreleased `master` branch; verified against the actually-installed `0.12.3` package before ruling it out.

Fixes INT-1306.

- Collapsed `_find_attachment`'s hand-duplicated check/invalidate/refetch/check/invalidate sequence into a two-attempt loop, per the reviewer's suggested shape — same behavior, one place instead of two near-identical inline blocks.

Review thread resolved.

## Test plan
- [x] `uv run pytest tests/runtime/test_tools.py -v --no-cov` — 155 passed: cache-hit skips pagination, isolation by room *and* by REST client (both verified to actually fail without that field in the key), a miss isn't negative-cached, an expired entry is refreshed once (served if the refresh shows a later deadline, evicted if still expired), a naive `expires_at` doesn't crash, a repeated cursor terminates the walk, a room with 60+ pages of history before the target still resolves it, settings maxsize is wired correctly and re-read lazily (not at import time) — real `ExecutionContext` + `from_context()` cross-turn test included.
- [x] `uv run pytest tests/ --ignore=tests/integration/ --ignore=tests/e2e/ -q --no-cov` — 5179 passed, 0 failed.
- [x] `uv run pyrefly check` — 0 errors.
- [x] Installed the real released `async-lru==2.3.0` and `aiocache==0.12.3` in this repo's venv and exercised their actual behavior (LRU eviction, exception-eviction, per-instance-key semantics, `@staticmethod` compatibility, lazy-singleton settings timing) directly before committing to the design — see commit history for what each check ruled in/out.

https://claude.ai/code/session_01UESgZ3R38fhf3SAi3vSUrd


src/band/runtime/tools.py
```diff
@@ -2689,39 +2787,44 @@ async def list_room_files(self, cursor: str | None = None) -> dict[str, Any]:
...
+        the Fern model doesn't enforce one) is treated as UTC, the platform's
+        only timezone, rather than raising on a naive/aware comparison."""
+        expires_at = attachment.expires_at
+        if expires_at is None:
+            return False
+        if expires_at.tzinfo is None:
+            expires_at = expires_at.replace(tzinfo=timezone.utc)
+        return expires_at <= datetime.now(timezone.utc)
 
     async def _find_attachment(self, file_id: str) -> "Attachment":
```
amit-gazal-band, line 2801: `_find_attachment`'s expired-cache handling hand-duplicates the invalidate+refetch pattern (check → invalidate → refetch → check → invalidate → raise) instead of a small bounded loop. Not a correctness issue, but a future change to the refresh policy would mean editing two near-identical inline blocks instead of one loop body. Could collapse to something like:

```python
cache = AgentTools._attachment_cache()
attachment = await cache(self.room_id, self.rest, file_id)
for attempt in range(2):
    if attempt:
        attachment = await cache(self.room_id, self.rest, file_id)
    if not self._attachment_expired(attachment):
        return attachment
    cache.cache_invalidate(self.room_id, self.rest, file_id)
raise BandToolError(FILE_UNAVAILABLE_MESSAGE)
```

amit-gazal-band (dismissed):
Solid, well-tested change. One minor cleanup suggestion inline, non-blocking.

src/band/runtime/tools.py
AlexanderZ-Band: Fixed in 32183c5 -- collapsed into the two-attempt loop you suggested. Full unit suite green (5179 passed).

amit-gazal-band (approved):
Fix in 32183c5 addresses the review comment. Approving.

PR #593, opened 2026-08-28
fix: BandLink.is_room_subscribed() can go stale after a Phoenix topic rejoin failure

AlexanderZ-Band:
## Summary
- `is_room_subscribed()`/agent-topic membership could go stale forever: a room or agent topic's Phoenix channel can fail to *rejoin* after a reconnect with no callback into anything BandLink-owned (PHXChannelsClient exposes only connection-level hooks, no topic-level one).
- Bumps `band-sdk-core` to `1.2.1` for `SubscriptionTracker`'s generation-ticketed rejoin API. The tracker now preserves the ticket that created each `Subscribed`/`Joined` entry; `room_rejoin_candidates()`/`agent_topic_rejoin_candidates()` return `(id, ticket)` pairs, and `mark_room_rejoin_failed`/`mark_agent_topic_rejoin_failed` take that ticket and no-op on a stale one -- closing a real gap where a delayed rejoin-failure report could otherwise clobber a room/topic that had since been unsubscribed and resubscribed. Also adds `AgentTopicKind`/`chat_room_topic`/`room_participants_topic` -- `client.py`/`link.py` build every topic string through these instead of local duplicates.
- Adds `WebSocketClient.joined_topics()` (a frozen live snapshot of the transport's subscription registry) and `_detect_room_rejoin_failures`/`_detect_agent_topic_rejoin_failures`, wired into `_on_reconnected` before `_drain_reconciliation` so a caught failure gets the same best-effort clean leave every other ambiguous-outcome case already gets in the same cycle. Both now report each candidate's own ticket rather than a bare id.
- Replaces the agent-topic cancellation workaround: on a cancelled/unresolved join, `_subscribe_agent_topic` now calls the new `record_agent_topic_join_ambiguous(topic, ticket)` instead of forcing `record_agent_topic_join(joined=False)` -- core itself resolves straight to `NeedsReconciliation` instead of `Absent`, since transport registration may have happened before the cancel. `subscribe_room`'s equivalent cancellation path now also gates its local reconciliation mark on the actual `RoomSubscribeResult` returned, matching the exception-path handling just above it instead of marking unconditionally.
- Fixes a latent gap in `FakePhoenixServer` found while writing the real-wire test for this: a rejected join only skipped *adding* to `joined_topics`, never discarded a stale entry from an earlier successful join on the same topic -- so a rejected *rejoin* (this exact scenario) left a topic the client had already dropped still showing as server-joined.
- Adds a drift-guard test (`test_wire_event_matches_band_sdk_core_event_type`) keeping `client.py`'s `WireEvent` literals in sync with `band_sdk_core.EventType`'s wire-name vocabulary, and notes the same parity for the local memory enums against `band_sdk_core`'s.
- New tests: room and agent-topic rejoin-failure detection through real `BandLink` behavior (`test_link.py`), and the agent-topic cancellation path now asserts core's own `agent_topic_status` reaches `NeedsReconciliation`, not just that BandLink's local block engages (`test_link_contacts.py`).
- Codifies a comment-style rule in `AGENTS.md`: comments state facts, never narration of how the code got there; a function needing a long comment to be understood should be split into named sub-functions instead.

Linear: INT-1326

- `_connected_ws()` raises `RuntimeError("Not connected")` instead of a bare `assert`.
- `_drain_agent_topic_reconciliation` raises explicitly when `from_wire_name` returns `None` for an unrecognized topic kind, instead of falling through the equality check.
- `_on_reconnected` takes one `ws.joined_topics()` snapshot per reconnect pass, threaded into `_recover_agent_control`/`_detect_room_rejoin_failures`/`_detect_agent_topic_rejoin_failures` instead of each calling it separately.
- Extracted `_join_agent_control_channel` as the one place `connect()` and `_recover_agent_control` wire up the control-channel join.
- Dropped leftover pre-refactor narration comments per this PR's own new `AGENTS.md` rule.
- `agent_control`'s lack of durable recovery after a rejected rejoin (one attempt, then only a log line) is scoped out to a followup: [INT-1339](https://linear.app/thenvoi/issue/INT-1339/agent-control-has-no-durable-recovery-after-a-rejected-rejoin).
- Merged `main` to resolve a `band-sdk-core` version-pin conflict from an unrelated PR (#592 landed `1.0.1`; this PR needs `1.2.1` for the generation-ticketed rejoin API above) — kept this PR's `1.2.1`.

All 6 review threads resolved. Full CI green (20/20 checks) as of the merge commit.

## Test plan
- [x] `uv run pytest tests/ --ignore=tests/integration/ --ignore=tests/e2e/` -- 5173 passed, 122 skipped (full suite, incl. new real-wire tests and BandLink-level tests for the room and agent-topic rejoin-failure/cancellation paths)
- [x] `uv run ruff check .` / `uv run ruff format --check .` -- clean
- [x] `uv run pyrefly check` -- 0 errors

https://claude.ai/code/session_01CaRd1WGx5GuUBBjC94cDR6


src/band/platform/link.py
```diff
@@ -560,51 +528,124 @@ def is_room_subscribed(self, room_id: str) -> bool:
         """Whether ``room_id`` is currently fully subscribed (both topics)."""
         return self._subscriptions.is_room_subscribed(room_id=room_id)
 
+    def _connected_ws(self) -> WebSocketClient:
+        """The active WebSocket client -- only ever called from the client's
+        own reconnect hook (``_on_reconnected``) or something it calls
+        synchronously from there, where a live connection is guaranteed."""
+        assert self._ws is not None
```
amit-gazal-band: `_connected_ws()` uses a bare `assert self._ws is not None`. With `_on_reconnected` now awaiting a real network round trip in `_recover_agent_control` before the next `_connected_ws()` call inside `_drain_reconciliation`, the window where a concurrent `disconnect()` (which sets `self._ws = None`) races this is wider than before. That race now surfaces as an unguarded `AssertionError` (and disappears entirely under `python -O`) instead of the `RuntimeError` this class uses everywhere else for a missing connection. Suggest raising a `RuntimeError` here instead of `assert`.

src/band/platform/link.py
```diff
@@ -560,51 +528,124 @@ def is_room_subscribed(self, room_id: str) -> bool:
...
+        ]
+        for topic, ticket in dropped:
+            if self._subscriptions.mark_agent_topic_rejoin_failed(
+                topic=topic, ticket=ticket
+            ):
+                self._mark_needing_reconciliation(
+                    topic, self._agent_topics_needing_reconciliation, ws
+                )
+
+    async def _recover_agent_control(self, ws: WebSocketClient) -> None:
```
amit-gazal-band: `_recover_agent_control`'s repair attempt lives entirely outside `SubscriptionTracker` and, on failure, only logs a warning — no pending set, retry ticket, or consumer-facing resubscribe call to unblock later, unlike the room/agent-topic reconciliation this same PR hardened. If the one recovery rejoin also fails, `agent_control` stays unsubscribed for the rest of the session: STOP/INTERRUPT/PLAY control pushes are silently dropped with no caller-visible signal beyond a log line. Worth giving this the same durable-recovery treatment as the other channels, or at least surfacing the failure so a caller can react.

src/band/platform/link.py
```diff
@@ -622,14 +663,14 @@ async def _drain_room_reconciliation(self, ws: WebSocketClient) -> None:
...
+        """Same stale-session handling as ``_drain_room_reconciliation``."""
         for topic in list(self._agent_topics_needing_reconciliation):
-            if self._ws is not ws:
+            if not self._is_current_session(ws):
                 return
             kind, _, agent_id = topic.partition(":")
             leave = (
                 ws.leave_agent_rooms_channel
-                if kind == _AGENT_ROOMS_KIND
+                if AgentTopicKind.from_wire_name(kind) == AgentTopicKind.Rooms
```
amit-gazal-band: `AgentTopicKind.from_wire_name(kind) == AgentTopicKind.Rooms` — `from_wire_name` returns `AgentTopicKind | None`. For any topic string it doesn't recognize, this returns `None`, and `None == AgentTopicKind.Rooms` is `False`, so the branch silently falls through to `leave_agent_contacts_channel` instead of raising. Currently dormant only because `_agent_topics_needing_reconciliation` happens to be populated exclusively from Rooms/Contacts topics by convention — nothing at the type level guarantees that. Consider handling the `None` case explicitly (e.g. raise or assert) rather than relying on the equality falling through safely.

src/band/platform/link.py
```diff
@@ -560,51 +528,124 @@ def is_room_subscribed(self, room_id: str) -> bool:
...
+        Runs rejoin-failure detection before draining so a rejoin failure
+        gets the same best-effort clean leave as every other ambiguous
+        outcome (a cancelled join, a failed rollback) already queued for
+        ``_drain_reconciliation`` -- this reconnect is the only point where
+        that ambiguity is safely resolvable (see design doc).
         """
         logger.info("WebSocket reconnected — reconciling room state")
         self._subscriptions.on_reconnected()
+        ws = self._connected_ws()
+        await self._recover_agent_control(ws)
```
amit-gazal-band: `_on_reconnected` calls `ws.joined_topics()` three separate times across `_recover_agent_control`, `_detect_room_rejoin_failures`, and `_detect_agent_topic_rejoin_failures`, each copying the full subscription registry (`get_current_subscriptions()` copies its dict on every call). This contradicts `joined_topics()`'s own docstring: "call once per detection pass." Since nothing mutates the registry between these three calls in this pass, suggest computing `joined = ws.joined_topics()` once here and threading it into all three helpers as a parameter.

src/band/platform/link.py
```diff
@@ -560,51 +528,124 @@ def is_room_subscribed(self, room_id: str) -> bool:
...
+        session, outside SubscriptionTracker entirely -- unlike the
+        tracker-owned channels, there is no consumer-facing resubscribe call
+        waiting to be unblocked, so a rejoin PHX itself rejected is repaired
+        here directly instead of only flagged for later."""
+        topic = AgentTopicKind.Control.topic(self.agent_id)
+        if topic in ws.joined_topics():
+            return
+        logger.warning("agent_control rejoin failed, attempting recovery")
+        try:
+            await ws.join_agent_control_channel(
```
amit-gazal-band: This `join_agent_control_channel(agent_id, on_supersede=self._on_supersede, on_control=self._on_control)` call duplicates the identical call already made in `connect()` verbatim, with no shared helper. A future change to the control-channel join signature or callback wiring only needs to be applied at one of the two call sites to pass review, silently diverging connect-time and reconnect-recovery behavior. Consider extracting a small shared helper both call sites use.

src/band/platform/link.py
```diff
@@ -560,51 +528,124 @@ def is_room_subscribed(self, room_id: str) -> bool:
...
+                self.agent_id,
+                on_supersede=self._on_supersede,
+                on_control=self._on_control,
+            )
+            logger.info("Recovered agent_control after a rejected rejoin")
+        except Exception as e:
+            logger.warning("Failed to recover agent_control: %s", e)
+
     # --- Event handlers (from BandAgent, unified into PlatformEvent) ---
```
amit-gazal-band: This PR adds an AGENTS.md rule banning comments that narrate extraction/session/line-number history ("no session/PR/ticket/line-number history — git already owns that history"), and cleans up several nearby instances of exactly this pattern — but this comment (`(from BandAgent, unified into PlatformEvent)`), and its siblings at lines 100, 108, 308, and 324 (`(from BandAgent._api_client)`, `(from BandAgent._ws_client)`, `(from lines 733-736)`, `(from lines 739-743)`), survive untouched. Worth cleaning these up too for consistency with the new rule this same PR introduces.

amit-gazal-band (changes requested):
Six issues found in review, all introduced by this PR (except the last, which is pre-existing content the PR's own new AGENTS.md rule now makes non-compliant).

src/band/platform/link.py
AlexanderZ-Band: Fixed in 60c751a — `_connected_ws()` now raises `RuntimeError("Not connected")` instead of asserting, matching the rest of the class.

src/band/platform/link.py
AlexanderZ-Band: Fixed in 60c751a — `_drain_agent_topic_reconciliation` now raises `RuntimeError` explicitly when `from_wire_name` returns `None` for an unrecognized topic kind, instead of falling through the equality check.

src/band/platform/link.py
AlexanderZ-Band: Fixed in 60c751a — `_on_reconnected` now takes one `ws.joined_topics()` snapshot per reconnect pass and threads it into `_recover_agent_control`, `_detect_room_rejoin_failures`, and `_detect_agent_topic_rejoin_failures`.

src/band/platform/link.py
AlexanderZ-Band: Fixed in 60c751a — extracted `_join_agent_control_channel(ws)` as the one place that wires up the control-channel join; both `connect()` and `_recover_agent_control` call it now.

src/band/platform/link.py
AlexanderZ-Band: Fixed in 60c751a — dropped the leftover "(from ...)" narration comments at this line and its siblings (lines 100, 108, 308, 324 in the pre-fix file).

src/band/platform/link.py
AlexanderZ-Band: Agreed this needs the same durable-recovery treatment, but that's a real design change (pending-set tracking, retry wiring through the reconciliation drain, and/or a consumer-facing failure signal) rather than a mechanical fix — scoping it out of this PR rather than bolting it on. Filed as a followup, assigned to me: https://linear.app/thenvoi/issue/INT-1339/agent-control-has-no-durable-recovery-after-a-rejected-rejoin

amit-gazal-band (dismissed):
5 of 6 review comments are addressed (bare assert, from_wire_name None handling, redundant joined_topics() calls, duplicated join call, stale narration comments). The one outstanding item -- agent_control recovery still only logs a warning on failure with no retry/pending mechanism -- is left as-is; not blocking, but noting it stays open for a future pass.

amit-gazal-band (approved):

Issue #53, opened 2026-01-05
Bug: Anthropic and Pydantic AI converters lose tool history on session restore


vlad-thenvoi:
## Summary

The `AnthropicHistoryConverter` and `PydanticAIHistoryConverter` filter out `tool_call` and `tool_result` events during session restore, causing tool execution history to be lost when `enable_execution_reporting=True`.

## Affected Components

| Component | File |
|-----------|------|
| Anthropic Converter | `src/thenvoi/converters/anthropic.py` |
| Pydantic AI Converter | `src/thenvoi/converters/pydantic_ai.py` |

### Emission vs Restoration Mismatch

| Adapter | Emits tool_call | Emits tool_result | Restores tool_call | Restores tool_result |
|---------|-----------------|-------------------|--------------------|--------------------|
| **Anthropic** | ✅ Yes | ✅ Yes | ❌ No | ❌ No |
| **Pydantic AI** | ✅ Yes | ✅ Yes | ❌ No | ❌ No |
| **Claude SDK** | ✅ Yes | ✅ Yes | ✅ Yes | ✅ Yes |
| **LangChain** | ✅ Yes | ✅ Yes | ✅ Yes | ✅ Yes |

### Root Cause

Both converters have a hard filter that skips all non-text messages:

```python
# Only convert text messages
if message_type != "text":
    continue  # ← Filters out tool_call, tool_result
```

### Impact

When `enable_execution_reporting=True`:

1. **Execution**: Tool events are emitted and stored on platform ✅
2. **Session Restore**: Converter filters out tool events ❌
3. **Result**: Agent loses tool execution context, may re-execute tools

## Expected Behavior

Converters should restore tool history like Claude SDK and LangChain do:

| Message Type | Should Include |
|-------------|----------------|
| `text` (own) | ❌ Skip (redundant) |
| `text` (others) | ✅ Include |
| `tool_call` | ✅ Include |
| `tool_result` | ✅ Include |
| `thought`, `error`, `task` | ❌ Skip (observability events) |

## Mitigation

Currently mitigated because `enable_execution_reporting=False` by default. Users who enable it will experience the bug.

## Proposed Fix

Update both converters to match the Claude SDK / LangChain pattern:

```python
if message_type == "text":
    # Handle text messages (skip own, include others)
    ...
elif message_type == "tool_call":
    # Include tool call for session restore
    ...
elif message_type == "tool_result":
    # Include tool result for session restore
    ...
[...]

PR #595, opened 2026-08-29
feat(websocket): add heartbeat dead-threshold watchdog

AlexanderZ-Band:
## Summary

Part of [INT-1323](https://linear.app/thenvoi/issue/INT-1323): wires `WebSocketClient` up to `band-sdk-core`'s shared `SessionPolicy` heartbeat/dead-threshold policy (band-ai/band-sdk-core#60, released as v2.0.0) and `phoenix-channels-python-client`'s new `on_heartbeat_ack`/`close_connection` surface (band-ai/phoenix-channels-python-client#53).

Dead-connection detection lives in a new `band.client.streaming.watchdog.HeartbeatWatchdog` class, not inline in `WebSocketClient`:

- `HeartbeatWatchdog` owns the deadline, the watchdog task, and the force-close decision. `WebSocketClient` holds one `self._watchdog` and delegates `start`/`stop`/`reset_deadline` to it.
- `heartbeat_interval_s` (from `SessionPolicy`, default or an injected override) is passed to `PHXChannelsClient`; `on_heartbeat_ack` and `on_reconnect` are wired to reset the watchdog's deadline.
- `HeartbeatWatchdog.start(client)` takes the specific `PHXChannelsClient` instance as an argument (not read dynamically later), so a stale watchdog from a superseded initial-connect attempt can never act on a later instance.
- A force-close, once `dead_threshold_s` passes with no ack, flows into the client's own existing reconnect path — no new disconnect-handling branch.
- Reconnects reset the deadline immediately (not just acks and the initial connect) — otherwise a reconnect can inherit a stale deadline from disconnected-state polling and get force-closed before its first heartbeat cycle completes.
- The watchdog task is started only once, on a successful connect, and stopped cleanly in `__aexit__`.
- A `close_connection` failure is caught and logged rather than killing the watchdog task, and the watchdog skips warning/closing when there's no live connection to avoid misleading log spam during backoff.

## Dependency floor: resolved

`phoenix-channels-python-client` PR #53 has since released as a real PyPI version (`0.2.4`) containing `on_heartbeat_ack`/`close_connection`. `dependencies` pins `phoenix-channels-python-client>=0.2.4` — a real, verified floor — and there is no `[tool.uv.sources]` git override. This PR is mergeable/releasable; the `packaging` job's "Install from wheel" step passes.

- `test_aenter_restores_reconnect_after_successful_initial_connect` called `__aenter__` without a matching `__aexit__`, leaking a watchdog task against a test double with no real `.connection` — once `dead_threshold_s` elapsed, `_force_close_if_stale` would raise `AttributeError` inside that fire-and-forget task on the session-scoped event loop, polluting whichever test ran when it fired. Gave the double a no-op `__aexit__` and added the matching cleanup call, same shape as the sibling retry-backoff test's own earlier fix. Checked every other bare `__aenter__()` call in the file: the rest all raise before reaching the watchdog-start line, so this was the only actual leak.

Review thread resolved.

## Test plan

- [x] `tests/websocket/test_watchdog.py` — 5 tests against `HeartbeatWatchdog` directly, no real socket needed: task cancels cleanly on `stop()`; a stale watchdog from a superseded attempt cannot close a newer connection; survives a `close_connection` failure and keeps enforcing the threshold; neither warns nor closes while already disconnected; a reset deadline lets a fresh socket survive a reconnect that lands late in a disconnected-state polling window.
- [x] `tests/websocket/test_client.py` — real-wire tests against an in-process Phoenix-protocol peer, covering the actual wiring: ack keeps the connection alive across several heartbeat cycles; a withheld ack forces close + reconnect at ~`dead_threshold_s`; `__aexit__` stops the watchdog; the successful-initial-connect test no longer leaks a watchdog task (`-W error` clean).
- [x] `uv run pytest tests/ --ignore=tests/integration/ --ignore=tests/e2e/ --all-packages` — 5175 passed, 122 skipped.
- [x] `uv run pyrefly check` (project-wide, matching CI) — 0 errors.
- [x] `uv run ruff check` / `ruff format --check` — clean.
- [x] Merged `main` to pick up an unrelated `band-sdk-core`/lockfile drift; resolved keeping this PR's `band-sdk-core==2.0.0` and `phoenix-channels-python-client>=0.2.4`. Full CI green (20/20 checks) as of the merge commit, including `packaging`.

https://claude.ai/code/session_011EuhCwwdkV3gYphyNFkw9Y

### DIFF
--- a/src/band/client/streaming/client.py
+++ b/src/band/client/streaming/client.py
@@ -1,7 +1,6 @@
 from __future__ import annotations
 
 import asyncio
-import contextlib
 from collections.abc import Awaitable, Callable
 from dataclasses import dataclass
 from enum import StrEnum
@@ -18,6 +17,7 @@ from phoenix_channels_python_client.client_types import ReconnectPolicy
 from phoenix_channels_python_client.exceptions import PHXConnectionError
 from phoenix_channels_python_client.phx_messages import PHXMessage
 from band.client.streaming.errors import classify_initial_upgrade_error
+from band.client.streaming.watchdog import HeartbeatWatchdog
 from band.client.streaming.wire import WirePayload
 from band.logging_config import core_issues, trace_context_extra
 
@@ -346,9 +346,7 @@ class WebSocketClient:
         self._on_disconnect = on_disconnect
         self._validation_error_count: int = 0
         self._last_disconnect_reason: WebSocketDisconnectReason | None = None
-        self._session_policy = session_policy or SessionPolicy.default()
-        self._watchdog_task: asyncio.Task[None] | None = None
-        self._watchdog_deadline: float = 0.0
+        self._watchdog = HeartbeatWatchdog(session_policy or SessionPolicy.default())
 
     @property
     def validation_error_count(self) -> int:
@@ -374,11 +372,6 @@ class WebSocketClient:
             raise RuntimeError("WebSocket client is not connected")
         return self.client
 
-    def _reset_watchdog_deadline(self) -> None:
-        self._watchdog_deadline = (
-            asyncio.get_running_loop().time() + self._session_policy.dead_threshold_s
-        )
-
     async def _handle_reconnect(self) -> None:
         """Reset the watchdog deadline on reconnect, then forward to the
         caller's own on_reconnect callback (if any).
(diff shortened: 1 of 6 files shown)


src/band/client/streaming/client.py
```diff
@@ -410,12 +428,14 @@ async def __aenter__(self):
                 await asyncio.sleep(delay)
             else:
                 self.client.auto_reconnect = True
+                self._watchdog.start(self.client)
```
amit-gazal-band, line 431: `__aenter__` unconditionally starts the watchdog task, but nothing guarantees cleanup for a caller that only invokes `__aenter__()` without a matching `__aexit__` — several existing tests do exactly this. Since the test event loop is session-scoped, a leaked task survives past its originating test.

Concretely, `test_aenter_restores_reconnect_after_successful_initial_connect` starts a watchdog against a test double (`SuccessfulPHXClient`) that has no `.connection` attribute. Once `SessionPolicy.default()`'s 60s `dead_threshold_s` elapses, `_force_close_if_stale` raises `AttributeError` inside that fire-and-forget task, polluting whatever test happens to be running when it fires.

amit-gazal-band (dismissed):
Two test-isolation issues from the new HeartbeatWatchdog, neither blocking merge but worth a fast follow-up.

**tests/websocket/test_client.py:417** (pre-existing line, not touched by this diff, so GitHub won't let me anchor an inline comment there): `test_aenter_retries_unclassified_initial_connection_errors` does `monkeypatch.setattr("band.client.streaming.client.asyncio.sleep", fake_sleep)`, which patches the real `asyncio` module globally, not just this module's usage. The new `HeartbeatWatchdog` also calls `asyncio.sleep` internally (in `_sleep_until_deadline`), so it gets patched too — and since `fake_sleep` returns instantly, the watchdog's sleep loop busy-spins forever once it starts. Confirmed by running the test directly: it now takes ~32s instead of near-instant, and logs an unhandled `Task exception was never retrieved` / pytest-timeout error rooted in `HeartbeatWatchdog._loop`.

src/band/client/streaming/client.py
AlexanderZ-Band: Fixed in de8673a -- gave SuccessfulPHXClient a no-op __aexit__ and added the matching client.__aexit__(None, None, None) call at the end of the test, same shape as the retry-backoff test's fix. Confirmed clean with -W error (no leaked-task warning). Checked every other bare __aenter__() call in this file: the rest all raise before reaching the watchdog-start line, so this was the only actual leak.

amit-gazal-band (approved):
Verified Alex's fixes cover both review comments — the watchdog leak in `client.py:431` and the monkeypatched `asyncio.sleep` spin in the retry-backoff test are both resolved (each affected test now stops the watchdog via `__aexit__`). Re-ran `tests/websocket/test_client.py` + `tests/websocket/test_watchdog.py`: 54 passed in 3.8s, no warnings or leaked tasks.

amit-gazal-band (approved):
Reviewed and confirmed: the leaked-watchdog-task fix in de8673a resolves the review comment — verified locally (tests/websocket/test_client.py, 49/49 passed, no leaked-task warnings). Note this PR is still marked not-mergeable pending the phoenix-channels-python-client dependency release per the PR description.

PR #600, opened 2026-08-31
feat: Capability.FILES support across every adapter

AlexanderZ-Band:
## Summary

Every registered framework adapter now supports `Capability.FILES` (`band_list_room_files`/`band_read_room_file`/`band_send_room_file`). Of the 16 registered adapters, 14 pass a small previewable image through to the model as real vision content instead of degrading it to a JSON-stringified blob (12 with a direct per-adapter probe, plus `letta` and `copilot_acp`, which inherit the fix transitively through the shared MCP engine and have no adapter-local code path to probe separately). The remaining two, `google_adk` and `parlant`, are confirmed unsupportable by their own frameworks, not unresearched.

Linear: INT-1350

## File support (`Capability.FILES`)

Every registered adapter: `claude_sdk`, `anthropic`, `opencode`, `gemini`, `langgraph`, `google_adk`, `agno`, `strands`, `codex`, `copilot_sdk`, `copilot_acp`, `letta`, `parlant`, `pydantic_ai`, `crewai`, `crewai_flow`.

Two wiring mechanisms:
- **Registry-driven** (`claude_sdk`, `anthropic`, `langgraph`, `gemini`, `google_adk`, `agno`, `strands`, `codex`, `copilot_sdk`, `opencode`, the ACP client adapter): tool schemas already derive generically from `iter_tool_definitions`/`get_openai_tool_schemas`/`get_anthropic_tool_schemas`, so declaring the capability was enough — no new plumbing.
- **Hand-rolled** (`parlant`, `pydantic_ai`, `crewai`/`crewai_flow`, which share `integrations/crewai/`): each mints one wrapper per platform tool instead of deriving from the registry, so each grew hand-written wrappers.
- `letta` routes tool execution entirely through the shared MCP engine (`integrations/mcp/engine.py`) rather than calling `execute_tool_call` directly, so it needed no adapter-side wiring at all.

`render_system_prompt` gained a `FILES_SECTION` so agents are told about the file tools when the capability is enabled.

## Image passthrough (`band_read_room_file` real vision content, not text)

| Adapter | Status |
|---|---|
| `claude_sdk`, `anthropic`, `opencode`, `gemini`, `langgraph`, `agno`, `strands`, `copilot_sdk` | Supported, directly probed. |
| `pydantic_ai` | Supported — returns `pydantic_ai.messages.BinaryContent` directly from the tool, pydantic-ai's own documented multimodal tool-result channel (`ToolReturnPart.content`). |
| `crewai`, `crewai_flow` | Supported — emits CrewAI's own `VISION_IMAGE:<media_type>:<base64_data>` sentinel string, parsed by `StepExecutor` (the default native-tool-calling execution path) into a real `image_url` content block before the LLM sees it. Both adapters share `integrations/crewai/`, so one fix covers both. |
| `codex` | Supported at the wire-shape level (`inputImage` content item with a `data:` URI, verified against the installed CLI's own generated JSON schema). Whether the Rust-side `codex-rs` binary actually accepts a `data:` URI, rather than only fetching `http(s)`, isn't verifiable from this repo's installed Python packages. |
| `letta`, `copilot_acp` | Supported transitively — both route through the same fixed MCP engine as `opencode`, so they inherit its fix and have no adapter-local path to probe on their own. |
| `google_adk` | **Not supportable today.** Installed `google-adk` 1.10.0's own `__build_response_event()` calls `Part.from_function_response()` without the `parts=` kwarg the underlying SDK supports, so no tool return value can reach a multimodal `FunctionResponse` through ADK's own flow. A framework limitation, not something fixable in this SDK. |
| `parlant` | **Not supportable today.** `parlant.core.tools.ToolResult` has no multimodal field, and Parlant's own MCP integration (`mcp_result_to_tool_result_data()`) discards image content blocks before they ever reach `ToolResult.data` — the same conclusion Parlant reaches on its own canonical external-tool-result path. |

Unit-level matrix proof for every supported row: `tests/framework_conformance/test_files_image_passthrough_matrix.py`. `crewai`/`crewai_flow` cases there need the isolated `dev-crewai` venv (listed explicitly in `ci.yml`'s crewai job — a bare `-k crewai` filter isn't statically verifiable by the job-coverage guard) and skip cleanly everywhere else.

## Live E2E coverage

Two live tests in `tests/e2e/baseline/smoke/matrix/test_capability_matrix.py`:
- `test_file_round_trip_across_files_adapters` — send → list → read round-trip with a text file, across every FILES-capable adapter in the baseline toolkit's matrix.
- `test_image_vision_passthrough_across_adapters` — a bystander peer uploads a randomly-colored generated PNG, the agent is asked to read and describe it, and an LLM judge checks the reply against the real uploaded color.

Both are gated behind `requires_file_transfer` (`pytest.mark.skipif` on `BaselineSettings().deployment.file_transfer`, i.e. the platform's `ff_file_transfer` flag), so they skip cleanly rather than fail on every deployment where that flag is off — which today is every SaaS and CI environment this repo runs against. Both were validated against a live deployment with the flag enabled during development (full-matrix dispatch run, 267 passed after fixing two real bugs this work surfaced: `letta`/`opencode`/the ACP client adapter caching their MCP tool registration from pre-negotiation features — see below — and a translation gap in `PeerActor.send_file`'s error path). Neither test currently executes in this repo's CI; they will the moment a lane points at a deployment with the flag on.

Two adapters are outside the baseline toolkit's matrix by design, not oversight:
- `parlant` registers no matrix adapter at all (self-hosted server per room, a fundamentally different lifecycle) — it gets a bespoke `@lane`-pinned smoke instead, per the baseline README.
- `crewai_flow`'s E2E builder uses a hardcoded echo flow with no Band tool loop, so there's no tool call for a file round-trip to drive.

Also fixed a capability-negotiation timing bug surfaced while wiring this: `letta`, `opencode`, and the ACP client adapter each cached their MCP tool registration once at construction time from pre-negotiation `self.features.capabilities`, so a deployment where negotiation *disabled* FILES after construction kept serving the file tools anyway. Each now overrides `apply_effective_features` to rebuild its cached registration from the negotiated features (`tests/adapters/test_negotiated_file_tools.py`).

## Single source of truth

- `READ_ROOM_FILE_TOOL_NAME`/`SEND_MESSAGE_TOOL_NAME` and the rest of the agent-surface tool vocabulary in `runtime/tools.py` are now `BandTool(StrEnum)` members, not loose string constants — every adapter's own `"band_read_room_file"` literal comparison was replaced with the enum member.
- `image_block_placeholder()`/`decode_image_block()` (`runtime/tools.py`) dedupe the placeholder string and base64-decode pattern repeated across `gemini`/`agno`/`strands`/`pydantic_ai`.
- `src/band/integrations/crewai/` was split into `reporting.py` (room context + tool-event reporters), `catalog.py` (the `@band_tool` bodies), and `tools.py` (build + run) — each of the 20 platform tools is now a single `@band_tool` async function instead of a hand-rolled `BaseTool` subclass, with the decorator owning the report_call/report_result/serialize lifecycle.
- `src/band/integrations/parlant/tools.py` was similarly rewritten to compose `require_session_tools`/`guard_failures`/`describe_from_master` instead of repeating the same guard-log-translate boilerplate in all 15 tools.
- `IMAGE_PASSTHROUGH_SUPPORTED_FRAMEWORK_IDS` derives from the lane-stable `Adapter` enum minus three `ExcludedAdapter(adapter, reason)` entries, rather than a hand-maintained allowlist.

## Test plan
- [x] `uv run pytest tests/ --ignore=tests/integration/ --ignore=tests/e2e/ -q --no-cov` — 5357 passed, 0 failed (`dev` extra); `parlant` suite separately green under `dev-parlant`; `crewai`/`crewai_flow` suites separately green under `dev-crewai`
- [x] `uv run ruff check .` / `uv run ruff format .`
- [x] CI green on all 20 checks (lint, test × 2 OS × 2 Python, test-crewai × 4, test-parlant × 4, packaging, CodeQL, release-baseline-gate)
- [x] Live E2E validated manually against a deployment with `ff_file_transfer` on (full-matrix `workflow_dispatch`) before landing the skip-gate for CI environments where it's off

https://claude.ai/code/session_0142oBtXk6Q7RJYK5f79UurZ

### DIFF
--- a/src/band/adapters/agno.py
+++ b/src/band/adapters/agno.py
@@ -33,6 +33,7 @@ from band.runtime.tools import (
     READ_ROOM_FILE_TOOL_NAME,
     SEND_MESSAGE_TOOL_NAME,
     get_band_tool_category,
+    image_block_placeholder,
     is_mcp_content_result,
 )
 
@@ -111,7 +112,7 @@ def _make_band_entrypoint(tool_name: str) -> Callable[..., Awaitable[Any]]:
                 for block in result["content"]
             ]
             return ToolResult(
-                content=f"<{len(images)} image content block(s)>", images=images
+                content=image_block_placeholder(len(images)), images=images
             )
         return result if isinstance(result, str) else json.dumps(result, default=str)
 
--- a/src/band/adapters/agno.py
+++ b/src/band/adapters/agno.py
@@ -30,6 +30,7 @@ from band.converters.agno import AgnoHistoryConverter, AgnoMessages
 from band.runtime.capabilities import with_hub_room_contacts
 from band.runtime.prompts import render_system_prompt
 from band.runtime.tools import (
+    READ_ROOM_FILE_TOOL_NAME,
     SEND_MESSAGE_TOOL_NAME,
     get_band_tool_category,
     is_mcp_content_result,
@@ -101,7 +102,7 @@ def _make_band_entrypoint(tool_name: str) -> Callable[..., Awaitable[Any]]:
         if active is None:
             return f"Error: no active Band context for tool {tool_name}"
         result = await active.execute_tool_call(tool_name, kwargs)
-        if tool_name == "band_read_room_file" and is_mcp_content_result(result):
+        if tool_name == READ_ROOM_FILE_TOOL_NAME and is_mcp_content_result(result):
             images = [
                 Image(
                     content=base64.b64decode(block["data"]),
--- a/src/band/adapters/agno.py
+++ b/src/band/adapters/agno.py
@@ -2,6 +2,7 @@
 
 from __future__ import annotations
 
+import base64
 import json
 import logging
 import warnings
@@ -10,6 +11,8 @@ from contextlib import contextmanager
 from contextvars import ContextVar
 from typing import TYPE_CHECKING, Any, ClassVar
 
+from agno.media import Image
+from agno.tools.function import ToolResult
 from typing_extensions import Unpack
 
 from band.core.protocols import AgentToolsProtocol
@@ -26,7 +29,11 @@ from band.core.types import (
 from band.converters.agno import AgnoHistoryConverter, AgnoMessages
 from band.runtime.capabilities import with_hub_room_contacts
 from band.runtime.prompts import render_system_prompt
-from band.runtime.tools import SEND_MESSAGE_TOOL_NAME, get_band_tool_category
+from band.runtime.tools import (
+    SEND_MESSAGE_TOOL_NAME,
+    get_band_tool_category,
+    is_mcp_content_result,
+)
 
 try:
     from agno.models.message import Message
@@ -88,12 +95,23 @@ def _tool_name(execution: Any) -> str:
     return getattr(execution, "tool_name", None) or ""
 
 
-def _make_band_entrypoint(tool_name: str) -> Callable[..., Awaitable[str]]:
-    async def _entrypoint(**kwargs: Any) -> str:
+def _make_band_entrypoint(tool_name: str) -> Callable[..., Awaitable[Any]]:
+    async def _entrypoint(**kwargs: Any) -> Any:
         active = _current_tools.get()
         if active is None:
             return f"Error: no active Band context for tool {tool_name}"
         result = await active.execute_tool_call(tool_name, kwargs)
+        if tool_name == "band_read_room_file" and is_mcp_content_result(result):
+            images = [
+                Image(
+                    content=base64.b64decode(block["data"]),
+                    mime_type=block["mimeType"],
+                )
+                for block in result["content"]
+            ]
+            return ToolResult(
+                content=f"<{len(images)} image content block(s)>", images=images
+            )
         return result if isinstance(result, str) else json.dumps(result, default=str)
 
     _entrypoint.__name__ = tool_name
--- a/src/band/adapters/agno.py
+++ b/src/band/adapters/agno.py
@@ -29,8 +29,7 @@ from band.converters.agno import AgnoHistoryConverter, AgnoMessages
 from band.runtime.capabilities import with_hub_room_contacts
 from band.runtime.prompts import render_system_prompt
 from band.runtime.tools import (
-    READ_ROOM_FILE_TOOL_NAME,
-    SEND_MESSAGE_TOOL_NAME,
+    BandTool,
     decode_image_block,
     get_band_tool_category,
     image_block_placeholder,
@@ -103,7 +102,7 @@ def _make_band_entrypoint(tool_name: str) -> Callable[..., Awaitable[str | ToolR
         if active is None:
             return f"Error: no active Band context for tool {tool_name}"
         result = await active.execute_tool_call(tool_name, kwargs)
-        if tool_name == READ_ROOM_FILE_TOOL_NAME and is_mcp_content_result(result):
+        if tool_name == BandTool.READ_ROOM_FILE and is_mcp_content_result(result):
             images = [
                 Image(content=data, mime_type=mime_type)
                 for data, mime_type in (
@@ -349,7 +348,7 @@ class AgnoAdapter(SimpleAdapter[AgnoMessages]):
         self._persist_turn(room_id, response)
 
         if not any(
-            _tool_name(execution) == SEND_MESSAGE_TOOL_NAME
+            _tool_name(execution) == BandTool.SEND_MESSAGE
             for execution in _tool_executions(response)
         ):
             logger.debug(
--- a/src/band/adapters/agno.py
+++ b/src/band/adapters/agno.py
@@ -97,8 +97,8 @@ def _tool_name(execution: Any) -> str:
     return getattr(execution, "tool_name", None) or ""
 
 
-def _make_band_entrypoint(tool_name: str) -> Callable[..., Awaitable[Any]]:
-    async def _entrypoint(**kwargs: Any) -> Any:
+def _make_band_entrypoint(tool_name: str) -> Callable[..., Awaitable[str | ToolResult]]:
+    async def _entrypoint(**kwargs: Any) -> str | ToolResult:
         active = _current_tools.get()
         if active is None:
             return f"Error: no active Band context for tool {tool_name}"
--- a/src/band/adapters/agno.py
+++ b/src/band/adapters/agno.py
@@ -33,7 +33,8 @@ from band.runtime.tools import (
     decode_image_block,
     get_band_tool_category,
     image_block_placeholder,
-    is_mcp_content_result,
+    is_image_passthrough_result,
+    redact_tool_call_args,
 )
 
 try:
@@ -102,13 +103,20 @@ def _make_band_entrypoint(tool_name: str) -> Callable[..., Awaitable[str | ToolR
         if active is None:
             return f"Error: no active Band context for tool {tool_name}"
         result = await active.execute_tool_call(tool_name, kwargs)
-        if tool_name == BandTool.READ_ROOM_FILE and is_mcp_content_result(result):
-            images = [
-                Image(content=data, mime_type=mime_type)
-                for data, mime_type in (
-                    decode_image_block(block) for block in result["content"]
-                )
-            ]
+        if is_image_passthrough_result(tool_name, result):
+            try:
+                images = [
+                    Image(content=data, mime_type=mime_type)
+                    for data, mime_type in (
+                        decode_image_block(block) for block in result["content"]
+                    )
+                ]
+            except Exception as error:
+                # A malformed or future-extended image block (see
+                # is_mcp_content_result's docstring) must degrade to the
+                # adapter's usual error string, not raise uncaught out of
+                # the tool entrypoint Agno invokes directly.
+                return f"Error reading room file: {error}"
             return ToolResult(
                 content=image_block_placeholder(len(images)), images=images
             )
@@ -753,7 +761,9 @@ class AgnoAdapter(SimpleAdapter[AgnoMessages]):
                 "tool_call",
                 {
                     ToolEventKey.NAME: ex.tool_name or "",
-                    ToolEventKey.ARGS: ex.tool_args or {},
+                    ToolEventKey.ARGS: redact_tool_call_args(
+                        ex.tool_name or "", ex.tool_args or {}
+                    ),
                     ToolEventKey.TOOL_CALL_ID: ex.tool_call_id or "",
                 },
                 room_id=room_id,
--- a/src/band/adapters/agno.py
+++ b/src/band/adapters/agno.py
@@ -2,7 +2,6 @@
 
 from __future__ import annotations
 
-import base64
 import json
 import logging
 import warnings
@@ -32,6 +31,7 @@ from band.runtime.prompts import render_system_prompt
 from band.runtime.tools import (
     READ_ROOM_FILE_TOOL_NAME,
     SEND_MESSAGE_TOOL_NAME,
+    decode_image_block,
     get_band_tool_category,
     image_block_placeholder,
     is_mcp_content_result,
@@ -105,11 +105,10 @@ def _make_band_entrypoint(tool_name: str) -> Callable[..., Awaitable[str | ToolR
         result = await active.execute_tool_call(tool_name, kwargs)
         if tool_name == READ_ROOM_FILE_TOOL_NAME and is_mcp_content_result(result):
             images = [
-                Image(
-                    content=base64.b64decode(block["data"]),
-                    mime_type=block["mimeType"],
+                Image(content=data, mime_type=mime_type)
+                for data, mime_type in (
+                    decode_image_block(block) for block in result["content"]
                 )
-                for block in result["content"]
             ]
             return ToolResult(
                 content=image_block_placeholder(len(images)), images=images
--- a/src/band/adapters/agno.py
+++ b/src/band/adapters/agno.py
@@ -132,7 +132,7 @@ class AgnoAdapter(SimpleAdapter[AgnoMessages]):
         {Emit.TOOL_CALLS, Emit.THOUGHTS, Emit.USAGE}
     )
     SUPPORTED_CAPABILITIES: ClassVar[frozenset[Capability]] = frozenset(
-        {Capability.MEMORY, Capability.CONTACTS}
+        {Capability.MEMORY, Capability.CONTACTS, Capability.FILES}
     )
 
     def __init__(
--- a/src/band/adapters/anthropic.py
+++ b/src/band/adapters/anthropic.py
@@ -57,7 +57,7 @@ class AnthropicAdapter(SimpleAdapter[AnthropicMessages]):
 
     SUPPORTED_EMIT: ClassVar[frozenset[Emit]] = frozenset({Emit.TOOL_CALLS, Emit.USAGE})
     SUPPORTED_CAPABILITIES: ClassVar[frozenset[Capability]] = frozenset(
-        {Capability.MEMORY, Capability.CONTACTS}
+        {Capability.MEMORY, Capability.CONTACTS, Capability.FILES}
     )
 
     def __init__(
--- a/src/band/adapters/anthropic.py
+++ b/src/band/adapters/anthropic.py
@@ -35,9 +35,9 @@ from band.runtime.custom_tools import (
 )
 from band.runtime.prompts import render_system_prompt
 from band.runtime.tools import (
-    BandTool,
     image_block_placeholder,
-    is_mcp_content_result,
+    is_image_passthrough_result,
+    redact_tool_call_args,
 )
 
 logger = logging.getLogger(__name__)
@@ -450,7 +450,9 @@ class AnthropicAdapter(SimpleAdapter[AnthropicMessages]):
                         content=json.dumps(
                             {
                                 ToolEventKey.NAME: tool_name,
-                                ToolEventKey.ARGS: tool_input,
+                                ToolEventKey.ARGS: redact_tool_call_args(
+                                    tool_name, tool_input
+                                ),
                                 ToolEventKey.TOOL_CALL_ID: tool_use_id,
                             }
                         ),
@@ -469,9 +471,7 @@ class AnthropicAdapter(SimpleAdapter[AnthropicMessages]):
                     result = await execute_custom_tool(custom_tool, tool_input)
                 else:
                     result = await tools.execute_tool_call(tool_name, tool_input)
-                if tool_name == BandTool.READ_ROOM_FILE and is_mcp_content_result(
-                    result
-                ):
+                if is_image_passthrough_result(tool_name, result):
                     content = _image_tool_result_content(result)
                     result_str = image_block_placeholder(len(content))
                 else:
--- a/src/band/adapters/anthropic.py
+++ b/src/band/adapters/anthropic.py
@@ -35,7 +35,7 @@ from band.runtime.custom_tools import (
 )
 from band.runtime.prompts import render_system_prompt
 from band.runtime.tools import (
-    READ_ROOM_FILE_TOOL_NAME,
+    BandTool,
     image_block_placeholder,
     is_mcp_content_result,
 )
@@ -44,13 +44,10 @@ logger = logging.getLogger(__name__)
 
 
 def _image_tool_result_content(result: dict[str, Any]) -> list[dict[str, Any]]:
-    """Convert an MCP-content-shaped ``band_read_room_file`` result into
-    Anthropic ``ImageBlockParam`` dicts for a ``tool_result``'s content list.
+    """An image tool result as Anthropic ``ImageBlockParam`` dicts.
 
-    Anthropic's ``media_type`` (jpeg/png/gif/webp) is exactly
-    ``PREVIEWABLE_IMAGE_CONTENT_TYPES`` -- the only content types
-    ``AgentTools.read_room_file`` ever inlines as an image -- so every
-    ``mimeType`` this sees is already a valid ``media_type``.
+    No media_type validation: Anthropic's accepted set is exactly
+    ``PREVIEWABLE_IMAGE_CONTENT_TYPES``, the only types read_room_file inlines.
     """
     return [
         {
@@ -472,7 +469,7 @@ class AnthropicAdapter(SimpleAdapter[AnthropicMessages]):
                     result = await execute_custom_tool(custom_tool, tool_input)
                 else:
                     result = await tools.execute_tool_call(tool_name, tool_input)
-                if tool_name == READ_ROOM_FILE_TOOL_NAME and is_mcp_content_result(
+                if tool_name == BandTool.READ_ROOM_FILE and is_mcp_content_result(
                     result
                 ):
                     content = _image_tool_result_content(result)
--- a/src/band/adapters/anthropic.py
+++ b/src/band/adapters/anthropic.py
@@ -34,7 +34,11 @@ from band.runtime.custom_tools import (
     find_custom_tool,
 )
 from band.runtime.prompts import render_system_prompt
-from band.runtime.tools import READ_ROOM_FILE_TOOL_NAME, is_mcp_content_result
+from band.runtime.tools import (
+    READ_ROOM_FILE_TOOL_NAME,
+    image_block_placeholder,
+    is_mcp_content_result,
+)
 
 logger = logging.getLogger(__name__)
 
@@ -472,7 +476,7 @@ class AnthropicAdapter(SimpleAdapter[AnthropicMessages]):
                     result
                 ):
                     content = _image_tool_result_content(result)
-                    result_str = f"<{len(content)} image content block(s)>"
+                    result_str = image_block_placeholder(len(content))
                 else:
                     content = result_str = (
                         json.dumps(result, default=str)
--- a/src/band/adapters/anthropic.py
+++ b/src/band/adapters/anthropic.py
@@ -34,8 +34,31 @@ from band.runtime.custom_tools import (
     find_custom_tool,
 )
 from band.runtime.prompts import render_system_prompt
+from band.runtime.tools import is_mcp_content_result
 
 logger = logging.getLogger(__name__)
+
+
+def _image_tool_result_content(result: dict[str, Any]) -> list[dict[str, Any]]:
+    """Convert an MCP-content-shaped ``band_read_room_file`` result into
+    Anthropic ``ImageBlockParam`` dicts for a ``tool_result``'s content list.
+
+    Anthropic's ``media_type`` (jpeg/png/gif/webp) is exactly
+    ``PREVIEWABLE_IMAGE_CONTENT_TYPES`` -- the only content types
+    ``AgentTools.read_room_file`` ever inlines as an image -- so every
+    ``mimeType`` this sees is already a valid ``media_type``.
+    """
+    return [
+        {
+            "type": "image",
+            "source": {
+                "type": "base64",
+                "media_type": block["mimeType"],
+                "data": block["data"],
+            },
+        }
+        for block in result["content"]
+    ]
 
 
 class AnthropicAdapter(SimpleAdapter[AnthropicMessages]):
@@ -445,14 +468,18 @@ class AnthropicAdapter(SimpleAdapter[AnthropicMessages]):
                     result = await execute_custom_tool(custom_tool, tool_input)
                 else:
                     result = await tools.execute_tool_call(tool_name, tool_input)
-                result_str = (
-                    json.dumps(result, default=str)
-                    if not isinstance(result, str)
-                    else result
-                )
+                if tool_name == "band_read_room_file" and is_mcp_content_result(result):
+                    content = _image_tool_result_content(result)
+                    result_str = f"<{len(content)} image content block(s)>"
+                else:
+                    content = result_str = (
+                        json.dumps(result, default=str)
+                        if not isinstance(result, str)
+                        else result
+                    )
                 is_error = False
             except Exception as e:
-                result_str = f"Error: {e}"
+                content = result_str = f"Error: {e}"
                 is_error = True
                 logger.error("Tool %s failed: %s", tool_name, e)
 
@@ -480,7 +507,7 @@ class AnthropicAdapter(SimpleAdapter[AnthropicMessages]):
                 {
                     "type": "tool_result",
                     "tool_use_id": tool_use_id,
-                    "content": result_str,
+                    "content": content,
                     "is_error": is_error,
                 }
             )
--- a/src/band/adapters/anthropic.py
+++ b/src/band/adapters/anthropic.py
@@ -34,7 +34,7 @@ from band.runtime.custom_tools import (
     find_custom_tool,
 )
 from band.runtime.prompts import render_system_prompt
-from band.runtime.tools import is_mcp_content_result
+from band.runtime.tools import READ_ROOM_FILE_TOOL_NAME, is_mcp_content_result
 
 logger = logging.getLogger(__name__)
 
@@ -468,7 +468,9 @@ class AnthropicAdapter(SimpleAdapter[AnthropicMessages]):
                     result = await execute_custom_tool(custom_tool, tool_input)
                 else:
                     result = await tools.execute_tool_call(tool_name, tool_input)
-                if tool_name == "band_read_room_file" and is_mcp_content_result(result):
+                if tool_name == READ_ROOM_FILE_TOOL_NAME and is_mcp_content_result(
+                    result
+                ):
                     content = _image_tool_result_content(result)
                     result_str = f"<{len(content)} image content block(s)>"
                 else:
--- a/src/band/adapters/copilot_sdk.py
+++ b/src/band/adapters/copilot_sdk.py
@@ -46,6 +46,7 @@ from band.runtime.tools import (
     CHAT_ID_FIELD_NAME,
     READ_ROOM_FILE_TOOL_NAME,
     get_band_tool_category,
+    image_block_placeholder,
     is_mcp_content_result,
     is_room_posting_tool,
 )
@@ -759,7 +760,7 @@ class CopilotSDKAdapter(SimpleAdapter[CopilotSDKSessionState]):
                 )
                 for block in result["content"]
             ]
-            text_result = f"<{len(binary_results)} image content block(s)>"
+            text_result = image_block_placeholder(len(binary_results))
         else:
             text_result = (
                 result if isinstance(result, str) else json.dumps(result, default=str)
--- a/src/band/adapters/copilot_sdk.py
+++ b/src/band/adapters/copilot_sdk.py
@@ -44,7 +44,7 @@ from band.runtime.custom_tools import (
 from band.runtime.prompts import render_system_prompt
 from band.runtime.tools import (
     CHAT_ID_FIELD_NAME,
-    READ_ROOM_FILE_TOOL_NAME,
+    BandTool,
     get_band_tool_category,
     image_block_placeholder,
     is_mcp_content_result,
@@ -753,7 +753,7 @@ class CopilotSDKAdapter(SimpleAdapter[CopilotSDKSessionState]):
             )
 
         binary_results: list[ToolBinaryResult] | None = None
-        if tool_name == READ_ROOM_FILE_TOOL_NAME and is_mcp_content_result(result):
+        if tool_name == BandTool.READ_ROOM_FILE and is_mcp_content_result(result):
             binary_results = [
                 ToolBinaryResult(
                     data=block["data"], mime_type=block["mimeType"], type="image"
--- a/src/band/adapters/copilot_sdk.py
+++ b/src/band/adapters/copilot_sdk.py
@@ -44,11 +44,11 @@ from band.runtime.custom_tools import (
 from band.runtime.prompts import render_system_prompt
 from band.runtime.tools import (
     CHAT_ID_FIELD_NAME,
-    BandTool,
     get_band_tool_category,
     image_block_placeholder,
-    is_mcp_content_result,
+    is_image_passthrough_result,
     is_room_posting_tool,
+    redact_tool_call_args,
 )
 
 try:
@@ -753,13 +753,22 @@ class CopilotSDKAdapter(SimpleAdapter[CopilotSDKSessionState]):
             )
 
         binary_results: list[ToolBinaryResult] | None = None
-        if tool_name == BandTool.READ_ROOM_FILE and is_mcp_content_result(result):
-            binary_results = [
-                ToolBinaryResult(
-                    data=block["data"], mime_type=block["mimeType"], type="image"
+        if is_image_passthrough_result(tool_name, result):
+            # Same failure path as the tool-execution try/except above: a
+            # malformed or future-extended content block must still route
+            # through _fail_tool_call, not raise uncaught past it.
+            try:
+                binary_results = [
+                    ToolBinaryResult(
+                        data=block["data"], mime_type=block["mimeType"], type="image"
+                    )
+                    for block in result["content"]
+                ]
+            except (KeyError, TypeError) as exc:
+                logger.exception("Malformed image content block for %s", tool_name)
+                return await self._fail_tool_call(
+                    room_tools, invocation, f"Error: {exc}", report=should_report
                 )
-                for block in result["content"]
-            ]
             text_result = image_block_placeholder(len(binary_results))
         else:
             text_result = (
@@ -804,7 +813,9 @@ class CopilotSDKAdapter(SimpleAdapter[CopilotSDKSessionState]):
             json.dumps(
                 {
                     ToolEventKey.NAME: invocation.tool_name,
-                    ToolEventKey.ARGS: arguments,
+                    ToolEventKey.ARGS: redact_tool_call_args(
+                        invocation.tool_name, arguments
+                    ),
                     ToolEventKey.TOOL_CALL_ID: invocation.tool_call_id,
                 }
             ),
--- a/src/band/adapters/copilot_sdk.py
+++ b/src/band/adapters/copilot_sdk.py
@@ -44,6 +44,7 @@ from band.runtime.custom_tools import (
 from band.runtime.prompts import render_system_prompt
 from band.runtime.tools import (
     CHAT_ID_FIELD_NAME,
+    READ_ROOM_FILE_TOOL_NAME,
     get_band_tool_category,
     is_mcp_content_result,
     is_room_posting_tool,
@@ -751,7 +752,7 @@ class CopilotSDKAdapter(SimpleAdapter[CopilotSDKSessionState]):
             )
 
         binary_results: list[ToolBinaryResult] | None = None
-        if tool_name == "band_read_room_file" and is_mcp_content_result(result):
+        if tool_name == READ_ROOM_FILE_TOOL_NAME and is_mcp_content_result(result):
             binary_results = [
                 ToolBinaryResult(
                     data=block["data"], mime_type=block["mimeType"], type="image"
--- a/src/band/adapters/copilot_sdk.py
+++ b/src/band/adapters/copilot_sdk.py
@@ -45,11 +45,18 @@ from band.runtime.prompts import render_system_prompt
 from band.runtime.tools import (
     CHAT_ID_FIELD_NAME,
     get_band_tool_category,
+    is_mcp_content_result,
     is_room_posting_tool,
 )
 
 try:
-    from copilot import CopilotClient, PermissionHandler, Tool, ToolResult
+    from copilot import (
+        CopilotClient,
+        PermissionHandler,
+        Tool,
+        ToolBinaryResult,
+        ToolResult,
+    )
     from copilot.generated.session_events import (
         AssistantReasoningData,
         AssistantUsageData,
@@ -743,14 +750,26 @@ class CopilotSDKAdapter(SimpleAdapter[CopilotSDKSessionState]):
                 room_tools, invocation, f"Error: {exc}", report=should_report
             )
 
-        text_result = (
-            result if isinstance(result, str) else json.dumps(result, default=str)
-        )
+        binary_results: list[ToolBinaryResult] | None = None
+        if tool_name == "band_read_room_file" and is_mcp_content_result(result):
+            binary_results = [
+                ToolBinaryResult(
+                    data=block["data"], mime_type=block["mimeType"], type="image"
+                )
+                for block in result["content"]
+            ]
+            text_result = f"<{len(binary_results)} image content block(s)>"
+        else:
+            text_result = (
+                result if isinstance(result, str) else json.dumps(result, default=str)
+            )
         if is_room_posting_tool(tool_name) and turn is not None:
             self._mark_replied_in_room(room_id, turn)
         if should_report:
             await self._report_tool_result(room_tools, invocation, text_result)
-        return ToolResult(text_result_for_llm=text_result)
+        return ToolResult(
+            text_result_for_llm=text_result, binary_results_for_llm=binary_results
+        )
 
     async def _fail_tool_call(
         self,
--- a/src/band/adapters/copilot_sdk.py
+++ b/src/band/adapters/copilot_sdk.py
@@ -214,7 +214,7 @@ class CopilotSDKAdapter(SimpleAdapter[CopilotSDKSessionState]):
         {Emit.TOOL_CALLS, Emit.THOUGHTS, Emit.USAGE}
     )
     SUPPORTED_CAPABILITIES: ClassVar[frozenset[Capability]] = frozenset(
-        {Capability.MEMORY, Capability.CONTACTS}
+        {Capability.MEMORY, Capability.CONTACTS, Capability.FILES}
     )
 
     def __init__(
--- a/src/band/adapters/google_adk.py
+++ b/src/band/adapters/google_adk.py
@@ -288,7 +288,7 @@ class GoogleADKAdapter(SimpleAdapter[GoogleADKMessages]):
 
     SUPPORTED_EMIT: ClassVar[frozenset[Emit]] = frozenset({Emit.TOOL_CALLS, Emit.USAGE})
     SUPPORTED_CAPABILITIES: ClassVar[frozenset[Capability]] = frozenset(
-        {Capability.MEMORY, Capability.CONTACTS}
+        {Capability.MEMORY, Capability.CONTACTS, Capability.FILES}
     )
 
     def __init__(
--- a/src/band/adapters/google_adk.py
+++ b/src/band/adapters/google_adk.py
@@ -37,6 +37,12 @@ from band.runtime.custom_tools import (
     find_custom_tool,
 )
 from band.runtime.prompts import render_system_prompt
+from band.runtime.tools import (
+    BandTool,
+    image_block_placeholder,
+    is_image_passthrough_result,
+    redact_tool_call_args,
+)
 
 if TYPE_CHECKING:
     from google.adk.runners import InMemoryRunner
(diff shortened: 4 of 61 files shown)


tests/e2e/baseline/toolkit/builders.py
```diff
@@ -36,9 +36,10 @@
 from tests.e2e.baseline.toolkit.tools import ToolSpec
 
 _LLM_TOOL_LOOP = (Capability.MEMORY, Capability.CONTACTS)
+_FILE_CAPABLE = (*_LLM_TOOL_LOOP, Capability.FILES)
```
AlexanderZ-Band: Is this indirection needed?

tests/e2e/baseline/toolkit/provisioning.py
```diff
@@ -170,6 +190,68 @@ async def send_message(
...
+        """Upload ``body`` as an attachment and post it as this peer; return the
+        message id.
+
+        Mirrors AgentTools.send_room_file's own upload call
+        (src/band/runtime/tools.py) -- same headers, same attach-then-message
+        two-step -- since the platform's file-upload endpoint is agent-scoped
+        (Agent API only; there is no user-side upload endpoint), so a live
+        vision-passthrough scenario needs a peer agent to play "someone already
+        shared a file here," not the test's own UserOps driver.
+
```
AlexanderZ-Band: Very long comment is it really needed or restating what the code already tells us?

tests/framework_conformance/test_adapter_conformance.py
```diff
@@ -212,3 +214,145 @@ def test_omitted_emit_defaults_to_supported_emit(self, adapter_config):
...
+
+
+# Every adapter used to hand-roll one wrapper function/class per platform
+# tool instead of deriving its tool surface from the central registry
+# (iter_tool_definitions/get_openai_tool_schemas/get_anthropic_tool_schemas).
+# Capability.FILES needed three new hand-written wrappers on each --
+# claude_sdk-style adapters get it generically, but parlant, pydantic_ai, and
+# crewai/crewai_flow (which share integrations/crewai/tools.py) all grew real
+# band_list_room_files/band_read_room_file/band_send_room_file wrappers.
+# Empty now; kept as the seam a genuinely new hand-rolled adapter joins.
```
AlexanderZ-Band: Why is it here? Cleanup

tests/framework_conformance/test_adapter_conformance.py
```diff
@@ -212,3 +214,145 @@ def test_omitted_emit_defaults_to_supported_emit(self, adapter_config):
...
+# google_adk stays out for the reason above. letta needs no adapter-side row: it
+# routes tool execution entirely through the already-fixed MCP engine (it never
+# calls execute_tool_call), so the image fix is already live for it via the
+# opencode/ACP-client row -- the one unverifiable hop is Letta's own backend
+# turning an MCP ImageContent into its ToolReturn.tool_return image part, which
+# isn't observable from the installed letta_client package. parlant stays out:
+# verified that parlant.core.tools.ToolResult has no multimodal field, and
+# Parlant's own MCP integration (mcp_result_to_tool_result_data()) discards
+# image content blocks before they ever reach ToolResult.data -- a confirmed
+# framework limitation, not an adapter gap.
```
AlexanderZ-Band: Not needed cleanup

tests/framework_conformance/test_adapter_conformance.py
```diff
@@ -212,3 +214,145 @@ def test_omitted_emit_defaults_to_supported_emit(self, adapter_config):
...
+IMAGE_PASSTHROUGH_SUPPORTED_FRAMEWORK_IDS = frozenset(
+    {
+        "claude_sdk",
+        "anthropic",
```
AlexanderZ-Band: Easter to say what to exclude , use decorators

tests/framework_conformance/test_files_image_passthrough_matrix.py
```diff
@@ -0,0 +1,374 @@
...
+        capabilities=frozenset({Capability.FILES}),
+    )
+    read_room_file = next(t for t in tools if t.name == "band_read_room_file")
+
+    result = read_room_file._run(file_id="file-1")
+
+    return result == vision_sentinel(_IMAGE_RESULT)
+
+
+IMAGE_PASSTHROUGH_PROBES: dict[str, Callable[[], Awaitable[bool]]] = {
```
AlexanderZ-Band, line 349: This pattern repeats itself scan cleanup

src/band/integrations/crewai/tools.py
```diff
@@ -236,6 +240,23 @@ def serialize_success_result(result: Any) -> str:
...
 
 
+def vision_sentinel(result: dict[str, Any]) -> str:
+    """Encode an MCP-content-shaped image result as CrewAI's native vision
+    sentinel string, so the model gets real image content instead of a
+    JSON-stringified blob.
+
+    CrewAI's own StepExecutor (crewai/agents/step_executor.py, on the default
+    native-tool-calling execution path) recognizes a tool result string
+    shaped ``VISION_IMAGE:<media_type>:<base64_data>`` and rewrites it into a
```
AlexanderZ-Band: To long a comment for such. A small pice of code makes no sense

src/band/integrations/crewai/tools.py
```diff
@@ -831,6 +863,81 @@ async def execute(tools: AgentToolsProtocol) -> str:
 
             return _exec("band_archive_memory", execute)
 
+    class ListRoomFilesTool(BaseTool):
```
AlexanderZ-Band: Consider refactoring into a separate file as well as using decorators to hide plumbing , common lifecycle code refactoring
Think how to make it less verbose

src/band/integrations/langgraph/langchain_tools.py
```diff
@@ -112,7 +114,19 @@ async def execute_definition(
             **kwargs: Any,
```
AlexanderZ-Band, line 126: Let's try to avoid any, can we define a common stronger type for similar classes

src/band/integrations/mcp/engine.py
```diff
@@ -77,6 +80,14 @@ class MCPToolRegistration:
     description: str
     input_model: type[BaseModel]
     execute: MCPToolExecutor
+    # Passed straight through to FastMCP's add_tool(structured_output=...).
+    # None (default) preserves today's auto-inferred behavior for every
```
AlexanderZ-Band: Very long narration comment

src/band/integrations/mcp/engine.py
```diff
@@ -606,6 +622,21 @@ def build_resolved_band_mcp_tool_registrations(
...
+def _mcp_content_blocks(result: dict[str, Any]) -> list[ImageContent]:
+    """Convert an MCP-content-shaped tool result into real MCP content blocks.
+
+    ``is_mcp_content_result`` already confirmed ``result["content"]`` is a
+    list of MCP content block dicts. FastMCP's own result conversion
+    recognizes an actual ``ContentBlock`` instance (or a list of them) and
+    passes it through to the client verbatim -- a plain dict does not match
+    and would otherwise fall through to JSON-text encoding, same as any
+    other tool result. Only "image" blocks exist today (the sole structured
+    branch of ``AgentTools.read_room_file``); a block of any other shape
```
AlexanderZ-Band: Very long comment on liner function

src/band/runtime/tools.py
```diff
@@ -851,6 +851,12 @@ class ListMyPeersInput(BaseModel):
 # Single source of truth so none of those re-type the literal independently.
 SEND_MESSAGE_TOOL_NAME = "band_send_message"
 
+# Every adapter's image-vision-passthrough branch gates on this exact tool
+# name (is_mcp_content_result(result) only means anything for this one tool's
+# result shape). Single source of truth so none of those re-type the literal
+# independently -- same rationale as SEND_MESSAGE_TOOL_NAME above.
+READ_ROOM_FILE_TOOL_NAME = "band_read_room_file"
```
AlexanderZ-Band: Feels like those should be strenum

src/band/runtime/tools.py
```diff
@@ -1782,6 +1788,47 @@ def serialize_tool_result(result: Any) -> Any:
...
+        isinstance(data, dict)
+        and isinstance(data.get("content"), list)
+        and bool(data["content"])
+        and all(
+            isinstance(block, dict) and "type" in block for block in data["content"]
+        )
+    )
+
+
+def image_block_placeholder(block_count: int) -> str:
```
AlexanderZ-Band, line 1864: Review the whole pr on liner function can't have long comments it makes no sense

src/band/integrations/crewai/tools.py
AlexanderZ-Band: Trimmed to one sentence: this is CrewAI's own undocumented internal StepExecutor sentinel protocol, verified against the installed package rather than the docs (d6456552).

src/band/integrations/crewai/tools.py
AlexanderZ-Band: Split crewai/tools.py into reporting.py (context+reporters), catalog.py (the 20 tool bodies), tools.py (build+run). Each tool body is now a single @band_tool async function; the decorator owns report_call/report_result/serialize lifecycle. 1119 -> 906 lines across the three files. Verified byte-identical advertised tool schema/description across all 8 capability combinations before and after (d6456552).

src/band/integrations/langgraph/langchain_tools.py
AlexanderZ-Band: Image branch now builds langchain_core's own ImageContentBlock TypedDict instead of a raw dict. execute_definition's return stays Any because the protocol it implements (AgentToolsProtocol.execute_tool_call) is itself declared -> Any -- narrowing here would be a lie the type checker can't verify. Filed as a follow-up: a real ToolResultValue alias on the protocol itself would let ~7 adapters drop Any, but that's a contract change beyond this PR's scope (d6456552).

src/band/integrations/mcp/engine.py
AlexanderZ-Band: Trimmed 7 lines to 2 (d6456552).

src/band/integrations/mcp/engine.py
AlexanderZ-Band: Trimmed 11 lines to 4, kept only the misfire-warning fact (d6456552).

src/band/runtime/tools.py
AlexanderZ-Band: Converted fully: BandTool(StrEnum) with all 20 agent-surface tool names, both loose constants deleted, 81 references across 18 files now use enum members. Added an import-time guard that frozenset(BandTool) matches the registry. Verified hash/dict-lookup/json.dumps/f-string all behave identically to the plain str (d6456552).

src/band/runtime/tools.py
AlexanderZ-Band: AST-scanned every PR-added function with a <=2-statement body and a >2-line docstring; trimmed image_block_placeholder, decode_image_block, is_mcp_content_result, anthropic._image_tool_result_content, codex._image_content_items, plus three test helpers and the parlant declarative-refactor helpers. Left the sample_agents.py prompt builders and IMAGE_PASSTHROUGH_EXCLUSIONS comments alone -- those document live-observed model behavior / a closed vocabulary, not a one-liner's mechanics (d6456552).

src/band/adapters/pydantic_ai.py
```diff
@@ -639,6 +642,58 @@ async def band_archive_memory(
...
+                try:
+                    result = await ctx.deps.read_room_file(file_id)
+                    if is_mcp_content_result(result):
+                        return [
+                            BinaryContent(data=data, media_type=mime_type)
+                            for data, mime_type in (
+                                decode_image_block(block) for block in result["content"]
+                            )
+                        ]
+                    return result
```
amit-gazal-band, line 676: Not a bug in this hunk, but a gap this PR should close nearby: at `pydantic_ai.py:841`, `ToolEventKey.OUTPUT: str(event.part.content)` embeds raw image bytes when `Capability.FILES` + `Emit.TOOL_CALLS` are enabled and the agent reads an image room attachment via `band_read_room_file`. This tool (added in this hunk) returns `list[BinaryContent]`, and `BinaryContent.__repr__` always includes its `data: bytes` field, so `str()` on that list puts the full raw image bytes into the emitted `tool_result` event. This is the same class of leak fixed for CrewAI in `cfc3da9d` (`read_room_file` -> bounded placeholder) — this adapter needs the equivalent fix at line 841.

src/band/integrations/crewai/catalog.py
```diff
@@ -0,0 +1,488 @@
...
+    await call.reporter.report_result(call.tools, BandTool.READ_ROOM_FILE, result)
+    return result
+
+
+@band_tool(
+    BandTool.SEND_ROOM_FILE,
+    args_schema=SEND_ROOM_FILE_ARGS_SCHEMA,
+    normalize=_posted_file_args,
+)
+async def _send_room_file(
```
amit-gazal-band, line 482: `_send_room_file` has no `reports=False`/bounded override the way `_read_room_file` does (which uses `image_block_placeholder`). With `Emit.TOOL_CALLS` enabled and a large `content` payload (up to `MAX_SEND_CONTENT_BYTES`, ~1MB), the default `reports=True` path calls `reporter.report_call` with the full args dict, and `EmitToolCallsReporter.report_call` `json.dumps()`s it with no size check — embedding up to ~1MB of raw file content into the `tool_call` event.

src/band/integrations/parlant/tools.py
```diff
@@ -169,507 +229,418 @@ def create_parlant_tools(features: AdapterFeatures | None = None) -> list[Any]:
...
+                )
+                return ToolResult(data=NO_SESSION_TOOLS_ERROR)
+            except Exception as exc:
+                context_phrase = failure.format(**call.arguments)
+                logger.error(
+                    "%s Error %s: %s", LOG_PREFIX, context_phrase, exc, exc_info=True
+                )
+                message = str(exc)
+                if mention_hints and isinstance(exc, (ValueError, BandToolError)):
+                    message = with_mention_handles(
```
amit-gazal-band: This re-fetches session tools independently of the outer lookup: if the session/room was torn down between the initial call and this except block (i.e. `set_session_tools(session_id, None)`), `get_session_tools(context.session_id)` returns `None`, and `with_mention_handles(message, None)` immediately does `None.participants`, raising an unhandled `AttributeError` instead of returning a `ToolResult` with the original error.

src/band/adapters/strands.py
```diff
@@ -96,6 +121,8 @@ def _result_text(result: ToolResult) -> str:
                 parts.append(text)
             case {"json": value}:
                 parts.append(_format_tool_output(value))
+            case {"image": _}:
+                parts.append(image_block_placeholder(1))
```
amit-gazal-band: This loop calls `image_block_placeholder(1)` once per image block instead of computing the total image count once. For a multi-image `read_room_file` result, `_tool_result` correctly builds one content block per image, but this produces N separate `"<1 image content block(s)>"` lines instead of one `"<N image content block(s)>"` line, misrepresenting the count in execution events/terminal-state logging.

src/band/integrations/letta/prompts.py
```diff
@@ -11,7 +11,7 @@
 # the enforcement prompt, silent-reporting set, and auto-relay detection all
 # follow whichever server is wired — first entry doubles as the fallback.
 SEND_MESSAGE_TOOL_NAMES: tuple[str, ...] = (
-    SEND_MESSAGE_TOOL_NAME,
+    BandTool.SEND_MESSAGE,
```
amit-gazal-band, line 14: `SEND_MESSAGE_TOOL_NAMES` was migrated here to `BandTool.SEND_MESSAGE`, but the sibling `SEND_EVENT_TOOL_NAMES` three lines below (line 18) still hardcodes the literal `"band_send_event"` instead of `BandTool.SEND_EVENT`. The whole point of `BandTool` is to prevent this constant from drifting from `runtime/tools.py` — that sibling tuple was missed.

src/band/integrations/parlant/tools.py
```diff
@@ -169,507 +229,418 @@ def create_parlant_tools(features: AdapterFeatures | None = None) -> list[Any]:
...
+            band_list_room_files,
+            band_read_room_file,
+            band_send_room_file,
+        ]
+
+    capabilities = features.capabilities if features else None
+    tools = chat_tools()
+    if capabilities is None or Capability.CONTACTS in capabilities:
+        tools += contact_tools()
+    if capabilities is None or Capability.FILES in capabilities:
```
amit-gazal-band, line 661: (Lower confidence) `create_parlant_tools()` registers file tools by default when `capabilities` is `None`, same as the pre-existing `CONTACTS` pattern above it. Given `Capability.FILES` gates an on-prem-only platform flag, worth confirming `ParlantAdapter`'s real constructor path never leaves `features` unset in production — if it can, room-file tools would be exposed without explicit opt-in.

amit-gazal-band (changes requested):
Automated code review: 5 correctness/consistency findings and 1 lower-confidence one, all inline below. Two comments are anchored to the nearest in-diff line since the actual bug lives outside this PR's diff hunks (GitHub can't attach inline comments there) — the comment body names the real location.

src/band/adapters/pydantic_ai.py
AlexanderZ-Band: Confirmed and fixed (2899e606): the tool_result event now checks whether event.part.content is a list[BinaryContent] and, if so, reports image_block_placeholder(len(output)) instead of str(output) -- str() on BinaryContent does embed the raw data bytes via its repr, verified live against the installed pydantic-ai package. Regression test: tests/adapters/test_pydantic_ai_adapter.py::TestExecutionReporting::test_tool_result_event_redacts_binary_content (fails on the old code, passes now).

src/band/integrations/crewai/catalog.py
AlexanderZ-Band: Confirmed and fixed (2899e606): _send_room_file now sets reports=False and manually reports a redacted args dict where content is replaced with a bounded file_content_placeholder(byte_count) (new helper next to image_block_placeholder), mirroring how _read_room_file already bounds its own result. Note the real cap is 16384 chars (AgentTools.send_event truncates every event), not the full ~1MB MAX_SEND_CONTENT_BYTES -- but embedding any real file bytes into a tool_call log event was still the wrong idiom here, so fixed regardless. A self-review pass also caught that the placeholder's byte_count must be len(content.encode('utf-8')), not len(content) -- a plain char count understates true size for non-ASCII content; fixed and covered by a non-ASCII regression case. Test: tests/integrations/test_crewai_tools.py::TestFileTools::test_send_room_file_reports_content_placeholder_not_raw_bytes.

src/band/integrations/parlant/tools.py
AlexanderZ-Band: Confirmed and fixed (2899e606): guard_failures now only fetches get_session_tools when mention_hints is set (restoring the original short-circuit -- get_session_tools also logs at debug level, so calling it unconditionally on every tool's every failure was wasted work too) and skips with_mention_handles entirely when the session already tore down, falling back to the plain error message instead of crashing on None.participants. Regression test: tests/integrations/parlant/test_tools.py::TestParlantToolFunctions::test_send_message_mention_hint_survives_session_teardown_race, which fails with the exact AttributeError on the pre-fix code.

src/band/adapters/strands.py
AlexanderZ-Band: Confirmed and fixed (2899e606): _result_text now tallies image blocks once and emits a single image_block_placeholder(image_count) at the position of the first image block (order-preserving, in case a result ever mixes image and text/json blocks), instead of one placeholder line per block. One nuance for the record: read_room_file's current contract (src/band/runtime/tools.py) only ever returns 0 or 1 image block per call (one file_id in, at most one MCP image content block out), so a >1-image result isn't reachable today -- the fix still lands since the loop is written defensively as if multi-block results were possible. Regression test: tests/adapters/test_strands_adapter.py::TestReadRoomFileImagePassthrough::test_multi_image_result_reports_one_placeholder_with_total_count (plus an order-preservation test for the mixed-content case).

src/band/integrations/letta/prompts.py
AlexanderZ-Band: Confirmed and fixed (2899e606): SEND_EVENT_TOOL_NAMES[0] now uses BandTool.SEND_EVENT instead of the hardcoded "band_send_event" literal, matching its sibling SEND_MESSAGE_TOOL_NAMES. Regression test: tests/adapters/test_letta_mcp.py::test_send_tool_names_use_band_tool_enum_member asserts both tuples' first entries are BandTool members, not bare strings (fails on the pre-fix literal).

src/band/integrations/parlant/tools.py
AlexanderZ-Band: Investigated, not a bug: AdapterFeatures is a frozen dataclass with no __bool__/__len__ override, so any real instance -- including one built with every field at its default empty value -- is truthy. ParlantAdapter's only call site (_prepare_server: self._tools = create_parlant_tools(self.features)) always passes the real self.features built by SimpleAdapter.__init__, which is never None, so `capabilities = features.capabilities if features else None` never actually falls into the `capabilities is None` branch in production -- capabilities is always the real (possibly empty) frozenset, and Capability.FILES gates it exactly like CONTACTS already does above it. The `features=None` default only matters for direct/test calls to create_parlant_tools() with no adapter behind it. No change made.

src/band/adapters/google_adk.py
```diff
@@ -288,7 +288,7 @@ class GoogleADKAdapter(SimpleAdapter[GoogleADKMessages]):
 
     SUPPORTED_EMIT: ClassVar[frozenset[Emit]] = frozenset({Emit.TOOL_CALLS, Emit.USAGE})
     SUPPORTED_CAPABILITIES: ClassVar[frozenset[Capability]] = frozenset(
-        {Capability.MEMORY, Capability.CONTACTS}
+        {Capability.MEMORY, Capability.CONTACTS, Capability.FILES}
```
amit-gazal-band, line 323: **Image content leaks into the platform-visible `tool_result` event, now reachable now that this adapter gains `Capability.FILES` here.**

(Anchoring on this line since the actual bug is in pre-existing, untouched code -- `_report_event`'s function-response reporting at `google_adk.py:681` -- that this one-line capability addition makes reachable for the first time.)

`ToolEventKey.OUTPUT: str(fr.response)` at line 681 serializes the full function response unconditionally, with no image-specific redaction. Every other FILES-enabled adapter in this PR (CrewAI, pydantic_ai, strands, ...) now special-cases the `band_read_room_file` image branch before reporting it as an event; this adapter doesn't.

Repro: an agent with `Capability.FILES` and `Emit.TOOL_CALLS` on calls `band_read_room_file` on a previewable image. `str(fr.response)` embeds the full base64 payload into a `tool_result` platform event -- the exact leak class fixed for CrewAI in 2899e606, missed here.

src/band/adapters/langgraph.py
```diff
@@ -80,7 +80,7 @@ def graph_factory(tools):
 
     SUPPORTED_EMIT: ClassVar[frozenset[Emit]] = frozenset({Emit.TOOL_CALLS, Emit.USAGE})
     SUPPORTED_CAPABILITIES: ClassVar[frozenset[Capability]] = frozenset(
-        {Capability.MEMORY, Capability.CONTACTS}
+        {Capability.MEMORY, Capability.CONTACTS, Capability.FILES}
```
amit-gazal-band, line 117: **Same image-leak class as the google_adk.py finding, now reachable now that this adapter gains `Capability.FILES` here.**

(Anchoring on this line since the actual bug is in pre-existing, untouched code -- the `on_tool_end` handler at `langgraph.py:405` -- that this one-line capability addition makes reachable for the first time.)

When `band_read_room_file`'s image branch returns a langchain `ToolMessage` wrapping the image content block, `data.get("output", "")` at line 405 puts that non-JSON-serializable object straight into `payload`, and the `json.dumps(payload, default=str)` a few lines down falls back to `str(ToolMessage(...))`, embedding the full base64 image into the `tool_result` event sent to the platform.

src/band/adapters/anthropic.py
```diff
@@ -445,14 +469,20 @@ async def _process_tool_calls(
                     result = await execute_custom_tool(custom_tool, tool_input)
                 else:
                     result = await tools.execute_tool_call(tool_name, tool_input)
-                result_str = (
-                    json.dumps(result, default=str)
-                    if not isinstance(result, str)
-                    else result
-                )
+                if tool_name == BandTool.READ_ROOM_FILE and is_mcp_content_result(
```
amit-gazal-band: **Unbounded file content in the `tool_call` event for `band_send_room_file`.**

(Anchoring here since this is the nearest line this PR actually touched -- the real issue is a few lines up, in pre-existing code at `anthropic.py:453`: `ToolEventKey.ARGS: tool_input` in the *tool_call* reporter, separate from this hunk's tool_*result* redaction.)

`ToolEventKey.ARGS: tool_input` json.dumps's the raw tool arguments -- including up to the full ~1MB `content` argument now that `Capability.FILES` exposes `band_send_room_file` here. `send_event` truncates to ~16KB (`_EVENT_CONTENT_MAX_LENGTH`) so this is bounded rather than a full leak, but real file content (not a placeholder) still lands in a platform-visible `tool_call` event.

Same pattern independently confirmed at `pydantic_ai.py:815` (`event.part.args`) and `strands.py:333` (`event.tool_use['input']`), and present in `codex.py`, `gemini.py`, `agno.py`, `copilot_sdk.py`, `letta.py`, and `google_adk.py` too. CrewAI's `_send_room_file` already fixes this with a bounded `file_content_placeholder()` (see `crewai/catalog.py:486`) -- that fix was never generalized to the other adapters that gained `band_send_room_file` in this PR.

src/band/integrations/crewai/catalog.py
```diff
@@ -0,0 +1,505 @@
...
+        args = {**self.declared, **known}
+        return self.normalize(args) if self.normalize else args
+
+    async def invoke(self, call: Invocation, kwargs: dict[str, Any]) -> str:
+        args = self.arguments(kwargs)
+        if self.reports:
+            await call.reporter.report_call(call.tools, self.name, args)
+        result = await self.body(call, **args)
+        if self.reports:
+            await call.reporter.report_result(call.tools, self.name, result)
```
amit-gazal-band, line 129: **Several tool bodies return raw Pydantic/Fern models to `report_result`, which silently drops the `tool_result` event.**

`ToolSpec.invoke` passes the tool body's raw return value here *before* `self.render(result)` serializes it for the LLM. `_lookup_peers`, `_list_contacts`, `_list_contact_requests`, `_list_memories`, `_get_memory`, `_supersede_memory`, and `_archive_memory` all `return await call.tools.X(...)` directly -- raw response models, not dicts.

`report_result` does `json.dumps({..., OUTPUT: result})` with no `default=str`; the resulting `TypeError` is caught inside `report_result`'s own try/except and only logged as a warning. For all 7 of these tools, no `tool_result` event is ever emitted when `Emit.TOOL_CALLS` is on, even though `render()` correctly serializes the same value for the LLM-facing text.

src/band/adapters/strands.py
```diff
@@ -80,22 +86,50 @@ def _format_tool_output(value: object) -> str:
...
     """Build the framework's typed result envelope at the Strands boundary."""
+    content: list[ToolResultContent]
+    if (
+        ok
+        and tool_use["name"] == BandTool.READ_ROOM_FILE
+        and is_mcp_content_result(value)
+    ):
+        content = []
+        for block in cast(dict[str, Any], value)["content"]:
+            data, mime_type = decode_image_block(block)
```
amit-gazal-band: **Uncaught exception on a malformed image content block.**

`decode_image_block(block)` runs here with no try/except, unlike `anthropic.py` / `gemini.py` / `codex.py`, which wrap the equivalent step inside their existing tool-failure error boundary. `is_mcp_content_result` (see `runtime/tools.py:1828`) only checks each block has a `"type"` key, not `"data"`/`"mimeType"` -- so a malformed or future-extended content block reaching this path raises `KeyError`/`binascii.Error` uncaught inside the synchronous `ToolResult` builder, instead of surfacing as the adapter's normal error result.

src/band/adapters/copilot_sdk.py
```diff
@@ -743,14 +752,26 @@ async def _execute_bridged_tool(
                 room_tools, invocation, f"Error: {exc}", report=should_report
             )
 
-        text_result = (
-            result if isinstance(result, str) else json.dumps(result, default=str)
-        )
+        binary_results: list[ToolBinaryResult] | None = None
```
amit-gazal-band, line 755: **`ToolBinaryResult` construction sits outside the method's error boundary.**

This block (`block["data"]`/`block["mimeType"]`) runs after the try/except above that funnels every other failure through `_fail_tool_call`, unlike anthropic/gemini/codex's equivalent branch. A malformed content block from `read_room_file` raises an uncaught `KeyError` here, bypassing `_fail_tool_call` -- the turn fails with a raw unhandled exception instead of the adapter's normal failure result.

src/band/adapters/agno.py
```diff
@@ -88,12 +96,22 @@ def _tool_name(execution: Any) -> str:
...
+    async def _entrypoint(**kwargs: Any) -> str | ToolResult:
         active = _current_tools.get()
         if active is None:
             return f"Error: no active Band context for tool {tool_name}"
         result = await active.execute_tool_call(tool_name, kwargs)
+        if tool_name == BandTool.READ_ROOM_FILE and is_mcp_content_result(result):
+            images = [
+                Image(content=data, mime_type=mime_type)
+                for data, mime_type in (
+                    decode_image_block(block) for block in result["content"]
```
amit-gazal-band: **Uncaught exception on a malformed image content block.**

`decode_image_block()` runs in this comprehension with zero try/except anywhere in `_entrypoint`, unlike `pydantic_ai.py`'s `band_read_room_file` handling, which wraps the identical step in `try/except Exception`. Any `decode_image_block` failure raises uncaught out of the tool entrypoint Agno invokes directly, instead of returning the adapter's usual `"Error: ..."` string.

src/band/integrations/parlant/tools.py
```diff
@@ -169,507 +229,418 @@ def create_parlant_tools(features: AdapterFeatures | None = None) -> list[Any]:
...
+        ``failure`` completes ``"Error {failure}: {exc}"`` and may template the
+        call's own arguments (e.g. ``"adding participant '{identifier}'"``);
+        ``mention_hints`` appends the room's available handles to a
+        mention-related failure. ``functools.wraps`` is load-bearing: Parlant
+        introspects ``__wrapped__``, so the registered signature is *func*'s own.
+        """
+
+        @functools.wraps(func)
+        async def run(context: Any, *args: Any, **kwargs: Any) -> Any:
+            call = inspect.signature(func).bind(context, *args, **kwargs)
```
amit-gazal-band: **New unguarded surface introduced by the `guard_failures` refactor.**

`inspect.signature(func).bind(context, *args, **kwargs)` runs *before* the `try:` block below it that's meant to turn failures into `ToolResult(data=f"Error ...")`. The old per-tool implementations had no separate bind step. If `.bind()` ever raises `TypeError` (a signature/argument-shape mismatch between what Parlant passes and `func`'s introspected signature), the exception propagates straight out of the registered tool coroutine instead of becoming the intended error result.

src/band/integrations/parlant/tools.py
```diff
@@ -169,507 +229,418 @@ def create_parlant_tools(features: AdapterFeatures | None = None) -> list[Any]:
...
+            description = field.description
+            if choices := _literal_choices(field.annotation):
+                description = description.rstrip() + f" One of: {', '.join(choices)}."
+            if param_overrides and param_name in param_overrides:
+                description = description.rstrip() + param_overrides[param_name]
+            func.__annotations__[param_name] = Annotated[
+                param.annotation, ToolParameterOptions(description=description)
+            ]
+
+    def guard_failures(
```
amit-gazal-band, line 267: **Per-tool INFO/WARNING logs were dropped when this refactor collapsed each tool body into `guard_failures`/`describe_from_master`.**

`guard_failures()` only logs the initial `"<tool> called: session=..."` line and, on failure, an ERROR line. Every old per-tool success/branch INFO log -- e.g. `"Message sent successfully via tool"`, `"'%s' is already in the room"`, `"Successfully added/removed '%s'"`, `"lookup_peers result: %s"`, `"get_participants result: %s"`, `"Created chatroom: %s"` -- and the WARNING-level `"send_message: No mentions provided"` log are gone.

Not user-facing (the LLM-visible `ToolResult` text is unchanged), but an operator grepping logs for confirmation that e.g. `band_add_participant` hit the already-in-room branch, or for the empty-mentions warning, will find nothing now.

src/band/integrations/claude_sdk/tools.py
```diff
@@ -148,7 +131,7 @@ def _format_success_payload(
     result: Any,
 ) -> dict[str, Any]:
     """Keep tool result payloads stable across Claude integrations."""
-    if tool_name == "band_read_room_file" and _is_mcp_content_result(result):
+    if tool_name == BandTool.READ_ROOM_FILE and is_mcp_content_result(result):
```
amit-gazal-band: **Only this one branch of seven in this function was migrated to the new `BandTool` enum.**

This `band_read_room_file` check uses `BandTool.READ_ROOM_FILE`, but the next six branches -- `"band_send_message"` (139), `"band_send_event"` (141), `"band_add_participant"` (143), `"band_remove_participant"` (151), `"band_get_participants"` (157), `"band_create_chatroom"` (168) -- stay raw string literals (all pre-existing, untouched by this diff). That contradicts the single-source-of-truth rule this PR's own `BandTool` StrEnum migration exists to enforce: a future rename of any of these six members silently desyncs this function (the import-time `frozenset(BandTool)` guard only checks the registry, not every string-literal call site), falling through to the generic envelope branch with no error and no drift check catching it.

src/band/runtime/tools.py
```diff
@@ -1782,6 +1825,39 @@ def serialize_tool_result(result: Any) -> Any:
     return result
 
 
+def is_mcp_content_result(data: Any) -> bool:
```
amit-gazal-band, line 1828: **This shared gate is the root cause of the strands/copilot_sdk/agno uncaught-KeyError findings on this PR.**

`is_mcp_content_result()` only checks that `data["content"]` is a non-empty list of dicts each carrying a `"type"` key -- it never checks `block["type"] == "image"` or that `"data"`/`"mimeType"` are present, even though 8+ call sites across this diff treat a `True` result as license to unconditionally read `block["data"]`/`block["mimeType"]`. Tightening the shape check here (or adding a shared decode helper that validates before decoding) would fix all three uncaught-exception call sites at once instead of three separate per-adapter try/except additions.

src/band/integrations/crewai/catalog.py
```diff
@@ -0,0 +1,505 @@
...
+    *,
+    content: str = "",
+    filename: str = "",
+    caption: str = "",
+    mentions: Any = None,
+) -> Any:
+    # reports=False: report_call's default behavior would json.dumps the raw
+    # file content into a tool_call event -- report a bounded placeholder for
+    # content instead, mirroring how _read_room_file bounds its own result.
+    await call.reporter.report_call(
```
amit-gazal-band, line 493: **Reporter's "must not raise" contract isn't structurally enforced.**

`reporter.report_call` is awaited before `call.tools.send_room_file` runs, with no local try/except here. The `CrewAIToolReporter` Protocol's documented contract (reporters must not raise) is honored by the two shipped reporters today but nothing enforces it. If any future/custom reporter's `report_call` raises, the file upload is silently skipped, and if `report_result` also raises on the same broken reporter in the except path, the exception escapes uncaught through `run_async` as a raw unhandled exception instead of a clean error JSON.

tests/adapters/test_anthropic_adapter.py
```diff
@@ -277,6 +277,74 @@ async def test_reports_tool_calls_when_enabled(self, mock_tools):
         # Should have sent tool_call and tool_result events
         assert mock_tools.send_event.call_count == 2
 
+    @pytest.mark.asyncio
+    async def test_read_room_file_image_result_passes_through_as_vision_content(
```
amit-gazal-band, line 281: **No assertion on the emitted `tool_call`/`tool_result` event content for an image `band_read_room_file` result.**

This test (and `test_gemini_adapter.py`, `agno/test_adapter.py`, `copilot_sdk/test_tool_bridging.py`, and `test_codex_adapter.py`) checks that the image reaches the model correctly, but none of the five assert on what gets reported to the platform -- unlike `test_pydantic_ai_adapter.py` and `test_strands_adapter.py`, which do assert on it directly, and CrewAI, which got test coverage alongside its own fix (2899e606).

A future regression that drops the image-redaction branch in any of these five adapters (see the google_adk.py/langgraph.py findings on this PR for exactly that failure mode) would pass every existing test for that adapter.

src/band/integrations/crewai/tools.py
```diff
@@ -281,662 +114,123 @@ async def _execute() -> str:
...
-                return serialize_success_result(result)
-
-            return _exec("band_list_contacts", execute)
-
-    class AddContactTool(BaseTool):
-        name: str = "band_add_contact"
-        description: str = get_tool_description("band_add_contact")
-        args_schema: type[BaseModel] = platform_args_schema("band_add_contact")
+    class PlatformTool(BaseTool):
+        name: str = spec.name
```
amit-gazal-band: **A `BandTool` (StrEnum) instance is assigned to a field pydantic types as `str`.**

CrewAI's `BaseTool` has no `validate_default`, so `tool.name` remains a real `BandTool` member at runtime, not a plain `str` (confirmed live: `type(tool.name) is BandTool`). Currently benign for `==`, hashing, `json.dumps`, and f-string formatting (StrEnum subclasses `str`), but `repr(tool.name)` differs from a plain string's repr, and any future `isinstance(x, Enum)` special-casing, `type(x) is str` check, or pickling of the tool would see different behavior than before this PR. Worth an explicit `str(spec.name)` here if `str` is the intended type.

src/band/integrations/langgraph/langchain_tools.py
```diff
@@ -106,13 +119,20 @@ def agent_tools_to_langchain(
...
+        # pass-through branch below cannot honestly be narrowed here.
         async def execute_definition(
             *,
             _tool_name: str = definition.name,
             **kwargs: Any,
         ) -> Any:
             try:
-                return await tools.execute_tool_call(_tool_name, kwargs)
+                result = await tools.execute_tool_call(_tool_name, kwargs)
+                if _tool_name == BandTool.READ_ROOM_FILE and is_mcp_content_result(
```
amit-gazal-band: **The image-passthrough guard condition is retyped at 7+ call sites instead of one shared predicate.**

`_tool_name == BandTool.READ_ROOM_FILE and is_mcp_content_result(result)` appears verbatim here and in `anthropic.py`, `gemini.py`, `agno.py`, `copilot_sdk.py`, `claude_sdk/tools.py`, `codex.py`, and `strands.py`. A future change to what counts as an image-passthrough result (e.g. gating on a second tool, or tightening `is_mcp_content_result`'s shape check) requires updating the same condition independently in all 7+ files; a missed site silently falls back to json-dumping the image blob as text in that one adapter only. Worth a shared `is_image_passthrough_result(tool_name, result)` helper in `runtime/tools.py`.

tests/e2e/baseline/smoke/matrix/test_capability_matrix.py
```diff
@@ -197,6 +219,137 @@ async def test_list_contacts_across_contacts_adapters(
...
+# tests/framework_conformance/test_adapter_conformance.py (that's the unit-level
+# source of truth; this list exists only because @per_adapter selects from the
+# separate tests.baseline.adapter.Adapter enum, a different registry). Differs
+# from that set by: -crewai_flow (its E2E builder is a hardcoded echo flow with
+# no Band tool loop, so there's no tool call for this to drive), -parlant (not
+# in Adapter at all -- confirmed unsupportable, no matrix cell), +copilot_acp
+# (wraps ACPClientAdapter, which shares the same already-fixed MCP engine as
+# opencode but isn't tracked in the unit-level set since it's not its own
+# ADAPTER_CONFIGS entry there).
+IMAGE_PASSTHROUGH_ADAPTERS = (
```
amit-gazal-band, line 271: **Hand-retyped vocabulary instead of a single source of truth.**

`IMAGE_PASSTHROUGH_ADAPTERS` hand-retypes the same "which adapters support image passthrough" list that `IMAGE_PASSTHROUGH_SUPPORTED_FRAMEWORK_IDS` already derives programmatically in `tests/framework_conformance/test_adapter_conformance.py`, with no runtime assertion tying the two together -- this PR otherwise follows the single-source-of-truth rule for every sibling list it introduces. When a future adapter gains image-passthrough support and the conformance-test list is updated, nothing forces a matching update here -- the E2E vision-passthrough smoke silently never runs for that adapter, and CI stays green while coverage quietly regresses.

amit-gazal-band (changes requested):
## Automated review: `/code-review 600`

All 6 of my original inline comments on this PR are addressed by 2899e606:

- `pydantic_ai.py` raw image bytes via `BinaryContent.__repr__` -- fixed
- `crewai/catalog.py` unbounded file content in `tool_call` -- fixed
- `parlant/tools.py` `None.participants` crash -- fixed
- `strands.py` N placeholder lines instead of 1 -- fixed
- `letta/prompts.py` hardcoded `"band_send_event"` literal -- fixed
- `parlant/tools.py` `features=None` capability leak -- reconfirmed no fix needed, `self.features` is always set

Requesting changes on the broader pass below: the CrewAI image/file-content leak fix wasn't generalized to `google_adk.py`/`langgraph.py` (full leak) and the other adapters' generic `tool_call` reporting (bounded but still real leak), plus three adapters (`strands.py`, `copilot_sdk.py`, `agno.py`) can now crash a turn on a malformed image content block since the decode step sits outside their error boundaries. Everything else below is lower-severity (DRY/single-source-of-truth misses, one dropped logging granularity, one test-coverage gap).

src/band/adapters/google_adk.py
AlexanderZ-Band: Confirmed and fixed (793119a3): added `_redacted_function_response_output()` in google_adk.py -- unwraps ADK's `{"result": <json str>}` envelope for band_read_room_file and reports `image_block_placeholder(...)` instead of `str(fr.response)` when it's an image result. Also applied the shared `redact_tool_call_args()` fix (see the anthropic.py thread) to this adapter's function-call ARGS reporting. Regression test: tests/adapters/test_google_adk_adapter.py (asserts the reported tool_result event uses the placeholder, not raw base64).

src/band/adapters/langgraph.py
AlexanderZ-Band: Confirmed and fixed (793119a3): added `_redacted_tool_end_output()` -- LangChain's ToolMessage.content for the image branch is a list[ImageContentBlock] (mime_type/base64 keys, a different shape than the platform's own MCP blocks), so on_tool_end/on_tool_error now checks for that shape and reports `image_block_placeholder(...)` instead of falling through to `json.dumps(..., default=str)` -> `str(ToolMessage(...))`. Also found and fixed an unredacted band_send_room_file ARGS leak at on_tool_start (not in the original findings list) using the same shared `redact_tool_call_args()` helper. Regression tests: tests/adapters/langgraph/test_stream_events.py.

src/band/adapters/anthropic.py
AlexanderZ-Band: Confirmed, and it turned out to be systemic rather than anthropic-specific: the same unguarded `ToolEventKey.ARGS: <raw args>` pattern for band_send_room_file was present in ~10 adapters (anthropic, pydantic_ai, strands, google_adk, copilot_sdk, gemini, agno, codex x2, letta, langgraph). Fixed once at the root (793119a3): new `redact_tool_call_args(tool_name, args)` in runtime/tools.py, applied at every one of those call sites -- mirrors the bounded-placeholder idiom CrewAI's `_send_room_file` already used. Regression test per adapter (e.g. tests/adapters/test_anthropic_adapter.py) asserts the reported tool_call ARGS is `file_content_placeholder(...)`, not raw content.

src/band/integrations/crewai/catalog.py
AlexanderZ-Band: Confirmed and fixed (793119a3): `ToolSpec.invoke()` now calls `serialize_tool_result(result)` once, right after the body returns, and reuses that serialized value for both `report_result` and `render()` -- previously it reported the raw pre-render `result` first, so the 7 tools that return a raw Pydantic/Fern model directly (`_lookup_peers`, `_list_contacts`, etc.) hit an uncaught TypeError inside `report_result`'s own try/except, silently dropping the tool_result event. Regression test: tests/integrations/test_crewai_tools.py::test_lookup_peers_reports_serialized_result_for_raw_model_return (fails on old code with exactly this warning, passes now).

src/band/adapters/strands.py
AlexanderZ-Band: Confirmed and fixed (793119a3): `_tool_result()`'s `decode_image_block` call is now wrapped in try/except, matching the guard anthropic.py/gemini.py already have around the equivalent step -- a malformed image block now returns a clean error result instead of raising uncaught out of the async generator. Regression test: tests/adapters/test_strands_adapter.py::TestReadRoomFileImagePassthrough::test_malformed_image_data_returns_error_result.

src/band/adapters/copilot_sdk.py
AlexanderZ-Band: Confirmed and fixed (793119a3): `ToolBinaryResult` construction is now wrapped in its own try/except routing through the same `_fail_tool_call` path every other failure in this method already uses, instead of sitting outside the existing error boundary. Worth noting: combined with the `is_mcp_content_result` tightening done for the runtime/tools.py:1828 thread (now requires data/mimeType keys), this specific KeyError isn't independently reachable anymore -- this fix is defense-in-depth matching the established idiom, not a provably-reachable bug on its own.

src/band/adapters/agno.py
AlexanderZ-Band: Confirmed and fixed (793119a3): `_entrypoint()`'s `decode_image_block` call (previously bare in a comprehension with zero try/except anywhere in the function) now mirrors pydantic_ai.py's exact idiom -- `try: ... except Exception as e: return f"Error reading room file: {e}"`. Regression test: tests/adapters/agno/test_adapter.py::TestBandEntrypointBinding::test_read_room_file_malformed_image_data_returns_error_string.

src/band/integrations/parlant/tools.py
AlexanderZ-Band: Confirmed and fixed (793119a3): `guard_failures()`'s `inspect.signature(func).bind(...)` now has its own try/except (TypeError -> a clean ToolResult error) separate from the call-handling try below it, so a signature/argument-shape mismatch no longer escapes uncaught before there's even a call.arguments to build the usual failure message from. Regression test: tests/integrations/parlant/test_tools.py::test_tool_returns_error_on_malformed_call_arguments.

src/band/integrations/parlant/tools.py
AlexanderZ-Band: Confirmed and fixed (793119a3): `guard_failures()` now logs one generic result line ("%s %s -> %s") after a successful call, restoring per-tool completion signal for all ~15 migrated tools with one line rather than resurrecting the 8 bespoke per-branch messages the refactor dropped. Regression test: tests/integrations/parlant/test_tools.py::test_tool_logs_result_on_success.

src/band/integrations/claude_sdk/tools.py
AlexanderZ-Band: Confirmed and fixed (793119a3): the six sibling branches in `_format_success_payload` (band_send_message, band_send_event, band_add_participant, band_remove_participant, band_get_participants, band_create_chatroom) now compare against their BandTool members instead of raw string literals, matching the one branch (READ_ROOM_FILE) that was already migrated. Behavior-identical (StrEnum members equal their string value), so no new test is meaningful here -- the existing 169 claude_sdk tests all still pass, confirming no regression.

src/band/runtime/tools.py
AlexanderZ-Band: Confirmed as a real (if not currently reachable) gap and fixed (793119a3): `is_mcp_content_result` now also requires `block.get("type") == "image"` and data/mimeType keys, not just `"type" in block` -- matches read_room_file's actual contract exactly, and closes the shape gap that let a malformed/future-extended block reach `decode_image_block` unguarded. Paired with real try/except additions at the three outlier call sites this predicate alone can't fully cover (strands.py, copilot_sdk.py, agno.py -- see their own threads), since a present-but-garbage base64 string still needs an actual decode attempt to catch.

src/band/integrations/crewai/catalog.py
AlexanderZ-Band: Investigated, not fixing: `CrewAIToolReporter`'s "must not raise" contract does hold for both shipped reporters (EmitToolCallsReporter, NoopReporter), and I traced the actual failure path -- `_execute_tool`'s inner `_execute()` wraps the whole tool body, including `report_call`, in try/except, so a raising `report_call` produces a clean `{"status":"error",...}` JSON, not an uncaught escape. The upload does get skipped (that part's right), but a genuinely uncaught exception needs a second independent failure: the except-branch's own `report_result(..., is_error=True)` also raising on the same already-contract-violating reporter. No shipped reporter can trigger that double-fault, so this is a hypothetical with no reachable trigger under the current contract -- leaving as-is.

tests/adapters/test_anthropic_adapter.py
AlexanderZ-Band: Confirmed and fixed (793119a3): added event-content assertions to the image-passthrough tests in test_anthropic_adapter.py, test_gemini_adapter.py, tests/adapters/agno/test_adapter.py, test_codex_adapter.py, and tests/adapters/copilot_sdk/test_tool_bridging.py -- each now asserts the reported tool_result event is the bounded placeholder, not raw base64, closing the gap where a future regression in any of these five adapters' redaction branch would have passed every existing test.

src/band/integrations/crewai/tools.py
AlexanderZ-Band: Confirmed and fixed (793119a3): `PlatformTool.name` is now `str(spec.name)` instead of the bare BandTool (StrEnum) default -- CrewAI's BaseTool has no validate_default, so pydantic never coerced it, confirmed live (`type(tool.name) is BandTool`, not str, before this fix). Regression test: tests/framework_conformance/test_files_image_passthrough_matrix.py::test_crewai_platform_tool_name_is_plain_str (run under the real crewai venv, fails on old code with exactly this type mismatch).

src/band/integrations/langgraph/langchain_tools.py
AlexanderZ-Band: Confirmed, same systemic issue as the anthropic.py thread: this exact compound predicate was duplicated verbatim at 9 call sites (langchain_tools.py, claude_sdk/tools.py x2, gemini.py, anthropic.py, copilot_sdk.py, agno.py, strands.py, codex.py), not the 7+ originally estimated. Fixed once at the root (793119a3): new `is_image_passthrough_result(tool_name, result)` in runtime/tools.py, single-sourced and swapped in at every site (preserving each site's own extra prefix conditions like `ok and`/`success and` where present). Behavior-identical refactor, so existing image-passthrough tests (72 across the langgraph integration suites) are the regression coverage -- all still pass.

tests/e2e/baseline/smoke/matrix/test_capability_matrix.py
AlexanderZ-Band: Confirmed and fixed (793119a3): added test_e2e_matrix_list_matches_this_unit_level_set in tests/framework_conformance/test_adapter_conformance.py (not in this file itself -- everything under tests/e2e/baseline/ is gated on E2E_TESTS_ENABLED, which would've skipped a pure offline list-equality check right along with the live ones). It asserts IMAGE_PASSTHROUGH_ADAPTERS == (IMAGE_PASSTHROUGH_SUPPORTED_FRAMEWORK_IDS - {crewai_flow}) | {copilot_acp, letta}. Also caught and fixed that the comment above IMAGE_PASSTHROUGH_ADAPTERS was itself incomplete (documented +copilot_acp but not +letta, which is excluded from the unit-level set for the same MCP-engine-inheritance reason).

amit-gazal-band (approved):
All findings from the previous CHANGES_REQUESTED review are resolved in 793119a3:

- Image/file-content leaks into platform tool_call/tool_result events: generalized via new shared `redact_tool_call_args`/`is_image_passthrough_result` helpers in `runtime/tools.py`, applied consistently across anthropic, gemini, codex, google_adk, langgraph, strands, copilot_sdk, agno, letta, and pydantic_ai.
- Uncaught exceptions on malformed image content blocks in `strands.py`, `copilot_sdk.py`, `agno.py`: all now degrade to the adapter's normal error result instead of crashing the turn, and `is_mcp_content_result` itself was tightened to require `type == "image"` + `data`/`mimeType`.
- `parlant/tools.py`'s `inspect.signature(...).bind()` is now inside its own try/except.
- `parlant/tools.py` logging: a generic post-call result log restores per-tool debuggability.
- `claude_sdk/tools.py`'s `_format_success_payload`: all seven branches now use the `BandTool` enum, not string literals.
- `crewai/catalog.py`'s `ToolSpec.invoke`: results are serialized via `serialize_tool_result` before reaching `report_result`, so raw Pydantic/Fern models no longer silently drop the tool_result event.
- `crewai/tools.py`'s `PlatformTool.name`: explicit `str(spec.name)` so it's a real `str` at runtime.
- The image-passthrough guard is now a single shared predicate instead of retyped at 7+ call sites.
- `IMAGE_PASSTHROUGH_ADAPTERS` in the E2E matrix is now cross-checked against the unit-level set by a real assertion (`test_e2e_matrix_list_matches_this_unit_level_set`).
- New tests cover the redaction/error paths in all five previously-uncovered adapters (anthropic, gemini, agno, copilot_sdk, codex).

One minor item from the earlier review (`crewai/catalog.py`'s `_send_room_file` reporter-ordering / "must not raise" contract not being structurally enforced) wasn't addressed, but it's speculative rather than an active bug and doesn't block this. CI is green across all jobs. Approving.

PR #603, opened 2026-09-01
refactor: integrate band-sdk-core Session into BandLink

AlexanderZ-Band:
## Summary

Implements INT-1303: replaces band-sdk-python's hand-rolled reconnect/backoff/disconnect-classification logic in `WebSocketClient`/`BandLink` with band_sdk_core's `Session` state machine, per the ticket's attached implementation plan. All 5 migration steps shipped, squashed into 5 commits matching the plan's own step boundaries:

1. Extract the duplicated `_fast_session_policy` test helper into `tests/websocket/conftest.py` (prep, test-only)
2. Wire `Session` into `WebSocketClient.__init__`/`__aexit__` (no behavior change)
3. Replace `__aenter__`'s retry loop with `Session`-driven backoff
4. Route `agent_control:supersede` through `Session.on_supersede`
5. Extend the CI wheel-smoke step with a `Session`/`SessionPolicy` call

## Design decisions (from the plan) — all 4 shipped as specified

1. 429/503 upgrade rejections on the initial-connect path are now retried automatically via `Session.on_upgrade_rejected` (using `Retry-After` as a delay floor), instead of raised immediately as before.
2. `Session.on_supersede`'s `Reconnecting` branch is implemented in `BandLink._on_supersede`, even though the platform hardcodes `retryable=False` on every real supersede today (only a synthetic regression test exercises it) — keeps the SDK correct about *not treating it as terminal* if that ever changes. It does not yet drive actual reconnect timing off that decision — see below.
3. `WebSocketDisconnectReason` gained `dead_reason`/`stale_reason` fields.
4. Those fields are populated from both the supersede path and `__aenter__`'s terminal-failure path (a new `_disconnect_reason_from_exception`, since there's no platform wire payload for that case) — widening `BandLink.last_disconnect_reason`'s trigger scenarios to include a permanently-failed initial connect, not just a supersede.

**Out of scope** (per the ticket, untouched): taking ownership of the vendored `PHXChannelsClient`'s internal `auto_reconnect` loop — separate, larger follow-up work, tracked as [INT-1355](https://linear.app/thenvoi/issue/INT-1355/take-over-phxchannelsclients-auto-reconnect-loop-so-session-drives) (filed during review: a retryable supersede's `retry_after_s` is classified but not enforced without that ownership change; not reachable today since the platform always sends `retryable=False`).

Other real, intentional behavior change from Step 3: the initial-connect give-up bound is now a rolling 5-minute rapid-disconnect window (10 by default) instead of a fixed 11-attempt cap.

INT-1303: https://linear.app/thenvoi/issue/INT-1303/integrate-band-sdk-core-session-connection-reliability-into-band-sdk

## Review history

Two rounds of external review, both verified and closed out:
- `/code-review high`: confirmed a real exception-chaining regression (`raise raise_exc` had dropped `from exc` for the classified-upgrade-error case) — fixed, with a regression assertion. A comment-style finding (narrating removed mechanisms) — fixed.
- `/my-here-is-a-review` (2 rounds) on the retryable-supersede `retry_after_s` gap: confirmed real but architecturally out of scope (needs the deferred `auto_reconnect`-ownership work) — filed as INT-1355, PR thread resolved.

## Test plan

- [x] `uv run ruff check .` / `uv run ruff format .` / `uv run pyrefly check` / full unit suite (`uv run pytest tests/ --ignore=tests/integration/ --ignore=tests/e2e/`) — clean
- [x] Every ticket-listed acceptance regression test passes explicitly (`test_aenter_restores_reconnect_after_successful_initial_connect`, `test_uses_retry_after_header_for_429_upgrade_error`, `test_supersede_event_records_terminal_reason_and_disables_reconnect`, `test_watchdog_forces_close_and_reconnect_when_ack_withheld`, `test_cancelled_connect_closes_the_half_opened_client_and_allows_retry`, `test_disconnect_after_supersede_still_cleans_up_websocket`, `test_close_without_supersede_leaves_disconnect_reason_empty`, all of `tests/websocket/test_watchdog.py`)
- [x] New tests added per the plan's test plan: rapid-disconnect-ladder trip to `Dead`, 429 upgrade rejection retried via the probe fallback then succeeding, retryable supersede leaves the connection non-terminal, `connect()` propagates a terminal initial-connect failure onto `last_disconnect_reason`, exception-chaining regression guards
- [x] All CI checks green (lint, packaging, full test matrix, CodeQL)

https://claude.ai/code/session_01FuwaxzwGje1eq94W7V8zrW

### DIFF
--- a/src/band/client/streaming/client.py
+++ b/src/band/client/streaming/client.py
@@ -24,7 +24,7 @@ from phoenix_channels_python_client.exceptions import PHXConnectionError
 from phoenix_channels_python_client.phx_messages import PHXMessage
 from band.client.streaming.errors import (
     WebSocketUpgradeError,
-    classify_initial_upgrade_error,
+    probe_upgrade_error,
 )
 from band.client.streaming.watchdog import HeartbeatWatchdog
 from band.client.streaming.wire import WirePayload
@@ -387,6 +387,8 @@ class WebSocketClient:
         self._last_disconnect_reason: WebSocketDisconnectReason | None = None
         self._watchdog = HeartbeatWatchdog(session_policy or SessionPolicy.default())
         self._session = Session(self._watchdog.policy)
+        self._connect_failure_probed: bool = False
+        self._cached_connect_failure: WebSocketUpgradeError | None = None
 
     @property
     def validation_error_count(self) -> int:
@@ -455,20 +457,39 @@ class WebSocketClient:
             client.channel_socket_url += f"&agent_id={self.agent_id}"
         return client
 
+    async def _classify_connect_failure(
+        self, exc: Exception
+    ) -> WebSocketUpgradeError | None:
+        """Classify one failed connect exception, reusing the previous
+        live-socket probe result for a repeat unclassifiable
+        PHXConnectionError.
+
+        probe_upgrade_error blocks for up to open_timeout=5s; without this
+        cache it reran on every backoff retry of the very same rejection,
+        stacking its own network latency on top of every computed delay.
+        """
+        upgrade_error = WebSocketUpgradeError.from_exception(exc)
+        if upgrade_error is not None or not isinstance(exc, PHXConnectionError):
+            return upgrade_error
+        if not self._connect_failure_probed:
+            self._cached_connect_failure = await probe_upgrade_error(
+                self._require_client().channel_socket_url
+            )
+            self._connect_failure_probed = True
+        return self._cached_connect_failure
+
     async def _resolve_failed_connect_attempt(
         self, exc: Exception, epoch: int
     ) -> float:
         """Classify one failed initial-connect attempt through Session and
         resolve it to a retry delay, or raise if Session now considers the
         session Dead."""
-        # Captured before the live-socket probe below (up to open_timeout=5s
-        # inside classify_initial_upgrade_error) -- charging the probe's own
-        # latency to Session's rapid-disconnect timing would understate how
-        # fast repeated failures are actually happening.
+        # Captured before the live-socket probe inside
+        # _classify_connect_failure (up to open_timeout=5s) -- charging the
+        # probe's own latency to Session's rapid-disconnect timing would
+        # understate how fast repeated failures are actually happening.
         now = asyncio.get_running_loop().time()
-        upgrade_error = await classify_initial_upgrade_error(
-            exc, self._require_client().channel_socket_url
-        )
+        upgrade_error = await self._classify_connect_failure(exc)
         match upgrade_error:
             case WebSocketUpgradeError():
                 outcome = self._session.on_upgrade_rejected(
@@ -555,6 +576,8 @@ class WebSocketClient:
         start a new connection lifecycle, not resume a now-Dead one.
         """
         self._session = Session(self._watchdog.policy)
+        self._connect_failure_probed = False
+        self._cached_connect_failure = None
         while True:
             now = asyncio.get_running_loop().time()
             epoch = self._session.begin_attempt(now)
--- a/src/band/client/streaming/client.py
+++ b/src/band/client/streaming/client.py
@@ -387,7 +387,7 @@ class WebSocketClient:
         self._last_disconnect_reason: WebSocketDisconnectReason | None = None
         self._watchdog = HeartbeatWatchdog(session_policy or SessionPolicy.default())
         self._session = Session(self._watchdog.policy)
-        self._connect_failure_probed: bool = False
+        self._probed_failure_message: str | None = None
         self._cached_connect_failure: WebSocketUpgradeError | None = None
 
     @property
@@ -460,22 +460,19 @@ class WebSocketClient:
     async def _classify_connect_failure(
         self, exc: Exception
     ) -> WebSocketUpgradeError | None:
-        """Classify one failed connect exception, reusing the previous
-        live-socket probe result for a repeat unclassifiable
-        PHXConnectionError.
-
-        probe_upgrade_error blocks for up to open_timeout=5s; without this
-        cache it reran on every backoff retry of the very same rejection,
-        stacking its own network latency on top of every computed delay.
-        """
+        """Classify one failed connect exception. Reuses the previous
+        live-socket probe result only while the wrapped PHXConnectionError's
+        message is unchanged -- a different message means the underlying
+        failure may have changed, so it must be probed again."""
         upgrade_error = WebSocketUpgradeError.from_exception(exc)
         if upgrade_error is not None or not isinstance(exc, PHXConnectionError):
             return upgrade_error
-        if not self._connect_failure_probed:
+        message = str(exc)
+        if message != self._probed_failure_message:
             self._cached_connect_failure = await probe_upgrade_error(
                 self._require_client().channel_socket_url
             )
-            self._connect_failure_probed = True
+            self._probed_failure_message = message
         return self._cached_connect_failure
 
     async def _resolve_failed_connect_attempt(
@@ -576,7 +573,7 @@ class WebSocketClient:
         start a new connection lifecycle, not resume a now-Dead one.
         """
         self._session = Session(self._watchdog.policy)
-        self._connect_failure_probed = False
+        self._probed_failure_message = None
         self._cached_connect_failure = None
         while True:
             now = asyncio.get_running_loop().time()
--- a/src/band/client/streaming/client.py
+++ b/src/band/client/streaming/client.py
@@ -461,10 +461,14 @@ class WebSocketClient:
         """Classify one failed initial-connect attempt through Session and
         resolve it to a retry delay, or raise if Session now considers the
         session Dead."""
+        # Captured before the live-socket probe below (up to open_timeout=5s
+        # inside classify_initial_upgrade_error) -- charging the probe's own
+        # latency to Session's rapid-disconnect timing would understate how
+        # fast repeated failures are actually happening.
+        now = asyncio.get_running_loop().time()
         upgrade_error = await classify_initial_upgrade_error(
             exc, self._require_client().channel_socket_url
         )
-        now = asyncio.get_running_loop().time()
         match upgrade_error:
             case WebSocketUpgradeError():
                 outcome = self._session.on_upgrade_rejected(
@@ -501,8 +505,12 @@ class WebSocketClient:
             raise raise_exc from exc
 
         # None only when Dead (handled above) or stale (unreachable here) --
-        # narrows the type for the caller's asyncio.sleep.
-        assert outcome.retry_after_s is not None
+        # fail loudly rather than trust a Session contract that assert would
+        # silently drop under python -O.
+        if outcome.retry_after_s is None:
+            raise RuntimeError(
+                "Session returned a non-Dead outcome with no retry_after_s"
+            )
         logger.warning(
             "Initial WebSocket connection failed; retrying in %.2fs: %s",
             outcome.retry_after_s,
@@ -553,7 +561,16 @@ class WebSocketClient:
             if epoch is None:
                 # Unreachable given this loop's own control flow, but
                 # Session's signature allows it -- fail loudly rather than
-                # silently proceed with a None epoch.
+                # silently proceed with a None epoch, and still populate
+                # last_disconnect_reason like every other Dead path in this
+                # class so BandLink.connect() sees a terminal reason too.
+                self.record_terminal_disconnect(
+                    WebSocketDisconnectReason(
+                        reason="session_ended",
+                        message="WebSocket session is no longer connectable",
+                        retryable=False,
+                    )
+                )
                 raise RuntimeError("WebSocket session is no longer connectable")
 
             self.client = self._build_phx_client()
--- a/src/band/client/streaming/client.py
+++ b/src/band/client/streaming/client.py
@@ -8,15 +8,24 @@ import logging
 import random
 from typing import Any, Literal
 
-from band_sdk_core import Session, SessionPolicy
+from band_sdk_core import (
+    DeadReason,
+    Session,
+    SessionOutcome,
+    SessionPolicy,
+    SessionState,
+    StaleReason,
+)
 from phoenix_channels_python_client.client import (
     PHXChannelsClient,
     PhoenixChannelsProtocolVersion,
 )
-from phoenix_channels_python_client.client_types import ReconnectPolicy
 from phoenix_channels_python_client.exceptions import PHXConnectionError
 from phoenix_channels_python_client.phx_messages import PHXMessage
-from band.client.streaming.errors import classify_initial_upgrade_error
+from band.client.streaming.errors import (
+    WebSocketUpgradeError,
+    classify_initial_upgrade_error,
+)
 from band.client.streaming.watchdog import HeartbeatWatchdog
 from band.client.streaming.wire import WirePayload
 from band.logging_config import core_issues, trace_context_extra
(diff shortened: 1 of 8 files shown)


src/band/platform/link.py
```diff
@@ -719,12 +726,37 @@ async def _drain_agent_topic_reconciliation(self, ws: WebSocketClient) -> None:
...
+        actually current (not a stale notification about an epoch this
+        connection has already moved past) and, in principle, whether a
+        retryable supersede should keep the connection reconnecting --
+        though the platform hardcodes retryable=False today, so in practice
+        this still always resolves to Dead.
+        """
+        if self._ws is not None:
+            outcome = self._ws.handle_supersede(payload.retryable, payload.retry_after)
+            reason = payload.to_disconnect_reason(outcome)
+            if outcome.state is not SessionState.Dead:
```
AlexanderZ-Band, line 740: **Resolved** — confirmed real: `Session.on_supersede`'s computed `retry_after_s` for a retryable supersede is classified but never applied here; if the vendored `PHXChannelsClient` reconnects afterward it uses its own `ReconnectPolicy`, blind to this value. Not reachable today (the platform hardcodes `retryable=False` on every real supersede), and the actual fix requires taking ownership of `PHXChannelsClient`'s `auto_reconnect` loop — the architecture change INT-1303 explicitly scoped out (see "Restated out-of-scope boundary" in the ticket). Filed as a tracked followup: [INT-1355](https://linear.app/thenvoi/issue/INT-1355/take-over-phxchannelsclients-auto-reconnect-loop-so-session-drives).

src/band/client/streaming/client.py
```diff
@@ -395,58 +430,147 @@ async def _handle_reconnect(self) -> None:
...
+    async def _resolve_failed_connect_attempt(
+        self, exc: Exception, epoch: int
+    ) -> float:
+        """Classify one failed initial-connect attempt through Session and
+        resolve it to a retry delay, or raise if Session now considers the
+        session Dead."""
+        upgrade_error = await classify_initial_upgrade_error(
+            exc, self._require_client().channel_socket_url
+        )
+        now = asyncio.get_running_loop().time()
```
amit-gazal-band, line 495: **Blocking:** `now` is captured *after* `classify_initial_upgrade_error`'s live-socket probe (up to `open_timeout=5s`), not at the moment the connect attempt actually failed.

A burst of true back-to-back connect failures each burn up to ~5s inside the probe before this timestamp is taken, stretching the recorded gap between failures relative to when they really happened. That understates how rapid the disconnects are, which can delay or prevent `SessionPolicy`'s rolling rapid-disconnect window from tripping Dead as intended — undermining the "fail fast on rapid disconnects" guarantee this migration exists to deliver. Pre-PR, the retry loop never consumed wall-clock time for its backoff decision, so this timing/probe interaction is a fresh risk from combining the pre-existing probe with the new time-sensitive `Session` classification.

Suggest capturing `now` before calling `classify_initial_upgrade_error`, so the probe's own latency doesn't get charged against the Session's rapid-disconnect timing.

src/band/client/streaming/client.py
```diff
@@ -395,58 +430,147 @@ async def _handle_reconnect(self) -> None:
...
+            client.channel_socket_url += f"&agent_id={self.agent_id}"
+        return client
+
+    async def _resolve_failed_connect_attempt(
+        self, exc: Exception, epoch: int
+    ) -> float:
+        """Classify one failed initial-connect attempt through Session and
+        resolve it to a retry delay, or raise if Session now considers the
+        session Dead."""
+        upgrade_error = await classify_initial_upgrade_error(
```
amit-gazal-band: **Blocking:** this live-socket probe (up to 5s) now reruns on every iteration of the retry loop while backing off from a classified 429/503, instead of running once before giving up.

Before this PR, a classified upgrade rejection was raised immediately after one probe. Now the same rejection is retried automatically (an intentional design change), so the same blocking probe reruns on every subsequent retry attempt for as long as the rolling rapid-disconnect window allows — stacking real network latency and extra connection attempts on top of every computed backoff delay. That cost isn't mentioned in the PR description.

Worth caching the classification (or skipping the live probe) across retries of the same failure, so backoff timing isn't inflated by repeated network round-trips.

src/band/client/streaming/client.py
```diff
@@ -395,58 +430,147 @@ async def _handle_reconnect(self) -> None:
...
+                # Re-raising the exception already being handled -- no new
+                # object, so no chaining to add (bare raise).
+                raise raise_exc
+            # A newly-built WebSocketUpgradeError -- chain it to the
+            # original exception, same as the pre-Session code did.
+            raise raise_exc from exc
+
+        # None only when Dead (handled above) or stale (unreachable here) --
+        # narrows the type for the caller's asyncio.sleep.
+        assert outcome.retry_after_s is not None
```
amit-gazal-band: `assert outcome.retry_after_s is not None` is the sole guard before `await asyncio.sleep(outcome.retry_after_s)` a few lines below, while the structurally identical "can't happen" check for `epoch is None` in `__aenter__` uses a real `if ... raise RuntimeError(...)` instead.

Assertions are stripped under `python -O` / `PYTHONOPTIMIZE=1`. If a non-Dead `SessionOutcome` ever carries `retry_after_s=None`, an optimized run turns this into `asyncio.sleep(None)` → unhandled `TypeError`, instead of the fail-loud `RuntimeError` the sibling check uses for an equally "unreachable" condition in the same function.

Suggest swapping this for `if outcome.retry_after_s is None: raise RuntimeError(...)` to match the sibling pattern and stay safe under `-O`.

src/band/client/streaming/client.py
```diff
@@ -395,58 +430,147 @@ async def _handle_reconnect(self) -> None:
...
+        """
+        self._session = Session(self._watchdog.policy)
+        while True:
+            now = asyncio.get_running_loop().time()
+            epoch = self._session.begin_attempt(now)
+            if epoch is None:
+                # Unreachable given this loop's own control flow, but
+                # Session's signature allows it -- fail loudly rather than
+                # silently proceed with a None epoch.
+                raise RuntimeError("WebSocket session is no longer connectable")
```
amit-gazal-band, line 602: When `Session.begin_attempt()` returns `None` (its own type signature allows this), `__aenter__` raises `RuntimeError` here without calling `self.record_terminal_disconnect(...)`, unlike every other Dead-classification path in this class (see `_resolve_failed_connect_attempt` above, which always calls it before raising).

If this path is ever hit, `BandLink.last_disconnect_reason` stays whatever it was before — the `if ws.last_disconnect_reason is not None:` check in `link.py`'s `connect()` won't see anything new — silently breaking the terminal-reason contract this refactor was built around. Suggest calling `record_terminal_disconnect(...)` with an appropriate reason before raising, for consistency with the other Dead paths in this class.

amit-gazal-band (changes requested):
Requesting changes on two correctness issues that undermine this PR's core "fail fast on rapid disconnects" guarantee, plus two smaller robustness gaps worth closing in the same pass. Everything else from the fuller review (efficiency/maintainability/style/test-coverage) is lower stakes and fine as follow-up.

src/band/client/streaming/client.py
AlexanderZ-Band: Confirmed and fixed in e839a97a.

`now` is now captured as the very first statement of `_resolve_failed_connect_attempt`, before the `await classify_initial_upgrade_error(...)` call that can block up to `open_timeout=5s`. It reflects the moment the connect attempt actually failed, not when classification finished, so the probe's own latency no longer gets charged against `Session`'s rapid-disconnect window.

Added `test_resolve_failed_connect_attempt_captures_now_before_probe_latency` (tests/websocket/test_client.py), which injects a slow `classify_initial_upgrade_error` and asserts the `now_s` Session actually receives predates the probe's delay -- fails on the pre-fix ordering, passes now.

src/band/client/streaming/client.py
AlexanderZ-Band: Confirmed real, and worse than a corner case: I traced `phoenix_channels_python_client/supervisor.py`'s connect loop -- with `auto_reconnect=False` (always true during this initial-connect loop, per `_build_phx_client`), *every* failed `connect()` gets wrapped into a bare `PHXConnectionError(f"Failed to connect to {url}: {exc}")`, discarding the original exception's `.response`/`.status_code`. So `WebSocketUpgradeError.from_exception(exc)`'s cheap path can never succeed here -- the live probe is not a rare fallback, it reruns on literally every retry for as long as the backend keeps rejecting.

I pushed back on landing a fix for this in this PR, though. The reviewer's own suggested fix -- caching the classification (or skipping the probe) across retries -- is a real design-contract change, not a local tweak: it trades the probe's current job (re-verifying the *true current* rejection reason/`Retry-After` on every attempt) for an assumption that a masked failure is still the same failure as last time, for the rest of the session's retry lifetime. That has its own correctness risk (a masked failure that's actually a *new, different* problem gets silently misclassified as the stale one) and needs a deliberate invalidation policy, not a bolt-on cache field. I don't think that belongs in a review-fix pass on top of an already-large migration.

The line-467 fix (this PR, e839a97a) does mean each individual probe's latency no longer corrupts the recorded rapid-disconnect gap, which removes part of the harm -- but it does not reduce how often the probe itself runs, which is what this comment is actually about.

I'd suggest tracking the caching/skip-probe design as its own follow-up (same pattern as INT-1355, already spun out of this PR for the retryable-supersede gap) rather than forcing it in here -- leaving this open for a call on that.

src/band/client/streaming/client.py
AlexanderZ-Band: Confirmed and fixed in e839a97a. Replaced `assert outcome.retry_after_s is not None` with `if outcome.retry_after_s is None: raise RuntimeError(...)`, matching the sibling `epoch is None` check.

Added `test_resolve_failed_connect_attempt_raises_when_retry_after_s_missing`, which fakes a non-Dead `SessionOutcome` with `retry_after_s=None` and asserts a `RuntimeError` (not an `AssertionError`, and not a bare `asyncio.sleep(None)` `TypeError`) is what surfaces.

src/band/client/streaming/client.py
AlexanderZ-Band: Confirmed and fixed in e839a97a. This `epoch is None` branch now calls `self.record_terminal_disconnect(WebSocketDisconnectReason(reason="session_ended", message="WebSocket session is no longer connectable", retryable=False))` before raising, matching `_resolve_failed_connect_attempt`'s Dead paths.

Added `test_aenter_records_terminal_disconnect_when_session_returns_no_epoch`, which fakes a `Session.begin_attempt()` returning `None` and asserts `last_disconnect_reason` is populated (not `None`) before the `RuntimeError` propagates -- fails on the pre-fix code, passes now.

src/band/client/streaming/client.py
AlexanderZ-Band: Fixed in da0b5931: split the live probe out of `classify_initial_upgrade_error` into `probe_upgrade_error`, and cache its result on `WebSocketClient` across repeat unclassifiable `PHXConnectionError`s within one `__aenter__` retry loop (reset alongside the fresh `Session` on each new entry). A regression test (`test_aenter_caches_probe_result_across_repeated_connect_failures`) fails on the old code (2 probes) and passes now (1 probe).

src/band/client/streaming/client.py
```diff
@@ -395,58 +432,182 @@ async def _handle_reconnect(self) -> None:
...
+    ) -> WebSocketUpgradeError | None:
+        """Classify one failed connect exception. Reuses the previous
+        live-socket probe result only while the wrapped PHXConnectionError's
+        message is unchanged -- a different message means the underlying
+        failure may have changed, so it must be probed again."""
+        upgrade_error = WebSocketUpgradeError.from_exception(exc)
+        if upgrade_error is not None or not isinstance(exc, PHXConnectionError):
+            return upgrade_error
+        message = str(exc)
+        if message != self._probed_failure_message:
```
AlexanderZ-Band: issue: This key cannot distinguish two 429 responses with different `Retry-After` headers. The Phoenix wrapper includes `str(InvalidStatus)`, which is only `HTTP 429`; after a probe sees `Retry-After: 5`, a later 429 asking for 60 seconds reuses the cached 5-second classification. `Session` then retries before the server's current delay floor, potentially continuing to hammer a rate-limited endpoint.

AlexanderZ-Band (commented):
Found one correctness issue in the initial-connect retry cache.

src/band/client/streaming/client.py
AlexanderZ-Band: Confirmed and fixed in 62c43d68.

Verified: the wrapped PHXConnectionError's message for an InvalidStatus 429 is always just "server rejected WebSocket connection: HTTP 429" (websockets' InvalidStatus.__str__ carries only the status code, no headers), so two real 429s with different Retry-After values hash to the same cache key -- the message-keyed fix from the prior round couldn't tell them apart.

Changed the cache-validity check: a cached classification that carries a retry_after now always forces a fresh probe on the next occurrence, regardless of message match. This is scoped to the actual unsafe property (a retry_after that can vary and matters for Session's backoff) rather than hardcoding "429", so it stays correct if another status ever starts carrying one too. Terminal 400/409 and the plain-None case are unaffected -- they still reuse the cache, since Session either ignores retry_after (Dead) or there's nothing to go stale.

Added test_aenter_reprobes_repeated_429_for_current_retry_after (tests/websocket/test_client.py), which scripts two identical-message 429 failures with Retry-After 5 then 60 and asserts both get probed and the second delay reflects 60, not the stale 5 -- fails on the pre-fix code (1 probe, stale 5s reused) and passes now. Also retargeted test_aenter_caches_probe_result_across_repeated_connect_failures from 429 to 503, since a repeated 429 is no longer the right example of the caching path it's demonstrating.

src/band/client/streaming/client.py
```diff
@@ -395,58 +432,189 @@ async def _handle_reconnect(self) -> None:
...
+            )
+            logger.warning(
+                "Initial WebSocket connection permanently failed (dead_reason=%s): %s",
+                outcome.dead_reason,
+                raise_exc,
+            )
+            if raise_exc is exc:
+                # Re-raising the exception already being handled -- no new
+                # object, so no chaining to add (bare raise).
+                raise raise_exc
```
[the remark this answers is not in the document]
AlexanderZ-Band: Confirmed and fixed in 9361f79a. Verified live that raise raise_exc (even with raise_exc is exc) adds one extra traceback frame versus a bare raise -- switched to bare raise.

tests/websocket/conftest.py
```diff
@@ -0,0 +1,139 @@
...
+
+import pytest
+
+from band_sdk_core import SessionPolicy
+
+SUCCEEDS = object()
+"""Sentinel meaning a scripted connect attempt or probe finds no error."""
+
+
+class _Script:
```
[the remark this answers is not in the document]
AlexanderZ-Band: Confirmed and fixed in 9361f79a. Renamed _Script to Script (both call sites updated) to match the repo's no-leading-underscore class naming convention.

tests/websocket/conftest.py
```diff
@@ -0,0 +1,139 @@
...
+    probe_upgrade_error) driven by a script of outcomes: each call raises
+    the next scripted exception, or finds a clean handshake on SUCCEEDS."""
+
+    def __init__(self, *script: Exception | object):
+        self._script = _Script(script)
+        self.probed_urls: list[tuple[str, int]] = []
+
+    def __call__(self, url: str, *, open_timeout: float) -> _ScriptedProbeConnection:
+        self.probed_urls.append((url, open_timeout))
+        return _ScriptedProbeConnection(self._script.next())
```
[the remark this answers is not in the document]
AlexanderZ-Band: Confirmed and fixed in 9361f79a. Renamed _ScriptedProbeConnection to ScriptedProbeConnection, and retyped probed_urls as list[tuple[str, float]] to match open_timeout: float.

amit-gazal-band (approved):

PR #605, opened 2026-09-01
test: fill A2A adapter unit test gaps

AlexanderZ-Band:
## Summary

- Fills unit test gaps across the outbound `A2AAdapter` and `A2AGatewayAdapter`:
  peer-not-found resolution, peer-list pagination, artifact-overwrite
  reduction, `TASK_STATE_AUTH_REQUIRED` handling, and gateway JSON-RPC edge
  cases (malformed body, non-scalar request id, `SendStreamingMessage`,
  missing-id task methods, partial room-creation state).
- Adds three live baseline E2E smokes under
  `tests/e2e/baseline/smoke/adapters/` (A2A is a protocol bridge, not a
  matrix adapter, so these follow the `test_parlant.py` pattern):
  - `test_a2a.py` — `A2AAdapter` against a scripted, non-Band A2A
    counterparty fixture (`a2aServer.py`, built on a2a-sdk's own server
    primitives, no LLM key needed).
  - `test_a2a_gateway.py` — the official a2a-sdk reference client driving a
    live `A2AGatewayAdapter` that exposes a real Anthropic-backed Band peer.
  - `test_a2a_roundtrip.py` — the full Band-to-Band shape: Band Agent A
    (`A2AAdapter`) → live gateway → Band Agent B (Anthropic) → reply
    relayed back.
- Extracts `ManagedUvicornServer` (`band.integrations.uvicorn_server`), a
  shared "run an ASGI app on a background uvicorn server, wait for
  readiness, tear it down" lifecycle used by `GatewayServer`,
  `LocalMCPServer`, and the A2A test counterparty fixture.
- Adds `tests/lifecycle.py` (`running`, `backgrounded`, `held_open`,
  `elapsed`) so tests that drive a background server/task read as "given a
  running X, assert Y" instead of hand-rolled setup/teardown.

## Bugs fixed

- `A2AGatewayAdapter._execute_a2a`: a REST failure while posting to Band
  left the remote A2A caller stuck on a WORKING task forever, since no
  terminal FAILED event was ever published on that path. Now calls
  `pending.fail(...)` with a stable, non-sensitive message before
  re-raising (the raw exception is logged, not sent to the external caller).
- `A2AAdapter.cleanup_all`: an exception from the A2A client's `close()`
  skipped closing the owned httpx transport, leaking it on shutdown. Now
  closed in a `finally`.
- `A2AAdapter`'s owned httpx client left `read` unbounded, so a genuinely
  hung remote peer (accepts the connection, then never responds) could
  block a room forever — nothing in `band/runtime/execution.py` bounds an
  in-flight adapter cycle. Bounded to a generous-but-finite
  `_SSE_READ_TIMEOUT_S` (300s) instead.
- `GatewayServer.start()` returned as soon as `serve()` was scheduled,
  before uvicorn was actually listening, so a caller dialing in right after
  (a real A2A client, or the new smokes) could hit connection-refused.
  `start()` now waits for uvicorn's ready signal, bounded by a timeout, and
  unwinds the task/socket if that wait fails or times out.
- The A2A gateway leaked `sse_starlette`'s process-global shutdown flag
  across instances: `sse_starlette`'s background watcher promotes whichever
  `uvicorn.Server` owns the process's SIGTERM slot to a process-global
  switch, and `GatewayServer.stop()` sets `should_exit` directly (never via
  a real signal) — so once any gateway in a process stopped, every later
  gateway's SSE streams in that process got cancelled immediately.
  `band.integrations.mcp.local_server` had already hit and fixed the
  identical bug for itself; the gateway now disables the same
  process-global switch directly instead of depending on that other module
  happening to be imported first.
- `tests.lifecycle.running()` called `start()` before its `try`/`finally`,
  so a failing `start()` skipped `stop()` entirely; `start()` now runs
  inside the `try` so cleanup always happens.
- `GatewayServer.start()` and `A2AGatewayAdapter.on_started` logged the
  *requested* port instead of the OS-resolved `bound_port`, which is
  misleading for an ephemeral (`port=0`) bind.

## Test plan

- `uv run ruff check .` / `uv run ruff format .`
- `uv run pyrefly check` (0 errors)
- `uv run pytest tests/ --ignore=tests/integration/ --ignore=tests/e2e/ -q`
  (5294 passed, 123 skipped, 0 failed)
- `E2E_TESTS_ENABLED=true uv run pytest tests/e2e/baseline/smoke/adapters/test_a2a.py tests/e2e/baseline/smoke/adapters/test_a2a_gateway.py tests/e2e/baseline/smoke/adapters/test_a2a_roundtrip.py -v -s --no-cov`
  against a live platform + `ANTHROPIC_API_KEY` — all 4 tests pass
- Reverted the gateway shutdown-leak fix and reran to confirm the
  regression test actually catches it (`RemoteProtocolError: incomplete
  chunked read`), then restored the fix

Closes [INT-1357](https://linear.app/thenvoi/issue/INT-1357/a2a-adapter-add-live-e2e-coverage-fill-unit-test-gaps).

https://claude.ai/code/session_01SnZ59Puicp5HazTjWGZd5A

### DIFF
--- a/src/band/integrations/a2a/adapter.py
+++ b/src/band/integrations/a2a/adapter.py
@@ -107,7 +107,13 @@ class A2AAdapter(SimpleAdapter[A2ASessionState]):
 
         headers = self.auth.to_headers() if self.auth else {}
 
-        self._http_client = httpx.AsyncClient(headers=headers)
+        # httpx's default 5s read timeout fires on the normal, multi-second
+        # gap between SSE events during a real remote turn (a live LLM call,
+        # a tool loop) -- not a hang. Leave read unbounded; connect/write/pool
+        # stay bounded so a genuinely dead peer still fails promptly.
+        self._http_client = httpx.AsyncClient(
+            headers=headers, timeout=httpx.Timeout(10.0, read=None)
+        )
         factory = ClientFactory(
             ClientConfig(streaming=self.streaming, httpx_client=self._http_client)
         )
--- a/src/band/integrations/a2a/adapter.py
+++ b/src/band/integrations/a2a/adapter.py
@@ -319,12 +319,14 @@ class A2AAdapter(SimpleAdapter[A2ASessionState]):
 
     async def cleanup_all(self) -> None:
         """Close the owned A2A client and its HTTP transport."""
-        if self._client is not None:
-            await self._client.close()
-            self._client = None
-        if self._http_client is not None:
-            await self._http_client.aclose()
-            self._http_client = None
+        client, self._client = self._client, None
+        http_client, self._http_client = self._http_client, None
+        try:
+            if client is not None:
+                await client.close()
+        finally:
+            if http_client is not None:
+                await http_client.aclose()
 
     async def _emit_task_event(
         self, tools: AgentToolsProtocol, task: Task, state: TaskState
--- a/src/band/integrations/a2a/adapter.py
+++ b/src/band/integrations/a2a/adapter.py
@@ -37,6 +37,13 @@ from band.integrations.a2a.types import A2AAuth, A2ASessionState
 
 logger = logging.getLogger(__name__)
 
+# httpx's read timeout resets on every chunk received, so this bounds the gap
+# between SSE events, not the turn as a whole. Generous enough for the
+# multi-second silences of a live LLM call or tool loop; still finite, so a
+# peer that accepts the connection and then hangs eventually fails the turn
+# instead of blocking the room forever.
+_SSE_READ_TIMEOUT_S = 300.0
+
 
 class A2AAdapter(SimpleAdapter[A2ASessionState]):
     """Adapter that forwards messages to a remote A2A agent.
@@ -109,10 +116,11 @@ class A2AAdapter(SimpleAdapter[A2ASessionState]):
 
         # httpx's default 5s read timeout fires on the normal, multi-second
         # gap between SSE events during a real remote turn (a live LLM call,
-        # a tool loop) -- not a hang. Leave read unbounded; connect/write/pool
-        # stay bounded so a genuinely dead peer still fails promptly.
+        # a tool loop) -- not a hang. Use a generous bound instead of the
+        # default so a genuinely dead peer still fails promptly.
         self._http_client = httpx.AsyncClient(
-            headers=headers, timeout=httpx.Timeout(10.0, read=None)
+            headers=headers,
+            timeout=httpx.Timeout(10.0, read=_SSE_READ_TIMEOUT_S),
         )
         factory = ClientFactory(
             ClientConfig(streaming=self.streaming, httpx_client=self._http_client)
--- a/src/band/integrations/a2a/gateway/server.py
+++ b/src/band/integrations/a2a/gateway/server.py
@@ -22,6 +22,7 @@ from starlette.requests import Request
 from starlette.responses import JSONResponse
 from starlette.routing import BaseRoute, Route
 
+from band.integrations.uvicorn_server import wait_until_started
 from band_rest import Peer
 
 logger = logging.getLogger(__name__)
@@ -253,7 +254,7 @@ class GatewayServer:
         import uvicorn
 
         self._app = self._build_app()
-        self._uvicorn = uvicorn.Server(
+        server = uvicorn.Server(
             uvicorn.Config(
                 self._app,
                 host="0.0.0.0",
@@ -262,24 +263,15 @@ class GatewayServer:
                 timeout_graceful_shutdown=SERVER_STOP_TIMEOUT_S,
             )
         )
-        self._server_task = asyncio.create_task(self._uvicorn.serve())
-        await self._wait_until_started()
+        server_task = asyncio.create_task(server.serve())
+        self._uvicorn = server
+        self._server_task = server_task
+        await wait_until_started(server, server_task, timeout_s=SERVER_START_TIMEOUT_S)
         logger.info(
             "Starting A2A Gateway server on port %d with %d peers",
             self.port,
             len(self.peers),
         )
-
-    async def _wait_until_started(self) -> None:
-        assert self._uvicorn is not None
-        deadline = asyncio.get_running_loop().time() + SERVER_START_TIMEOUT_S
-        while not self._uvicorn.started:
-            if asyncio.get_running_loop().time() > deadline:
-                raise RuntimeError(
-                    f"A2A Gateway server did not start within "
-                    f"{SERVER_START_TIMEOUT_S}s on port {self.port}"
-                )
-            await asyncio.sleep(0.05)
 
     async def stop(self) -> None:
         if self._uvicorn is None or self._server_task is None:
--- a/src/band/integrations/a2a/gateway/server.py
+++ b/src/band/integrations/a2a/gateway/server.py
@@ -266,7 +266,15 @@ class GatewayServer:
         server_task = asyncio.create_task(server.serve())
         self._uvicorn = server
         self._server_task = server_task
-        await wait_until_started(server, server_task, timeout_s=SERVER_START_TIMEOUT_S)
+        try:
+            await wait_until_started(
+                server, server_task, timeout_s=SERVER_START_TIMEOUT_S
+            )
+        except BaseException:
+            # A failed/timed-out startup still leaves server_task running and
+            # the socket bound; stop() already unwinds both.
+            await self.stop()
+            raise
         logger.info(
             "Starting A2A Gateway server on port %d with %d peers",
             self.port,
--- a/src/band/integrations/a2a/gateway/server.py
+++ b/src/band/integrations/a2a/gateway/server.py
@@ -20,7 +20,11 @@ from starlette.requests import Request
 from starlette.responses import JSONResponse
 from starlette.routing import BaseRoute, Route
 
-from band.integrations.uvicorn_server import ManagedUvicornServer
+from band.integrations.uvicorn_server import (
+    SERVER_START_TIMEOUT_S,
+    SERVER_STOP_TIMEOUT_S,
+    ManagedUvicornServer,
+)
 from band_rest import Peer
 
 logger = logging.getLogger(__name__)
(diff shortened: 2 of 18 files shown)


src/band/integrations/a2a/adapter.py
```diff
@@ -107,7 +107,13 @@ async def on_started(self, agent_name: str, agent_description: str) -> None:
 
         headers = self.auth.to_headers() if self.auth else {}
 
-        self._http_client = httpx.AsyncClient(headers=headers)
+        # httpx's default 5s read timeout fires on the normal, multi-second
+        # gap between SSE events during a real remote turn (a live LLM call,
+        # a tool loop) -- not a hang. Leave read unbounded; connect/write/pool
+        # stay bounded so a genuinely dead peer still fails promptly.
+        self._http_client = httpx.AsyncClient(
+            headers=headers, timeout=httpx.Timeout(10.0, read=None)
```
amit-gazal-band: `httpx.Timeout(10.0, read=None)` removes read-timeout protection entirely, not just for the SSE inter-event gap this is meant to fix. If a remote A2A peer accepts the connection and then genuinely hangs (deadlocks, hung internally) after the request is sent, `on_message()`'s `async for event in self._client.send_message(...)` loop has no timeout wrapper, and nothing in `band/runtime/execution.py` bounds an in-flight adapter cycle either. The turn now hangs forever instead of failing promptly, which contradicts the comment above this call ("a genuinely dead peer still fails promptly"). Consider a bounded-but-generous read timeout (or an explicit per-turn `asyncio.wait_for` around the `send_message` loop) instead of `read=None`.

src/band/integrations/a2a/gateway/server.py
```diff
@@ -251,12 +263,24 @@ async def start(self) -> None:
...
         logger.info(
             "Starting A2A Gateway server on port %d with %d peers",
             self.port,
             len(self.peers),
         )
 
+    async def _wait_until_started(self) -> None:
+        assert self._uvicorn is not None
+        deadline = asyncio.get_running_loop().time() + SERVER_START_TIMEOUT_S
+        while not self._uvicorn.started:
```
amit-gazal-band: `_wait_until_started()` only polls `self._uvicorn.started` and never checks whether `self._server_task` already failed. If `uvicorn.serve()` fails fast (e.g. port already in use), `.started` never flips to `True`, so this busy-waits the full `SERVER_START_TIMEOUT_S` and then raises a generic "did not start within Ns" `RuntimeError` instead of surfacing the real `OSError`/exception — and the failed task's exception is never retrieved (asyncio will log "Task exception was never retrieved"). The sibling implementation in `src/band/integrations/mcp/local_server.py:352-356` already handles this correctly:
```python
while not self._uvicorn_server.started:
    if self._serve_task.done():
        await self._serve_task
    ...
```
Worth mirroring that check here.

src/band/integrations/a2a/gateway/server.py
```diff
@@ -251,12 +263,24 @@ async def start(self) -> None:
...
         )
         self._server_task = asyncio.create_task(self._uvicorn.serve())
+        await self._wait_until_started()
         logger.info(
             "Starting A2A Gateway server on port %d with %d peers",
             self.port,
             len(self.peers),
         )
 
+    async def _wait_until_started(self) -> None:
```
amit-gazal-band: This uvicorn-readiness busy-poll loop is now implemented three times with no shared helper: `GatewayServer._wait_until_started` (here), `LocalMCPServer._wait_until_started` (`src/band/integrations/mcp/local_server.py:347`), and a third inline copy added in `tests/e2e/baseline/smoke/adapters/a2aServer.py`. The `mcp/local_server.py` version is the correct one (it checks for a failed serve task — see the other comment on line 276); the other two are missing that check. A future fix or tuning of this logic (timeout, poll interval, or this exact correctness gap) has to be hand-applied in three places, and as this review shows, that already went wrong twice. Consider extracting a shared `wait_until_started(uvicorn_server, serve_task, timeout_s)` helper both adapters (and the test fixture) can call.

amit-gazal-band (changes requested):
Flagging three issues from review, focused on the SSE-timeout fix and the new gateway startup-wait logic.

src/band/integrations/a2a/adapter.py
AlexanderZ-Band: Confirmed -- verified live that `on_message()`'s `async for event in self._client.send_message(...)` loop has no timeout wrapper and `band/runtime/execution.py` does not bound an in-flight adapter cycle either, so `read=None` really did mean a peer that hangs after accepting the connection would block the room forever.

Fixed by swapping `read=None` for a generous-but-finite bound (`_SSE_READ_TIMEOUT_S = 300.0`) instead of adding a separate `asyncio.wait_for`/per-turn timeout layer -- httpx already resets its read timeout on every chunk received, so this is exactly the "bounded but generous" mechanism you suggested, applied at the one place it needs to live. Updated the regression test (`test_owned_http_client_has_a_generous_bounded_read_timeout`) to assert the new bound instead of `None`.

See `src/band/integrations/a2a/adapter.py` (commit 43b0557e).

src/band/integrations/a2a/gateway/server.py
AlexanderZ-Band: Confirmed -- `_wait_until_started` only polled `self._uvicorn.started` and never checked whether `self._server_task` had already died, so a fast startup failure (e.g. port already in use) would busy-wait the full `SERVER_START_TIMEOUT_S` and then raise a generic `RuntimeError`, leaking the real exception into an unretrieved task.

Fixed by adopting `mcp/local_server.py`'s already-correct check (`if serve_task.done(): await serve_task`, re-raising the real failure immediately) -- see the reply on the sibling thread for how this and the duplication finding were fixed together via one shared helper, `band.integrations.uvicorn_server.wait_until_started`.

See `src/band/integrations/a2a/gateway/server.py` (commit 43b0557e).

src/band/integrations/a2a/gateway/server.py
AlexanderZ-Band: Agreed -- extracted the correct version (the one with the serve-task-failure check, per the other thread) into `band.integrations.uvicorn_server.wait_until_started`, a new module directly under `integrations/` (a shared-across-integrations location, alongside the existing `integrations/base.py`, so neither `a2a` nor `mcp` depends on the other).

All three copies now call it:
- `mcp/local_server.py` (was already correct -- now the shared implementation)
- `a2a/gateway/server.py` (was missing the failure check -- fixed by adopting it)
- `tests/e2e/baseline/smoke/adapters/a2aServer.py` (was a third duplicate -- now calls the same helper)

Added focused tests for the shared helper (`tests/integrations/test_uvicorn_server.py`): happy-path readiness, timeout, and the regression case (a serve task that fails immediately must surface its real exception well under the timeout window, not a generic one after busy-waiting it). Ran the fix-reverted version of the helper against that regression test to confirm it actually hangs until pytest's timeout without the fix.

See `src/band/integrations/uvicorn_server.py`, `src/band/integrations/mcp/local_server.py`, `src/band/integrations/a2a/gateway/server.py`, `tests/e2e/baseline/smoke/adapters/a2aServer.py`, `tests/integrations/test_uvicorn_server.py` (commit 43b0557e).

src/band/integrations/a2a/gateway/server.py
```diff
@@ -250,7 +263,10 @@ async def start(self) -> None:
                 timeout_graceful_shutdown=SERVER_STOP_TIMEOUT_S,
             )
         )
-        self._server_task = asyncio.create_task(self._uvicorn.serve())
+        server_task = asyncio.create_task(server.serve())
+        self._uvicorn = server
+        self._server_task = server_task
+        await wait_until_started(server, server_task, timeout_s=SERVER_START_TIMEOUT_S)
         logger.info(
```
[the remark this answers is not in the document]
AlexanderZ-Band: Fixed — `start()` now wraps `wait_until_started(...)` in try/except and calls `stop()` on failure, so a failed/timed-out startup no longer leaks the task/socket. Later consolidated into `ManagedUvicornServer.start()`, which owns this same try/except once for every embedder (`GatewayServer`, `LocalMCPServer`'s sibling, and the A2A test fixture).

src/band/integrations/a2a/adapter.py
```diff
@@ -107,7 +114,14 @@ async def on_started(self, agent_name: str, agent_description: str) -> None:
...
 
-        self._http_client = httpx.AsyncClient(headers=headers)
+        # httpx's default 5s read timeout fires on the normal, multi-second
+        # gap between SSE events during a real remote turn (a live LLM call,
+        # a tool loop) -- not a hang. Use a generous bound instead of the
+        # default so a genuinely dead peer still fails promptly.
+        self._http_client = httpx.AsyncClient(
+            headers=headers,
+            timeout=httpx.Timeout(10.0, read=_SSE_READ_TIMEOUT_S),
+        )
```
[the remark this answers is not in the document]
AlexanderZ-Band: Fixed — description now says the bound is a generous-but-finite `_SSE_READ_TIMEOUT_S` (300s), not unbounded.

tests/lifecycle.py
```diff
@@ -0,0 +1,60 @@
...
+
+
+@asynccontextmanager
+async def running(server: Startable) -> AsyncIterator[Startable]:
+    """Start ``server``, yield it, and always stop it -- even on failure."""
+    await server.start()
+    try:
+        yield server
+    finally:
+        await server.stop()
```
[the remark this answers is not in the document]
AlexanderZ-Band: Fixed — `start()` now runs inside the `try`, so a failed/timed-out start still hits the `finally: await server.stop()`. Also added a docstring on `Startable` stating the no-op-on-not-started contract this relies on, since every current implementer honors it but the protocol didn't say so.

src/band/integrations/a2a/gateway/server.py
```diff
@@ -237,38 +231,33 @@ async def _handle_list_peers(self, _request: Request) -> JSONResponse:
...
+            start_timeout_s=SERVER_START_TIMEOUT_S,
+            stop_timeout_s=SERVER_STOP_TIMEOUT_S,
         )
-        self._server_task = asyncio.create_task(self._uvicorn.serve())
+        await self._runtime.start()
         logger.info(
             "Starting A2A Gateway server on port %d with %d peers",
             self.port,
             len(self.peers),
         )
```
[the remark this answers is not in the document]
AlexanderZ-Band: Fixed — now logs `self.bound_port` instead of `self.port`. Also found and fixed the identical bug one layer up in `A2AGatewayAdapter.on_started`, which had the same requested-vs-bound-port mislog.

amit-gazal-band (approved):

PR #606, opened 2026-09-03
feat: Add room-task-board capability to band-sdk-python

AlexanderZ-Band:
## Summary

Adds `Capability.TASKS`: 7 agent-surface tools (`band_list_tasks`, `band_create_task`, `band_get_task`, `band_update_task`, `band_get_task_history`, `band_get_board`, `band_set_board`) backed by `band-client-rest`'s `agent_api_chat_tasks` client.

- New `band.core.task_types` vocabulary: `TaskListState`, `TaskLifecycleState`, `TaskAssignmentStatus`, `TaskIncludeOption`.
- `Capability.TASKS` wired into all 16 adapters, a `tasks` `--tools` group on the `band-mcp` CLI, and hand-written wrappers for the three adapters (CrewAI/CrewAIFlow, Pydantic AI, Parlant) that don't consume the central tool registry generically.
- `tool_filter.sanitize_tool_schema` converts a single-value `Literal` field's JSON-Schema `const` to `enum` (Gemini's restricted JSON-Schema subset rejects `const`), applied in the MCP engine's registration path and the shared `platform_args_schema()` chokepoint. Pydantic AI's and Parlant's own signature-based schema builders can't take a `Literal` directly, so their `include` parameter is typed `str` and cast back to `Literal["history"]` at the `AgentToolsProtocol` boundary; `AgentTools.get_task`/`get_board` validate the value at runtime regardless of which path a caller takes.
- `UpdateTaskInput`, `SetBoardInput`, and `RemoveMyContactInput` each require at least one settable field, enforced by a shared `at_least_one_of()` helper (`band.core.validation`); `AgentTools.update_task`/`set_board` re-check the same rule directly, so adapters that hand-register tools instead of constructing the Pydantic input model are covered too.
- Every REST call in the new task/board `AgentTools` methods passes `request_options=DEFAULT_REQUEST_OPTIONS` (3x retry on transient failure), matching every other `AgentTools` method.
- `src/band/runtime/tools.py` (4000+ lines) split into a `src/band/runtime/tools/` package (`types`, `registry`, `schema`, `agent`, `human`, `inputs/*`); the public import surface is unchanged.
- Live E2E coverage: a multi-agent task-board delegation test, plus a per-adapter task-lifecycle test (create → claim → complete → list/get/get-history/get-board) across every `Capability.TASKS` adapter.

## Also included: reject blank/whitespace-only content at the source

Stacked on this branch because its core file, `src/band/runtime/tools/agent.py`, only exists after the `tools.py` → `tools/` split above. Companion fix: [band-sdk-typescript#174](https://github.com/band-ai/band-sdk-typescript/pull/174).

- `band.core.content.has_visible_content` — shared predicate mirroring the platform's own visible-content rule (Unicode categories L/N/P/S; whitespace/zero-width/bidi-mark-only strings don't count).
- A Pydantic field validator on every tool-facing `content` field (`SendMessageInput`, `SendEventInput`, `SendEventWideInput`, `SendMyChatMessageInput`).
- `band.platform.posting.post_message`/`post_event` — the one choke point every chat-send REST call in `src/band` routes through; refuses blank content with a log warning instead of hitting the platform. Enforced by an AST guardrail, `tests/platform/test_posting_boundary.py`.
- `send_room_file` rejects a whitespace-only caption instead of raising `AttributeError` after the file upload succeeds. `ACPServerAdapter.handle_prompt` and `A2AGatewayAdapter._send_to_band` fail fast on blank content instead of hanging out the full response timeout. `FakeAgentTools` rejects blank content the same way the real tools do.

## Test plan

- [x] `uv run pyrefly check`
- [x] `uv run pytest tests/ --ignore=tests/integration/ --ignore=tests/e2e/ -v` (dev venv, full suite green)
- [x] `UV_PROJECT_ENVIRONMENT=.venv-crewai uv run pytest ...` (crewai-specific suites)
- [x] `UV_PROJECT_ENVIRONMENT=.venv-parlant uv run pytest ...` (parlant-specific suites)
- [x] Unit tests: `tests/runtime/test_task_tools.py`, `tests/core/test_validation.py`, `tests/framework_conformance/test_task_tool_conformance.py`
- [x] Extended: `tests/runtime/test_tool_definitions.py`, `tests/mcp/test_wire_contract.py`, `test_config.py`, `test_standalone_spec.py`, `tests/integrations/parlant/test_tools.py`, `tests/framework_conformance/test_tool_name_drift.py`
- [x] Live E2E baseline: `E2E_TESTS_ENABLED=true uv run pytest tests/e2e/baseline/smoke/behavior/test_multi_agent_collaboration.py -k task_board -v -s --no-cov` and `... test_capability_matrix.py -k task -v -s --no-cov`, both against the real dev platform.
- [x] Blank-content fix: `tests/core/test_content.py`, `tests/platform/test_posting.py`, `tests/platform/test_posting_boundary.py`; extended `tests/runtime/test_tools.py`, `tests/testing/test_fake_tools.py`, `tests/integrations/acp/test_room_emitter.py`, `tests/integrations/acp/test_server_adapter.py`, `tests/mcp/test_engine.py`; `docs/rest-client.md` gained a runnable markdown-docs snippet.

Followup filed: [INT-1375](https://linear.app/thenvoi/issue/INT-1375/genericize-pydantic-ai-adapter-tool-registration-via-a-schema-driven) — genericize `pydantic_ai.py`'s per-tool hand-written registration into a schema-driven factory (assigned to Alexander Zaikman).

https://claude.ai/code/session_01LBTVbqCTPwkbEM9WKgYZAj


src/band/adapters/pydantic_ai.py
```diff
@@ -644,6 +645,145 @@ async def band_archive_memory(
...
+            @platform_tool
+            async def band_get_task(
+                ctx: RunContext[AgentToolsProtocol],
+                id: str,
+                # str, not Literal["history"] | None: pydantic-ai's own schema
+                # builder emits an unsanitized JSON-Schema `const` for a
+                # single-value Literal (unlike the master model/MCP paths,
+                # which run sanitize_tool_schema()), which providers with a
+                # restricted JSON-Schema subset (e.g. Gemini) reject.
+                include: str | None = None,
```
amit-gazal-band, line 696: `include` is cast from `str` to `Literal["history"] | None` purely to dodge pydantic-ai's schema-builder issue with single-value `Literal`s (per the comment above), but the cast is type-checker-only -- nothing validates the runtime value. Unlike every other constrained field in this file, an LLM can pass `include="anything"` here (and in `band_get_board` below, line 757) and it goes straight to the live REST API with no local rejection. Might be worth a small runtime check (`if include not in (None, "history"): raise ValueError(...)`) alongside the cast.

src/band/runtime/tools/inputs/human_contacts.py
```diff
@@ -0,0 +1,115 @@
...
+    """Look up an entity by handle.
+
+    Resolves a handle to its entity details. Use this to verify a handle
+    exists before sending a contact request.
+    """
+
+    handle: str = Field(..., description="The handle to resolve (required).")
+
+
+class RemoveMyContactInput(BaseModel):
```
amit-gazal-band, line 102: This docstring promises "At least one must be provided," but there's no validator enforcing it -- `RemoveMyContactInput(contact_id=None, handle=None)` validates fine today. `UpdateTaskInput`/`SetBoardInput` in `tasks.py` cover this exact constraint shape with a `model_validator`; this class is missing the equivalent.

src/band/integrations/parlant/tools.py
```diff
@@ -654,10 +664,128 @@ async def band_send_room_file(
...
+                    subject, detail=detail, supersedes_id=supersedes_id
+                )
+            )
+            return ToolResult(data=json.dumps(data, default=str))
+
+        @band_tool("getting task '{id}'")
+        async def band_get_task(
+            context: ToolContext,
+            id: str,
+            include: str | None = None,
```
amit-gazal-band, line 704: This cast mirrors pydantic_ai.py's `band_get_task`/`band_get_board`, but without the explanatory comment pydantic_ai and crewai's `catalog.py` carry for why it's `str` instead of `Literal["history"] | None`. Is Parlant's own tool-schema builder subject to the same single-value-`Literal` issue, or was this copied over without checking? If it's not needed here, `Literal["history"] | None` would be simpler and get real validation for free; if it is needed, worth adding the same comment (same question applies to `band_get_board` at line 753).

src/band/runtime/tools/inputs/tasks.py
```diff
@@ -0,0 +1,166 @@
...
+        ),
+    )
+    subject: str | None = Field(None, description="Edit the task subject")
+    detail: str | None = Field(None, description="Edit the task detail")
+    state: TaskLifecycleState | None = Field(
+        None,
+        description="Lifecycle: cancel, archive, or restore ('active' un-archives)",
+    )
+
+    @model_validator(mode="after")
```
amit-gazal-band, line 106: `UpdateTaskInput.validate_at_least_one_field` (here) and `SetBoardInput.validate_at_least_one_field` (line 162) are two independently hand-written "raise if all of these fields are None" checks. `b5cfd9b1` just added two more copies of the same logic directly in `AgentTools.update_task`/`set_board` (`agent.py`) as a backstop for callers that bypass this model, so the same constraint is now hand-written in four places. Worth factoring into one shared `at_least_one_of(**fields)` helper that both the Pydantic validators and the `AgentTools` methods can call.

amit-gazal-band (changes requested):
A few correctness gaps and consistency nits from reviewing the room task-board changes. Note: the pydantic_ai/parlant at-least-one-field bypass on `update_task`/`set_board`, and the `response.data` guard in `agent.py`'s task methods, were both already resolved by `b5cfd9b1` (the guard turned out to be load-bearing, not dead code), so neither is repeated below. Two other blank-content-swallowed-as-success bugs (CrewAI's `_send_message` in `catalog.py`, pydantic_ai's `band_send_message`/`band_send_event`) are real but predate this PR -- that code hasn't changed since well before this branch, so they're not part of this diff; flagging separately rather than as inline comments here.

src/band/adapters/pydantic_ai.py
AlexanderZ-Band: Fixed at the root: added a `validate_include()` backstop directly in `AgentTools.get_task`/`get_board` (agent.py), the one place every adapter's call converges — same pattern as the existing `at_least_one_of` backstop for `update_task`/`set_board`. No change needed in this file; the cast here is still required (schema-builder workaround), but the runtime value is now validated regardless of which adapter calls through.

src/band/runtime/tools/inputs/human_contacts.py
AlexanderZ-Band: Fixed: added the missing `model_validator` on `RemoveMyContactInput`, using a new shared `at_least_one_of()` helper (see the reply on the other thread). Note this is reachable in practice, not just theoretical — the standalone `band-mcp` server validates every human-surface tool's input model (`build_tool_registration` → `validate_tool_arguments`) before dispatch, so this closes a real gap for that path. `HumanTools.remove_my_contact`'s own inline guard is untouched (it's intentionally preserved, tested behavior matching the pre-split MCP handler's exact output string).

src/band/integrations/parlant/tools.py
AlexanderZ-Band: Verified against the actual installed dependency (`parlant==3.3.2`, matching this repo's own `parlant>=3.3.2` pin): the cast is necessary, but not for the same reason as pydantic-ai. Parlant's own signature-based schema builder (`parlant.core.services.tools.plugins._describe_parameters`/`_resolve_param_info`) has no special case for `Literal` at all — it falls into the "generic container" branch, sees `get_args(Literal["history"])` is non-empty, and raises `Exception("Only \`list\` is supported as a generic container in parameters")` at *tool-registration* time (i.e. this would crash on import, not just get rejected by Gemini like pydantic-ai's case). So this wasn't copied without checking — it's required, for a harder failure mode. Added a comment stating that.

src/band/runtime/tools/inputs/tasks.py
AlexanderZ-Band: Fixed: extracted a shared `at_least_one_of(**fields)` helper (`src/band/core/validation.py`) and wired it into `UpdateTaskInput`/`SetBoardInput`'s validators, `AgentTools.update_task`/`set_board`'s backstops, and `RemoveMyContactInput`'s new validator (see the other thread) — one implementation for what was 4 (now 5) hand-written copies.

amit-gazal-band (approved):

PR #610, opened 2026-09-04
fix: Isolate self-hosted Letta MCP tool storage per adapter instance

AlexanderZ-Band:
## Summary

Self-hosted Letta dedupes MCP-discovered `Tool` rows by `(name, organization_id)`. With no `user_id` header set, every `LettaAdapter` instance on a shared self-hosted server resolves to the same default org/actor — so a second instance's MCP registration silently re-points the first instance's `band_send_message` tool row's `metadata_.mcp.server_id` to itself. When the first instance's agent later calls that tool, Letta routes the call into the second instance's in-process MCP server, posting the reply under the wrong Band agent's identity. Root-caused and verified live against `letta-ai/letta@0.16.8` (see INT-985's Investigation section).

- Provision a dedicated Letta organization + user per self-hosted `LettaAdapter` instance and send its `user_id` in every request (`AsyncLetta(default_headers={"user_id": ...})`), isolating MCP server + tool storage between instances. Default-on for self-hosted, no-op on Cloud (`org_scoped=True` against Cloud is rejected at config construction — Cloud doesn't expose the admin API this needs).
- New `src/band/integrations/letta/orgscope.py`: a small raw-`httpx` client for Letta's self-hosted-only admin API (`/v1/admin/orgs/`, `/v1/admin/users/` — not exposed by the `letta_client` SDK), following the `opencode/client.py` precedent.
- Force-seed base tools (`tools.list()`) for a freshly-provisioned org in `on_started`, since a fresh org otherwise has zero `Tool` rows and the first `agents.create(include_base_tools=True)` would raise.
- CI: set `LETTA_NO_DEFAULT_ACTOR=true` on the test Letta server so an unresolvable `user_id` fails loud instead of silently falling back to the default org (closing a residual gap that would otherwise mask a single-instance scoping bug).
- Removed the `Adapter.LETTA` exclusion from the rehydration E2E test — this fix is exactly what that exclusion was waiting on.

**Bug found and fixed while live-testing this against a real Letta server:** the org/user pagination helper's `after`-cursor walk stalled forever once it landed on Letta's own seeded default user, whose `created_at` is `null` — Letta's own pagination code silently drops the boundary filter for a null-`created_at` cursor, so the "next page" comes back as the same unfiltered list forever. Fixed with stall detection (treat a non-advancing cursor as exhaustion) plus a permanent regression test.

## Test plan

- [x] `uv run pytest tests/adapters/test_letta_orgscope.py tests/adapters/test_letta_mcp.py tests/adapters/test_letta_adapter.py tests/adapters/test_deprecation_shims.py -v` — all pass (including 6 new unit/wiring tests + 1 pagination-stall regression test)
- [x] Full unit suite: `uv run pytest tests/ --ignore=tests/integration/ --ignore=tests/e2e/` — 5402 passed, 0 failed
- [x] `uv run ruff check . && uv run ruff format --check . && uv run pyrefly check` — clean
- [x] Live integration tests against a real self-hosted Letta (Docker): `uv run pytest tests/integration/test_letta_live.py -v -s --no-cov` — 4/4 pass, including the new `test_two_instances_stay_isolated_in_shared_org` (org/user isolation, MCP tool metadata non-collision, cross-org 404 checks, fresh-org agent creation)
- [x] E2E baseline `letta` lane against the live Band platform + LLMs: `BAND_E2E_LANE=letta E2E_TESTS_ENABLED=true uv run pytest tests/e2e/baseline/ -v -s --no-cov` — 60 passed, 0 failed, including `test_partial_reboot_preserves_context_and_peer[letta]` (the test this fix was specifically meant to unblock)

https://claude.ai/code/session_018j6NRbrMzkT4jTP6dBChsn

### DIFF
--- a/src/band/adapters/letta.py
+++ b/src/band/adapters/letta.py
@@ -24,10 +24,10 @@ from band.core.types import (
     TurnUsage,
 )
 from band.integrations.letta.config import (
-    LETTA_CLOUD_BASE_URL,
     LettaAdapterConfig,
     LettaMCPConfig,
     MCPTransport,
+    is_letta_cloud_url,
 )
 from band.integrations.letta.mcp import LettaMCPBridge, bounded_teardown
 from band.integrations.letta.orgscope import resolve_org_scoped_headers
@@ -213,7 +213,7 @@ class LettaAdapter(SimpleAdapter[LettaSessionState]):
         org_scoped = (
             self.config.org_scoped
             if self.config.org_scoped is not None
-            else self.config.base_url.rstrip("/") != LETTA_CLOUD_BASE_URL.rstrip("/")
+            else not is_letta_cloud_url(self.config.base_url)
         )
         if org_scoped:
             client_kwargs["default_headers"] = await resolve_org_scoped_headers(
--- a/src/band/adapters/letta.py
+++ b/src/band/adapters/letta.py
@@ -187,6 +187,30 @@ class LettaAdapter(SimpleAdapter[LettaSessionState]):
             features=self.features,
         )
 
+        self._client = await self._build_client(agent_name)
+
+        # Fail loud at startup if the tool path cannot be wired — the adapter
+        # is useless without it.
+        await self._mcp.ensure_ready(self._client)
+
+        logger.info(
+            "Letta adapter started for agent: %s (mode=%s, mcp=%s)",
+            agent_name,
+            self.config.mode,
+            self.config.mcp.mode,
+        )
+
+    async def _build_client(self, agent_name: str) -> Any:
+        """Construct the AsyncLetta SDK client, org-scoped for self-hosted servers.
+
+        Self-hosted Letta dedupes MCP-discovered Tool rows by
+        (name, organization_id): with no user_id header, every instance
+        resolves to the same default org, and a second instance's MCP
+        registration silently re-points the first instance's tool row to
+        its own server. Scoping each instance to its own org+user closes
+        that collision. Cloud is never scoped (org_scoped=True against it
+        is already rejected at config construction — see LettaAdapterConfig).
+        """
         try:
             from letta_client import AsyncLetta  # type: ignore[import-not-found]  # optional dependency
         except ImportError:
@@ -203,13 +227,6 @@ class LettaAdapter(SimpleAdapter[LettaSessionState]):
         if self.config.project:
             client_kwargs["project"] = self.config.project
 
-        # Self-hosted Letta dedupes MCP-discovered Tool rows by
-        # (name, organization_id): with no user_id header, every instance
-        # resolves to the same default org, and a second instance's MCP
-        # registration silently re-points the first instance's tool row to
-        # its own server. Scoping each instance to its own org+user closes
-        # that collision. Cloud is never scoped (org_scoped=True against it
-        # is already rejected at config construction — see LettaAdapterConfig).
         org_scoped = (
             self.config.org_scoped
             if self.config.org_scoped is not None
@@ -221,25 +238,16 @@ class LettaAdapter(SimpleAdapter[LettaSessionState]):
                 agent_name=agent_name,
                 bearer_token=self.config.provider_key,
             )
-        self._client = AsyncLetta(**client_kwargs)
+
+        client = AsyncLetta(**client_kwargs)
         if org_scoped:
             # A fresh org has zero Tool rows; base tools (send_message, etc.)
             # are seeded lazily only by this list call (GET /v1/tools/ with
             # upsert_base_tools). Nothing else in this adapter's flow reaches
             # it, so the first agents.create(include_base_tools=True) in a
             # fresh org would otherwise raise before the agent exists.
-            await self._client.tools.list()
-
-        # Fail loud at startup if the tool path cannot be wired — the adapter
-        # is useless without it.
-        await self._mcp.ensure_ready(self._client)
-
-        logger.info(
-            "Letta adapter started for agent: %s (mode=%s, mcp=%s)",
-            agent_name,
-            self.config.mode,
-            self.config.mcp.mode,
-        )
+            await client.tools.list()
+        return client
 
     # ------------------------------------------------------------------
     # Message handling
--- a/src/band/adapters/letta.py
+++ b/src/band/adapters/letta.py
@@ -24,11 +24,13 @@ from band.core.types import (
     TurnUsage,
 )
 from band.integrations.letta.config import (
+    LETTA_CLOUD_BASE_URL,
     LettaAdapterConfig,
     LettaMCPConfig,
     MCPTransport,
 )
 from band.integrations.letta.mcp import LettaMCPBridge, bounded_teardown
+from band.integrations.letta.orgscope import resolve_org_scoped_headers
 from band.integrations.letta.prompts import render_tool_enforcement
 from band.runtime.prompts import render_system_prompt
 from band.runtime.tools import (
@@ -200,7 +202,33 @@ class LettaAdapter(SimpleAdapter[LettaSessionState]):
             client_kwargs["api_key"] = self.config.provider_key
         if self.config.project:
             client_kwargs["project"] = self.config.project
+
+        # Self-hosted Letta dedupes MCP-discovered Tool rows by
+        # (name, organization_id): with no user_id header, every instance
+        # resolves to the same default org, and a second instance's MCP
+        # registration silently re-points the first instance's tool row to
+        # its own server. Scoping each instance to its own org+user closes
+        # that collision. Cloud is never scoped (org_scoped=True against it
+        # is already rejected at config construction — see LettaAdapterConfig).
+        org_scoped = (
+            self.config.org_scoped
+            if self.config.org_scoped is not None
+            else self.config.base_url.rstrip("/") != LETTA_CLOUD_BASE_URL.rstrip("/")
+        )
+        if org_scoped:
+            client_kwargs["default_headers"] = await resolve_org_scoped_headers(
+                base_url=self.config.base_url,
+                agent_name=agent_name,
+                bearer_token=self.config.provider_key,
+            )
         self._client = AsyncLetta(**client_kwargs)
+        if org_scoped:
+            # A fresh org has zero Tool rows; base tools (send_message, etc.)
+            # are seeded lazily only by this list call (GET /v1/tools/ with
+            # upsert_base_tools). Nothing else in this adapter's flow reaches
+            # it, so the first agents.create(include_base_tools=True) in a
+            # fresh org would otherwise raise before the agent exists.
+            await self._client.tools.list()
 
         # Fail loud at startup if the tool path cannot be wired — the adapter
         # is useless without it.
--- a/src/band/integrations/letta/config.py
+++ b/src/band/integrations/letta/config.py
@@ -14,6 +14,11 @@ from band.integrations.mcp.local_server import LOCAL_MCP_HOST
 
 MCPTransport = Literal["sse", "streamable_http"]
 
+# Single source of truth for the Letta Cloud endpoint: both the base_url
+# field default and self-hosted detection (org_scoped auto-detection, the
+# Cloud+org_scoped=True construction guard) read this constant.
+LETTA_CLOUD_BASE_URL = "https://api.letta.com"
+
 
 @dataclass
 class LettaMCPConfig:
@@ -102,7 +107,7 @@ class LettaAdapterConfig(BaseSettings):
     provider_key: str | None = Field(
         default=None, validation_alias=AliasChoices("LETTA_API_KEY")
     )  # Required for Letta Cloud; omit for self-hosted
-    base_url: str = "https://api.letta.com"
+    base_url: str = LETTA_CLOUD_BASE_URL
     custom_section: str = ""
     include_base_instructions: bool = True
     persona: str | None = None
@@ -112,6 +117,19 @@ class LettaAdapterConfig(BaseSettings):
 
     # Letta Cloud project scoping (ignored for self-hosted)
     project: str | None = None
+
+    # Self-hosted only: provision a dedicated Letta organization + user per
+    # adapter instance so MCP tool storage never collides between instances
+    # sharing one server (Letta dedupes MCP-discovered Tool rows by
+    # (name, organization_id); with no scoping, a second instance silently
+    # re-points the first instance's tool to its own MCP server). None
+    # (default) auto-enables for a self-hosted base_url and stays off for
+    # Letta Cloud. Explicit False is the only meaningful override — it opts
+    # a self-hosted deployment back out. Explicit True against Letta Cloud
+    # raises BandConfigError at construction (see _reject_org_scoped_cloud
+    # below): Cloud does not expose the admin API this needs, so honoring it
+    # would fail deep inside on_started instead of failing loud up front.
+    org_scoped: bool | None = None
 
     # Embedding model passed on agent create. Letta's Docker server requires
     # one; Cloud picks its own default when None.
@@ -190,4 +208,17 @@ class LettaAdapterConfig(BaseSettings):
             )
             self.mcp_server_url = None
             self.mcp_server_name = None
+        return self
+
+    @model_validator(mode="after")
+    def _reject_org_scoped_cloud(self) -> LettaAdapterConfig:
+        is_cloud = self.base_url.rstrip("/") == LETTA_CLOUD_BASE_URL.rstrip("/")
+        if self.org_scoped and is_cloud:
+            raise BandConfigError(
+                "org_scoped=True is not supported against Letta Cloud "
+                f"(base_url={self.base_url!r}) — it provisions organizations "
+                "and users via a self-hosted-only admin API that Letta Cloud "
+                "does not expose. Leave org_scoped unset (Cloud stays "
+                "unscoped) or set base_url to a self-hosted server."
+            )
         return self
--- a/src/band/integrations/letta/config.py
+++ b/src/band/integrations/letta/config.py
@@ -5,6 +5,7 @@ from __future__ import annotations
 import warnings
 from dataclasses import dataclass
 from typing import Any, Literal
+from urllib.parse import urlsplit
 
 from pydantic import AliasChoices, Field, model_validator
 from pydantic_settings import BaseSettings, SettingsConfigDict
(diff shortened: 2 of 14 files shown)


src/band/integrations/letta/orgscope.py
```diff
@@ -0,0 +1,152 @@
...
+        against ``letta/orm/sqlalchemy_base.py``'s ``_list_preprocess``) --
+        the next "page" comes back as the same unfiltered list again, always
+        ending on that same row once more. Advancing ``after`` to an
+        unchanged value is the signal that pagination has stalled; treating
+        it as exhaustion avoids looping forever instead of trying to detect
+        the null-created_at row directly (an implementation detail of a
+        response shape this client does not otherwise need to know).
+        """
+        after: str | None = None
+        while True:
```
AlexanderZ-Band: Should have circuit breaker or loop on kill signal in any case while true is a bad patter

tests/adapters/test_deprecation_shims.py
```diff
@@ -158,6 +158,35 @@ def test_letta_provider_key_and_api_key_conflict(self) -> None:
...
+        ],
+    )
+    def test_letta_org_scoped_and_cloud_conflict(self, base_url: str) -> None:
+        from band.adapters.letta import LettaAdapterConfig
+
+        with pytest.raises(BandConfigError, match="org_scoped=True"):
+            LettaAdapterConfig(base_url=base_url, org_scoped=True)
+
+    def test_letta_org_scoped_true_on_self_hosted_is_accepted(self) -> None:
+        from band.adapters.letta import LettaAdapterConfig
```
AlexanderZ-Band, line 185: Top level imports of needed we have lazy import function for this, see open pr carful not to collide

tests/adapters/test_letta_mcp.py
```diff
@@ -192,6 +193,178 @@ async def test_on_started_mcp_registration_failure_raises(self) -> None:
...
+        base_url = "http://localhost:8283"
+        adapter = LettaAdapter(
+            config=LettaAdapterConfig(
+                base_url=base_url, mcp=LettaMCPConfig(mode="external")
+            )
+        )
+
+        httpx_mock.add_response(method="GET", url=f"{base_url}/v1/admin/orgs/", json=[])
+        httpx_mock.add_response(
+            method="POST",
```
AlexanderZ-Band: Feels like this should be a common fixture or tool, check if we have something like this already

tests/adapters/test_letta_mcp.py
```diff
@@ -192,6 +193,178 @@ async def test_on_started_mcp_registration_failure_raises(self) -> None:
...
+        base_url = "http://localhost:8283"
+
+        def make_adapter() -> LettaAdapter:
+            return LettaAdapter(
+                config=LettaAdapterConfig(
+                    base_url=base_url, mcp=LettaMCPConfig(mode="external")
+                )
+            )
+
+        async def fake_resolve(
```
AlexanderZ-Band: Avoid over faking, see what tools we have if needed extend improve or build new

tests/adapters/test_letta_mcp.py
```diff
@@ -192,6 +193,178 @@ async def test_on_started_mcp_registration_failure_raises(self) -> None:
...
+    async def test_on_started_two_instances_resolve_distinct_user_ids(
+        self,
+    ) -> None:
+        """Two adapter instances against the same self-hosted server must
+        resolve distinct org-scoped identities — the actual collision-
+        avoidance property this fix exists to provide."""
+        base_url = "http://localhost:8283"
+
+        def make_adapter() -> LettaAdapter:
+            return LettaAdapter(
```
AlexanderZ-Band: Tests should be declarative smells of plumbing make sure to show intent and hide plumbing

tests/adapters/test_letta_mcp.py
```diff
@@ -192,6 +193,178 @@ async def test_on_started_mcp_registration_failure_raises(self) -> None:
...
+
+        async def fake_resolve(
+            *, base_url: str, agent_name: str, bearer_token: str | None
+        ) -> dict[str, str]:
+            return {"user_id": f"user-{agent_name}"}
+
+        mock_client_a = AsyncMock()
+        mock_client_a.mcp_servers.create.return_value = make_mock_mcp_server()
+        mock_client_a.mcp_servers.tools.list.return_value = []
+        mock_client_b = AsyncMock()
```
AlexanderZ-Band: Feels over mocked

tests/adapters/test_letta_orgscope.py
```diff
@@ -0,0 +1,267 @@
...
+)
+
+_BASE_URL = "http://localhost:8283"
+
+
+def _client(*, bearer_token: str | None = None) -> LettaOrgScopeClient:
+    return LettaOrgScopeClient(base_url=_BASE_URL, bearer_token=bearer_token)
+
+
+class TestFindOrCreateOrganization:
```
AlexanderZ-Band, line 30: Prefer common fixtures conftest at the correct level

tests/adapters/test_letta_orgscope.py
```diff
@@ -0,0 +1,267 @@
...
+        created_at, and GET /v1/admin/users/ silently drops the after
+        boundary filter whenever after resolves to that row (confirmed live
+        against letta/letta:0.16.8) -- the "next page" comes back as the
+        exact same unfiltered list, forever. Without stall detection this
+        hangs instead of concluding "not found".
+        """
+        stalled_page = [
+            {
+                "id": "user-default",
+                "name": "default_user",
```
AlexanderZ-Band, line 173: Keep code clean and dry

tests/integration/test_letta_live.py
```diff
@@ -126,10 +154,95 @@ async def test_adapter_self_hosted_mcp_registration() -> None:
...
+        user_id_b = adapter_b._client.default_headers["user_id"]
+        assert user_id_a != user_id_b
+
+        # B registering its own MCP server must not repoint A's tool row —
+        # the exact collision this fix exists to prevent.
+        meta_a_after = await send_message_tool_metadata(adapter_a)
+        assert meta_a_before["mcp"]["server_id"] == meta_a_after["mcp"]["server_id"]
+        assert meta_a_after["mcp"]["server_id"] != meta_b["mcp"]["server_id"]
+
+        # Fresh-org agent creation succeeds for both — proves the tools.list()
```
AlexanderZ-Band, line 222: Code should speak seems over commented or not declarative enough

tests/runtime/test_execution.py
```diff
@@ -964,7 +969,9 @@ async def test_pending_next_message_present_in_context_still_executes(
         )
 
         await ctx.start()
-        await asyncio.sleep(0.2)
+        await wait_for_condition(
```
AlexanderZ-Band, line 972: Wait for probably needs some kind of circuit breaker

src/band/adapters/letta.py
```diff
@@ -200,7 +202,33 @@ async def on_started(self, agent_name: str, agent_description: str) -> None:
...
             client_kwargs["project"] = self.config.project
+
+        # Self-hosted Letta dedupes MCP-discovered Tool rows by
+        # (name, organization_id): with no user_id header, every instance
+        # resolves to the same default org, and a second instance's MCP
+        # registration silently re-points the first instance's tool row to
+        # its own server. Scoping each instance to its own org+user closes
+        # that collision. Cloud is never scoped (org_scoped=True against it
+        # is already rejected at config construction — see LettaAdapterConfig).
+        org_scoped = (
```
AlexanderZ-Band, line 230: Single concern functions

src/band/integrations/letta/orgscope.py
AlexanderZ-Band: Good catch. Added a page-count circuit breaker (`_MAX_PAGINATION_PAGES = 10_000`) that raises a `RuntimeError` instead of looping forever if a second, undiscovered pagination quirk in Letta's admin API ever surfaces — same surface that already produced the null-created_at stall this code works around. Regression test: `test_pagination_gives_up_after_max_pages`.

tests/adapters/test_deprecation_shims.py
AlexanderZ-Band: Checked this against the rest of the file: every other test class here (Anthropic, Gemini, and the existing `TestLettaApiKeyShim`/`TestLettaMCPKwargShim` right above/below this one) does the exact same per-test local import — `letta-client` is in the `dev` extra alongside `anthropic`/`google-genai` in this same venv, so it's not the extras-gating exception either. I couldn't find an existing lazy-import helper in the codebase, or an open PR touching this file. Left it matching the surrounding convention — happy to revisit if there's a specific helper or PR you had in mind.

tests/adapters/test_letta_mcp.py
AlexanderZ-Band: You're right — this exact 4-call org+user provisioning block was duplicated verbatim in `test_letta_orgscope.py`'s `test_end_to_end_provisions_org_then_user`. Extracted `mock_org_user_provisioned()` into `lettakit.py` (the shared factories module for these tests) and both now use it.

tests/adapters/test_letta_mcp.py
AlexanderZ-Band: Deleted this test. It mocked `resolve_org_scoped_headers` entirely and juggled two separate fake Letta clients to prove headers thread through independently per instance — but there's no shared state between `LettaAdapter` instances that could actually collide there, so it wasn't protecting anything real beyond what's already covered: the actual collision-avoidance property (distinct org/user per agent name) is proven by orgscope's name-based dedup unit tests, and end-to-end by the live `test_two_instances_stay_isolated_in_shared_org`.

tests/adapters/test_letta_mcp.py
AlexanderZ-Band: Same fix as the sibling comment on this test — deleted it rather than trying to make the plumbing more declarative, since it wasn't testing anything the orgscope unit tests and the live isolation test don't already cover. See the reply on the "avoid over faking" comment for the full reasoning.

tests/adapters/test_letta_mcp.py
AlexanderZ-Band: Agreed it was heavy — deleted the test. Details in the reply on the neighboring comment on this same test.

tests/adapters/test_letta_orgscope.py
AlexanderZ-Band: `_client()` is called with a different `bearer_token` per test, so a fixture here would just be a fixture-wrapped factory — same shape, more machinery — and there's no `tests/adapters/conftest.py` yet to host it, so I left it as a plain helper. That said, you were onto something real nearby: the org+user httpx_mock provisioning block in `TestResolveOrgScopedHeaders` was duplicated with `test_letta_mcp.py`, so I extracted that into a shared `mock_org_user_provisioned()` in `lettakit.py` (see the reply on the "common fixture or tool" comment).

tests/adapters/test_letta_orgscope.py
AlexanderZ-Band: Addressed via the same fix as the "common fixture or tool" comment on `test_letta_mcp.py`: extracted the duplicated org+user provisioning wiring into `mock_org_user_provisioned()` in `lettakit.py`, used by both this file's `test_end_to_end_provisions_org_then_user` and `test_letta_mcp.py`'s `test_on_started_self_hosted_org_scoped_by_default`. The stalled-cursor test this line sits in already reuses its own `stalled_page` literal for both mocked calls, so left that part inline.

tests/integration/test_letta_live.py
AlexanderZ-Band: This is a 3-line why-comment (a non-obvious invariant: base tools must be seeded before agent create in a fresh org, or it raises) rather than what-narration, which matches this repo's comment-style convention. Left it as-is — happy to trim further if you can point at the specific line(s) that feel redundant.

tests/runtime/test_execution.py
AlexanderZ-Band: `wait_for_condition` (`tests/runtime/conftest.py`) already has a `timeout` deadline and calls `pytest.fail("Timed out waiting for condition")` on expiry, so there's already a circuit breaker here. Let me know if you saw a call site that skips the default timeout — happy to take another look.

src/band/adapters/letta.py
AlexanderZ-Band: Agreed — `on_started` was doing prompt-building, client construction (import/kwargs/org-scope resolution/base-tool seeding), and MCP wiring all inline. Extracted the client-construction piece into a new `_build_client(agent_name)` helper; `on_started` now just orchestrates prompt → client → MCP.

src/band/integrations/letta/config.py
```diff
@@ -14,6 +15,25 @@
...
+
+def is_letta_cloud_url(base_url: str) -> bool:
+    """Whether ``base_url`` points at Letta Cloud rather than a self-hosted server.
+
+    Compares hostnames, not raw strings, so casing and a trailing slash never
+    cause a Cloud URL to be misclassified as self-hosted; ``.strip()`` closes
+    the same gap for incidental whitespace (e.g. an unquoted .env value).
+    """
+    try:
+        return urlsplit(base_url.strip()).hostname == _LETTA_CLOUD_HOSTNAME
```
amit-gazal-band: `urlsplit()` only parses a hostname when a scheme/authority is present. If `LETTA_BASE_URL` is set without a scheme (e.g. `api.letta.com`, a plausible config typo), `urlsplit(base_url.strip()).hostname` is `None`, which never equals `_LETTA_CLOUD_HOSTNAME` — so this misclassifies actual Letta Cloud as self-hosted. `org_scoped` then auto-enables and the Cloud+org_scoped construction guard never fires, so the adapter ends up calling self-hosted-only admin endpoints against Letta Cloud and fails deep inside `on_started` instead of loudly at construction. Worth normalizing the URL (e.g. defaulting/adding a scheme, or using `urlsplit` with a scheme fallback) before comparing hostnames.

src/band/integrations/letta/orgscope.py
```diff
@@ -0,0 +1,162 @@
...
+        bearer_token: str | None,
+        timeout_s: float = _DEFAULT_TIMEOUT_S,
+    ) -> None:
+        self._base_url = base_url.rstrip("/")
+        self._headers = (
+            {"Authorization": f"Bearer {bearer_token}"} if bearer_token else {}
+        )
+        self._timeout_s = timeout_s
+
+    async def find_or_create_organization(self, name: str) -> str:
```
amit-gazal-band, line 53: `find_or_create_organization` and `find_or_create_user` share the identical find-or-create-with-logging shape (GET/paginated-find, return if matched, else POST + log), differing only in the match predicate and create payload. A generic `_find_or_create(path, match, create_payload, entity_label)` helper would collapse both to one-line callers and avoid the two copies drifting (e.g. one getting a retry/timeout tweak the other doesn't). Not blocking, just a simplification opportunity.

src/band/integrations/letta/orgscope.py
```diff
@@ -0,0 +1,162 @@
...
+            user = response.json()
+            logger.info(
+                "Created Letta user %r (id=%s) in organization %s",
+                name,
+                user["id"],
+                organization_id,
+            )
+            return user["id"]
+
+    def _client(self) -> httpx.AsyncClient:
```
amit-gazal-band: `_client()` opens a brand-new `httpx.AsyncClient` (fresh connection) on every call. `resolve_org_scoped_headers` calls `find_or_create_organization` then `find_or_create_user` back-to-back, so every org-scoped adapter startup pays double connection-setup overhead. Consider holding one client for the `LettaOrgScopeClient`'s lifetime instead (mirrors the pattern in `band.adapters.opencode`'s HTTP client wrapper).

tests/adapters/test_letta_orgscope.py
```diff
@@ -0,0 +1,278 @@
+"""Unit tests for Letta self-hosted org/user provisioning (INT-985).
```
amit-gazal-band: CLAUDE.md: "Never put issue-tracker references in code — no Linear issue IDs (e.g. `INT-123`), Linear URLs, or ticket numbers in comments, docstrings, or strings." This docstring embeds `INT-985` directly; same issue also appears in `tests/integration/test_letta_live.py:161` ("Regression test for the org+user scoping fix (INT-985)") and `tests/e2e/baseline/smoke/matrix/test_rehydration_partial.py:46` ("INT-985 fixed this by provisioning..."). Could rephrase all three to state the rationale in plain terms instead of the ticket ID.

amit-gazal-band (dismissed):
Approving — solid fix for the MCP org/user isolation issue. Left a few inline notes: a scheme-less base_url edge case in Cloud detection, a Linear ticket ID that snuck into code comments/docstrings (repo convention keeps those out of code), and two optional cleanup ideas (duplicated find-or-create shape, per-call AsyncClient re-creation).

src/band/integrations/letta/config.py
AlexanderZ-Band: Good catch — a scheme-less `LETTA_BASE_URL` (e.g. `api.letta.com`) made `urlsplit(...).hostname` come back `None`, which never equals the Cloud hostname. Normalized by prepending `//` when no `://` is present before splitting, so a scheme-less value still parses its hostname correctly. Added `"api.letta.com"` to `test_letta_org_scoped_and_cloud_conflict`'s parametrize list in `test_deprecation_shims.py` as a regression guard — it now raises `BandConfigError` instead of silently misclassifying Cloud as self-hosted.

src/band/integrations/letta/orgscope.py
AlexanderZ-Band: Extracted a shared `_find_or_create(path, match, payload, log_label)` helper — both `find_or_create_organization` and `find_or_create_user` are now one-line callers that differ only in the match predicate, POST payload, and log label.

src/band/integrations/letta/orgscope.py
AlexanderZ-Band: Agreed — `LettaOrgScopeClient` now opens its `httpx.AsyncClient` once in `__init__` and holds it for the instance's lifetime (mirrors `HttpOpencodeClient`'s pattern), with an explicit `aclose()`. `resolve_org_scoped_headers` (the sole call site) closes it in a `finally` after both provisioning calls, so the two sequential admin calls now share one connection instead of paying setup twice.

tests/adapters/test_letta_orgscope.py
AlexanderZ-Band: Fixed — rephrased all three in plain terms with no ticket ID: this file's module docstring (which also had a stale "per-call `httpx.AsyncClient`" claim, now corrected to match the held-client change from the other comment), `test_letta_live.py:161`, and `test_rehydration_partial.py:46`.

amit-gazal-band (approved):

PR #623, opened 2026-09-09
refactor: Extract subscription management out of BandLink into its own module

AlexanderZ-Band:
## Summary
- Extracts room/agent-topic subscription tracking, join/leave orchestration, and reconnect reconciliation out of `BandLink` into a new `SubscriptionManager` class (`src/band/platform/subscriptions.py`), mirroring the `MessageLifecycle` precedent from INT-1301.
- `BandLink` now composes a `SubscriptionManager` instance and keeps thin delegate methods with unchanged public signatures; the `on_*` event-dispatch callbacks stay owned by `BandLink` and are passed in per call.
- Consolidates the duplicated `mock_ws_client` test fixture (`test_link.py` vs. `test_link_contacts.py`) into `tests/platform/conftest.py` as a preceding, independently-verified commit.

Implements the finalized plan attached to [INT-1319](https://linear.app/thenvoi/issue/INT-1319/extract-subscription-management-out-of-bandlink-into-its-own-module) exactly as written — no behavior change.

## Test plan
- [x] `uv run ruff check .`
- [x] `uv run ruff format .`
- [x] `uv run pyrefly check`
- [x] `uv run --all-packages pytest tests/ --ignore=tests/integration/ --ignore=tests/e2e/ -v` (5589 passed, 146 skipped)
- [x] `tests/platform/` and `tests/mcp/test_import_boundary.py` specifically confirmed green

https://claude.ai/code/session_01DUqQ9My1w5BjmRRaciJxLs

### DIFF
--- a/src/band/platform/link.py
+++ b/src/band/platform/link.py
@@ -246,15 +246,7 @@ class BandLink:
         )
 
     async def subscribe_room(self, room_id: str) -> None:
-        """Subscribe to room messages and participants.
-
-        Blocked (a no-op, logged) while ``room_id`` is in
-        ``SubscriptionManager``'s reconciliation set — a prior
-        cancelled/ambiguous attempt's outcome is unresolved and must not be
-        retried on the same socket. Stays blocked until the next reconnect
-        drains it (see ``_drain_reconciliation``); see the design doc for
-        why.
-        """
+        """Subscribe to room messages and participants."""
         if not self._ws:
             raise RuntimeError("Not connected")
         await self._subscriptions_manager.subscribe_room(
@@ -290,12 +282,7 @@ class BandLink:
         await self._subscriptions_manager.unsubscribe_room(self._ws, room_id)
 
     async def unsubscribe_agent_contacts(self) -> None:
-        """Unsubscribe from agent contacts channel.
-
-        A true no-op when the topic was never joined — the tracker's
-        ``leave_agent_topic`` returns ``None`` in that case rather than
-        issuing a leave the transport would just reject.
-        """
+        """Unsubscribe from agent contacts channel."""
         if not self._ws:
             return
         await self._subscriptions_manager.unsubscribe_agent_contacts(
--- a/src/band/platform/link.py
+++ b/src/band/platform/link.py
@@ -259,12 +259,7 @@ class BandLink:
         )
 
     async def subscribe_agent_contacts(self, agent_id: str) -> None:
-        """
-        Subscribe to agent contact events.
-
-        Events: contact_request_received, contact_request_updated,
-                contact_added, contact_removed
-        """
+        """Subscribe to agent contact events."""
         if not self._ws:
             raise RuntimeError("Not connected")
         await self._subscriptions_manager.subscribe_agent_contacts(
--- a/src/band/platform/link.py
+++ b/src/band/platform/link.py
@@ -17,16 +17,9 @@ from band.config.settings import DEFAULT_REST_URL, DEFAULT_WS_URL
 from band.client.streaming import WebSocketClient, WebSocketDisconnectReason
 from band.core.types import PlatformConnection
 from band.platform.message_lifecycle import MessageLifecycle
+from band.platform.subscriptions import SubscriptionManager
 from band.runtime.types import PlatformMessage
-from band_sdk_core import (
-    AgentTopicKind,
-    LeaveOutcome,
-    RoomSubscribeResult,
-    SessionState,
-    SubscriptionTracker,
-    chat_room_topic,
-    room_participants_topic,
-)
+from band_sdk_core import AgentTopicKind, SessionState
 
 from band.platform.event import (
     MessageEvent,
@@ -113,13 +106,11 @@ class BandLink:
         # concurrent connect() sees it even mid-handshake (see connect()).
         self._connecting = False
 
-        # Subscription tracking (band_sdk_core.SubscriptionTracker) plus local
-        # bookkeeping for claims whose real-world outcome is ambiguous (a
-        # cancelled join, a failed rollback, a non-clean leave) — drained only
-        # at the next reconnect boundary, see _drain_reconciliation.
-        self._subscriptions = SubscriptionTracker()
-        self._rooms_needing_reconciliation: set[str] = set()
-        self._agent_topics_needing_reconciliation: set[str] = set()
+        # Room/agent-topic subscription tracking and reconnect reconciliation --
+        # lives in its own class since it needs the live WebSocketClient, unlike
+        # MessageLifecycle. `current_ws` is a getter, never a cached value, so
+        # BandLink stays the single owner of `_ws` across reconnects.
+        self._subscriptions_manager = SubscriptionManager(current_ws=lambda: self._ws)
 
         # Event queue for async iteration
         self._event_queue: asyncio.Queue[PlatformEvent] = asyncio.Queue(maxsize=1000)
@@ -223,9 +214,7 @@ class BandLink:
         await self._ws.__aexit__(None, None, None)
         self._ws = None
         self._is_connected = False
-        self._subscriptions.end_session()
-        self._rooms_needing_reconciliation.clear()
-        self._agent_topics_needing_reconciliation.clear()
+        self._subscriptions_manager.end_session()
         logger.info("Disconnected from platform")
 
     async def run_forever(self) -> None:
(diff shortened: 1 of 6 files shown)


src/band/platform/link.py
```diff
@@ -113,13 +106,11 @@ def __init__(
...
-        # cancelled join, a failed rollback, a non-clean leave) — drained only
-        # at the next reconnect boundary, see _drain_reconciliation.
-        self._subscriptions = SubscriptionTracker()
-        self._rooms_needing_reconciliation: set[str] = set()
-        self._agent_topics_needing_reconciliation: set[str] = set()
+        # Room/agent-topic subscription tracking and reconnect reconciliation --
+        # lives in its own class since it needs the live WebSocketClient, unlike
+        # MessageLifecycle. `current_ws` is a getter, never a cached value, so
+        # BandLink stays the single owner of `_ws` across reconnects.
+        self._subscriptions_manager = SubscriptionManager(current_ws=lambda: self._ws)
```
AlexanderZ-Band, line 113: nit: `MessageLifecycle` is composed as `self._messages` (line 100), but `SubscriptionManager` is composed as `self._subscriptions_manager` here — a different naming idiom for the same composition pattern (a sub-object holding live-client-per-call state, per both classes' own docstrings).

src/band/platform/link.py
```diff
@@ -245,168 +234,37 @@ async def _join_agent_control_channel(self, ws: WebSocketClient) -> None:
...
-            ws,
+        await self._subscriptions_manager.subscribe_agent_rooms(
+            self._ws,
+            agent_id,
+            on_room_added=self._on_room_added,
+            on_room_removed=self._on_room_removed,
         )
 
     async def subscribe_room(self, room_id: str) -> None:
         """Subscribe to room messages and participants.
```
AlexanderZ-Band: nit: This docstring duplicates the full reconciliation-blocking explanation that now actually lives on `SubscriptionManager.subscribe_room` (subscriptions.py:138-144) — near-identical prose, just the attribute/method names swapped. The same duplication exists for `unsubscribe_agent_contacts` (link.py:292-298 vs subscriptions.py:334-339). Now that these are pure delegating wrappers, the two copies have to be kept in sync by hand. The sibling wrappers `subscribe_agent_rooms`, `unsubscribe_room`, and `is_room_subscribed` were left with only a one-line docstring or none, suggesting this wasn't a deliberate choice.

tests/platform/test_link.py
```diff
@@ -655,8 +595,12 @@ async def test_drain_reconciliation_bails_on_ws_swap_mid_drain(
 
         link = BandLink(agent_id="agent-123", api_key="test-key")
         await link.connect()
-        link._rooms_needing_reconciliation.update({"room-1", "room-2"})
-        link._agent_topics_needing_reconciliation.add("agent_rooms:agent-123")
+        link._subscriptions_manager._rooms_needing_reconciliation.update(
```
AlexanderZ-Band, line 598: suggestion: This test reaches `link._subscriptions_manager._rooms_needing_reconciliation` / `._agent_topics_needing_reconciliation` directly (also lines 601, 617, 623) — one level deeper than before the extraction (`link._rooms_needing_reconciliation`). This is white-box testing of an internal race condition (which room's leave pair completes before a mid-drain ws swap), so the deeper reach may be a reasonable trade-off rather than something to fix.

tests/platform/test_link.py
```diff
@@ -791,7 +737,10 @@ async def test_agent_topic_rejoin_failure_is_detected_and_drained_on_reconnect(
         mock_ws_client.joined_topics.return_value = AllTopicsExcept({topic})
         await link._on_reconnected()
         mock_ws_client.leave_agent_rooms_channel.assert_called_once_with("agent-123")
-        assert link._subscriptions.agent_topic_status(topic) == AgentTopicStatus.Absent
+        assert (
+            link._subscriptions_manager._subscriptions.agent_topic_status(topic)
```
AlexanderZ-Band, line 741: nit: `link._subscriptions_manager._subscriptions.agent_topic_status(...)` now reaches through two private attributes where it previously reached through one (`link._subscriptions.agent_topic_status(...)`). The same pattern recurs at tests/platform/test_link_contacts.py:203 and tests/platform/test_link_real_wire.py:122,133 (`is_agent_topic_joined`).

AlexanderZ-Band (commented):
Automated pass (my-code-review, 6 angle-groups: Style & Hygiene, Design & Reuse, Runtime Safety & Observability, Test Quality, Simplification, Logical Bugs). This is a faithful, plan-driven mechanical extraction — Runtime Safety, Simplification, and Logical Bugs groups found nothing (no over-engineering, no proven behavioral regression, concurrency guarantee re-verified against the existing ws-swap-mid-drain test). 4 non-blocking nits/suggestions below from Style/Design/Test-Quality.

One additional note that doesn't land inside a diff hunk so can't be an inline comment: `tests/platform/test_link_real_wire.py:108-112`'s docstring still reads "...so this reaches into `link._subscriptions`...", but the actual call two lines below (line 122) now reads `link._subscriptions_manager._subscriptions.is_agent_topic_joined(...)` — the docstring names a path that no longer exists after the extraction.

src/band/platform/link.py
AlexanderZ-Band: Fixed in 38211502: trimmed both BandLink wrapper docstrings (subscribe_room here, and unsubscribe_agent_contacts at link.py:292) to a one-liner, matching the sibling wrappers (subscribe_agent_rooms, unsubscribe_room, is_room_subscribed). The full reconciliation-blocking explanation now lives only on SubscriptionManager's own methods (subscriptions.py) — single source of truth.

src/band/platform/link.py
AlexanderZ-Band: Declining this one. `self._subscriptions_manager` is the exact name specified in INT-1319's finalized, pre-verified implementation plan, which this PR follows deliberately rather than re-deriving. There's also a concrete reason to keep it distinct from `_messages`: renaming to `self._subscriptions` would create a stutter against `SubscriptionManager`'s own internal `_subscriptions` (the SubscriptionTracker) — call sites already read `link._subscriptions_manager._subscriptions`, and collapsing the outer name would make that `link._subscriptions._subscriptions`, which is worse, not better. Leaving open (not fixed) rather than resolving.

tests/platform/test_link.py
AlexanderZ-Band: Declining this one, agreeing with the reviewer's own caveat: this test deliberately exercises an internal race (which room's leave pair completes before a mid-drain ws swap), so reaching into the reconciliation-pending sets directly is appropriate white-box testing, not a design smell to fix. No accessor added. Leaving open (not fixed) rather than resolving.

tests/platform/test_link.py
AlexanderZ-Band: Declining this one. Adding a public read accessor for agent-topic status was explicitly considered and rejected in INT-1319's finalized implementation plan, as unrequested scope creep beyond a pure structural move with no behavior change — the reach-in pattern (and its justification: no real caller besides these tests) predates this PR and is orthogonal to this ticket's scope, per test_link_real_wire.py's own docstring. This PR only makes an existing, already-justified reach one level deeper, consistently at all 3 sites. Leaving open (not fixed) rather than resolving.

src/band/platform/link.py
```diff
@@ -245,168 +234,29 @@ async def _join_agent_control_channel(self, ws: WebSocketClient) -> None:
...
+            self._ws,
+            room_id,
+            on_message_created=lambda msg: self._on_message_created(room_id, msg),
+            on_participant_added=lambda p: self._on_participant_added(room_id, p),
+            on_participant_removed=lambda p: self._on_participant_removed(room_id, p),
+            on_room_deleted=lambda p: self._on_room_deleted(room_id, p),
+        )
 
     async def subscribe_agent_contacts(self, agent_id: str) -> None:
         """
```
AlexanderZ-Band: nit: This wrapper still carries the full multi-line docstring duplicated verbatim from `SubscriptionManager.subscribe_agent_contacts` (subscriptions.py:229-234) — the same class of issue already fixed for subscribe_room and unsubscribe_agent_contacts in this PR, but this sibling was missed.

src/band/platform/subscriptions.py
```diff
@@ -0,0 +1,499 @@
...
+involved.
+"""
+
+from __future__ import annotations
+
+import logging
+from collections.abc import Awaitable, Callable
+from typing import TYPE_CHECKING
+
+from band.client.streaming import WebSocketClient
```
AlexanderZ-Band, line 12: nit: `WebSocketClient` is imported at module level here, but every use in this file is annotation-only (the `current_ws` callable's return type, and every `ws: WebSocketClient` parameter) — nothing instantiates or isinstance-checks it. The file already TYPE_CHECKING-guards the payload types from the same `band.client.streaming` module for exactly this reason, so this one import is inconsistent with the file's own stated convention.

AlexanderZ-Band (commented):
Cycle 2 of the automated pass (my-code-review). 4 of 6 groups (B, C, D, E) came back clean — nothing new since cycle 1. Group A found 2 genuinely new nits below.

src/band/platform/link.py
AlexanderZ-Band: Fixed in a4158beb: trimmed this to a one-liner, matching the pattern already applied to subscribe_room and unsubscribe_agent_contacts. The full explanation now lives only on SubscriptionManager.subscribe_agent_contacts (subscriptions.py).

src/band/platform/subscriptions.py
AlexanderZ-Band: Declining this one. Checked the precedent it's modeled on: message_lifecycle.py imports its own primary collaborator type, AsyncRestClient, the same way -- at module level, not TYPE_CHECKING-gated -- even though MessageLifecycle's own methods use it purely as a passed-in-instance parameter, same as WebSocketClient here. SubscriptionManager is explicitly designed to mirror MessageLifecycle's shape, so keeping this import consistent with that sibling class (rather than diverging from it) is the more single-source-of-truth-consistent choice, even though gating it under TYPE_CHECKING would work in isolation. Leaving open (not fixed) rather than resolving.

amit-gazal-band (approved):

PR #624, opened 2026-09-09
fix: retry CrewAI's first-call empty completion before failing

AlexanderZ-Band:
## Summary
- CrewAI's native OpenAI tool-calling path can return an empty completion (`finish_reason='stop'`, no content, no tool_calls) on a turn's very first LLM call, before any tool has run — reproduced 8/8 times across local live runs, root-caused and filed upstream at [crewAIInc/crewAI#7349](https://github.com/crewAIInc/crewAI/issues/7349).
- That case is indistinguishable from a genuine provider outage from the caller's side, so `CrewAIAdapter` always failed the whole delivery — even though nothing ran yet this turn, so nothing is duplicated by retrying.
- Adds one bounded, in-process retry (`_EMPTY_RESPONSE_FIRST_CALL_RETRIES = 1`) via a new `_kickoff_with_empty_response_retry` helper, only for `_is_empty_llm_response(e) and not reply_tracker.any_tool_ran`. Any other failure, or exhausting the retry, still reports the error and raises exactly as before.
- No change to the existing `any_tool_ran` swallow logic (PR #621).

Traced from last night's failed nightly E2E run ([34307490909](https://github.com/band-ai/band-sdk-python/actions/runs/34307490909)) → root-caused → planned and independently fact-verified → implemented, per [INT-1427](https://linear.app/thenvoi/issue/INT-1427/crewai-adapter-still-fails-a-turn-when-its-very-first-llm-call-comes) (plan attached to the ticket).

## Test plan
- [x] `uv run --extra dev-crewai pytest tests/adapters/test_crewai_adapter.py -v` — 82 passed
  - Updated `test_empty_answer_with_no_tool_call_still_raises` to assert the retry happens (`kickoff_async.call_count == 2`) before the final raise
  - Added `test_empty_first_call_recovers_on_retry`: first call empty, second call replies — turn completes, no error event
- [x] `uv run pytest tests/ --ignore=tests/integration/ --ignore=tests/e2e/` — 5590 passed, 146 skipped

https://claude.ai/code/session_01Fw8AZwnoa3Fen6HBhHEQ7H

### DIFF
--- a/src/band/adapters/crewai.py
+++ b/src/band/adapters/crewai.py
@@ -491,11 +491,18 @@ class CrewAIAdapter(SimpleAdapter[CrewAIMessages]):
             )
             try:
                 return await agent.kickoff_async(prompt)
-            except Exception:
-                logger.warning(
-                    "Room %s: CrewAI's retry also came back empty; giving up",
-                    room_id,
-                )
+            except Exception as retry_exc:
+                if _is_empty_llm_response(retry_exc):
+                    logger.warning(
+                        "Room %s: CrewAI's retry also came back empty; giving up",
+                        room_id,
+                    )
+                else:
+                    logger.warning(
+                        "Room %s: CrewAI's retry failed with a different error: %s",
+                        room_id,
+                        retry_exc,
+                    )
                 raise
 
     async def _report_error(self, tools: AgentToolsProtocol, error: str) -> None:
--- a/src/band/adapters/crewai.py
+++ b/src/band/adapters/crewai.py
@@ -56,6 +56,11 @@ _reply_tracker_var: ContextVar[ReplyTracker | None] = ContextVar(
 # is the only discriminator. One definition, matched here and faked in tests.
 EMPTY_LLM_RESPONSE_MARKER = "Invalid response from LLM call"
 
+# Bound on the in-process retry for a turn's very first LLM call coming back
+# empty (see _kickoff_with_empty_response_retry) -- one immediate retry, no
+# backoff, since nothing has run yet this turn and so nothing is duplicated.
+_EMPTY_RESPONSE_FIRST_CALL_RETRIES = 1
+
 
 def _is_empty_llm_response(exc: Exception) -> bool:
     """Whether ``exc`` is CrewAI reporting that an LLM call came back empty.
@@ -405,7 +410,9 @@ class CrewAIAdapter(SimpleAdapter[CrewAIMessages]):
             # "already handled" / "act on this now" markers, matches what
             # CrewAI actually does with the input either way.
             prompt = "\n\n".join(sections)
-            result = await self._crewai_agent.kickoff_async(prompt)
+            result = await self._kickoff_with_empty_response_retry(
+                self._crewai_agent, prompt, reply_tracker, room_id
+            )
 
         except Exception as e:
             # An empty response is benign only once some tool ran this turn --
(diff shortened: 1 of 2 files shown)


tests/adapters/test_crewai_adapter.py
```diff
@@ -741,6 +742,52 @@ async def test_empty_answer_with_no_tool_call_still_raises(
             )
 
         mock_tools.send_event.assert_awaited_once()
+        assert mock_crewai_agent.kickoff_async.call_count == 2
```
AlexanderZ-Band, line 747: nit: `call_count == 2` hardcodes the retry bound as a bare literal instead of deriving it from `_EMPTY_RESPONSE_FIRST_CALL_RETRIES`, the constant this PR defines in `src/band/adapters/crewai.py`. The identical literal is repeated at line 790. A future change to the constant silently desyncs these two expected call counts from the value that actually governs the retry.

src/band/adapters/crewai.py
```diff
@@ -56,6 +56,11 @@
 # is the only discriminator. One definition, matched here and faked in tests.
 EMPTY_LLM_RESPONSE_MARKER = "Invalid response from LLM call"
 
+# Bound on the in-process retry for a turn's very first LLM call coming back
+# empty (see _kickoff_with_empty_response_retry) -- one immediate retry, no
+# backoff, since nothing has run yet this turn and so nothing is duplicated.
```
AlexanderZ-Band: nit: This clause repeats the rationale the `_kickoff_with_empty_response_retry` docstring already gives in full a few lines down ("has run no tool yet, so there is nothing to duplicate: one immediate retry absorbs..."), which this very comment already points to via "(see _kickoff_with_empty_response_retry)". The one mechanism's why is now spelled out twice.

src/band/adapters/crewai.py
```diff
@@ -461,6 +468,43 @@ async def on_cleanup(self, room_id: str) -> None:
...
+        """Retry a first-call empty completion once before giving up.
+
+        CrewAI raises the identical ValueError whether the model correctly has
+        nothing left to say (fine once a tool has run -- handled by the
+        caller) or the turn's very first call came back empty. The second case
+        has run no tool yet, so there is nothing to duplicate: one immediate
+        retry absorbs a single-call fluke instead of failing the whole
+        delivery and leaving CrewAI to improvise on a cold redelivery.
+        """
+        attempt = 0
```
AlexanderZ-Band: suggestion: `_kickoff_with_empty_response_retry` uses a `while True` loop with a manually incremented `attempt` counter -- machinery sized for a general N-retry problem -- but the bound is fixed at exactly 1 (`_EMPTY_RESPONSE_FIRST_CALL_RETRIES = 1`) and nothing in this ticket's scope calls for more. A single inline try/except that retries once achieves the same behavior with no loop and no counter to reason about. Applying this simplification in a follow-up commit on this PR.

AlexanderZ-Band (commented):
Automated multi-lens review (my-code-review, 6 groups: Style & Hygiene, Design & Reuse, Runtime Safety, Test Quality, Simplification, Logical Bugs; `--fix`).

3 inline comments below. One additional finding couldn't be placed inline (it's on pre-existing code outside this diff's hunks, per GitHub's line-anchoring rules):

suggestion: `src/band/adapters/crewai.py:423` (and the debug log at `:428`) -- when the retry is exhausted, `logger.error("Error processing message: %s", e, exc_info=True)` logs the same text CrewAI would raise on a plain first-call failure, with no attempt count. The only other evidence a retry happened is the `logger.info` at line 500, which fires before the attempt and isn't tied to the eventual outcome -- so a log pipeline filtered to WARNING/ERROR can't tell, from line 423 alone, whether a delivery failed after one attempt or after the retry fired and failed again.

No blocking findings. Logical Bugs (Group F) built 3 targeted repros against the exact head SHA (retry-cap off-by-one, tool-runs-mid-first-call, exception-type confusion) -- none reproduced. Design & Reuse (Group B) confirmed the retry-loop shape matches this repo's existing precedent for small bounded retries (gemini.py, codex/rpc_base.py, crewai_flow.py).

tests/adapters/test_crewai_adapter.py
AlexanderZ-Band: Fixed differently than suggested: the governing constant (`_EMPTY_RESPONSE_FIRST_CALL_RETRIES`) is gone now that the retry is a single inline try/except instead of a loop bounded by a comparable constant (see the sibling over-engineering thread) -- there's nothing left to derive `2` from. Kept the literal with a short comment explaining why it's 2 (first call + the one retry) instead of deriving from a no-longer-existent single source of truth.

src/band/adapters/crewai.py
AlexanderZ-Band: Fixed by removing this comment block entirely along with the constant it described -- the retry is now a single inline try/except (see the sibling thread), so there's no separate bound value needing its own comment; the docstring is now the only place stating the rationale.

src/band/adapters/crewai.py
AlexanderZ-Band: Agreed and applied -- replaced the while/attempt-counter loop with a single inline try/except that retries once. Also removed the now-unused `_EMPTY_RESPONSE_FIRST_CALL_RETRIES` constant, and added a distinct `logger.warning` at the point the retry itself fails (this also addresses the logging-gap finding in the review body -- that ambiguity is gone now that the retry's own failure is logged separately from the outer handler).

amit-gazal-band (approved):

PR #629, opened 2026-09-12
feat: isolate coding agent processes by room

AlexanderZ-Band:
## Summary
- assign each room an absolute workspace via `workspace_for_room(room_id)`, defaulting to an isolated `./.band-workspaces/<room-id>` directory when unset
- lazily create and own one local stdio coding-agent process and protocol state per room
- pass the resolved room workspace to Codex `thread/start` and ACP subprocess startup
- enforce one live room per resolved workspace, including symlink-equivalent paths
- make workspace/runtime creation atomic; preserve model overrides, release failed-start reservations, and make multi-room cleanup best-effort
- migrate shipped stdio callers and update ACP/Codex documentation and examples
- reject remote/shared/injected transports that cannot preserve room isolation

## Validation
- focused workspace/ACP/Copilot regression suite: 139 passed
- non-MCP unit/integration suite: 5,469 passed, 146 skipped
- ruff check and format check: passed
- compileall for changed Python trees: passed
- default full pytest/pyrefly remain blocked by the checkout pre-existing `band_mcp` package import/configuration boundary; no changed file is involved

## Follow-up
Room-safe remote ACP/TCP and Codex WebSocket transport support is tracked in [INT-1492](https://linear.app/thenvoi/issue/INT-1492/restore-room-safe-remote-acp-and-codex-transport-support).

### DIFF
--- a/docs/adapters/codex.md
+++ b/docs/adapters/codex.md
@@ -39,15 +39,8 @@ adapter = CodexAdapter(
     config=CodexAdapterConfig(model="gpt-5.5"),
 )
 
-agent = Agent.create(
-    adapter=adapter,
-    agent_id="your-agent-uuid",
-    api_key="your-band-api-key",
-    ws_url="wss://app.band.ai/api/v1/socket/websocket",
-    rest_url="https://app.band.ai",
-)
-
-asyncio.run(agent.run())
+async with Agent.from_config("my_agent", adapter=adapter) as agent:
+    await agent.run_forever()
 ```
 
 ## Where Parameters Go
--- a/docs/adapters/codex.md
+++ b/docs/adapters/codex.md
@@ -45,7 +45,7 @@ from band.adapters.codex import CodexAdapter, CodexAdapterConfig
 
 adapter = CodexAdapter(
     config=CodexAdapterConfig(
-        cwd=os.getcwd(),
+        workspace_for_room=lambda room_id: os.path.join("/workspaces", room_id),
         model="gpt-5.5",
     ),
 )
--- a/docs/adapters/codex.md
+++ b/docs/adapters/codex.md
@@ -31,16 +31,18 @@ Credentials for Band can also be loaded from `agent_config.yaml` with `Agent.fro
 ## Quick Start
 
 ```python
-import asyncio
 from band import Agent
 from band.adapters.codex import CodexAdapter, CodexAdapterConfig
 
 adapter = CodexAdapter(
     config=CodexAdapterConfig(model="gpt-5.5"),
 )
-
-async with Agent.from_config("my_agent", adapter=adapter) as agent:
-    await agent.run_forever()
+agent = Agent.create(
+    adapter=adapter,
+    agent_id="your-agent-uuid",
+    api_key="your-band-api-key",
+)
+assert adapter.config.model == "gpt-5.5"
 ```
 
 ## Where Parameters Go
--- a/docs/adapters/codex.md
+++ b/docs/adapters/codex.md
@@ -1,6 +1,6 @@
 # Codex Adapter
 
-[OpenAI Codex](https://openai.com/codex) is a coding agent runtime that can inspect files, edit files, run commands, and manage approval workflows. The Band Codex adapter connects a Codex process to Band rooms over stdio or WebSocket so it can take part in conversations as a coding collaborator.
+[OpenAI Codex](https://openai.com/codex) is a coding agent runtime that can inspect files, edit files, run commands, and manage approval workflows. The Band Codex adapter connects one room-owned Codex process to each Band room over stdio.
 
 Use this adapter when you want an OpenAI-powered coding agent with configurable sandboxing, approval commands, command/file-change telemetry, reasoning visibility, and task lifecycle events. Use the [Claude SDK adapter](claude_sdk.md) for Claude Code based coding agents, the [Anthropic adapter](anthropic.md) for direct Claude API chat/tool agents, or the [LangGraph adapter](langgraph.md) for custom graph workflows.
 
@@ -25,12 +25,6 @@ You need two credentials or auth contexts:
 
 - A Band platform API key for `Agent.create(api_key=...)`.
 - Codex authentication for the Codex process. Use `codex login`, or set `OPENAI_API_KEY` if that is how your Codex environment is configured.
-
-For `transport="ws"`, start the Codex app server separately:
-
-```bash
-codex app-server --listen ws://127.0.0.1:8765
-```
 
 Credentials for Band can also be loaded from `agent_config.yaml` with `Agent.from_config("my_agent", adapter=adapter)`.
 
@@ -65,8 +59,8 @@ asyncio.run(agent.run())
 
 Codex has three setup layers:
 
-- `CodexAdapterConfig(...)` configures the Codex runtime: transport, model, working directory, sandbox, approval behavior, prompts, context injection, and streaming/telemetry detail.
-- `CodexAdapter(...)` wraps that runtime config for Band and adds adapter-level settings: feature flags, custom tools, history conversion, and advanced client injection.
+- `CodexAdapterConfig(...)` configures the Codex runtime: a room workspace resolver, model, sandbox, approval behavior, prompts, context injection, and streaming/telemetry detail.
+- `CodexAdapter(...)` wraps that runtime config for Band and adds adapter-level settings: feature flags, custom tools, and history conversion.
 - `Agent.create(...)` connects the configured adapter to Band. Use it for the Band agent identity, Band API key, platform URLs, session settings, contact-event handling, callbacks, and preprocessing.
 
 Codex authentication is handled by `codex login`, `OPENAI_API_KEY`, or the Codex process environment. `Agent.create(api_key=...)` is only the Band platform key.
@@ -104,7 +98,10 @@ from band.core.types import Emit
 from band.adapters.codex import CodexAdapter, CodexAdapterConfig
 
 adapter = CodexAdapter(
-    config=CodexAdapterConfig(cwd="/repo", sandbox="workspace-write"),
+    config=CodexAdapterConfig(
+        workspace_for_room=lambda room_id: f"/workspaces/{room_id}",
+        sandbox="workspace-write",
+    ),
     emit=Emit.TOOL_CALLS | Emit.TASK_EVENTS,
 )
 ```
@@ -115,12 +112,11 @@ Pass these to `CodexAdapterConfig(...)`:
 
 | Parameter | Type | Default | Description |
 |-----------|------|---------|-------------|
-| `transport` | `"stdio" \| "ws"` | `"stdio"` | How the adapter connects to Codex. Use `"stdio"` to spawn a process, or `"ws"` to connect to `codex app-server`. |
+| `workspace_for_room` | `Callable[[str], str]` | required | Returns the absolute workspace for a room. Every live room must receive a distinct workspace. |
 | `model` | `str \| None` | `None` | Model to use. When unset, the adapter asks Codex for visible models and uses the first visible model, or the adapter default if discovery fails or returns no usable model. |
 | `reasoning_effort` | `"none" \| "minimal" \| "low" \| "medium" \| "high" \| "xhigh" \| None` | `None` | Reasoning effort for models that support it. |
 | `reasoning_summary` | `"auto" \| "concise" \| "detailed" \| "none" \| None` | `None` | How Codex summarizes reasoning in responses. |
 | `personality` | `"friendly" \| "pragmatic" \| "none"` | `"pragmatic"` | Codex response style. |
-| `cwd` | `str \| None` | `None` | Working directory for Codex sessions. |
 | `turn_timeout_s` | `float` | `180.0` | Maximum seconds to wait for one Codex turn. |
 
 ### Safety, Sandbox, and Approvals
@@ -170,7 +166,6 @@ Pass these to `CodexAdapterConfig(...)`:
 |-----------|------|---------|-------------|
 | `codex_command` | `tuple[str, ...] \| None` | `None` | Custom command used to launch Codex for stdio transport. |
 | `codex_env` | `dict[str, str] \| None` | `None` | Extra environment variables for the Codex process. |
-| `codex_ws_url` | `str` | `"ws://127.0.0.1:8765"` | WebSocket URL for `transport="ws"`. |
 | `experimental_api` | `bool` | `True` | Use experimental Codex API features. |
 | `enable_self_config_tools` | `bool` | `False` | Expose tools that let Codex change its own model and reasoning settings. Use only in trusted rooms. |
 | `additional_dynamic_tools` | `list[dict]` | `[]` | Extra dynamic tool schemas registered with the Codex client. |
@@ -187,7 +182,6 @@ Pass these directly to `CodexAdapter(...)`:
 | `capabilities` | `Capability \| Iterable[Capability] \| None` | none | Optional Band tool categories exposed to the model. Opt-in: omitted, defaults to empty. |
 | `additional_tools` | `list[CustomToolDef] \| None` | `None` | Custom tools as `(PydanticModel, callable)` tuples. |
 | `history_converter` | `CodexHistoryConverter \| None` | auto | Advanced escape hatch for replacing the default history/thread-metadata converter. |
-| `client_factory` | callable | `None` | Test/advanced injection point for a custom Codex client. |
 
 ## Feature flags: Capabilities and Emit
 
--- a/docs/adapters/codex.md
+++ b/docs/adapters/codex.md
@@ -32,16 +32,11 @@ Credentials for Band can also be loaded from `agent_config.yaml` with `Agent.fro
 
 ```python
 import asyncio
-import os
-
 from band import Agent
 from band.adapters.codex import CodexAdapter, CodexAdapterConfig
 
 adapter = CodexAdapter(
-    config=CodexAdapterConfig(
-        workspace_for_room=lambda room_id: os.path.join("/workspaces", room_id),
-        model="gpt-5.5",
-    ),
+    config=CodexAdapterConfig(model="gpt-5.5"),
 )
 
 agent = Agent.create(
@@ -112,7 +107,7 @@ Pass these to `CodexAdapterConfig(...)`:
 
 | Parameter | Type | Default | Description |
 |-----------|------|---------|-------------|
-| `workspace_for_room` | `Callable[[str], str]` | required | Returns the absolute workspace for a room. Every live room must receive a distinct workspace. |
+| `workspace_for_room` | `Callable[[str], str] | None` | `None` | Optional override for a room workspace. By default, the adapter creates `./.band-workspaces/<room-id>`. |
 | `model` | `str \| None` | `None` | Model to use. When unset, the adapter asks Codex for visible models and uses the first visible model, or the adapter default if discovery fails or returns no usable model. |
 | `reasoning_effort` | `"none" \| "minimal" \| "low" \| "medium" \| "high" \| "xhigh" \| None` | `None` | Reasoning effort for models that support it. |
 | `reasoning_summary` | `"auto" \| "concise" \| "detailed" \| "none" \| None` | `None` | How Codex summarizes reasoning in responses. |
@@ -211,10 +206,7 @@ from band.core.types import Capability, Emit
 from band.adapters.codex import CodexAdapter, CodexAdapterConfig
 
 adapter = CodexAdapter(
-    config=CodexAdapterConfig(
-        workspace_for_room=lambda room_id: f"/workspaces/{room_id}",
-        model="gpt-5.5",
-    ),
+    config=CodexAdapterConfig(model="gpt-5.5"),
     capabilities=Capability.CONTACTS | Capability.MEMORY,
     emit=Emit.TOOL_CALLS | Emit.THOUGHTS | Emit.TASK_EVENTS,
 )
@@ -293,10 +285,7 @@ def get_weather(args: WeatherInput) -> str:
 
 
 adapter = CodexAdapter(
-    config=CodexAdapterConfig(
-        workspace_for_room=lambda room_id: f"/workspaces/{room_id}",
-        model="gpt-5.5",
-    ),
+    config=CodexAdapterConfig(model="gpt-5.5"),
     additional_tools=[(WeatherInput, get_weather)],
 )
 ```
--- a/docs/adapters/codex.md
+++ b/docs/adapters/codex.md
@@ -211,7 +211,10 @@ from band.core.types import Capability, Emit
 from band.adapters.codex import CodexAdapter, CodexAdapterConfig
 
 adapter = CodexAdapter(
-    config=CodexAdapterConfig(model="gpt-5.5"),
+    config=CodexAdapterConfig(
+        workspace_for_room=lambda room_id: f"/workspaces/{room_id}",
+        model="gpt-5.5",
+    ),
     capabilities=Capability.CONTACTS | Capability.MEMORY,
     emit=Emit.TOOL_CALLS | Emit.THOUGHTS | Emit.TASK_EVENTS,
 )
@@ -290,7 +293,10 @@ def get_weather(args: WeatherInput) -> str:
 
 
 adapter = CodexAdapter(
-    config=CodexAdapterConfig(model="gpt-5.5"),
+    config=CodexAdapterConfig(
+        workspace_for_room=lambda room_id: f"/workspaces/{room_id}",
+        model="gpt-5.5",
+    ),
     additional_tools=[(WeatherInput, get_weather)],
 )
 ```
--- a/examples/acp/clients/bridge_architecture.py
+++ b/examples/acp/clients/bridge_architecture.py
@@ -45,7 +45,6 @@ Run with:
 from __future__ import annotations
 
 import asyncio
-import os
 import logging
 import shlex
 
@@ -84,7 +83,6 @@ async def main() -> None:
     settings = Settings()
 
     command = shlex.split(settings.acp_agent_command)
-    cwd = settings.acp_agent_cwd
     auth_method = settings.acp_auth_method or None
     inject_band_tools = settings.acp_inject_band_tools
     profile_name = settings.acp_client_profile.strip().lower()
@@ -92,7 +90,6 @@ async def main() -> None:
 
     adapter = ACPClientAdapter(
         command=command,
-        workspace_for_room=lambda room_id: os.path.join(cwd, room_id),
         inject_band_tools=inject_band_tools,
         auth_method=auth_method,
         profile=profile,
--- a/examples/acp/clients/bridge_architecture.py
+++ b/examples/acp/clients/bridge_architecture.py
@@ -45,6 +45,7 @@ Run with:
 from __future__ import annotations
 
 import asyncio
+import os
 import logging
 import shlex
 
@@ -91,7 +92,7 @@ async def main() -> None:
 
     adapter = ACPClientAdapter(
         command=command,
-        cwd=cwd,
+        workspace_for_room=lambda room_id: os.path.join(cwd, room_id),
         inject_band_tools=inject_band_tools,
         auth_method=auth_method,
         profile=profile,
--- a/examples/acp/clients/bridge_architecture.py
+++ b/examples/acp/clients/bridge_architecture.py
@@ -53,7 +53,7 @@ from pydantic_settings import BaseSettings, SettingsConfigDict
 
 from band import Agent, configure_logging
 from band.adapters import ACPClientAdapter
-from band.integrations.acp.client_profiles import CursorACPClientProfile
+from band.integrations.acp.client_profiles import resolve_acp_client_profile
 
 configure_logging(
     level=logging.INFO,
@@ -85,8 +85,7 @@ async def main() -> None:
     command = shlex.split(settings.acp_agent_command)
     auth_method = settings.acp_auth_method or None
     inject_band_tools = settings.acp_inject_band_tools
-    profile_name = settings.acp_client_profile.strip().lower()
-    profile = CursorACPClientProfile() if profile_name == "cursor" else None
+    profile = resolve_acp_client_profile(settings.acp_client_profile)
 
     adapter = ACPClientAdapter(
         command=command,
--- a/examples/codex/01_basic_agent.py
+++ b/examples/codex/01_basic_agent.py
@@ -31,7 +31,6 @@ from __future__ import annotations
 
 import asyncio
 import logging
-import os
 from pathlib import Path
 
 from dotenv import load_dotenv
@@ -63,12 +62,6 @@ class Settings(BaseSettings):
     codex_role: str = ""
 
 
-def workspace_for_room(room_id: str) -> str:
-    workspace = Path(os.getenv("CODEX_WORKSPACE_ROOT", ".band-workspaces")) / room_id
-    workspace.mkdir(parents=True, exist_ok=True)
-    return str(workspace.resolve())
-
-
 async def main() -> None:
     load_dotenv()
     settings = Settings()
@@ -93,7 +86,6 @@ async def main() -> None:
     # docstring) when omitted here.
     adapter = CodexAdapter(
         config=CodexAdapterConfig(
-            workspace_for_room=workspace_for_room,
             personality="pragmatic",
             custom_section=custom_section,
             include_base_instructions=True,
--- a/examples/codex/01_basic_agent.py
+++ b/examples/codex/01_basic_agent.py
@@ -13,17 +13,14 @@ Runs a Band agent backed by Codex app-server.
 Prerequisites:
 1. OAuth login:
    codex login
-2. For stdio mode (default), no extra process is needed.
-3. For ws mode, start app-server separately:
-   codex app-server --listen ws://127.0.0.1:8765
+2. The adapter starts one local stdio process for each Band room.
 
 Run:
     uv run examples/codex/01_basic_agent.py
 
 Optional env overrides:
     AGENT_KEY=darter
-    CODEX_TRANSPORT=stdio|ws
-    CODEX_WS_URL=ws://127.0.0.1:8765
+    CODEX_WORKSPACE_ROOT=.band-workspaces
     CODEX_ROLE=coding|planner|reviewer
     CODEX_MODEL=gpt-5.5
     CODEX_APPROVAL_MODE=manual|auto_accept|auto_decline
@@ -34,6 +31,7 @@ from __future__ import annotations
 
 import asyncio
 import logging
+import os
 from pathlib import Path
 
 from dotenv import load_dotenv
@@ -65,6 +63,12 @@ class Settings(BaseSettings):
     codex_role: str = ""
 
 
+def workspace_for_room(room_id: str) -> str:
+    workspace = Path(os.getenv("CODEX_WORKSPACE_ROOT", ".band-workspaces")) / room_id
+    workspace.mkdir(parents=True, exist_ok=True)
+    return str(workspace.resolve())
+
+
 async def main() -> None:
     load_dotenv()
     settings = Settings()
@@ -84,11 +88,12 @@ async def main() -> None:
                 "Role '%s' specified but no prompt file at %s", codex_role, prompt_file
             )
 
-    # transport/codex_ws_url/model/cwd/approval_policy/approval_mode/
+    # model/approval_policy/approval_mode/
     # emit_turn_task_markers all self-source from CODEX_* env vars (see module
     # docstring) when omitted here.
     adapter = CodexAdapter(
         config=CodexAdapterConfig(
+            workspace_for_room=workspace_for_room,
             personality="pragmatic",
             custom_section=custom_section,
             include_base_instructions=True,
--- a/src/band/adapters/codex.py
+++ b/src/band/adapters/codex.py
@@ -301,7 +301,7 @@ class CodexAdapterConfig(BaseSettings):
         Literal["none", "minimal", "low", "medium", "high", "xhigh"] | None
     ) = None
     reasoning_summary: Literal["auto", "concise", "detailed", "none"] | None = None
-    cwd: str = Field(default_factory=os.getcwd)
+    cwd: str | None = None
     workspace_for_room: Callable[[str], str] | None = Field(default=None, exclude=True)
     approval_policy: str = "never"
     personality: Literal["friendly", "pragmatic", "none"] = "pragmatic"
@@ -434,6 +434,8 @@ class CodexAdapter(SimpleAdapter[CodexSessionState]):
             self._custom_tools.extend(self._build_self_config_tools())
         if self.config.workspace_for_room is None:
             raise ValueError("workspace_for_room is required for Codex room isolation")
+        if self.config.cwd is not None:
+            raise ValueError("cwd is not supported; use workspace_for_room")
         if self.config.transport != "stdio":
             raise ValueError("only stdio Codex transport guarantees room process isolation")
         if client_factory is not None:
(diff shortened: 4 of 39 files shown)


docs/adapters/codex.md
```diff
@@ -45,7 +45,7 @@ from band.adapters.codex import CodexAdapter, CodexAdapterConfig
 
 adapter = CodexAdapter(
     config=CodexAdapterConfig(
-        cwd=os.getcwd(),
+        workspace_for_room=lambda room_id: os.path.join("/workspaces", room_id),
```
AlexanderZ-Band: no magic strings

examples/acp/clients/bridge_architecture.py
```diff
@@ -91,7 +92,7 @@ async def main() -> None:
 
     adapter = ACPClientAdapter(
         command=command,
-        cwd=cwd,
+        workspace_for_room=lambda room_id: os.path.join(cwd, room_id),
```
AlexanderZ-Band: DRY principal

src/band/adapters/codex.py
```diff
@@ -417,14 +432,25 @@ def __init__(
         self._custom_tools: list[CustomToolDef] = list(additional_tools or [])
         if self.config.enable_self_config_tools:
             self._custom_tools.extend(self._build_self_config_tools())
+        if self.config.workspace_for_room is None:
+            raise ValueError("workspace_for_room is required for Codex room isolation")
+        if self.config.transport != "stdio":
+            raise ValueError("only stdio Codex transport guarantees room process isolation")
```
AlexanderZ-Band: error or warnning?

src/band/adapters/codex.py
```diff
@@ -445,7 +471,78 @@ def __init__(
...
+    @_initialized.setter
+    def _initialized(self, value: bool) -> None:
+        state = self._active_client_state()
+        if state is not None:
+            state.initialized = value
+        else:
+            self._fallback_initialized = value
+
+    @property
+    def _selected_model(self) -> str | None:
```
AlexanderZ-Band, line 532: why leading _?

src/band/integrations/acp/client_adapter.py
```diff
@@ -163,18 +165,27 @@ def __init__(
             history_converter=ACPClientHistoryConverter(),
             **features,
         )
+        if workspace_for_room is None:
+            raise ValueError("workspace_for_room is required for ACP client isolation")
+        if cwd is not None:
+            raise ValueError("cwd is not supported; use workspace_for_room")
+        if host is not None or port is not None:
+            raise ValueError("TCP ACP transport cannot guarantee room process isolation")
```
AlexanderZ-Band: the check seems to repeat keep DRY

src/band/integrations/acp/client_adapter.py
```diff
@@ -348,18 +371,18 @@ async def on_message(
...
-                await self._runtime.prompt(
+                await runtime.prompt(
                     session_id=session_id,
                     prompt_text=prompt_text,
                     on_chunk=emitter.emit,
                 )
         except Exception as e:
             logger.exception("ACP agent error: %s", e)
-            await self.stop()
+            await self.on_cleanup(room_id)
```
AlexanderZ-Band, line 362: recourse manager?

src/band/integrations/acp/client_adapter.py
```diff
@@ -473,10 +496,9 @@ def _build_system_context(self, room_id: str, msg: PlatformMessage) -> str:
         return f"[System Context]\n{system_prompt}\n{room_context}"
 
     def _build_local_mcp_server_config(
-        self,
-        local_server: LocalMCPServer,
+        self, local_server: LocalMCPServer, transport: str
     ) -> LocalMcpServerConfig:
-        if self._runtime._agent_mcp_transport == "sse":
+        if transport == "sse":
```
AlexanderZ-Band, line 451: single source of truth no magic strings

src/band/integrations/acp/client_adapter.py
```diff
@@ -535,21 +557,24 @@ async def _ensure_band_mcp_backend(self) -> BandMCPBackend:
                 self._band_mcp_backend = None
             if self._band_mcp_backend is None:
                 backend = await create_band_mcp_backend(
-                    kind=self._runtime._agent_mcp_transport,
+                    kind="http",
```
AlexanderZ-Band: magic stirng

src/band/integrations/acp/client_adapter.py
```diff
@@ -703,7 +736,7 @@ async def cleanup_all(self, *, final: bool = True) -> None:
             # None and start a fresh backend while this one is mid-teardown.
             if backend is not None:
                 await backend.stop()
-        await self._runtime.stop()
+        await asyncio.gather(*(runtime.stop() for runtime in runtimes))
```
AlexanderZ-Band: deserves helper function

AlexanderZ-Band (commented):
Review pass: fixed room-workspace ownership, removal of shared Codex fallback state, failed Codex startup cleanup, stale transport documentation, and focused regression coverage. The legacy ACP/Codex fixture suite still targets the intentionally removed shared-process and injected-client APIs; it remains the reason this PR stays draft.

docs/adapters/codex.md
AlexanderZ-Band: Quick start now uses `Agent.from_config` so Band URLs come from config instead of hardcoded strings in the doc snippet.

examples/acp/clients/bridge_architecture.py
AlexanderZ-Band: Extracted `resolve_acp_client_profile()` in `client_profiles.py`; bridge example now calls that helper instead of inlining the cursor check.

src/band/adapters/codex.py
AlexanderZ-Band: Kept as `ValueError`: non-stdio transport cannot guarantee room isolation, so construction must fail loudly rather than warn and proceed unsafely.

src/band/adapters/codex.py
AlexanderZ-Band: These `_client` / `_selected_model` properties are room-scoped shims over `RoomCodexClient`; the leading `_` marks them as turn-context internals, not public adapter config.

src/band/integrations/acp/client_adapter.py
AlexanderZ-Band: Pulled the duplicated workspace-collision guard into `claim_room_workspace()` in `workspaces.py`; both Codex and ACP adapters call it.

src/band/integrations/acp/client_adapter.py
AlexanderZ-Band: `_ensure_band_mcp_backend()` already serializes start/stop under `_mcp_backend_lock`; added `_stop_runtimes()` helper for the runtime teardown half of `cleanup_all()`.

src/band/integrations/acp/client_adapter.py
AlexanderZ-Band: `CHAT_ID_FIELD_NAME` and `BAND_MCP_SERVER_NAME` already come from `band.runtime.tools`; MCP backend kind now uses `DEFAULT_BAND_MCP_BACKEND_KIND`.

src/band/integrations/acp/client_adapter.py
AlexanderZ-Band: Replaced the bare `"http"` backend kind literal with `DEFAULT_BAND_MCP_BACKEND_KIND`; transport typing uses `MCPTransportKind`.

src/band/integrations/acp/client_adapter.py
AlexanderZ-Band: Extracted `_stop_runtimes()` from `cleanup_all()` so runtime teardown is a named helper.

src/band/workspaces.py
```diff
@@ -0,0 +1,54 @@
...
+
+def resolve_room_workspace(
+    room_id: str, workspace_for_room: WorkspaceResolver | None
+) -> str:
+    """Return a room's absolute workspace, creating the safe default on demand."""
+    if workspace_for_room is not None:
+        workspace = workspace_for_room(room_id)
+        if not isinstance(workspace, str) or not os.path.isabs(workspace):
+            raise ValueError("workspace_for_room must return an absolute path")
+        return os.path.realpath(workspace)
```
amit-gazal-band: **The custom-resolver path never creates the directory.** A caller-supplied `workspace_for_room` result is only validated as absolute and `realpath`'d; the `mkdir(parents=True, exist_ok=True)` lives exclusively in `create_room_workspace_resolver`, i.e. only on the default branch below.

Concretely: the example this PR adds at `docs/adapters/codex.md:92` (`workspace_for_room=lambda room_id: f"/workspaces/{room_id}"`) hands a non-existent path to `CodexStdioClient(cwd=...)` / ACP `new_session(cwd=...)`, and the room's first turn dies on subprocess spawn with a bare `FileNotFoundError`/`NotADirectoryError` rather than anything that points at the resolver.

Either `Path(workspace).mkdir(parents=True, exist_ok=True)` here too, or document in both adapter docs that a custom resolver must create the directory itself.

src/band/adapters/codex.py
```diff
@@ -2622,7 +2731,7 @@ async def _handle_local_command(
                     )
                 return True
 
-            self.config.model = model_arg
+            # Model selection belongs to the room's process, never adapter-wide config.
```
amit-gazal-band, line 2756: **A room's `/model` override no longer survives a client rebuild.** Previously `/model` wrote `self.config.model`, which `_select_model()` reads first; now it writes only `state.selected_model`.

Failure path: a turn hits `transport/closed` (line 1059 sets `_initialized = False`), the next message runs `_ensure_client_ready`, which re-runs `_selected_model = await self._select_model()` — and `_select_model` returns `config.model` (or auto-discovery). The user's override is silently reverted mid-conversation, with the room still showing the "set for subsequent turns" confirmation.

`_handle_set_model` (line 563) has the same hole. Note the asymmetry: `reasoning_effort`/`reasoning_summary` *do* survive a rebuild because they live on `RoomCodexClient`, which outlives the client — so this looks unintentional rather than a deliberate narrowing. Storing the override on `RoomCodexClient` as e.g. `model_override` and having `_select_model` prefer it would restore the old semantics per-room.

src/band/adapters/codex.py
```diff
@@ -1055,42 +1153,53 @@ async def on_cleanup(self, room_id: str) -> None:
...
-                self._token_usage.clear()
-                self._approval_audit.clear()
-                self._session_approved.clear()
-                self._sandbox_overrides.clear()
+                self._room_clients.pop(room_id, None)
+                self._workspace_rooms.pop(room.workspace, None)
+
+    async def cleanup_all(self) -> None:
+        """Close every room-owned Codex process during agent shutdown."""
+        await asyncio.gather(
```
amit-gazal-band: **`asyncio.gather` without `return_exceptions=True` abandons the other rooms' teardown.** `on_cleanup` can raise — `client.close()` is only guarded against `asyncio.TimeoutError`, so any other exception (a broken pipe on a dead stdio transport is the obvious one) propagates.

When it does, `gather` re-raises immediately while the sibling tasks are still pending; `Agent.stop()` catches and logs (`agent.py:337`) and shutdown continues, leaving those `codex` subprocesses unreaped. Since the whole point of this PR is one process per room, a partial reap leaks in proportion to room count — exactly the failure mode the old single-client design couldn't have.

`return_exceptions=True` plus a log of the failures would make shutdown best-effort across all rooms.

src/band/adapters/codex.py
```diff
@@ -1055,42 +1153,53 @@ async def on_cleanup(self, room_id: str) -> None:
...
+            await client.connect()
+            await client.initialize(
                 client_name=self.config.client_name,
                 client_title=self.config.client_title,
                 client_version=self.config.client_version,
                 experimental_api=self.config.experimental_api,
             )
             self._selected_model = await self._select_model()
             self._initialized = True
+        except Exception:
```
amit-gazal-band, line 1197: Minor, but the PR description says this path "evict[s] the failed client" — it evicts the *client*, not the room. On a failed `connect()`/`initialize()`, `_room_clients[room_id]` survives with `client=None` and its `_workspace_rooms[workspace]` claim is still held.

Retry works (the next message rebuilds the client), so this isn't fatal — but a room that never successfully started still holds its workspace against every other room, so a resolver that would legitimately hand that workspace to a different room now fails the `claim_room_workspace` check for a process that doesn't exist.

src/band/integrations/acp/client_adapter.py
```diff
@@ -778,7 +769,7 @@ async def _fetch_replay(
         raw = messages_before(context.get("data") or [], msg.id)
         return build_replay_messages([m for m in raw if m.get("id") != msg.id])
 
-    async def _ensure_connection(self) -> ACPConnectionProtocol:
-        return await self._runtime.ensure_connection(
+    async def _ensure_connection(self, runtime: ACPRuntime) -> ACPConnectionProtocol:
+        return await runtime.ensure_connection(
             can_respawn=bool(self.agent_name),
```
amit-gazal-band: With `on_started`'s eager `await self._spawn_process()` removed, `ensure_connection(can_respawn=bool(self.agent_name))` is now the **only** spawn path in the adapter. If `agent_name` is ever falsy, `ACPRuntime.ensure_connection` takes the `else` branch (`client_runtime.py:754`) and every turn fails with `"ACP client not initialized. Call on_started first."` — a state that previously could not arise, because `on_started` spawned unconditionally regardless of the name. The guard's original meaning was "only *re*spawn once we've been started"; it now doubles as the initial-spawn gate, which is a different question.

Related cosmetic fallout on the same path: the very first spawn of every room now logs `"Respawning ACP agent subprocess"`, because `ensure_connection` passes `respawn=True` (`client_runtime.py:752`).

src/band/integrations/acp/client_adapter.py
```diff
@@ -108,6 +110,7 @@ def new_message_marker() -> str:
 # stdio and TCP are the built-in transports; injecting one (e.g. docker exec / ssh,
 # or a fake in tests) is the supported extension point.
 SpawnProcess = Callable[..., object]
+WorkspaceResolver = Callable[[str], str]
```
amit-gazal-band: `WorkspaceResolver = Callable[[str], str]` is re-declared here, but this same file already imports from `band.workspaces` on line 68 — which defines the identical alias. Two definitions of one closed type; per CLAUDE.md's single-source-of-truth rule, import it alongside `claim_room_workspace`/`resolve_room_workspace` instead.

src/band/integrations/acp/client_adapter.py
```diff
@@ -108,6 +110,7 @@ def new_message_marker() -> str:
 # stdio and TCP are the built-in transports; injecting one (e.g. docker exec / ssh,
 # or a fake in tests) is the supported extension point.
 SpawnProcess = Callable[..., object]
```
amit-gazal-band, line 116: Stale after this PR: the comment above says "stdio and TCP are the built-in transports; injecting one (e.g. docker exec / ssh, or a fake in tests) is the supported extension point" — but `__init__` now rejects both `host`/`port` and `spawn_process`, so neither is reachable and `SpawnProcess` has no remaining production caller (`_build_runtime` hardcodes `spawn_agent_process`).

Same issue in `docs/acp.md:24`, which still advertises `tcp_spawn_process` as the "TCP connect seam" in the key-files table while the client side can no longer reach it.

examples/codex/01_basic_agent.py
```diff
@@ -13,17 +13,14 @@
...
+2. The adapter starts one local stdio process for each Band room.
 
 Run:
     uv run examples/codex/01_basic_agent.py
 
 Optional env overrides:
     AGENT_KEY=darter
-    CODEX_TRANSPORT=stdio|ws
-    CODEX_WS_URL=ws://127.0.0.1:8765
+    CODEX_WORKSPACE_ROOT=.band-workspaces
```
amit-gazal-band, line 23: `CODEX_WORKSPACE_ROOT` is documented here as an env override but nothing reads it — `grep -r CODEX_WORKSPACE_ROOT` finds only this line. `CodexAdapterConfig` has no such field, and with `extra="forbid"` + `env_prefix="CODEX_"` it isn't even settable as a passthrough. Either wire it up (e.g. a field that feeds `create_room_workspace_resolver`) or drop the line, since it's the only workspace configuration this example advertises.

amit-gazal-band (changes requested):
The core refactor reads soundly — per-room workspaces, per-room processes, and a per-room `_rpc_lock` are a real improvement over the shared-client design, and the `ContextVar` plumbing is safe under `asyncio.gather` (each coroutine is wrapped in a Task that copies the context). Two blocking themes:

**1. The new constructor validations are fatal, and the in-repo callers were not migrated.** `cwd=`, `transport="ws"`, `host`/`port`, `client_factory=` and `spawn_process=` now `raise ValueError` at construction. Several shipped callers still pass them, so they fail before the first message.

**2. A few behavioral regressions** around model-override persistence, partial teardown, and workspace directory creation — flagged inline.

### Not inline (unchanged files)

These five have no hunks in this PR, so they can't be anchored:

- **`docker/codex/runner.py:248` — blocking. The shipped Docker entrypoint cannot start.** It always passes `cwd=codex_cwd`; the value chain (`CODEX_CWD` env → `config["workspace"]` → `repo_init.repo_path` → literal `"/workspace/repo"`) is never `None`, so every containerized Codex agent dies on `ValueError: cwd is not supported`. Line 247 also forwards `transport=codex_transport`, which `_parse_transport` still allows to be `ws` — also now rejected. This needs a `workspace_for_room=create_room_workspace_resolver(codex_cwd)` migration in the same PR.

- **`tests/e2e/baseline/toolkit/builders.py:340` — blocking.** `codex_config_kwargs` sets `"cwd": s.backends.codex_cwd` unconditionally, so every codex E2E matrix cell in the `backends` lane fails at adapter construction with the same `ValueError`. The `Dep.CODEX_CWD` gate and `settings.codex_cwd` have no replacement either — nothing maps them onto `workspace_for_room`.

- **`tests/e2e/baseline/toolkit/builders.py:477` — blocking.** The `copilot_acp` builder sets `"cwd": sandbox` (a per-cell `tempfile.mkdtemp`), which reaches `ACPClientAdapter.__init__` and now raises. Beyond the hard failure, the *intent* of that cwd — the comment on line 468 says it keeps the repo's own `.claude/` skills from leaking into the agent under test — has no replacement, because `CopilotACPAdapterConfig` exposes no `workspace_for_room`.

- **`src/band/adapters/copilot_acp.py:115`.** `CopilotACPAdapterConfig` still declares `cwd`, `host` and `port`, and `__init__` still forwards `"cwd": config.cwd` / `host=` / `port=` into the base adapter. All three are now unconditionally rejected — dead but fatal config knobs — and with no `workspace_for_room` passthrough, Copilot ACP users have no way to choose a workspace at all.

- **`examples/coding_agents/docker-compose.yml:27`** (also `examples/codex/docker-compose.yml:11`, `docker-compose.multi.yml:11,64`, `docker-compose.plan-review.yml:11,68`). These all set `CODEX_CWD`, which `CodexAdapterConfig` still sources via `env_prefix="CODEX_"` into the now-rejected `cwd` field. Any existing deployment or `.env` carrying `CODEX_CWD` (or `CODEX_TRANSPORT=ws`) now fails at startup with no documented migration path.

### Checked and found OK

Recording so it isn't re-litigated: `async with self._rpc_lock` in `on_cleanup` binds the lock object before `_room_clients.pop`, so `__aexit__` releases the right lock; ACP's `_runtime_for` is never called while `_session_lock` is held, so there's no re-entrant deadlock; and hardcoding `DEFAULT_BAND_MCP_BACKEND_KIND = "http"` while `_build_local_mcp_server_config` still branches on the negotiated transport is benign, because `LocalMCPServer` serves both `/sse` and `/mcp` regardless of `kind` and `BandMCPBackend.kind` has no downstream reader.

src/band/adapters/codex.py
AlexanderZ-Band: Fixed in 81ecf44: cleanup_all gathers room teardowns with return_exceptions=True and logs each failure, so one broken transport does not abandon sibling room cleanup. The regression test verifies both clients are closed.

src/band/workspaces.py
AlexanderZ-Band: Fixed in 81ecf44: resolve_room_workspace now creates the resolved custom directory, with a regression test covering nested custom paths.

src/band/adapters/codex.py
AlexanderZ-Band: Fixed in 81ecf44: RoomCodexClient now stores model_override and _select_model prefers it after a client rebuild. The regression test exercises a rebuild and verifies the override remains selected.

src/band/adapters/codex.py
AlexanderZ-Band: Fixed in 81ecf44: failed startup closes the partial client and releases the workspace reservation when close succeeds, while retaining room state for retry. The regression test verifies another room can claim the workspace.

examples/codex/01_basic_agent.py
AlexanderZ-Band: Fixed in 81ecf44: the example now reads CODEX_WORKSPACE_ROOT and passes it through create_room_workspace_resolver.

src/band/integrations/acp/client_adapter.py
AlexanderZ-Band: Fixed in 81ecf44: the stale adapter comment and docs/acp.md key-files table now describe the reachable local stdio boundary. Remote ACP/TCP support is intentionally out of scope for this PR and is tracked in INT-1492.

src/band/integrations/acp/client_adapter.py
AlexanderZ-Band: Fixed in 81ecf44: WorkspaceResolver is imported from band.workspaces as the single shared type definition.

src/band/integrations/acp/client_adapter.py
AlexanderZ-Band: Fixed in 81ecf44: the adapter now permits the room runtime to perform its initial lazy spawn even when agent_name is empty, and the runtime records the initial spawn as non-respawn. ACP spawn coverage verifies the room cwd.

src/band/adapters/codex.py
```diff
@@ -1055,46 +1157,75 @@ async def on_cleanup(self, room_id: str) -> None:
...
+                except Exception:
+                    closed_cleanly = False
+                    logger.debug(
+                        "Failed to close unsuccessfully initialized Codex client",
+                        exc_info=True,
+                    )
+                if closed_cleanly:
+                    state = self._active_client_state()
+                    if state is not None:
+                        self._workspace_rooms.pop(state.workspace, None)
```
AlexanderZ-Band: issue: After a failed initial Codex connect, this except block releases the workspace reservation (`_workspace_rooms.pop(state.workspace, None)`) but leaves the room's `RoomCodexClient` cached in `self._room_clients[room_id]` with `client=None`. That stale cache entry causes `on_cleanup` to later corrupt a *different* room's live workspace claim.

Concretely: room-a's connect fails and releases workspace W. A different room-b, mapped to the same W (a configuration this PR explicitly supports/tests via `tests/adapters/test_room_workspace_isolation.py::test_adapters_reject_a_custom_workspace_shared_by_live_rooms`), then claims W and is actively using it. The platform later calls `on_cleanup("room-a")` for an unrelated reason (room-a's chat session simply ends) - since `self._room_clients["room-a"]` still exists with `client is None`, `on_cleanup` takes the early-return branch a few lines below and unconditionally does `self._workspace_rooms.pop(room.workspace, None)` for room-a's cached `workspace` value W - **without checking that `_workspace_rooms[W]` still equals `"room-a"`**. This silently deletes room-b's live claim on W, even though room-b's Codex process is still actively running there. A subsequent room can now claim W too, breaking the exact "one live room per resolved workspace" invariant this PR exists to provide (two rooms' Codex processes could concurrently read/write the same directory).

Reproduced against this exact code: after this sequence, `adapter._workspace_rooms` ends up empty even though room-b is still live, instead of retaining `{W: "room-b"}`.

fix: Evict the room from `self._room_clients` here too (mirroring `on_cleanup`'s full eviction), so a failed connect fully releases room-a's state instead of leaving a half-torn-down entry that a later, unrelated `on_cleanup` call can use to clobber someone else's claim.

src/band/adapters/codex.py
```diff
@@ -1055,46 +1157,75 @@ async def on_cleanup(self, room_id: str) -> None:
...
+        except Exception:
+            if self._client is client:
+                self._client = None
+                closed_cleanly = True
+                try:
+                    await client.close()
+                except Exception:
+                    closed_cleanly = False
+                    logger.debug(
+                        "Failed to close unsuccessfully initialized Codex client",
```
AlexanderZ-Band, line 1355: nit: This close failure (after a failed init) logs at DEBUG, but the structurally similar close-timeout failure in `on_cleanup` (line ~1149) logs at WARNING. Both indicate the Codex subprocess may not have terminated cleanly - a potential leaked process - so DEBUG makes this invisible by default exactly when it would be useful to notice.

fix: Raise this to `logger.warning` (or align both to the same level) so a failed close during the connect-retry path is as visible as one during `on_cleanup`.

AlexanderZ-Band (commented):
my-code-review pass 1/3: scope statement - this PR gives each Band room its own isolated filesystem workspace and its own local stdio coding-agent process for both ACPClientAdapter and CodexAdapter, enforces one live room per resolved workspace, and rejects transports that can't preserve room isolation. All 17 pre-existing review threads are resolved and I verified their fixes are genuinely present in the current code. One new confirmed bug (Edge-Case Flow Analysis) and one logging nit (Runtime Safety) below.

src/band/adapters/codex.py
```diff
@@ -287,7 +302,8 @@ class CodexAdapterConfig(BaseSettings):
         Literal["none", "minimal", "low", "medium", "high", "xhigh"] | None
     ) = None
     reasoning_summary: Literal["auto", "concise", "detailed", "none"] | None = None
-    cwd: str = Field(default_factory=os.getcwd)
+    cwd: str | None = None
+    workspace_for_room: Callable[[str], str] | None = Field(default=None, exclude=True)
```
AlexanderZ-Band: nit: this redeclares the exact shape of `WorkspaceResolver = Callable[[str], str]` (`src/band/workspaces.py`) inline instead of importing that alias. `ACPClientAdapter.__init__` and `CopilotACPAdapterConfig` both use `WorkspaceResolver` from `band.workspaces` for this identical field -- two definitions of one closed concept, the same class of issue already fixed elsewhere in this PR's own review threads (the client_adapter.py `WorkspaceResolver` de-dup).

src/band/integrations/acp/client_adapter.py
```diff
@@ -535,21 +511,24 @@ async def _ensure_band_mcp_backend(self) -> BandMCPBackend:
...
+    async def _get_or_start_band_mcp_server(self, room_id: str) -> LocalMcpServerConfig:
         backend = await self._ensure_band_mcp_backend()
         local_server = backend.local_server
         if local_server is None:
             raise RuntimeError("ACP MCP backend did not create a local server")
 
-        return self._build_local_mcp_server_config(local_server)
+        runtime = await self._runtime_for(room_id)
+        return self._build_local_mcp_server_config(
+            local_server, runtime._agent_mcp_transport
```
AlexanderZ-Band: nit: `runtime._agent_mcp_transport` reaches into `ACPRuntime`'s private attribute from outside the class (this line moved from the old adapter-wide `self._runtime._agent_mcp_transport` to the per-room `runtime` as part of this PR's refactor). `ACPRuntime` already exposes `client` as a public property (`client_runtime.py` ~line 861); `_agent_mcp_transport` deserves the same treatment instead of the underscore reach-through.

src/band/adapters/codex.py
```diff
@@ -1055,46 +1157,75 @@ async def on_cleanup(self, room_id: str) -> None:
...
+                closed_cleanly = True
+                try:
+                    await client.close()
+                except Exception:
+                    closed_cleanly = False
+                    logger.debug(
+                        "Failed to close unsuccessfully initialized Codex client",
+                        exc_info=True,
+                    )
+                if closed_cleanly:
```
AlexanderZ-Band: issue: on a failed `connect()`/`initialize()`, the workspace reservation in `self._workspace_rooms` is released only when `client.close()` itself also succeeds (`closed_cleanly`). If `close()` also raises, the reservation is never released even though the room's client is now dead (`self._client is None`).

Reproduced: a `FailingClient` whose `connect()` and `close()` both raise -> `_room_client('room-a')` claims workspace W -> `_ensure_client_ready()` fails -> a second room `room-b` whose `workspace_for_room` resolves to the same/symlink-equivalent path W then hits `claim_room_workspace`'s `ValueError("workspace_for_room assigned ... to both 'room-a' and 'room-b'")` -- permanently, until `on_cleanup('room-a')`/`cleanup_all()` runs. `test_failed_codex_start_releases_workspace_reservation` only covers the `close()`-succeeds branch.

Fix: release `self._workspace_rooms.pop(state.workspace, None)` whenever `self._client` is reset to `None` here, not only when `closed_cleanly` -- the reservation should track "is this room's connection dead", not "did teardown itself also succeed".

AlexanderZ-Band (commented):
my-code-review (personal checklist), pass 1/3.

Scope: gives each Band room its own isolated filesystem workspace and its own local stdio coding-agent process/protocol state (ACPClientAdapter + CodexAdapter), enforces one live room per resolved workspace (incl. symlink-equivalent paths), and rejects transports that can't preserve room isolation. All 17 prior review threads are resolved and verified still fixed at HEAD.

3 new findings this pass (2 nits, 1 issue with a reproduced flow bug). Will apply fixes and resolve threads.

src/band/adapters/codex.py
AlexanderZ-Band: Fixed: _ensure_client_ready's except block now evicts the room from self._room_clients (not just the workspace reservation) when the close succeeds cleanly, mirroring on_cleanup's full eviction. Added a regression test (test_cleanup_after_failed_start_does_not_evict_a_new_owners_claim) that fails on the pre-fix code (mutation-verified) and passes now: a failed connect no longer leaves a stale room entry that a later, unrelated on_cleanup call can use to pop a different room's live workspace claim.

src/band/adapters/codex.py
AlexanderZ-Band: Fixed: raised this to logger.warning to match on_cleanup's close-timeout log level, so a failed close during the connect-retry path is equally visible.

src/band/adapters/codex.py
AlexanderZ-Band: nit: `workspace_for_room: Callable[[str], str] | None` redeclares the exact shape of `WorkspaceResolver` (`band/workspaces.py`) inline instead of importing that type alias. `ACPClientAdapter.__init__` (client_adapter.py) and `CopilotACPAdapterConfig` (copilot_acp.py) both correctly use `WorkspaceResolver` for the identical field -- this is the same duplicated-shape issue already fixed elsewhere in this PR (see the resolved "WorkspaceResolver re-declared" thread on client_adapter.py), just missed here.

src/band/integrations/acp/client_adapter.py
AlexanderZ-Band: nit: `runtime._agent_mcp_transport` reaches into `ACPRuntime`'s private attribute from outside the class. `ACPRuntime` already exposes `client` as a public property (client_runtime.py) -- worth adding an equivalent `agent_mcp_transport` property instead, especially since this call site was already touched by this refactor (`self._runtime._agent_mcp_transport` -> `runtime._agent_mcp_transport`).

tests/integrations/acp/test_client_adapter.py
```diff
@@ -468,34 +499,34 @@ def test_build_system_context_defers_to_external_mcp_tool_schema(self) -> None:
...
 
-    These inject a :class:`FakeSpawn` transport (the ``make_acp_transport`` fixture)
-    through the adapter's ``spawn_process`` seam rather than patching module globals,
-    so the real ACPRuntime start path runs against a scripted connection.
+    Spawn/transport tests inject :class:`FakeSpawn` via ``inject_acp_spawn`` and
+    start the room runtime explicitly — ``on_started`` no longer spawns.
     """
 
     @pytest.mark.asyncio
     async def test_on_started_spawns_process(self, make_acp_transport) -> None:
```
AlexanderZ-Band: nit: this test's name and docstring ("Should spawn ACP process and initialize connection") now contradict the class docstring two lines above ("`on_started` no longer spawns") and the test's own rewritten body, which spawns via an explicit `runtime.start()` call, not `on_started`. Worth renaming/rewording now that this test (and the surrounding class) were rewritten for lazy per-room spawning.

src/band/adapters/codex.py
AlexanderZ-Band: issue: this only releases the room's workspace reservation `if closed_cleanly` -- i.e. only when `client.close()` itself doesn't also raise. If `close()` raises too, `self._workspace_rooms[state.workspace]` keeps naming this (now-dead) room as the owner forever, until an explicit `on_cleanup`/`cleanup_all` runs. A different room whose `workspace_for_room` resolves to the same/symlink-equivalent path then gets permanently blocked by `claim_room_workspace`'s `ValueError`, even though this room's own client is already dead and retries fine on its own.

Reproduced: a `FailingClient` whose `connect()` and `close()` both raise leaves `_workspace_rooms` pointing at the dead room, and a second room aliasing the same path raises `ValueError` indefinitely.

Fix: release `self._workspace_rooms.pop(state.workspace, None)` unconditionally once the client is evicted -- `closed_cleanly` should only affect the log line, not this bookkeeping.

docker/codex/runner.py
```diff
@@ -245,7 +245,7 @@ async def main() -> None:
     adapter = CodexAdapter(
         config=CodexAdapterConfig(
             transport=codex_transport,
-            cwd=codex_cwd,
+            workspace_for_room=create_room_workspace_resolver(codex_cwd),
```
AlexanderZ-Band: issue: wrapping `codex_cwd` in `create_room_workspace_resolver` sends every room into a brand-new, empty subdirectory (`codex_cwd/<room_id>`) instead of `codex_cwd` itself -- but `codex_cwd` here is very often `repo_init.repo_path`, the exact directory `initialize_repo()` (above) just cloned/validated a real repository into.

Reproduced: `create_room_workspace_resolver(repo_path)('room-1')` returns `repo_path/room-1`, an empty directory containing none of the cloned repo's files.

This container is built around one pre-populated repo per deployment (`REQUIRED_MOUNTS`, `initialize_repo`) -- every room should operate directly in `codex_cwd`, e.g. `workspace_for_room=lambda _room_id: codex_cwd`, not get its own isolated subdirectory under it.

AlexanderZ-Band (commented):
Automated my-code-review pass 1/3. 5 findings from manual deep-dive review (2 confirmed issues with runnable repros, 3 style/reuse nits). Two background review agents (Style & Hygiene, Design & Reuse) were also dispatched but the system-wide concurrent-subagent limit prevented most parallel groups from completing this pass; findings below reflect direct verification against the code at HEAD.

src/band/adapters/codex.py
AlexanderZ-Band: Fixed: codex.py now imports WorkspaceResolver from band.workspaces and uses it as the field's type annotation instead of the inline Callable[[str], str].

src/band/integrations/acp/client_adapter.py
AlexanderZ-Band: Fixed: added a public agent_mcp_transport property to ACPRuntime (mirroring the existing client property) and updated the call site to use it instead of the private attribute.

tests/integrations/acp/test_client_adapter.py
AlexanderZ-Band: Fixed: renamed to test_room_runtime_start_spawns_and_initializes with an updated docstring describing that the room runtime's own start() performs the spawn/initialize, not on_started.

src/band/adapters/codex.py
AlexanderZ-Band: Fixed: the except block now evicts the room's whole RoomCodexClient entry (not just the workspace reservation) unconditionally once the client is evicted, regardless of whether close() itself also raised. This also closes a related latent bug: leaving a stale _room_clients entry behind (releasing only the reservation) would let a later on_cleanup for this same room pop a *different* room's since-claimed reservation on the same workspace path (see the new test_cleanup_after_failed_start_does_not_evict_a_new_owners_claim). Two new regression tests added in tests/adapters/test_room_workspace_isolation.py cover the close()-also-fails path and that TOCTOU.

docker/codex/runner.py
AlexanderZ-Band: Fixed: added a small _shared_room_workspace(codex_cwd) helper (next to the file's other small config helpers) that maps every room to the same codex_cwd directory, and used it in place of create_room_workspace_resolver(codex_cwd). This runner's mount contract (REQUIRED_MOUNTS) guarantees exactly one pre-populated repo per container, so every room now operates directly in that repo again, matching the pre-PR shared-workspace behavior for this specific deployment. Verified by loading the module directly and confirming the resolver maps distinct room ids to the same codex_cwd path; this file has no existing pytest coverage to extend (docker/codex/runner.py is a standalone entrypoint script, not under src/band or an existing tests/ anchor).

AlexanderZ-Band: Pass-2 follow-up: confirmed this fix's tradeoff explicitly -- since claim_room_workspace enforces one live room per resolved workspace, a second concurrent room sharing this container now fails loudly on its first message (ValueError) until the first room's session is cleaned up, rather than silently running against a stale/empty directory. This is the same fail-loud-over-silently-unsafe pattern the rest of this PR uses (e.g. TCP transport rejection), so kept as the fix, but documented the tradeoff directly in _shared_room_workspace's docstring so it's not a surprise.

src/band/adapters/codex.py
AlexanderZ-Band: Pass-2 follow-up (supersedes my earlier reply): the fix evolved to a better root cause. A follow-up Runtime Safety pass on this exact code found that fully evicting the room's RoomCodexClient (my first fix) silently dropped a room's /model or /reasoning override on ANY failed reconnect, not just the close()-also-fails case -- a real regression against this PR's own "preserve model overrides across client rebuilds" guarantee (proven with a FlakyClient whose first connect succeeds, a later rebuild's connect fails, and close() also fails).

Root-cause fix: added release_room_workspace(room_id, workspace, workspace_rooms) to workspaces.py (mirroring claim_room_workspace) -- it only releases a workspace claim while the given room_id is still its current owner, a no-op otherwise. _ensure_client_ready's except block now calls this unconditionally (regardless of whether close() itself also raised) instead of evicting the whole room, so the room's state survives for retry. The same ownership check also closes a latent TOCTOU: a stale room left behind after a failed connect can never evict a different room's since-claimed reservation on the same workspace path when on_cleanup later runs for it.

New regression test test_room_state_survives_a_failed_rebuild_attempt covers this exactly.

AlexanderZ-Band: Pass 2 follow-up: the fix above (full eviction of the room's RoomCodexClient) closed the original bug and a related TOCTOU (a stale entry's on_cleanup could otherwise evict a *different* room's since-claimed reservation -- see test_cleanup_after_failed_start_does_not_evict_a_new_owners_claim), but on closer look it reintroduced a narrower regression: transport/closed (around line 1062) resets _client/_initialized to force a rebuild on the next message, and if that rebuild's connect()+close() both fail, full eviction would silently drop the room's already-set /model override -- contradicting this PR's own 'preserve model overrides across client rebuilds' goal.

Replaced full eviction with an ownership-guarded release: added _release_room_workspace(room, room_id), which only pops the workspace reservation while the given room_id is still its current owner in _workspace_rooms. Used in on_cleanup's two release sites and in _ensure_client_ready's except block, which now keeps the room's RoomCodexClient (so model_override/reasoning settings survive a failed rebuild) instead of popping it. This is safe against the same TOCTOU the full-eviction fix guarded against, since a stale entry's later on_cleanup can no longer evict a different room's live claim (the ownership check fails).

Updated the two existing regression tests' now-stale `_room_clients == {}` assertions to `_room_clients["room-a"].client is None` (the entry now intentionally persists), and added test_room_state_survives_a_failed_rebuild_attempt proving a model override survives a failed *reconnect*, not just a failed first connect.

amit-gazal-band (approved):

PR #635, opened 2026-09-14
feat: Add E2E tests for the coding-agent adapters (claude_sdk, codex)

AlexanderZ-Band:
## Summary
- `test_codex.py`: regression-guard smoke restored (deleted in `fc62742c`).
- `test_claude_sdk.py`: new smoke proving Band's chat-based approval relay gates a real Claude Code native tool call end-to-end.
- `_build_claude_sdk`: per-cell disposable `cwd` (matrix cells previously ran Claude Code against this repo's own checkout).
- `claude_sdk.py`: approval prompt/resolution strings extracted into named constants (`APPROVAL_REQUESTED_TEMPLATE`, `APPROVAL_RESOLVED_TEMPLATE`), mirroring `opencode/approvals.py`.
- `_pre_tool_use_continue_hook`: returns `hookSpecificOutput.permissionDecision="ask"` instead of `{"continue_": True}`, which carried no permission decision — native tool calls (Bash/Write/Edit) were never reaching `can_use_tool` under the adapter's default `permission_mode`.
- `on_message`: a turn now runs as a detached task and only awaits its early release (mirrors `OpencodeAdapter`'s `release_turn_wait`/`turn_future` split). Band's runtime processes one room-message cycle at a time; a manual approval must release that cycle so the room can dispatch the reply that resolves it.

## Test plan
- [x] `ruff check .` / `ruff format --check .` / `pyrefly check`
- [x] `uv run pytest tests/ --ignore=tests/integration/ --ignore=tests/e2e/ -v` — 5592 passed
- [x] `E2E_TESTS_ENABLED=true BAND_E2E_LANE=core uv run pytest tests/e2e/baseline/smoke/adapters/test_claude_sdk.py -v -s --no-cov`
- [x] `E2E_TESTS_ENABLED=true BAND_E2E_LANE=backends uv run pytest tests/e2e/baseline/smoke/adapters/test_codex.py -v -s --no-cov`
- [x] `tests/e2e/baseline/guards/test_adapter_registry.py` — 8 passed

Closes INT-1448.

https://claude.ai/code/session_01PycgygL3XZiSheL4CpquPM


tests/e2e/baseline/toolkit/builders.py
```diff
@@ -76,9 +76,15 @@ def _build_claude_sdk(
 ) -> SimpleAdapter[Any]:
     from band.adapters.claude_sdk import ClaudeSDKAdapter  # noqa: PLC0415 -- isolates the claude_sdk extra from the other frameworks this file builds
 
+    # Claude Code gets real Bash/filesystem tools; an unset cwd falls back to
+    # the process cwd (this repo's own checkout). Mirrors _build_copilot_acp's
+    # per-cell disposable sandbox.
+    sandbox = tempfile.mkdtemp(prefix="band-e2e-claude-sdk-")
```
AlexanderZ-Band: issue: This allocates a new Claude working directory for every `build_adapter` call but hands only its path to `ClaudeSDKAdapter`; neither `running_agent` nor `ResourceManager.reap_all()` owns or removes it. The generic matrix constructs fresh adapters across its cells, so each run permanently leaves a `band-e2e-claude-sdk-*` directory behind (and the bespoke manual-approval smoke repeats the same unowned allocation at `test_claude_sdk.py:94`). The intended disposable sandbox is therefore not lifecycle-managed.

src/band/adapters/claude_sdk.py
```diff
@@ -678,32 +722,94 @@ async def on_message(
...
-            await self._report_error(tools, str(e))
-            raise
+            except Exception as e:
+                logger.exception("Error processing message: %s", e)
+                await self._report_error(tools, str(e))
+                raise
+
+            logger.debug("Message %s processed successfully", msg_id)
+        finally:
+            self._release_turn(room_id)
```
AlexanderZ-Band: issue: A detached prior turn releases whichever future is currently stored for the room, rather than the future it was created with. After cleanup removes the old entry, a rejoin can install a new turn's future; when the old task unwinds, this `finally` resolves the new future. `on_message` returns while that new query is still running, allowing another same-room turn to race it.

AlexanderZ-Band (commented):
Two lifecycle defects found: E2E workspaces leak, and an old detached turn can release a replacement turn after cleanup/rejoin.

tests/e2e/baseline/toolkit/builders.py
AlexanderZ-Band: Fixed in b060ba1: the bespoke smoke now scopes its directory with TemporaryDirectory, and matrix-created adapters retain a TemporaryDirectory cleanup finalizer.

src/band/adapters/claude_sdk.py
AlexanderZ-Band: Fixed in b060ba1: detached turns release only their captured future; cleanup cancels owned detached tasks, and a deterministic cleanup/rejoin regression test covers the ordering.

src/band/adapters/claude_sdk.py
```diff
@@ -180,12 +190,23 @@ async def _pre_tool_use_continue_hook(
...
+    the CLI falls back to its own ``permission_mode``-driven default instead
+    of ever consulting ``can_use_tool`` -- silently skipping chat-based manual
+    approval for native tools (Bash/Write/Edit) under the adapter's default
+    ``permission_mode="acceptEdits"``.
     """
-    return {"continue_": True}
+    return {
+        "hookSpecificOutput": {
+            "hookEventName": "PreToolUse",
+            "permissionDecision": "ask",
```
amit-gazal-band, line 211: **Blocking:** this hook is registered with no matcher (`HookMatcher(hooks=[_pre_tool_use_continue_hook])`, line 493), so it fires for *every* tool call — including `mcp__band__*`.

The installed SDK is explicit that a `PreToolUse` hook overrides `allowed_tools` (`claude_agent_sdk/types.py`: *"An allowed_tools entry that allows a whole tool auto-approves it before the callback is consulted. To gate every tool call, use a PreToolUse hook"*). Previously `{"continue_": True}` carried no decision, so `_mcp_backend.allowed_tools` (whole-tool entries `mcp__band__<name>`) auto-approved every Band tool before `can_use_tool` was consulted. With `"ask"` they all route into `_resolve_tool_permission`, which has no band-tool bypass.

Concretely:
- `approval_mode="manual"` → the agent cannot send a single chat reply without a human typing `/approve <token>` for the `band_send_message` call.
- `approval_mode="auto_decline"` → every Band tool is denied by policy, so the agent can never reply at all.

Fix by scoping the `HookMatcher` to native tools, or short-circuiting `MCP_TOOL_PREFIX`-prefixed names at the top of `_resolve_tool_permission`.

Also, while here: the comment at line 487 still says the hook returns `{"continue_": True}` "so the SDK delegates to can_use_tool" — i.e. it now describes the exact behavior this PR's docstring says is broken. Per the repo's comment-style rule it should state what the hook does today.

src/band/adapters/claude_sdk.py
```diff
@@ -678,32 +722,98 @@ async def on_message(
...
+        # early (see _resolve_manual_approval / _release_turn) so on_message
+        # can return and Band's strictly-sequential per-room message loop can
+        # dispatch the reply that will eventually resolve the approval. When
+        # nothing needs a human, the turn finishes before release fires and
+        # this is equivalent to the previous synchronous await.
+        release_future: asyncio.Future[None] = (
+            asyncio.get_running_loop().create_future()
+        )
+        self._turn_release[room_id] = release_future
+        turn_task = asyncio.create_task(
```
amit-gazal-band, line 742: No `CancelledError` handling around the detached turn. `OpencodeAdapter.on_message`, which this change mirrors, has an explicit `except asyncio.CancelledError:` that tears down turn state and aborts the session — precisely because "the runtime interrupts a turn by cancelling this coroutine, but the watcher runs detached".

Here, when the runtime cancels `on_message` (user interrupt, room teardown, shutdown), `turn_task` is left running: it keeps streaming Claude's response and posting messages into the room the user just stopped, and holds the new `_turn_tasks` busy-guard until it finishes.

src/band/adapters/claude_sdk.py
```diff
@@ -678,32 +722,98 @@ async def on_message(
...
+            self._run_turn(client, room_id, tools, full_message, msg.id, release_future)
+        )
+        self._turn_tasks[room_id] = turn_task
```
amit-gazal-band, line 745: `_turn_tasks` entries are never pruned on normal completion — only `on_cleanup`/`cleanup_all` remove them. So every room a turn ever ran in permanently retains a completed `Task`, keeping its `ClaudeSDKClient`, `tools` and the full prompt string alive.

`_run_turn`'s `finally` should drop the entry when it's still its own task (same identity check as the `_turn_release` cleanup just below).

src/band/adapters/claude_sdk.py
```diff
@@ -678,32 +722,98 @@ async def on_message(
...
+        Idempotent: a turn with several gated tool calls only needs the
+        first manual approval to release on_message, and the turn's own
+        completion (see _run_turn's finally) must release it too when
+        nothing ever blocked on a human.
+        """
+        current_release = self._turn_release.get(room_id)
+        if current_release is None or current_release.done():
+            return
+        if release_future is None or current_release is release_future:
+            current_release.set_result(None)
```
amit-gazal-band: This resolves via the `_turn_release[room_id]` dict, so when that entry is already gone it returns without ever touching the `release_future` it was handed.

Both cleanup paths hit that:
- `on_cleanup` pops `_turn_release[room_id]` (line 1240) and *then* cancels the turn task.
- `cleanup_all` clears the dict right after `cancel()` — cancellation is only scheduled, the task's `finally` runs later.

In both cases `_run_turn`'s `finally` finds no entry and never calls `set_result`, so an `on_message` still parked on `await release_future` (line 742) never returns and that room's message loop is wedged.

`_run_turn` should resolve its own `release_future` unconditionally — the dict identity check is only needed for the *dict cleanup*, not for the set.

src/band/adapters/claude_sdk.py
```diff
@@ -1125,6 +1235,10 @@ async def on_cleanup(self, room_id: str) -> None:
         self._room_last_sender.pop(room_id, None)
         self._notified_declines.pop(room_id, None)
         self._pending_tool_names.pop(room_id, None)
+        self._turn_release.pop(room_id, None)
+        turn_task = self._turn_tasks.pop(room_id, None)
+        if turn_task is not None:
+            turn_task.cancel()
```
amit-gazal-band: Ordering: `await self._session_manager.cleanup_session(room_id)` (line 1231) runs — and yields — while the detached `_run_turn` is still streaming. The turn then sees `CLIConnectionError`, calls `_invalidate_session`, and `_report_error` posts an `Error: ...` event into the room the agent has just left.

Cancel (and await) the turn task *before* closing the session.

src/band/adapters/claude_sdk.py
```diff
@@ -1152,6 +1266,10 @@ async def cleanup_all(self) -> None:
         self._room_last_sender.clear()
         self._notified_declines.clear()
         self._pending_tool_names.clear()
+        for turn_task in self._turn_tasks.values():
+            turn_task.cancel()
```
amit-gazal-band: Same ordering problem as `on_cleanup`: `await self._session_manager.stop()` (line 1258) runs before these cancels, so in-flight turns see a dead connection and report errors into their rooms on the way out. Cancel and await the turn tasks first, then stop the session manager.

tests/e2e/baseline/smoke/adapters/test_claude_sdk.py
```diff
@@ -0,0 +1,136 @@
+"""Claude SDK showcase smoke — the manual chat-based approval round trip.
+
+The generic matrix runs claude_sdk with no ``approval_mode`` set (the SDK's own
+``permission_mode`` governs instead; see ``toolkit/builders.py``), so the chat-based
+manual relay -- Claude Code's native ``Bash``/``Write``/``Edit`` tools gated through
+``can_use_tool``, the adapter posting an ``/approve <token>`` prompt to the room, a
+human replying, the turn resuming -- is otherwise never exercised live. Band's own
+MCP tools (and any caller ``additional_tools``) are swept into ``allowed_tools`` and
```
amit-gazal-band, line 8: This states the opposite of what the SDK documents — Band's MCP tools are *not* bypassing `can_use_tool` any more once the hook returns `"ask"` (see the comment on `claude_sdk.py:207`).

Worth noting the smoke can't catch the difference either: it only asserts on messages the adapter posts through `tools.send_message` directly, never on a model reply that would have gone through `band_send_message`. A case that forces the model to reply via the Band tool would turn this into real coverage.

tests/e2e/baseline/smoke/adapters/test_codex.py
```diff
@@ -0,0 +1,95 @@
...
+                room_id,
+                reasoning_joke_instruction(name),
+                mention_id=identity.id,
+                mention_name=identity.name,
+            )
+            replies = await capture.wait_for_reply(mid, identity.id)
+            thoughts = await capture.thoughts(sender_id=identity.id)
+
+    replies.assert_contains_any([name])
+    thoughts.assert_at_least(1)
```
amit-gazal-band: `thoughts.assert_at_least(1)` is model-dependent, and `@flaky_infra` is scoped to "transient live-turn timeout" — so a turn where the model returns no reasoning summary (legitimate even with `reasoning_summary="auto"`) hard-fails the backends lane instead of retrying.

Suggest making the vacuity guard tolerant here, and keeping the deterministic checks (`replies.assert_contains_any`, `thoughts.assert_contains_none(PLACEHOLDER_THOUGHTS)`) as the fail-loud part.

amit-gazal-band (changes requested):
Requesting changes — one blocking issue plus a cluster of turn-lifecycle races introduced by running the turn detached.

**Blocking:** switching the `PreToolUse` hook from `{"continue_": True}` to `permissionDecision: "ask"` also un-shadows Band's *own* MCP tools, because the hook is registered with no matcher. Under `approval_mode="manual"` the agent can no longer reply without a human approving `band_send_message`; under `auto_decline` it can never reply at all.

The rest: `_release_turn` can leave `on_message` parked forever, there's no `CancelledError` handling around the detached turn (the adapter this mirrors, `OpencodeAdapter`, has one), and both cleanup paths close the session before cancelling the turn. Details inline.

src/band/adapters/claude_sdk.py
AlexanderZ-Band: Fixed in 4ea04f0. _run_turn now removes its task only when the room map still points to that exact task, so completed turns do not retain the client/tools/prompt and a successor cannot be removed accidentally. Added a completion-pruning regression test.

src/band/adapters/claude_sdk.py
AlexanderZ-Band: Fixed in 4ea04f0. on_message now catches cancellation, cancels the detached turn, awaits its termination, and re-raises so a stopped runtime callback cannot leave Claude streaming or retain the busy guard. Added a deterministic cancellation regression test.

src/band/adapters/claude_sdk.py
AlexanderZ-Band: Fixed in 4ea04f0. _release_turn resolves the future passed by the turn even when cleanup has already removed the room entry; the map lookup is used only when no future is supplied. Added a regression test for a removed room state.

src/band/adapters/claude_sdk.py
AlexanderZ-Band: Fixed in 4ea04f0. cleanup_all now cancels and awaits every detached turn before stopping the session manager. Added an ordering regression test.

tests/e2e/baseline/smoke/adapters/test_codex.py
AlexanderZ-Band: Fixed in 4ea04f0. The Codex smoke no longer requires a model-dependent reasoning event when reasoning_summary=auto; it keeps the completed-reply and no-placeholder assertions, so empty summaries remain valid without weakening the regression guard.

src/band/adapters/claude_sdk.py
AlexanderZ-Band: Fixed in 4ea04f0. on_cleanup now cancels and awaits the detached turn before awaiting cleanup_session, preventing a dead-session error from being posted after room departure. Added an ordering regression test.

tests/e2e/baseline/smoke/adapters/test_claude_sdk.py
AlexanderZ-Band: Addressed in 4ea04f0. The live smoke remains intentionally focused on native Bash approval because Band MCP tools are meant to bypass this approval mode; a deterministic adapter test now guards that SDK matcher boundary, which the prior unscoped hook violated. The separate caller-supplied additional_tools policy is already tracked in INT-1455.

src/band/adapters/claude_sdk.py
AlexanderZ-Band: Fixed in 4ea04f0. The PreToolUse HookMatcher now uses a native-tool-only matcher, so mcp__band__* calls retain their intentional allowed-tools bypass while Bash/Write/Edit still reach can_use_tool. The stale comment was corrected, and the adapter test verifies Bash matches while band_send_message does not.

src/band/adapters/claude_sdk.py
```diff
@@ -591,6 +628,19 @@ async def on_message(
                     )
                     return
 
+        # A prior turn's background task (see _run_turn) may still be running
+        # this room's ClaudeSDKClient -- most commonly while it's parked on a
+        # manual approval. A second concurrent client.query() on the same
+        # session is unsafe, so refuse rather than race it (mirrors
+        # OpencodeAdapter's own "still processing" guard).
+        running_turn = self._turn_tasks.get(room_id)
+        if running_turn is not None and not running_turn.done():
```
AlexanderZ-Band, line 646: suggestion: this new "still processing" refusal guard (a second `on_message` call while `_turn_tasks[room_id]` is not done gets refused rather than starting a second turn) and `_log_turn_task_exception` (src/band/adapters/claude_sdk.py:834) are both new behavior introduced by this diff with no unit test coverage. Both are cheap to test: seed `_turn_tasks[room_id]` with a not-done task and assert a second `on_message` call sends "Still processing the previous request in this room." without starting a new turn (`get_or_create_session` never called); and drive `_log_turn_task_exception` with a task that raised, confirming the exception is retrieved.

AlexanderZ-Band (commented):
my-code-review pass 1 (Style/Design/Runtime/Test/Edge-case/Logic angles applied to the current diff): one non-blocking test-coverage suggestion. Everything else checked clean — most of the risky surface (turn detachment, cancellation, cleanup ordering, the native-tool hook matcher) was already hardened by the prior review round on this PR.

src/band/adapters/claude_sdk.py
AlexanderZ-Band: Fixed: added `test_refuses_second_message_while_a_turn_is_running` (TestOnMessage) covering the still-processing refusal guard, and `test_log_turn_task_exception_skips_cancelled_tasks` / `test_log_turn_task_exception_retrieves_a_failed_turns_exception` (TestOnCleanup) covering `_log_turn_task_exception`. Full adapter test file passes (133 passed).

eric-descourtis-thenvoi:
One gap in the new cancel path: `_cancel_turn()` (and the `except asyncio.CancelledError` in `on_message`) cancels the Python turn task, but it never calls `ClaudeSDKClient.interrupt()`. The CLI therefore keeps running the turn after a room Stop/Interrupt (or a `max_cycle_seconds` timeout from #637).

We hit this on a self-hosted agent running 3.1.1, which has the same behavior without the detached task. After Stop, the abandoned turn kept calling `band_send_message` and got `403 Agent execution is stopped`. After Play, the redelivered message was matched to the abandoned turn's `ResultMessage` and the room got a spurious "finished a turn without calling band_send_message". The detached task in this PR makes a CLI turn running with nobody reading it more likely, not less.

Suggestion: before or instead of `turn_task.cancel()`, `await client.interrupt()`, then let `_run_turn` read until the interrupted turn's `ResultMessage`, with a timeout of about 10s. If it times out, `_invalidate_session(room_id)`. Full write-up, with the related Monitor/background-task deadlock, is in #655.

amit-gazal-band (approved):

PR #636, opened 2026-09-14
ci: add python-core-coverage gate for band-sdk-core-core

AlexanderZ-Band:
## Summary

Adds a CI gate proving this repo's real test suite exercises at least 80% of `band-sdk-core-core`'s lines through the PyO3 binding — separate from this SDK's own `coverage.py` numbers, from `band-sdk-core`'s own Rust-only floor, and from TypeScript/wasm coverage.

- `.github/workflows/python-core-coverage.yml`: checks out this repo, `uv sync`'s it, resolves the pinned `band-sdk-core` version via `importlib.metadata`, checks that exact `band-sdk-core-core-v{version}` tag out as a sibling (never silently tests against `band-sdk-core`'s `main`, which is already ahead), and runs `just coverage-python-consumer ../band-sdk-python` from there. Uploads the LCOV/HTML/summary artifact and writes a per-module job summary.
- Triggers on PRs to `main` touching `src/**`, `tests/**`, `pyproject.toml`, or `uv.lock`, plus `merge_group` — same shape as this repo's existing `ci.yml`.

## Baseline

**84.75% lines**, measured against a real checkout at `b060ba19` (5593 tests passed, 610 skipped). Full per-module breakdown and the `classify_close`/`classify_upgrade` finding are on [INT-1304](https://linear.app/thenvoi/issue/INT-1304/add-80percent-band-sdk-core-coverage-gate-from-band-sdk-python-tests).

No ratchet: all six Python-adoption tickets (INT-1236, INT-1245, INT-1300, INT-1301, INT-1302, INT-1303) are Done and tested, so 80% is achievable today.

## Follow-up (repo admin, out of scope here)

Whether this becomes a *required* status check is a branch-protection setting, not something this PR can set — flagging for whoever owns `main`'s protection rules once this has run green a few times.

## Test plan

- [x] Companion PR (band-ai/band-sdk-core#71) implements and verifies `coverage-python-consumer` directly: local dry run against a real band-sdk-python worktree returns 84.75% lines and passes the gate; `--fail-under-lines 99` against the same data returns a non-zero exit, confirming the gate is load-bearing.
- [ ] This workflow itself needs one real PR run to confirm the cross-repo checkout/tag-resolution steps work under actual GitHub Actions (not just the local dry run) — watch this PR's own run.

Companion PR: band-ai/band-sdk-core#71.

Closes INT-1304.

https://claude.ai/code/session_01TmtjU5tgPv9Y6bd238JSBL


.github/workflows/python-core-coverage.yml
```diff
@@ -0,0 +1,115 @@
...
+  coverage:
+    runs-on: ubuntu-latest
+    timeout-minutes: 30
+    steps:
+      - name: Checkout band-sdk-python
+        uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7
+        with:
+          path: band-sdk-python
+
+      - name: Configure git to use HTTPS for GitHub
```
AlexanderZ-Band, line 47: suggestion: This `git config --global url."https://github.com/".insteadOf "git@github.com:"` line duplicates a line already extracted into `.github/actions/setup-e2e/action.yml` (used by every `e2e.yml` job) — now a 6th copy of this literal in the repo. Consider pulling just this line into its own tiny composite action so a future change to GitHub HTTPS auth needs one edit, not six.

.github/workflows/python-core-coverage.yml
```diff
@@ -0,0 +1,115 @@
...
+
+      - name: Run band-sdk-python's tests against instrumented band-sdk-core-core
+        working-directory: band-sdk-core
+        run: just coverage-python-consumer ../band-sdk-python
+
+      - name: Write coverage summary
+        if: always()
+        working-directory: band-sdk-core
+        run: |
+          summary=target/python-consumer-coverage/report/summary.txt
```
AlexanderZ-Band: nit: `target/python-consumer-coverage/report/` is hardcoded 4 times in this file (here and 3x in the Upload step's `path:` block below). Set a job-level `env: REPORT_DIR: ...` and reference `${{ env.REPORT_DIR }}` in both places so the directory lives in one spot.

.github/workflows/python-core-coverage.yml
```diff
@@ -0,0 +1,115 @@
...
+          if [ -f "$summary" ]; then
```
AlexanderZ-Band, line 133: issue: This step only handles the case where `summary.txt` exists (`if [ -f "$summary" ]; then ... fi`) — no else branch. If the coverage step fails before producing a summary (failed pytest, a tripped guard, a failed maturin build), this always-run step silently no-ops: nothing is appended to the job summary and it exits 0. A reviewer checking the PR's job summary first sees a blank coverage section with no pointer to why. Add an else branch appending a short note ("No coverage summary produced — see the coverage step's log above") so the summary itself carries a diagnostic trail.

.github/workflows/python-core-coverage.yml
```diff
@@ -0,0 +1,115 @@
...
+          client-id: ${{ secrets.APP_CLIENT_ID }}
+          private-key: ${{ secrets.APP_PRIVATE_KEY }}
+          owner: band-ai
+          repositories: band-sdk-core
+
+      - name: Checkout band-sdk-core at the pinned version
+        uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7
+        with:
+          repository: band-ai/band-sdk-core
+          ref: band-sdk-core-core-v${{ steps.pin.outputs.version }}
```
AlexanderZ-Band: issue: The tag-ref interpolation (`ref: band-sdk-core-core-v${{ steps.pin.outputs.version }}`, built from the `pin` step's `version` output) has no regression coverage, even though a lightweight structural test for exactly this shape is feasible and already has precedent in this repo — `tests/docker/test_kit_publish_workflow.py` parses workflow YAML with `yaml.BaseLoader` (via `tests/paths.py`'s `REPO_ROOT`) and asserts on interpolated run-script strings built from step outputs. Without it, a future step-id rename, output-key rename, or tag-prefix typo is caught only by a live cross-repo Actions run against real GitHub App auth, far slower feedback than a local test.

AlexanderZ-Band: issue: This ref is built directly from `importlib.metadata.version('band-sdk-core')` with no check that the resolved version is a final release. band-sdk-core's `publish.yml` dev mode publishes PEP 440 dev pre-releases (e.g. `2.5.0.dev3`) to PyPI that are never committed or tagged. If `pyproject.toml` is ever pinned to one of these (e.g. to validate an unreleased core change before its real release — this workflow's own `paths: pyproject.toml` trigger fires on exactly that edit), the cross-repo checkout fails on a nonexistent ref with a generic git error, not an explanation that dev-pinned versions aren't supported by this gate's tag-based resolution.

AlexanderZ-Band: issue: Proven via `git ls-remote`: none of band-sdk-core's six actually-published PyPI dev pre-releases (`0.2.1.dev1/2/4`, `0.4.0.dev0/1/2`) has a matching `band-sdk-core-core-v*` tag — this isn't hypothetical, the dev-publish path has fired for real with exactly this consequence. Reject/skip with a clear message when the resolved version is a pre-release (e.g. `packaging.version.Version(resolved).is_prerelease`), before constructing the checkout ref, rather than letting `actions/checkout` fail opaquely on a missing ref.

AlexanderZ-Band (commented):
Automated review (my-code-review): 6 findings — 4 issue, 1 nit, 1 suggestion. See inline comments. (Submitted as COMMENT, not REQUEST_CHANGES: GitHub blocks requesting changes on one's own PR.)

.github/workflows/python-core-coverage.yml
AlexanderZ-Band: Fixed in e61dcddc: extracted into `.github/actions/git-https-auth`, a one-step composite action, used from this workflow. One place to change GitHub HTTPS auth from here on.

.github/workflows/python-core-coverage.yml
AlexanderZ-Band: Fixed in e61dcddc: added a job-level `env: REPORT_DIR: target/python-consumer-coverage/report`, referenced via `$REPORT_DIR` (shell) / `${{ env.REPORT_DIR }}` (workflow expressions) everywhere it previously appeared as a literal.

.github/workflows/python-core-coverage.yml
AlexanderZ-Band: Fixed in e61dcddc: added an else branch that appends "No coverage summary produced -- see the coverage step's log above." to $GITHUB_STEP_SUMMARY when summary.txt is missing.

.github/workflows/python-core-coverage.yml
AlexanderZ-Band: Fixed in e61dcddc: added tests/test_python_core_coverage_workflow.py (mirrors tests/docker/test_kit_publish_workflow.py's yaml.BaseLoader pattern) asserting the pin step's shape, the checkout ref interpolation, the new pre-release guard, and step ordering. Verified it fails against the pre-fix workflow and passes against the fixed one.

AlexanderZ-Band: Fixed in e61dcddc: the "Resolve pinned band-sdk-core version" step now checks `packaging.version.Version(...).is_prerelease` and exits with a clear message before the checkout step ever runs, instead of letting the cross-repo checkout fail opaquely on a nonexistent ref. (Same root cause as the sibling Logical-Bugs finding below -- fixed together, one guard.)

AlexanderZ-Band: Fixed in e61dcddc, together with the sibling Edge-Case-Flow finding above (same root cause, one guard): the pin-resolution step now rejects any pre-release version before constructing the checkout ref. Verified locally: the guard exits 0 for a stable version (2.2.0) and exits 1 with the intended message for a dev pre-release (2.5.0.dev3), including under `set -e` semantics matching the real step (confirmed the fix doesn't fall into the classic bash pitfall where a failed command substitution inside an assignment doesn't trip `-e` -- the guard runs as its own statement, not inside the $(...) assignment).

.github/workflows/python-core-coverage.yml
```diff
@@ -0,0 +1,142 @@
...
+        with:
+          tool: just,cargo-llvm-cov
+
+      - name: Run band-sdk-python's tests against instrumented band-sdk-core-core
+        working-directory: band-sdk-core
+        run: just coverage-python-consumer ../band-sdk-python
+
+      - name: Write coverage summary
+        if: always()
+        working-directory: band-sdk-core
```
AlexanderZ-Band: issue: When the pre-release guard (lines 76-87) trips, or any earlier failure prevents "Checkout band-sdk-core at the pinned version" (lines 100-106) from running, `band-sdk-core/` never exists on disk. "Write coverage summary" still runs (`if: always()`, line 118) with `working-directory: band-sdk-core` (line 119) — GitHub Actions fails a step outright when its working-directory doesn't exist, before the run script body executes. So instead of reaching the intended else-branch message ("No coverage summary produced..."), the step errors out with a working-directory-not-found failure, defeating the graceful-degradation purpose the else branch exists for — and the new pre-release guard makes this routine (any dev pinning a prerelease band-sdk-core now reliably hits it).

Fix: give "Checkout band-sdk-core at the pinned version" an `id` and gate the summary/upload steps on `steps.<id>.outcome == 'success'` (combined with `always()`), or drop this step's `working-directory` and guard with `[ -d band-sdk-core ]` before referencing the path directly.

tests/test_python_core_coverage_workflow.py
```diff
@@ -0,0 +1,47 @@
...
+
+def test_pin_step_resolves_version_into_github_output() -> None:
+    step = _step(load_workflow(), "Resolve pinned band-sdk-core version")
+    assert step["id"] == "pin"
+    assert "importlib.metadata.version('band-sdk-core')" in step["run"]
+    assert '>> "$GITHUB_OUTPUT"' in step["run"]
+    assert "version=$version" in step["run"]
+
+
+def test_pin_step_rejects_prerelease_versions_before_checkout() -> None:
```
AlexanderZ-Band: issue: `test_pin_step_rejects_prerelease_versions_before_checkout` only checks string presence of `is_prerelease`/`packaging.version` inside the step's `run:` text — it doesn't execute the guard, so it would still pass even if the condition were inverted or the `sys.exit` were dropped. The guard logic itself is correct (independently verified), but this test doesn't prove it.

Fix: extract the embedded Python snippet from the loaded YAML and run it via `subprocess.run([sys.executable, "-c", script], ...)` with a substituted stable and a substituted pre-release version string, asserting the real exit codes (0 and 1).

.github/workflows/python-core-coverage.yml
```diff
@@ -0,0 +1,142 @@
...
+
+          # A failed command substitution inside an assignment does not trip
+          # bash's `set -e`, so the pre-release guard below runs as its own
+          # statement -- a non-zero exit there aborts the step for real,
+          # instead of silently continuing with an empty $version.
+          uv run python -c "
+          import sys
+          from packaging.version import Version
+
+          if Version('$version').is_prerelease:
```
AlexanderZ-Band: issue: `$version` is spliced via bash double-quote expansion directly into three embedded Python string literals (line 80, and the two message fragments at lines 82-83) before the text reaches the Python parser. Traced every place `$version` can originate (band-sdk-core's release-please Cargo.toml bump + `set_dev_version.py`'s `re.fullmatch(r"\d+\.\d+\.\d+-dev\.\d+")` guard, plus PEP 440's own PyPI-enforced grammar) — today this is provably not exploitable, since nothing in that chain can produce a quote, backslash, or whitespace. But the guarantee lives entirely in a different repository's scripts, untested by this workflow; if band-sdk-core's version format or validation ever changes, nothing here would catch a value capable of breaking out of the embedded literal.

Fix: pass `$version` through the environment instead of string-splicing it, e.g. `env: VERSION: ${{ ... }}` / `VERSION="$version"` and `Version(os.environ['VERSION'])` inside the -c script — removes the injection class entirely regardless of future version-format changes, at no cost to the existing guard.

.github/workflows/python-core-coverage.yml
```diff
@@ -0,0 +1,142 @@
...
+              sys.exit(
+                  'band-sdk-core==$version is a pre-release: band-sdk-core-core-v$version '
```
AlexanderZ-Band: nit: the tag prefix `band-sdk-core-core-v` is now duplicated — once in this guard's error message (lines 82-83), again independently in the checkout `ref:` at line 104. Same class of duplication the REPORT_DIR fix already addressed elsewhere in this file, not applied here.

Fix: hoist into the job-level `env:` block, e.g. `CORE_TAG_PREFIX: band-sdk-core-core-v`, and reference `${{ env.CORE_TAG_PREFIX }}` in both places.

.github/workflows/python-core-coverage.yml
```diff
@@ -0,0 +1,142 @@
...
+        id: pin
+        working-directory: band-sdk-python
+        run: |
+          version=$(uv run python -c "import importlib.metadata; print(importlib.metadata.version('band-sdk-core'))")
+
+          # A failed command substitution inside an assignment does not trip
+          # bash's `set -e`, so the pre-release guard below runs as its own
+          # statement -- a non-zero exit there aborts the step for real,
+          # instead of silently continuing with an empty $version.
+          uv run python -c "
```
AlexanderZ-Band: suggestion: `$version` is bash-interpolated into the embedded Python three separate times (lines 80, 82, 83). Interpolating once into a local Python variable and referencing that for the rest of the message would be a smaller, more readable diff than the env-var fix for the finding above — worth doing together if that fix lands.

tests/test_python_core_coverage_workflow.py
```diff
@@ -0,0 +1,47 @@
+"""Structural contracts for python-core-coverage.yml."""
```
AlexanderZ-Band: nit: this file's docstring ("Structural contracts for python-core-coverage.yml") reads as covering the whole workflow, but doesn't test the REPORT_DIR single-source-of-truth or the "Write coverage summary" else-branch fallback message added in cycle 1. Consider 1-2 more small structural assertions, or narrow the docstring's claimed scope.

.github/actions/git-https-auth/action.yml
```diff
@@ -0,0 +1,11 @@
...
+description: >-
+  Rewrites git@github.com: URLs to https://github.com/ so a later checkout
+  or clone authenticates over HTTPS (e.g. with a token) instead of SSH.
+
+runs:
+  using: composite
+  steps:
+    - name: Configure git to use HTTPS for GitHub
+      shell: bash
+      run: git config --global url."https://github.com/".insteadOf "git@github.com:"
```
AlexanderZ-Band, line 11: nit: this composite's purpose (deduplicating the git-HTTPS-auth line) isn't yet achieved repo-wide — the literal line still exists inline in 6 other places (`.github/actions/setup-e2e/action.yml:14`, `.github/workflows/ci.yml:23,79,127,190,236`). Out of scope for this PR (those sites predate it), but flagging since the "single source of truth" this action implies doesn't exist in practice yet.

Optional followup: point setup-e2e and ci.yml's inline copies at `uses: ./.github/actions/git-https-auth`.

.github/workflows/python-core-coverage.yml
```diff
@@ -0,0 +1,142 @@
...
+          import sys
+          from packaging.version import Version
```
AlexanderZ-Band, line 83: nit: `from packaging.version import Version` relies on `packaging`, which isn't declared anywhere in band-sdk-python's own manifest (neither base `dependencies` nor the `dev` extra) — it's only present transitively today (via pytest and others, confirmed in uv.lock). This gating step depends on that transitive presence with no comment recording the assumption, unlike this repo's other transitive-dependency notes (e.g. the `mcp` floor-pin comments in pyproject.toml).

Fix: either add `packaging` as an explicit dev-extra dependency, or leave a one-line comment noting it rides in transitively so a future dependency bump that drops it fails loud instead of with a bare ImportError.

AlexanderZ-Band (commented):
Cycle 2 of /my-code-review (fixed-criteria pass, second cycle after cycle-1's 6 fixes). 8 new findings: 1 real bug (E, graceful-degradation break when the checkout is skipped), 1 test-quality gap (D), 1 runtime-safety defense-in-depth note (C), 3 style/hygiene (A), 2 design/reuse (B). Logical Bugs (F) re-verified guard polarity, the injection surface, and multi-line -c argument passing empirically and found nothing further. (Submitted as COMMENT, not REQUEST_CHANGES: GitHub blocks requesting changes on one's own PR.)

.github/workflows/python-core-coverage.yml
AlexanderZ-Band: Fixed: dropped the step's `working-directory` and reference the path directly as `band-sdk-core/$REPORT_DIR/summary.txt`. Verified live: with the checkout present, the summary writes correctly; with it absent (simulating the guard/checkout failing), the step now exits 0 and writes the intended 'No coverage summary produced' fallback instead of erroring on a missing cwd. Regression-guarded by `test_write_coverage_summary_has_no_working_directory`.

tests/test_python_core_coverage_workflow.py
AlexanderZ-Band: Fixed: added `_run_pin_guard`, which extracts the real guard script from the loaded workflow YAML and executes it via subprocess against literal version strings, asserting real exit codes and stderr messages -- `test_pin_guard_passes_for_a_stable_version`, `test_pin_guard_rejects_dev_prerelease_versions`, `test_pin_guard_rejects_cargo_style_dev_versions`. Confirmed these fail against the pre-fix workflow (via git stash) and pass against the current one.

.github/workflows/python-core-coverage.yml
AlexanderZ-Band: Fixed: $version is now passed to the embedded Python via `VERSION="$version" uv run python -c "..."` and read with `os.environ['VERSION']`, instead of being bash-spliced into the script text -- removes the injection class entirely regardless of future band-sdk-core version-format changes. Live-verified against stable, PEP-440 dev, and Cargo-style dev version strings (all produce correct exit codes and messages).

.github/workflows/python-core-coverage.yml
AlexanderZ-Band: Fixed: hoisted the tag prefix into a job-level `env: CORE_TAG_PREFIX: band-sdk-core-core-v`, referenced from both the guard's error message (via `os.environ['CORE_TAG_PREFIX']`) and the checkout step's `ref:`. Regression-guarded by `test_core_tag_prefix_is_a_single_source_of_truth`.

.github/workflows/python-core-coverage.yml
AlexanderZ-Band: Fixed as part of the same change as the injection-surface fix above: $version is now interpolated into the embedded Python exactly once (`os.environ['VERSION']` assigned to a local `version` variable), not three separate bash splices.

tests/test_python_core_coverage_workflow.py
AlexanderZ-Band: Fixed by adding coverage rather than narrowing the docstring: `test_report_dir_is_a_single_source_of_truth`, `test_write_coverage_summary_has_no_working_directory`, `test_write_coverage_summary_falls_back_when_missing`, and `test_core_tag_prefix_is_a_single_source_of_truth` now cover the REPORT_DIR single-source-of-truth and the else-branch fallback this file's docstring already claimed.

.github/actions/git-https-auth/action.yml
AlexanderZ-Band: Agreed, and out of scope for this PR (none of the 6 duplicate sites are touched by this diff). Filed as a followup: https://linear.app/thenvoi/issue/INT-1470/adopt-githubactionsgit-https-auth-composite-across-existing-duplicated

.github/workflows/python-core-coverage.yml
AlexanderZ-Band: Fixed: added `packaging>=23.0` as an explicit `dev`-extra dependency in pyproject.toml (with a comment noting it's for the CI coverage-gate version guard) and regenerated uv.lock, instead of relying on its transitive presence via pytest.

.github/workflows/python-core-coverage.yml
```diff
@@ -0,0 +1,185 @@
...
+
+          echo "version=$version" >> "$GITHUB_OUTPUT"
+
+      - name: Checkout band-sdk-core at the pinned version
+        uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7
+        with:
+          repository: band-ai/band-sdk-core
+          ref: ${{ env.CORE_TAG_PREFIX }}${{ steps.pin.outputs.version }}
+          path: band-sdk-core
+          token: ${{ secrets.CORE_SDK_READ_KEY }}
```
amit-gazal-band, line 111: **Blocking: every Dependabot PR will fail here.**

Dependabot-triggered runs resolve secrets from GitHub's *separate* Dependabot secret store, not the Actions store. `gh secret list --app dependabot` on this repo returns empty, while `CORE_SDK_READ_KEY` exists only in the Actions store — so on a Dependabot run this resolves to an empty string and the checkout of the private `band-ai/band-sdk-core` fails.

And those PRs *will* reach this workflow: `.github/dependabot.yml` has a `uv` ecosystem at `/`, so every weekly bump touches `pyproject.toml` + `uv.lock` — exactly the `paths` filter on lines 20-24. Result: every Monday's dependency PRs go red on a failure that has nothing to do with the bump.

Either mirror `CORE_SDK_READ_KEY` into the Dependabot secret store, or skip the job for Dependabot:

```yaml
  coverage:
    if: github.actor != 'dependabot[bot]'
```

.github/workflows/python-core-coverage.yml
```diff
@@ -0,0 +1,185 @@
...
+  # observes main's current coverage without adding PR notifications.
+  schedule:
+    - cron: "33 20 * * 0" # Sunday night: 23:33 Israel summer time, 22:33 winter time
+  pull_request:
+    branches: [main]
+    paths:
+      - 'src/**'
+      - 'tests/**'
+      - 'pyproject.toml'
+      - 'uv.lock'
```
amit-gazal-band, line 27: This filter can't gate its own tooling: `.github/workflows/python-core-coverage.yml`, `.github/scripts/post-core-coverage-digest.py`, and `.github/actions/git-https-auth/action.yml` are all absent from it. A follow-up PR touching only those files merges without this workflow ever running — which is exactly the shape of PR most likely to follow this one.

```yaml
      - '.github/workflows/python-core-coverage.yml'
      - '.github/scripts/post-core-coverage-digest.py'
      - '.github/actions/git-https-auth/**'
```

.github/workflows/python-core-coverage.yml
```diff
@@ -0,0 +1,185 @@
...
+        working-directory: band-sdk-python
+        run: |
+          version=$(uv run python -c "import importlib.metadata; print(importlib.metadata.version('band-sdk-core'))")
+
+          # A failed command substitution inside an assignment does not trip
+          # bash's `set -e`, so the pre-release guard below runs as its own
+          # statement -- a non-zero exit there aborts the step for real,
+          # instead of silently continuing with an empty $version. $version
+          # is passed via the environment, not spliced into the script text,
+          # so it can never break out of the embedded Python literal.
```
amit-gazal-band: The premise here isn't right. For a simple command consisting only of assignments, the exit status *is* that of the last command substitution, and `set -e` does abort on it — GitHub's default shell is `bash --noprofile --norc -eo pipefail`, so it applies:

```
$ bash -eo pipefail -c 'version=$(false); echo REACHED'
$ echo $?
1
```

`REACHED` never prints. The code is fine either way, but the comment documents a hazard that doesn't exist, and would mislead the next editor into thinking `$version` can silently arrive empty. (`local version=$(false)` *is* the real exception — there the status is `local`'s — but that's not what's happening at line 76.)

Suggest trimming to the part that's both true and non-obvious: that `$version` is passed via the environment rather than spliced into the script text, so it can't break out of the embedded Python literal.

.github/workflows/python-core-coverage.yml
```diff
@@ -0,0 +1,185 @@
...
+    permissions:
+      contents: write
+    env:
+      GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+      REPO: ${{ github.repository }}
+      SHA: ${{ github.sha }}
+      RUN_URL: ${{ github.server_url }}/${{ github.repository }}/actions/runs/${{ github.run_id }}
+      WORKFLOW_RESULT: ${{ needs.coverage.result }}
+      REPORT_LABEL: band-sdk-core Python bindings
+      LCOV_PATH: artifacts/python-consumer.lcov
```
amit-gazal-band, line 173: Minor/optional: this path depends on an implicit calculation. `upload-artifact` roots the archive at the least-common-ancestor of the *matched* files, which resolves to `.../report/` today only because `python-consumer.lcov` and `summary.txt` sit directly beside `html/` (lines 145-147).

If a future `coverage-python-consumer` ever emits only the HTML tree, or moves the lcov one level down, the LCA shifts and `artifacts/python-consumer.lcov` silently stops existing. The digest then renders its "⚠️ Coverage unavailable" branch even though the artifact has data, and nothing fails. Staging the three files into a fixed directory before upload would pin the layout.

.github/workflows/python-core-coverage.yml
```diff
@@ -0,0 +1,185 @@
...
+      - name: Checkout code
+        uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7
+
+      - name: Download coverage report
+        continue-on-error: true
+        uses: actions/download-artifact@3e5f45b2cfb9172054b4087a40e8e0b5a5461e7c # v8
+        with:
+          name: python-core-coverage
+          path: artifacts
+          if-no-files-found: warn
```
amit-gazal-band: `if-no-files-found` is an input of `upload-artifact` only — `download-artifact`'s inputs are `name`, `artifact-ids`, `path`, `pattern`, `merge-multiple`, `github-token`, `repository`, `run-id`, `skip-decompress`, `digest-mismatch`.

So this does nothing except emit an "Unexpected input(s)" warning annotation on every weekly run. The thing actually keeping the job alive when the coverage job produced no artifact is `continue-on-error: true` on line 170. Worth deleting so it doesn't read as a working guard.

.github/scripts/post-core-coverage-digest.py
```diff
@@ -0,0 +1,202 @@
...
+        else:
+            ranges.append(str(start) if start == previous else f"{start}-{previous}")
+            start = previous = number
+    if start is not None:
+        ranges.append(str(start) if start == previous else f"{start}-{previous}")
+    shown = ", ".join(ranges[:MAX_MISSED_LINE_RANGES])
+    return (
+        shown
+        if len(ranges) <= MAX_MISSED_LINE_RANGES
+        else f"{shown}, … ({len(numbers)} missed)"
```
amit-gazal-band: The truncation suffix reports the wrong quantity: `numbers` is the full `missed_lines` tuple, so `len(numbers)` is the file's *total* missed-line count, not the number of elided ranges.

Rendered at line 161 it lands right next to `{item.missed} lines missing` in the same cell — a file with 43 missed lines spread over 20 ranges prints `…, (43 missed)` beside `43 lines missing`. The same number twice, and a reader reasonably parses the parenthetical as "43 *more*".

```suggestion
        else f"{shown}, … ({len(ranges) - MAX_MISSED_LINE_RANGES} more ranges)"
```

.github/actions/git-https-auth/action.yml
amit-gazal-band, line 11: Extracting this is the right instinct, but nothing adopts it — the identical line still appears inline at `ci.yml:23,79,127,190,236` and `.github/actions/setup-e2e/action.yml:14`. Introducing the composite action while leaving those six in place makes it *seven* definitions rather than one, which is the opposite of the single-source-of-truth intent.

Either switch those six callers to `uses: ./.github/actions/git-https-auth`, or drop the new action and inline the one line in `python-core-coverage.yml` like everywhere else. Adopting it is the better direction, and it's a mechanical change.

amit-gazal-band (changes requested):
Nice gate — the pin-guard-before-checkout ordering and the `Write coverage summary` step deliberately having no `working-directory` are both good calls, and the real run on this PR proves the cross-repo checkout works (3m35s cold, so the 30-minute timeout has plenty of headroom).

One blocker plus a handful of small ones, details inline.

**Blocking:** the Dependabot secret store is empty, so every weekly `uv` bump PR will fail this gate at the private-repo checkout.

AlexanderZ-Band (commented):
Personal review pass 1 (my-code-review, six angle-groups). 6 new findings beyond the 21 already-resolved threads (all independently re-verified as genuinely fixed, including the digest truncation-count fix and the download-artifact if-no-files-found removal against the real actions/download-artifact@v8 action.yml). 2 issue / 1 suggestion / 3 nit. Proceeding to fix the actionable ones.

.github/scripts/post-core-coverage-digest.py
```diff
@@ -0,0 +1,202 @@
...
+        shown
+        if len(ranges) <= MAX_MISSED_LINE_RANGES
+        else f"{shown}, … ({len(ranges) - MAX_MISSED_LINE_RANGES} more ranges)"
+    )
+
+
+def coverage_marker(percent: float) -> str:
+    if percent >= LOW_COVERAGE_PERCENT:
+        return "🟢"
+    if percent >= 50:
```
AlexanderZ-Band: nit: `coverage_marker`'s orange threshold is a bare `50` literal, while its green threshold reuses the named `LOW_COVERAGE_PERCENT` constant defined above for the exact same kind of percent comparison. Extract a `MID_COVERAGE_PERCENT = 50.0` constant for consistency.

.github/workflows/python-core-coverage.yml
```diff
@@ -0,0 +1,193 @@
...
+              echo "## band-sdk-core-core coverage from band-sdk-python's tests"
+              echo '```'
+              cat "$summary"
+              echo '```'
+            } >> "$GITHUB_STEP_SUMMARY"
+          else
+            echo "No coverage summary produced -- see the coverage step's log above." >> "$GITHUB_STEP_SUMMARY"
+          fi
+
+      - name: Stage coverage report
```
AlexanderZ-Band, line 142: issue: The "Stage coverage report" step's three conditional copies (lcov/summary.txt/html) and "Write coverage summary"'s if/else fallback are only checked by grepping for `$REPORT_DIR`/`$ARTIFACT_DIR` substrings in the step text (`test_report_dir_is_a_single_source_of_truth`) — neither script is ever actually executed against a real directory tree. This file already proves the pattern out for the pin-guard step via real subprocess execution (`_run_pin_guard`); a wrong path, bad quoting, or inverted `-f`/`-d` check here would today silently drop the artifact or feed the weekly digest an empty/wrong LCOV with no test failing.

.github/scripts/post-core-coverage-digest.py
```diff
@@ -0,0 +1,202 @@
...
+            found = hit = functions_found = functions_hit = 0
+            missed_lines = []
+    return records
+
+
+def render_digest(
+    *, lcov_path: Path, label: str, recipients: str, run_url: str, result: str
+) -> str:
+    header = "## 📊 Weekly Core coverage"
+    if not lcov_path.is_file():
```
AlexanderZ-Band, line 115: issue: `render_digest`'s `if not lcov_path.is_file()` branch — the path that fires when the coverage job fails, or the artifact-download step (`continue-on-error: true`) comes up empty — has zero test coverage. `test_weekly_digest_identifies_low_and_completely_uncovered_files` only exercises the happy path with a real lcov file present, but this degraded path is the most likely real-world failure mode this script exists to report.

.github/scripts/post-core-coverage-digest.py
```diff
@@ -0,0 +1,202 @@
...
+        ranges.append(str(start) if start == previous else f"{start}-{previous}")
+    shown = ", ".join(ranges[:MAX_MISSED_LINE_RANGES])
+    return (
+        shown
+        if len(ranges) <= MAX_MISSED_LINE_RANGES
+        else f"{shown}, … ({len(ranges) - MAX_MISSED_LINE_RANGES} more ranges)"
+    )
+
+
+def coverage_marker(percent: float) -> str:
```
AlexanderZ-Band, line 66: suggestion: `coverage_marker` is never called directly, and none of the percentages reached indirectly through the one digest test (50.00%, 60.00%, 33.33%, 20.00%, 0.00%) land at or above `LOW_COVERAGE_PERCENT` (80.0) — so the 🟢 branch, and both boundary values (49.99/50.0, 79.99/80.0), are never checked.

.github/scripts/post-core-coverage-digest.py
```diff
@@ -0,0 +1,202 @@
...
+    @property
+    def percent(self) -> float:
+        return 100 * self.hit / self.found if self.found else 100.0
+
+    @property
+    def missed(self) -> int:
+        return self.found - self.hit
+
+
+def display_path(path: str) -> str:
```
AlexanderZ-Band, line 42: nit: `display_path`'s fallback branch (`Path(path).name`, used when `/crates/` isn't in the source path) is never exercised — the test fixture's lcov paths all contain `/crates/`.

tests/test_python_core_coverage_workflow.py
```diff
@@ -0,0 +1,265 @@
...
+    assert "$REPORT_DIR" in stage_run
+    assert "$ARTIFACT_DIR" in stage_run
+    assert upload_paths == "${{ env.ARTIFACT_DIR }}"
+    assert report_dir not in summary_run
+    assert report_dir not in stage_run
+
+
+def test_coverage_artifact_has_a_single_staging_root() -> None:
+    coverage = load_workflow()["jobs"]["coverage"]
+    assert coverage["env"]["ARTIFACT_DIR"] == "python-core-coverage-artifact"
```
AlexanderZ-Band: nit: This test only asserts `ARTIFACT_DIR == "python-core-coverage-artifact"` — a literal-equals-itself check that can only fail on a rename. `ARTIFACT_DIR`'s actual cross-step usage is already independently verified by `test_report_dir_is_a_single_source_of_truth`, so this one restates a constant rather than protecting real behavior.

.github/scripts/post-core-coverage-digest.py
AlexanderZ-Band: Fixed: promoted the bare `50` to a `MID_COVERAGE_PERCENT = 50.0` constant alongside `LOW_COVERAGE_PERCENT`, and used it in `coverage_marker`.

.github/workflows/python-core-coverage.yml
AlexanderZ-Band: Fixed: added test_stage_coverage_report_copies_only_files_that_exist and test_stage_coverage_report_tolerates_a_missing_checkout, which extract this step's real run: text and execute it via subprocess (mirroring _run_pin_guard's pattern, factored into a shared _run_bash helper) against a tmp_path, in both the all-files-present and no-checkout-at-all cases.

.github/scripts/post-core-coverage-digest.py
AlexanderZ-Band: Fixed: added test_render_digest_reports_unavailable_coverage_when_lcov_is_missing, asserting the "Coverage unavailable" header, the workflow-result interpolation, and the run link when lcov_path doesn't exist.

.github/scripts/post-core-coverage-digest.py
AlexanderZ-Band: Fixed: added test_coverage_marker_thresholds, asserting the marker at both real boundaries (LOW_COVERAGE_PERCENT/MID_COVERAGE_PERCENT and just below each) plus 100.0 and 0.0, so the green branch and both edges are now exercised directly.

.github/scripts/post-core-coverage-digest.py
AlexanderZ-Band: Fixed: added test_display_path_falls_back_to_basename_without_a_crates_segment, calling display_path directly on a path with no /crates/ segment.

tests/test_python_core_coverage_workflow.py
AlexanderZ-Band: Fixed: replaced this with a real single-source-of-truth check -- test_report_dir_and_artifact_dir_are_single_sources_of_truth now also asserts the ARTIFACT_DIR literal isn't hardcoded in the stage/upload steps (mirroring the existing REPORT_DIR check), instead of just echoing the constant back at itself.

PR #637, opened 2026-09-15
fix: band_lookup_peers tool description + ExecutionContext watchdog timeout

AlexanderZ-Band:
## Summary
Root-causes two of the three `e2e (core / ubuntu)` failures from the 2026-09-15 nightly ([run 34925408252](https://github.com/band-ai/band-sdk-python/actions/runs/34925408252)), found via a 4-way parallel investigation (adapter/product bug, test-case assumptions, toolkit bug, LLM-flakiness), each validated with live reproduction against the real platform + Anthropic API. Full writeup in [INT-1471](https://linear.app/thenvoi/issue/INT-1471/nightly-e2e-core-failures-band-lookup-peers-tool-description-gap).

- **`band_lookup_peers` tool description gap**: `LookupPeersInput`'s docstring framed the tool purely as delegation discovery, never telling the model it also answers "who else is invitable" — reproduced live, the model calls `band_get_participants` but never `band_lookup_peers`, and hedges instead of naming the invitable peer. Explains the recurring `test_reports_identity_and_roster` flake across strands/langgraph/pydantic_ai/crewai (all on the same cheap `gpt-5.4-mini` baseline model; no Anthropic-model adapter has shown this in 6 nights). Widened the docstring to cover both use cases.
- **`ExecutionContext` missing per-cycle watchdog**: no timeout bounded a reasoning cycle on the normal processing path — the only timeout in the class fired inside `stop()`, after the fact. A stuck handler (matching the recurring `claude_sdk` "did not process message... last delivery status: processing" nightly timeout) left the message in `processing` forever. Added opt-in `SessionConfig.max_cycle_seconds`; on expiry `_run_cycle` cancels the cycle and lets `TimeoutError` flow through the existing `mark_failed`+retry path, same as any other handler error. Default `None` (unbounded) — zero behavior change unless a caller opts in. (Ruled out an initial approval-gate-timeout theory: the E2E harness uses `approval_mode="auto_accept"` for claude_sdk, which returns immediately with no wait path.)

## Test plan
- [x] `uv run ruff check` / `ruff format --check` on changed files
- [x] New tests: `tests/runtime/test_execution_interrupt.py::TestCycleWatchdog` (watchdog cancels + marks failed + loop stays alive; unset default never cancels a slow handler)
- [x] `tests/runtime/` + `tests/framework_conformance/test_tool_text_drift.py` — 844 passed
- [ ] Nightly E2E baseline (`core` lane) — will confirm on next scheduled run; the baseline harness doesn't yet set `max_cycle_seconds`, so this PR hardens the gap without itself forcing the harness to exercise it (follow-up noted in the ticket)

---

https://claude.ai/code/session_017N8P1RXCfL6sSiF7wXDQmC


src/band/runtime/execution.py
```diff
@@ -1688,7 +1688,26 @@ async def _run_cycle(self, event: PlatformEvent, msg_id: str | None) -> bool:
 
         self._active_cycle_task = asyncio.create_task(self._invoke_handler(event))
         try:
-            await self._active_cycle_task
+            if self.config.max_cycle_seconds is not None:
```
AlexanderZ-Band, line 1704: nit: The `if self.config.max_cycle_seconds is not None: ... else: await self._active_cycle_task` split duplicates control flow that `asyncio.wait_for` already handles — per its documented behavior, `timeout=None` blocks until the future completes, identical to a bare `await`. The `else` branch is an unnecessary second code path for the same operation.

Fix: always route through the timeout wrapper (which is `None`-safe), keeping only the exception handling for the warning log.

src/band/runtime/execution.py
```diff
@@ -1688,7 +1688,26 @@ async def _run_cycle(self, event: PlatformEvent, msg_id: str | None) -> bool:
...
+                try:
+                    await asyncio.wait_for(
```
AlexanderZ-Band: issue: This file already has a documented, deliberate ban on `asyncio.wait_for` for exactly this shape of call (an awaited task subject to external cancellation), fixed in commit `8640e8f2` ("fix: Windows support for band-sdk-python [INT-947]"). See the sibling comment at `execution.py:1092-1098`: *"asyncio.timeout() (not wait_for): wait_for wraps the awaitable in a child task, and on Python 3.11 cancelling this loop while it is parked there can be lost — the cancel is recorded but never delivered, wedging shutdown (a routine race on the Windows Proactor loop)."* That fix switched Phase 2's `queue.get()` wait from `wait_for` to `asyncio.timeout()` for this reason, and the same PR added Windows to the E2E CI matrix specifically to catch this class of bug.

`_run_cycle` is itself subject to external cancellation via `interrupt()`/`stop_room()` (the `except asyncio.CancelledError` branch a few lines below), so it's a plausible target for the same Windows/3.11 cancel-loss race the file already had to fix once. Reading CPython 3.12's `asyncio.tasks.wait_for` source shows it's now implemented via `asyncio.timeout()` internally (which is why the new code happens to be safe on 3.12) — but this repo targets `>=3.11` and CI-tests Windows on that version too, where `wait_for` still uses the older manual-waiter implementation the sibling comment is about.

Fix: mirror the existing Phase-2 idiom instead of a new `wait_for` call — `async with asyncio.timeout(self.config.max_cycle_seconds): await self._active_cycle_task`, inside the same try/except. This also resolves the redundant `if`/`else` above for free, since `asyncio.timeout(None)` is documented as unbounded.

src/band/runtime/execution.py
```diff
@@ -1688,7 +1688,26 @@ async def _run_cycle(self, event: PlatformEvent, msg_id: str | None) -> bool:
...
+                        self._active_cycle_task, timeout=self.config.max_cycle_seconds
+                    )
+                except TimeoutError:
```
AlexanderZ-Band: issue: `self._interrupt_kind` leaks across cycles through this branch. `interrupt()`/`stop_room()` cancel `self._active_cycle_task` directly (the same task `wait_for` is also racing to cancel on expiry) and set `self._interrupt_kind = kind` first. If that lands while `wait_for`'s own cancel-and-await is in flight, the resulting exception surfaces here as `TimeoutError`, not `asyncio.CancelledError` — and unlike the `except asyncio.CancelledError:` branch below (which reads-and-clears `self._interrupt_kind` specifically to prevent this), this branch does neither before re-raising.

Proven with a deterministic two-step repro (Logical Bugs skeptic pass): (1) a cycle that lands in this `except TimeoutError:` branch while `ctx._interrupt_kind` is already set leaves it non-`None` afterward; (2) a *subsequent*, unrelated cycle genuinely cancelled by real shutdown (no new `interrupt()` call) then reads that stale `kind` in the `except asyncio.CancelledError:` handler and calls `_abort_cycle(kind, msg_id)` instead of letting `CancelledError` propagate — silently swallowing the shutdown signal for that cycle, exactly the misclassification the method's own docstring says must never happen.

Fix: read-and-clear `self._interrupt_kind` in this branch too (mirroring the `except asyncio.CancelledError:` branch) before re-raising.

AlexanderZ-Band: issue (Logical Bugs, proven): this branch assumes any `TimeoutError` surfacing from `wait_for` means the watchdog cancelled the task. It doesn't distinguish that from the awaited task completing on its own with a *handler-raised* bare `TimeoutError` (e.g. an internal REST/socket timeout) well inside the budget — `asyncio.wait_for` just re-raises that original exception unchanged in that case.

Proof (repro against this exact code): a handler that raises `TimeoutError("downstream call timed out")` immediately, with `max_cycle_seconds=5.0`, completes in ~1ms (never cancelled — confirmed via `task.cancelled() is False`), yet the emitted log is `"ExecutionContext room-123: cycle for message h1 exceeded max_cycle_seconds=5.0; cancelled"` — a false diagnostic claim that would mislead on-call debugging.

Fix: disambiguate using `asyncio.timeout()`'s own `.expired()` state (see the sibling comment on line 1693 recommending that idiom) instead of inferring "the watchdog fired" from exception type alone — `expired()` is true iff the *timeout itself* elapsed, regardless of what the awaited code did in response.

src/band/runtime/execution.py
AlexanderZ-Band, line 1704: issue (Logical Bugs, proven): the watchdog's cancellation is silently defeated if the handler catches `CancelledError` and returns normally instead of re-raising — a pattern this same file already documents as possible two lines below ("A handler may suppress CancelledError and return normally"). When that happens, `asyncio.wait_for` returns the task's normal result with no exception at all, so execution falls through to `self._interrupt_kind = None; return True` and the caller calls `mark_processed` — treating a cycle that was genuinely cancelled past its deadline as a clean success, with no warning logged.

Proof: a handler `try: await asyncio.sleep(60) except asyncio.CancelledError: return`, with `max_cycle_seconds=0.05`, resolves in ~0.05s (proving it really was cancelled at the deadline, not the full 60s sleep) — yet `mark_processed` is awaited once and `mark_failed` is never called. This directly contradicts `SessionConfig.max_cycle_seconds`'s own docstring promise that exceeding it "DOES cancel the cycle" and routes through `mark_failed`+retry.

Fix: same `asyncio.timeout()` + `.expired()` redesign as the sibling findings on lines 1693/1696 — check `expired()` after the `async with` block regardless of whether it exited via exception or normally, since `expired()` reflects whether the deadline passed independent of what the wrapped code did with the resulting cancellation.

src/band/runtime/types.py
```diff
@@ -161,6 +171,12 @@ def __post_init__(self) -> None:
                     % self.max_working_state_seconds
                 )
 
+        if self.max_cycle_seconds is not None and self.max_cycle_seconds <= 0:
```
AlexanderZ-Band: nit: this guard shape (`if x is not None and x <= 0: raise ValueError("<name> must be > 0 when set (got %s)" % x)`) is structurally identical to the pre-existing `max_working_state_seconds` check a few lines above — same shape, same message template, differing only by field name. Consider factoring a small `_require_positive_when_set(name, value)` helper (or a loop over the `(name, value)` pairs) and calling it for both fields, so a future third optional-timeout field doesn't add a third copy.

AlexanderZ-Band: issue: this new validation branch has zero test coverage. Grepped `tests/runtime/` and every `SessionConfig(` call site in the suite — no test constructs `SessionConfig(max_cycle_seconds=0)` or a negative value and asserts the `ValueError`. This is new code the PR is directly responsible for, not a pre-existing gap.

Fix: add a unit test asserting `pytest.raises(ValueError, match="max_cycle_seconds")` for `max_cycle_seconds=0` and a negative value (e.g. alongside `TestCycleWatchdog` in `tests/runtime/test_execution_interrupt.py`).

tests/runtime/test_execution_interrupt.py
```diff
@@ -621,3 +621,51 @@ async def test_backlog_path_cancellation_during_mark_processed(self, mock_link):
         assert handler.invocations == 1
         assert ctx.claims.is_completed("room-123", "bk-cancel-ack")
         assert not ctx.claims.is_ack_pending("room-123", "bk-cancel-ack")
+
+
+class TestCycleWatchdog:
+    """``max_cycle_seconds`` is the hang-killer: unlike interrupt/stop (an
```
AlexanderZ-Band: nit: this class docstring ("a wedged adapter subprocess... otherwise leave a message in 'processing' forever") repeats, almost word-for-word, the rationale already stated in the `max_cycle_seconds` field comment in `src/band/runtime/types.py`. Per this repo's comment-style guidance, state the "why" once — consider trimming this to what's test-specific (what the class verifies) rather than re-explaining the feature's motivation a second time.

tests/runtime/test_execution_interrupt.py
```diff
@@ -621,3 +621,51 @@ async def test_backlog_path_cancellation_during_mark_processed(self, mock_link):
         assert handler.invocations == 1
         assert ctx.claims.is_completed("room-123", "bk-cancel-ack")
         assert not ctx.claims.is_ack_pending("room-123", "bk-cancel-ack")
+
+
+class TestCycleWatchdog:
```
AlexanderZ-Band, line 632: suggestion: both new tests exercise the watchdog firing in isolation; neither drives a concurrent `interrupt()`/`stop_room()` call while the watchdog's own cancel-and-await is in flight, so the `_interrupt_kind`-leak race flagged on `execution.py:1696` has no regression coverage. Once that's fixed, consider adding a deterministic test for it — set `ctx._interrupt_kind` directly to simulate the race window (rather than trying to hit the real timing), then run a second, genuinely-shutdown-cancelled cycle and assert `_interrupt_kind` is `None` for it.

tests/runtime/test_execution_interrupt.py
```diff
@@ -621,3 +621,51 @@ async def test_backlog_path_cancellation_during_mark_processed(self, mock_link):
...
+            config=SessionConfig(max_cycle_seconds=0.05),
+        )
+
+        result = await ctx._process_event(make_message_event(msg_id="watchdog-1"))
+
+        assert result is True  # loop continues; the watchdog is a handled error
+        assert handler.cancelled.is_set()  # the stuck cycle was actually cancelled
+        mock_link.mark_processed.assert_not_awaited()
+        mock_link.mark_failed.assert_awaited_once()
+        assert mock_link.mark_failed.await_args.args[:2] == ("room-123", "watchdog-1")
```
AlexanderZ-Band: issue: `mark_failed.await_args.args[:2] == ("room-123", "watchdog-1")` deliberately ignores the third positional argument — the error label actually passed to `mark_failed` (`_error_label(e)`, which for a bare `TimeoutError()` resolves to the literal string `"TimeoutError"`). The point of this feature is a *diagnosable* failure, but nothing here asserts the failure reason is meaningful or timeout-related — this test would still pass if a future refactor caused `_run_cycle` to propagate a different, unrelated exception as long as room/message-id plumbing stayed correct.

Fix: also assert on `mock_link.mark_failed.await_args.args[2]`, e.g. `assert "TimeoutError" in mark_failed.await_args.args[2]`.

src/band/runtime/tools/inputs/chat.py
```diff
@@ -109,8 +109,9 @@ class LookupPeersInput(BaseModel):
 
     Automatically excludes peers already in the room.
     Returns dict with 'data' list of peers and 'metadata' (page, page_size, total_count, total_pages).
-    Use this to find specialized agents (e.g., Weather Agent) when you cannot answer
-    a question directly.
+    Use this both to find specialized agents (e.g., Weather Agent) to delegate to when
+    you cannot answer a question directly, and to answer questions about who else is
```
AlexanderZ-Band, line 113: suggestion: this docstring widening has no unit-level coverage anywhere. Verified: `tests/framework_conformance/test_tool_text_drift.py` only diffs per-argument `Field(description=...)` text, never the class docstring; `tests/runtime/test_tool_definitions.py::TestLookupPeersInput` only checks `page`/`page_size` defaults/validation, not `__doc__`. Relying solely on the live E2E `test_reports_identity_and_roster` to prove this wording in practice is reasonable for the behavioral claim, but a cheap unit assertion (e.g. `"invitable" in LookupPeersInput.__doc__`) would guard the literal wording against an accidental future edit/revert independent of a comparatively expensive, non-deterministic live LLM call.

AlexanderZ-Band (commented):
Personal review pass (my-code-review) across six angles: Style & Hygiene, Design & Reuse, Runtime Safety & Observability, Test Quality, Edge-Case Flow Analysis, and Logical Bugs (explorer + skeptic, each hypothesis proven with a repro before being posted).

The `band_lookup_peers` docstring change is clean. The `ExecutionContext` watchdog's core idea (route a stuck cycle through the existing mark_failed/retry path) is right, but the `asyncio.wait_for`-based implementation has real correctness gaps, all converging on the same root cause: inferring "did the watchdog actually fire" from exception type alone is lossy. Three independent angles (Runtime Safety, Edge-Case Flow Analysis, and the Logical Bugs skeptic via a deterministic repro) converged on the same `_interrupt_kind` cross-cycle leak. The Logical Bugs skeptic additionally proved two more failure modes with executed repros: a misleading log when a handler's own bare `TimeoutError` is mistaken for the watchdog firing, and — more seriously — the watchdog being silently defeated (reported as success) when a handler catches and swallows `CancelledError`. Design & Reuse independently flagged that this file already banned `asyncio.wait_for` for exactly this kind of cancellable-task await (commit `8640e8f2`, Windows/3.11), in favor of `asyncio.timeout()` — which also happens to be the fix for the other three issues via its `.expired()` state.

Requesting changes given the confirmed correctness issues; the fix is one coherent redesign of the new branch in `_run_cycle`, not several unrelated patches.

(Submitted as COMMENT, not REQUEST_CHANGES — GitHub disallows requesting changes on one's own PR. Treat the issue-severity findings above as blocking; proceeding to fix them.)

src/band/runtime/execution.py
AlexanderZ-Band: Fixed — folded into the `asyncio.timeout()` redesign, which is `None`-safe by itself so the branch is no longer needed. See commit 1297a0b7.

src/band/runtime/execution.py
AlexanderZ-Band: Fixed — switched to `asyncio.timeout()`, matching the established idiom from commit `8640e8f2`. Its `.expired()` state also turned out to be the fix for the three correctness issues below it. See commit 1297a0b7.

src/band/runtime/execution.py
AlexanderZ-Band: Fixed — the `except TimeoutError:` branch now clears `self._interrupt_kind` before re-raising a genuine watchdog expiry, mirroring the `except asyncio.CancelledError:` branch. Added a deterministic regression test (`test_watchdog_expiry_does_not_leak_interrupt_kind_into_shutdown`) that simulates the race directly rather than chasing real timing, then confirms a subsequent genuine-shutdown cycle is not misclassified. See commit 1297a0b7.

AlexanderZ-Band: Fixed — the branch now checks `asyncio.timeout()`'s `.expired()` (true iff the deadline itself passed) instead of inferring "the watchdog fired" from exception type alone, so a handler's own bare `TimeoutError` well inside budget now propagates unchanged with no misleading log. Added a regression test (`test_handlers_own_timeout_error_is_not_mistaken_for_the_watchdog`) reproducing exactly the scenario from your proof. See commit 1297a0b7.

src/band/runtime/execution.py
AlexanderZ-Band: Fixed — after the `async with asyncio.timeout(...)` block, `expired()` is checked regardless of whether the body exited normally or via exception, so a handler that swallows `CancelledError` and returns still forces the mark_failed/retry path instead of a silent mark_processed. Added a regression test (`test_watchdog_cancellation_is_not_defeated_by_a_swallowed_cancel`) reproducing your proof scenario. See commit 1297a0b7.

src/band/runtime/types.py
AlexanderZ-Band: Fixed — factored a shared `_require_positive_when_set(name, value)` helper and used it for both `max_working_state_seconds` and `max_cycle_seconds`. See commit 1297a0b7.

AlexanderZ-Band: Fixed — added `test_max_cycle_seconds_if_set_must_be_positive` alongside the existing `SessionConfig` guard tests in `tests/runtime/test_execution_working_state.py::TestSessionConfigGuards` (covers both `0` and a negative value). See commit 1297a0b7.

tests/runtime/test_execution_interrupt.py
AlexanderZ-Band: Fixed — trimmed `TestCycleWatchdog`'s docstring to what's test-specific; the rationale now lives only in `SessionConfig.max_cycle_seconds`'s own comment. See commit 1297a0b7.

tests/runtime/test_execution_interrupt.py
AlexanderZ-Band: Fixed — added `test_watchdog_expiry_does_not_leak_interrupt_kind_into_shutdown`, which simulates the race deterministically (setting `_interrupt_kind` directly rather than trying to hit the real timing window) and then drives a second, genuinely shutdown-cancelled cycle to confirm it isn't misclassified. See commit 1297a0b7.

tests/runtime/test_execution_interrupt.py
AlexanderZ-Band: Fixed — the test (and the two new sibling tests) now unpack all three `mark_failed` args and assert the label content itself (e.g. `"max_cycle_seconds" in label`), not just room/message id. This also drove a real fix: the old code's bare re-`raise` could surface asyncio.timeout's own message-less `TimeoutError`, so genuine watchdog expiries now raise a `TimeoutError` with an explicit, diagnosable message instead. See commit 1297a0b7.

src/band/runtime/tools/inputs/chat.py
AlexanderZ-Band: Fixed — added `test_description_covers_reporting_invitable_peers` to `tests/runtime/test_tool_definitions.py::TestLookupPeersInput`, asserting `"invitable" in LookupPeersInput.__doc__`. See commit 1297a0b7.

src/band/runtime/execution.py
```diff
@@ -1688,7 +1688,74 @@ async def _run_cycle(self, event: PlatformEvent, msg_id: str | None) -> bool:
...
+                        # The child itself was never actually cancelled, so
+                        # its outcome is real -- recover it instead of
+                        # reporting a cycle that genuinely finished in time
+                        # as a watchdog failure.
+                        self._active_cycle_task.result()
+                    else:
+                        message = (
+                            f"cycle exceeded max_cycle_seconds="
+                            f"{self.config.max_cycle_seconds}"
+                        )
```
AlexanderZ-Band, line 1755: nit (Style & Hygiene): at the previous head (`1297a0b7`) this branch built two independently-worded descriptions of the same fact — the `logger.warning("...exceeded max_cycle_seconds=%s; cancelled", ...)` call and the separately-worded `TimeoutError(f"cycle exceeded max_cycle_seconds={...}")` a few lines below. Two hand-written renderings of the same diagnostic invite silent drift (e.g. a future edit to one phrasing without the other).

Fixed in this commit: both now read from the single `message` local computed once here, so the log line and the exception the caller sees say exactly the same thing by construction.

src/band/runtime/execution.py
```diff
@@ -1688,7 +1688,74 @@ async def _run_cycle(self, event: PlatformEvent, msg_id: str | None) -> bool:
...
+                        # A concurrent interrupt()/stop() takes priority over
+                        # the watchdog's own (coincidental) expiry: it also
+                        # cancels this same task, so it can race in here
+                        # instead of the CancelledError branch below. Honor
+                        # its documented contract (mirrors that branch) rather
+                        # than reporting the user's own stop/interrupt as a
+                        # timeout failure.
+                        kind = self._interrupt_kind
```
AlexanderZ-Band: issue (Runtime Safety & Observability + Edge-Case Flow Analysis, merged — both angles converged on the same defect): at the previous head, a concurrent `interrupt()`/`stop_room()` landing on `self._active_cycle_task` in the same window as the watchdog's own deadline expiry was indistinguishable from a genuine timeout — `_interrupt_kind` was read-and-cleared, but the branch always fell through to logging the watchdog warning and raising `TimeoutError`, reporting the user's own interrupt/stop as a cycle failure via `mark_failed` instead of honoring it via `_abort_cycle` (the contract the sibling `except asyncio.CancelledError:` branch already upholds a few lines below).

Fixed in this commit: `kind = self._interrupt_kind` is checked before treating the expiry as a genuine timeout; if a control signal is present, it now routes through `_abort_cycle(kind, msg_id)` exactly like the CancelledError branch, so the interrupt/stop wins the race and the message is never spuriously marked failed. Regression test: `TestCycleWatchdog::test_watchdog_expiry_honors_a_racing_interrupt_and_does_not_leak_it` (asserts `mark_processed` is called, `mark_failed` is not, and `_interrupt_kind` doesn't leak into a subsequent genuine-shutdown cycle).

src/band/runtime/execution.py
```diff
@@ -1688,7 +1688,74 @@ async def _run_cycle(self, event: PlatformEvent, msg_id: str | None) -> bool:
...
+                            raise TimeoutError
+                except TimeoutError as exc:
+                    # expired() also disambiguates a handler's own bare
+                    # TimeoutError (unrelated to this deadline, well inside
+                    # budget) from this deadline actually firing -- both raise
+                    # the same exception type, but only the timeout arms
+                    # cancellation, so only it flips expired() to True.
+                    if cycle_deadline is None or not cycle_deadline.expired():
+                        raise
+                    if self._active_cycle_task.cancelling() == 0:
```
AlexanderZ-Band, line 1731: issue (Logical Bugs, proven): `asyncio.timeout()`'s deadline cancels the *outer* task (this coroutine), not `self._active_cycle_task` directly. If the child task already completed successfully in the same event-loop tick the deadline fires, CPython's `Task.__step`/`_must_cancel` mechanism can still fabricate a `CancelledError` for the outer task's own resumption — discarding a real, successful result — even though the child itself was never actually cancel-requested. At the previous head this branch had no way to tell that case apart from a real watchdog cancellation; it would report a cycle that genuinely finished in time as a watchdog failure.

Verified against the real C-accelerated `_asyncio.Task` source (`Modules/_asynciomodule.c`) and reproduced deterministically with a fake expired-deadline context manager.

Fixed in this commit: `self._active_cycle_task.cancelling()` (public API since Python 3.11) discriminates the two cases — `== 0` means the child was never actually cancel-requested, so its real outcome is recovered via `.result()` instead of being reported as a timeout. Regression test: `TestCycleWatchdog::test_child_completing_at_the_deadline_boundary_is_not_misreported`.

tests/runtime/test_tool_definitions.py
```diff
@@ -207,6 +207,14 @@ def test_page_size_max(self):
         with pytest.raises(ValidationError):
             LookupPeersInput(page_size=101)
 
+    def test_description_covers_reporting_invitable_peers(self):
+        """The tool description must nudge the model to use this tool when
+        asked who else is invitable, not only to delegate when stuck --
+        otherwise a weaker model has no reason to call it for that question."""
+        assert LookupPeersInput.__doc__ is not None
```
AlexanderZ-Band, line 214: issue (Test Quality): `assert "invitable" in LookupPeersInput.__doc__` is too weak to guard the actual behavioral claim — the bare substring `"invitable"` could appear in an unrelated sentence and still pass, without actually confirming the docstring tells the model to use this tool for the "who else is available/invitable" question.

Fixed in this commit: the assertion now normalizes whitespace and checks the full guiding phrase, `"who else is available or invitable"`, so an edit that drops the actual instruction (while leaving some other mention of the word) fails the test.

AlexanderZ-Band (commented):
Cycle 2 of a personal --fix code-review pass. 4 findings from this cycle (1 diagnostic-message-duplication nit, 1 merged interrupt/stop-vs-watchdog race issue from Runtime Safety + Edge-Case Flow, 1 proven Logical Bugs boundary-completion race, 1 Test Quality weak-assertion issue) — all already fixed in commit 10a0bd4d, pushed before this review was posted. Submitting as COMMENT since GitHub disallows REQUEST_CHANGES on my own PR.

src/band/runtime/execution.py
AlexanderZ-Band: Fixed: the log message and the raised TimeoutError now both read from a single `message` local computed once, so they can't drift apart.

src/band/runtime/execution.py
AlexanderZ-Band: Fixed: `kind = self._interrupt_kind` is now checked before treating the expiry as a genuine timeout; when set, the cycle routes through `_abort_cycle(kind, msg_id)` instead of `mark_failed`. Regression test added: test_watchdog_expiry_honors_a_racing_interrupt_and_does_not_leak_it.

src/band/runtime/execution.py
AlexanderZ-Band: Fixed: `self._active_cycle_task.cancelling() == 0` now discriminates a genuinely-completed child from a real watchdog cancellation, recovering `.result()` in the former case instead of misreporting it as a timeout. Regression test added: test_child_completing_at_the_deadline_boundary_is_not_misreported.

tests/runtime/test_tool_definitions.py
AlexanderZ-Band: Fixed: the assertion now normalizes whitespace and checks the full guiding phrase "who else is available or invitable" instead of the bare substring "invitable".

src/band/runtime/execution.py
```diff
@@ -1675,6 +1683,11 @@ async def _run_cycle(self, event: PlatformEvent, msg_id: str | None) -> bool:
             bypasses the callers' ``except Exception`` and reaches the loop's
             ``except asyncio.CancelledError``; we only swallow it for an
             interrupt/stop, never for shutdown.
+            TimeoutError: when ``max_cycle_seconds`` is set and the cycle
```
AlexanderZ-Band, line 1686: nit (Style & Hygiene): `_run_cycle`'s `Raises:` docstring section only documented `asyncio.CancelledError`, but the watchdog branch this PR added also propagates a `TimeoutError` on genuine expiry — an incomplete/stale description of what the function actually raises.

Fixed in this commit: added a `TimeoutError` entry to the `Raises:` section.

src/band/runtime/execution.py
```diff
@@ -1656,6 +1656,14 @@ async def _execute_message_cycle(self, event: PlatformEvent) -> None:
         finally:
             await self._working_reporter.stop()
 
+    def _take_interrupt_kind(self) -> ControlMode | None:
```
AlexanderZ-Band, line 1659: suggestion (Style & Hygiene): the atomic 'read-and-clear `_interrupt_kind`' idiom was hand-duplicated at two call sites (the watchdog-expiry branch and the outer `except asyncio.CancelledError:` branch) — exactly the class of drift an earlier review cycle already had to fix once in this same function.

Fixed in this commit: factored both into a `_take_interrupt_kind()` helper, used at both sites.

src/band/runtime/execution.py
```diff
@@ -1688,19 +1701,91 @@ async def _run_cycle(self, event: PlatformEvent, msg_id: str | None) -> bool:
...
+                        # still fabricate a CancelledError for the outer
+                        # task's own resumption, discarding a real result.
+                        # The child itself was never actually cancelled, so
+                        # its outcome is real -- recover it instead of
+                        # reporting a cycle that genuinely finished in time
+                        # as a watchdog failure.
+                        logger.debug(
+                            "ExecutionContext %s: cycle for message %s "
+                            "completed right at the max_cycle_seconds=%s "
+                            "deadline boundary; recovering its real result",
```
AlexanderZ-Band, line 1745: suggestion (Runtime Safety & Observability): the boundary-completion recovery branch (`cancelling() == 0` → `.result()`) had no log line at all, unlike its sibling genuine-timeout branch a few lines below. If this rare CPython Task/Timeout race fires in production, there was previously zero diagnostic trace that it happened.

Fixed in this commit: added a `logger.debug` line noting the room/message id and that the cycle completed right at the deadline boundary.

tests/runtime/test_execution_interrupt.py
```diff
@@ -621,3 +622,209 @@ async def test_backlog_path_cancellation_during_mark_processed(self, mock_link):
...
+
+        assert result is True
+        mock_link.mark_processed.assert_awaited_once_with("room-123", "boundary")
+        mock_link.mark_failed.assert_not_awaited()
+        # The outcome above is also what a normal, un-raced completion looks
+        # like -- assert the recovery branch itself actually ran (not just a
+        # fall-through that never hit it), or a regression that silently
+        # removes the recovery logic would still pass this test.
+        assert any(
+            "deadline boundary; recovering its real result" in r.message
```
AlexanderZ-Band, line 828: suggestion (Test Quality, proven via mutation): `test_child_completing_at_the_deadline_boundary_is_not_misreported` only asserted black-box outcomes (`result is True`, `mark_processed` called, `mark_failed` not called) — indistinguishable from what a normal, un-raced completion looks like. Verified by deleting the `if cycle_deadline.expired(): raise TimeoutError` trigger the recovery branch depends on and rerunning: the test still passed, silently no longer covering the behavior it claims to guard.

Fixed in this commit: the test now also asserts the recovery branch's new debug log line fired (paired with the Runtime Safety fix above), which discriminates the real recovery path from a coincidental fall-through. Re-verified the same mutation now fails this test.

AlexanderZ-Band (commented):
Rerun of the --fix --cycles 2 review (previous run's 2 cycles + 15 findings already resolved; this is a fresh pass against unchanged code, per explicit user request). Cycle 1 of this rerun: 4 new findings (2 Style & Hygiene, 1 Runtime Safety & Observability, 1 Test Quality — proven via mutation testing), all non-blocking (nit/suggestion), all fixed in commit 0b34c915. Design & Reuse, Edge-Case Flow Analysis, and Logical Bugs found nothing new — the Logical Bugs explorer's 2 hypotheses were both independently reproduced/checked by the skeptic and dropped as pre-existing/out-of-scope, not new defects this diff introduced. Submitting as COMMENT since GitHub disallows REQUEST_CHANGES on my own PR.

src/band/runtime/execution.py
AlexanderZ-Band: Fixed: added a TimeoutError entry to the Raises: section.

src/band/runtime/execution.py
AlexanderZ-Band: Fixed: factored both read-and-clear sites into a new _take_interrupt_kind() helper.

src/band/runtime/execution.py
AlexanderZ-Band: Fixed: added a logger.debug line in the recovery branch noting the room/message id and that the cycle completed at the deadline boundary.

tests/runtime/test_execution_interrupt.py
AlexanderZ-Band: Fixed: the test now also asserts the new recovery-branch debug log fired, which discriminates the real recovery path from a coincidental fall-through. Re-verified via mutation (deleting the deadline-expired trigger) that this now fails the test.

tests/runtime/test_execution_interrupt.py
```diff
@@ -621,3 +631,197 @@ async def test_backlog_path_cancellation_during_mark_processed(self, mock_link):
         assert handler.invocations == 1
         assert ctx.claims.is_completed("room-123", "bk-cancel-ack")
         assert not ctx.claims.is_ack_pending("room-123", "bk-cancel-ack")
+
+
+# Short enough to fire immediately against a handler that blocks; not tied to
+# any real-world budget, just "small" for these deterministic tests.
+_WATCHDOG_TEST_DEADLINE = 0.05
+# Long enough that a handler blocking on it never finishes naturally within a test.
+_NEVER_RETURNS_SECONDS = 60
```
AlexanderZ-Band, line 640: nit (Style & Hygiene): `TestCycleWatchdog` extracted `_WATCHDOG_TEST_DEADLINE` as a named constant for its "fires immediately" sentinel, but two other same-purpose sentinel literals in the same class were left duplicated raw: `BlockingHandler(block_seconds=60)` (2x) and `SessionConfig(max_cycle_seconds=5.0)` (2x) -- an inconsistent application of the DRY treatment demonstrated one line above.

Fixed in this commit: extracted `_NEVER_RETURNS_SECONDS = 60` and `_AMPLE_CYCLE_BUDGET_SECONDS = 5.0` alongside `_WATCHDOG_TEST_DEADLINE`, used at all 4 call sites.

tests/runtime/test_execution_interrupt.py
```diff
@@ -43,6 +44,29 @@ def mock_link():
     return link
 
 
+async def _assert_fresh_cycle_still_propagates_shutdown_cancel(
```
AlexanderZ-Band, line 47: suggestion (Test Quality): the shutdown-cancel-propagation assertion block (a fresh cycle genuinely cancelled by shutdown must still propagate `CancelledError`, not get misclassified via a leaked `_interrupt_kind`) was near-verbatim duplicated between `test_swallowed_interrupt_does_not_leak_into_shutdown` and the watchdog race test, differing only by `msg_id`. As `TestCycleWatchdog` keeps growing, this was a candidate for a shared helper rather than being retyped a third time.

Fixed in this commit: factored into `_assert_fresh_cycle_still_propagates_shutdown_cancel(ctx, msg_id)`, called from both tests.

tests/runtime/test_execution_working_state.py
```diff
@@ -211,9 +211,22 @@ def test_max_duration_if_set_must_be_positive(self):
...
+            SessionConfig(max_cycle_seconds=-1.0)
+
     def test_disabled_skips_guards(self):
         # When disabled, odd values must not raise.
         cfg = SessionConfig(
             enable_working_state=False, working_keep_alive_seconds=100.0
         )
         assert cfg.enable_working_state is False
+
+    def test_max_cycle_seconds_guard_is_independent_of_working_state(self):
```
AlexanderZ-Band, line 227: suggestion (Test Quality): `_require_positive_when_set("max_cycle_seconds", ...)` in `SessionConfig.__post_init__` was deliberately placed *outside* the `if self.enable_working_state:` block, since `max_cycle_seconds` is an independent hang-killer unrelated to working-state reporting -- a real behavioral property this diff introduced, but no test proved it. A future refactor that accidentally re-nested the call inside that `if` block would have passed every existing test.

Fixed in this commit: added `test_max_cycle_seconds_guard_is_independent_of_working_state`, asserting `SessionConfig(enable_working_state=False, max_cycle_seconds=0)` still raises.

src/band/runtime/execution.py
```diff
@@ -1688,19 +1701,91 @@ async def _run_cycle(self, event: PlatformEvent, msg_id: str | None) -> bool:
...
+                            raise TimeoutError
+                except TimeoutError as exc:
+                    # expired() also disambiguates a handler's own bare
+                    # TimeoutError (unrelated to this deadline, well inside
+                    # budget) from this deadline actually firing -- both raise
+                    # the same exception type, but only the timeout arms
+                    # cancellation, so only it flips expired() to True.
+                    if cycle_deadline is None or not cycle_deadline.expired():
+                        raise
+                    if self._active_cycle_task.cancelling() == 0:
```
AlexanderZ-Band: Runtime Safety & Observability angle raised (as `issue:`): a full `ExecutionContext.stop()` racing the watchdog's own expiry could be misclassified as a genuine cycle timeout, since `stop()` cancels `_active_cycle_task` directly but never sets `_interrupt_kind`, so the `cancelling() != 0` branch would fall through to `logger.warning`/`raise TimeoutError` instead of a clean shutdown.

**Verified and rejected, not a real defect.** `stop()` cancels *two* things sequentially: the child task directly (`cycle_task.cancel()`), and separately `self._process_loop_task.cancel()` -- and `_run_cycle`'s `asyncio.timeout()` block is entered on that exact same loop task (`_run_cycle` is awaited directly from `_process_loop`, never as a separate task). I reproduced the precise race with a live probe: an outer task inside `asyncio.timeout()`, its deadline firing, plus one *additional*, independent `.cancel()` on that same outer task (modeling `stop()`'s own `_process_loop_task.cancel()`) -- confirmed CPython's `Timeout.__aexit__` `uncancel() <= self._cancelling` bookkeeping correctly detects the extra outstanding cancel request and propagates the genuine `CancelledError` as-is, **never** converting to `TimeoutError`. So `stop()` racing the watchdog is already handled correctly by the existing `except asyncio.CancelledError:` branch below, via `_take_interrupt_kind()` returning `None` -> clean re-raise -- no fix needed.

AlexanderZ-Band (commented):
Second rerun of --fix --cycles 2 (prior runs' 19 findings already resolved; this is another fresh pass against unchanged code, per explicit user request). Cycle 1: 4 findings — 1 Style & Hygiene nit and 2 Test Quality suggestions, all fixed in commit 21f37917; 1 Runtime Safety issue that I verified via a live repro of the exact CPython asyncio.timeout()/Task.cancelling() race it described and found does NOT reproduce as claimed — rejected with evidence rather than fixed. Design & Reuse, Edge-Case Flow Analysis, and Logical Bugs found nothing new. Submitting as COMMENT since GitHub disallows REQUEST_CHANGES on my own PR.

tests/runtime/test_execution_interrupt.py
AlexanderZ-Band: Fixed: extracted _NEVER_RETURNS_SECONDS and _AMPLE_CYCLE_BUDGET_SECONDS, used at all 4 sites.

tests/runtime/test_execution_interrupt.py
AlexanderZ-Band: Fixed: factored into _assert_fresh_cycle_still_propagates_shutdown_cancel(ctx, msg_id), called from both tests.

tests/runtime/test_execution_working_state.py
AlexanderZ-Band: Fixed: added test_max_cycle_seconds_guard_is_independent_of_working_state.

src/band/runtime/execution.py
AlexanderZ-Band: Verified via a live probe of the real CPython asyncio.timeout()/Task cancellation-count bookkeeping — the described misclassification does not occur (stop() cancels the same outer/loop task the timeout is armed on, so uncancel() correctly detects the extra cancel and CancelledError propagates, not TimeoutError). No code change; rejecting with evidence recorded above.

src/band/runtime/execution.py
```diff
@@ -1688,19 +1701,88 @@ async def _run_cycle(self, event: PlatformEvent, msg_id: str | None) -> bool:
...
+                            f"{self.config.max_cycle_seconds}"
+                        )
+                        # A concurrent interrupt()/stop() takes priority over
+                        # the watchdog's own (coincidental) expiry: it also
+                        # cancels this same task, so it can race in here
+                        # instead of the CancelledError branch below. Honor
+                        # its documented contract (mirrors that branch) rather
+                        # than reporting the user's own stop/interrupt as a
+                        # timeout failure.
+                        kind = self._take_interrupt_kind()
```
AlexanderZ-Band, line 1815: issue (Runtime Safety & Observability, verified empirically): every genuine `max_cycle_seconds` trip was logged twice -- a `logger.warning` right here, then an ERROR-with-traceback at the caller's existing `except Exception` catch-all, which already reports the same message. Confirmed by running the watchdog test with `--log-cli-level=WARNING`: both a WARNING and a full-traceback ERROR appeared for the same single event. Every other exception on this path (handler errors, a handler's own bare TimeoutError) logs exactly once, at the catch-all only.

Fixed in this commit: dropped the intermediate `logger.warning`, keeping only `raise TimeoutError(message) from exc` -- the message still reaches the log, once, via the caller's catch-all.

tests/runtime/test_execution_interrupt.py
```diff
@@ -621,3 +631,197 @@ async def test_backlog_path_cancellation_during_mark_processed(self, mock_link):
...
+        ctx._on_execute = handler2
+        result2 = await ctx._process_event(make_message_event(msg_id="watchdog-2"))
+        assert result2 is True
+        assert handler2.completed == ["watchdog-2"]
+
+    async def test_unset_max_cycle_seconds_never_cancels_a_slow_handler(
+        self, mock_link
+    ):
+        """Default (unbounded) behavior is unchanged: no watchdog fires."""
+        handler = BlockingHandler(block_seconds=0.05)
```
AlexanderZ-Band, line 685: Style & Hygiene angle raised (as a nit): `test_unset_max_cycle_seconds_never_cancels_a_slow_handler`'s `BlockingHandler(block_seconds=0.05)` hardcodes the same value as `_WATCHDOG_TEST_DEADLINE` instead of referencing that constant.

**Considered and declined, not fixed.** These are two semantically distinct concepts that only coincidentally share a numeric value today: `_WATCHDOG_TEST_DEADLINE` is a *deadline* (used everywhere else as `max_cycle_seconds=...`), while this test's `block_seconds=0.05` is an arbitrary short *handler runtime* in a test that configures no deadline at all (`max_cycle_seconds` is unset here -- that's the point of the test). Reusing the deadline constant as a handler-duration value would misleadingly imply a coupling between the two that doesn't exist, and this project's own conventions (CLAUDE.md) explicitly call out keeping genuinely distinct vocabularies separate even when they happen to share a value today. Left as a plain literal.

AlexanderZ-Band (commented):
Cycle 2 of the second rerun: 1 Runtime Safety issue (genuine watchdog trip double-logged, empirically verified) — fixed in commit fa298ef6. 1 Style & Hygiene nit (0.05 literal duplicate) — considered and declined, reasoning attached to the thread. Design & Reuse, Test Quality, Edge-Case Flow Analysis, and Logical Bugs found nothing new. Submitting as COMMENT since GitHub disallows REQUEST_CHANGES on my own PR.

src/band/runtime/execution.py
AlexanderZ-Band: Fixed: dropped the intermediate logger.warning; the message still reaches the log once, via the caller's existing catch-all.

tests/runtime/test_execution_interrupt.py
AlexanderZ-Band: Declined — these are two semantically distinct concepts (a deadline vs. an unrelated handler runtime) that only coincidentally share a value today; reusing the deadline constant here would misleadingly couple them. Left as a plain literal, per the project's own guidance to keep genuinely distinct vocabularies separate.

src/band/runtime/execution.py
```diff
@@ -1688,19 +1701,88 @@ async def _run_cycle(self, event: PlatformEvent, msg_id: str | None) -> bool:
 
         self._active_cycle_task = asyncio.create_task(self._invoke_handler(event))
         try:
-            await self._active_cycle_task
+            if self.config.max_cycle_seconds is not None:
```
amit-gazal-band, line 1755: **The watchdog issues exactly one cancel, then waits unboundedly — so it doesn't actually bound the cycle.**

`asyncio.timeout`'s `_on_timeout` cancels the outer task once, which cancels the child. The outer task then stays parked on `await self._active_cycle_task` until the child finishes unwinding, and `_timeout_handler` is already consumed, so nothing fires a second time.

If the handler's cancellation path itself blocks — a `finally: await proc.wait()` on a wedged `claude_sdk`/`codex`/`opencode` subprocess, or any `finally` awaiting a hung platform call — `_run_cycle` awaits forever: the room's process loop stays wedged and the message stays in `processing` indefinitely. That is exactly the failure `max_cycle_seconds` is documented to kill.

Reproduced against CPython 3.14: with `max_cycle_seconds=0.1` and a handler whose `finally` sleeps 60s, the `except TimeoutError` body is never reached and the cycle is still blocked 2s later.

For the hang-killer claim to hold, the drain of the child needs its own bound — e.g. `asyncio.wait_for(asyncio.shield(self._active_cycle_task), grace)` after the cancel, or detach the child and move on once the grace period elapses.

src/band/runtime/types.py
```diff
@@ -119,6 +126,16 @@ class SessionConfig:
     # cancel the reasoning. None = unbounded. NOT a hang-killer.
     max_working_state_seconds: float | None = None
 
+    # Upper bound on one reasoning cycle (the handler invoked by _run_cycle).
+    # Unlike max_working_state_seconds, exceeding this DOES cancel the cycle —
+    # it is the hang-killer: a handler stuck awaiting an external call (e.g. a
```
amit-gazal-band: This overstates the guarantee a reader will rely on when sizing the budget. Setting aside the single-cancel gap flagged in `execution.py`, `max_cycle_seconds` only wraps the handler inside `_run_cycle` — the other awaits on the same path in `_process_event_body` (`_ensure_fresh_context()`, `link.mark_processing`, `link.mark_processed`) stay unbounded. A wedge in any of those still leaves the message in `processing` forever.

Worth narrowing the claim to what the field actually covers: the handler, not the whole message lifecycle.

tests/runtime/test_execution_interrupt.py
```diff
@@ -621,3 +631,197 @@ async def test_backlog_path_cancellation_during_mark_processed(self, mock_link):
...
+            def expired(self) -> bool:
+                return True
+
+        @contextlib.asynccontextmanager
+        async def fake_timeout(_seconds: float) -> AsyncIterator[_FakeExpiredDeadline]:
+            # Stands in for the exact race: expired() reports True even though
+            # nothing was ever actually cancelled.
+            yield _FakeExpiredDeadline()
+
+        with patch("band.runtime.execution.asyncio.timeout", fake_timeout):
```
amit-gazal-band: This patch is process-wide, not module-local. `band.runtime.execution` does `import asyncio`, not `from asyncio import timeout`, so `band.runtime.execution.asyncio` resolves to the stdlib `asyncio` **module object** — `patch` replaces `asyncio.timeout` for everything on the interpreter for the duration of the block.

Any other coroutine alive on the loop in that window (pytest-asyncio internals, an anyio/httpx timeout in a background task from another fixture) silently gets a context manager that never expires. Patching a module-local alias the module owns keeps the blast radius to the code under test.

amit-gazal-band (dismissed):
Approving. The watchdog design is careful — the `expired()` disambiguation of a handler's own `TimeoutError` and the boundary-race result recovery are both right, and the `_require_positive_when_set` refactor correctly keeps `max_working_state_seconds` inside the `enable_working_state` guard while `max_cycle_seconds` is validated unconditionally.

Three things worth a follow-up, one of them substantive (the single-cancel gap in the watchdog — it doesn't currently bound a cycle whose cancellation path blocks). Details inline.

PR #638, opened 2026-09-16
refactor: genericize Pydantic AI tool registration

AlexanderZ-Band:
## Summary

- Adds registry-driven Pydantic AI built-in tool registration from the master tool schemas.
- Keeps adapter lifecycle and custom-tool registration separate; preserves validation, feature filtering, file results, and contact-error reporting.
- Adds execution and conformance coverage for the generic registrar.

Resolves INT-1375.


src/band/integrations/pydantic_ai/tools.py
```diff
@@ -0,0 +1,165 @@
...
+            result = await method(**kwargs)
+        except Exception as error:
+            logger.error("%s failed: %s", definition.name, error)
+            # The "Error " prefix is load-bearing: band_tool_errored reads it to
+            # tell a failed Band tool from productive work.
+            message = f"Error executing {definition.name}: {error}"
+            if definition.name == BandTool.RESPOND_CONTACT_REQUEST:
+                await _send_contact_request_error_event(ctx.deps, message)
+            return message
+        return _normalized_result(definition, result)
```
AlexanderZ-Band: issue: `_normalized_result(definition, result)` runs outside `dispatch`'s `try/except Exception`, which only wraps the `await method(**kwargs)` call above. For `band_read_room_file`, this calls `decode_image_block` on every image block once `is_image_passthrough_result` passes — but that check only verifies `data`/`mimeType` keys exist, not that `data` is valid base64. A malformed image block (e.g. `{"content": [{"type": "image", "data": "not-valid-base64!!", "mimeType": "image/png"}]}`) makes `base64.b64decode` raise `binascii.Error` uncaught here. Verified live with a real `Agent`+`FunctionModel` run: pydantic-ai 2.18.0's tool-execution machinery only catches `ToolFailed`/`ModelRetry`/`ToolRetryError`/`ToolFailedError`, so this crashes `agent.run()` itself instead of producing an LLM-visible error (traceback: `tools.py:129 -> tools.py:143-144 -> registry.py:946 decode_image_block -> base64.b64decode`). This is a regression from the pre-PR wrapper, which called `decode_image_block` inside the same `try/except Exception` as the method call and returned `f"Error reading room file: {e}"` on failure.

fix: move the `_normalized_result` call inside the existing `try` block (or wrap just the decode step) so a decode failure produces the same `"Error executing {tool}: {error}"` string dispatch already returns for method-call failures.

src/band/integrations/pydantic_ai/tools.py
```diff
@@ -0,0 +1,165 @@
...
+def _dispatcher(
+    definition: ToolDefinition, schema: type[BaseModel]
+) -> Callable[..., Coroutine[Any, Any, Any]]:
+    """The generic tool body: validate, call the bound method, normalize."""
+
+    async def dispatch(ctx: RunContext[AgentToolsProtocol], **arguments: Any) -> Any:
+        kwargs = _validated_kwargs(definition, schema, arguments)
+        # Resolved outside the try: a missing method is a registry bug, not
+        # something to hand the model as a tool error it could act on.
+        method = getattr(ctx.deps, definition.method_name)
```
AlexanderZ-Band: suggestion: `getattr(ctx.deps, definition.method_name)` has no default, so a stale/typo'd registry `method_name` raises a raw `AttributeError` instead of an actionable message. `resolve_tool_method()` in `src/band/integrations/mcp/engine.py` already solves this exact lookup, raising `RuntimeError(f"{name}: method '{method_name}' not found on {type(tools).__name__}")` — cross-importing from `mcp.engine` is already an established pattern here (`claude_sdk/tools.py` does it). Worth reusing instead of re-deriving the lookup.

src/band/integrations/pydantic_ai/tools.py
```diff
@@ -0,0 +1,165 @@
...
+    JSON schema rather than a function for it to introspect — which is what
+    lets one generic dispatcher stand in for a hand-written function per tool.
+    """
+    schema = platform_args_schema(definition.name)
+    return Tool.from_schema(
+        _dispatcher(definition, schema),
+        # str(): ``ToolDefinition.name`` carries a ``BandTool`` member, and it
+        # becomes a registry key pydantic-ai and its callers compare as a name.
+        name=str(definition.name),
+        description=(schema.__doc__ or "").strip(),
```
AlexanderZ-Band: nit: the description is built from `(schema.__doc__ or "").strip()` instead of calling `get_tool_description(definition.name)`, the function `runtime/tools/schema.py` documents as the single source of LLM-facing tool descriptions and that `crewai/tools.py`/`claude_sdk/tools.py` both call. `platform_args_schema()` already copies `model.__doc__` onto the schema, so these happen to agree for every current tool, but this bypasses the documented source-of-truth function and its `f"Execute {name}"` fallback for any future tool model that lacks a docstring.

src/band/integrations/pydantic_ai/tools.py
```diff
@@ -0,0 +1,165 @@
...
+
+    async def dispatch(ctx: RunContext[AgentToolsProtocol], **arguments: Any) -> Any:
+        kwargs = _validated_kwargs(definition, schema, arguments)
+        # Resolved outside the try: a missing method is a registry bug, not
+        # something to hand the model as a tool error it could act on.
+        method = getattr(ctx.deps, definition.method_name)
+        try:
+            result = await method(**kwargs)
+        except Exception as error:
+            logger.error("%s failed: %s", definition.name, error)
```
AlexanderZ-Band: suggestion: `logger.error("%s failed: %s", definition.name, error)` is now the only failure log for ~30 built-in tools (previously only the contact-request path logged anything). It omits `ctx.tool_call_id` (available on `RunContext`), so concurrent invocations of the same tool in one run can't be told apart in logs, and it logs `str(error)` rather than `exc_info=True`/`logger.exception`, discarding the traceback for every dispatch failure. Worth adding both for debuggability.

AlexanderZ-Band: nit: this one log line now covers every built-in tool's failures, including ones whose arguments/exception text can carry contact handles, memory content, or file/message content. `redact_tool_call_args` already exists in this codebase to keep raw content out of tool-call logs; if any `AgentToolsProtocol` implementation ever echoes a raw argument value into its exception message, it now reaches logs for a much wider surface than before this diff. Worth confirming that's safe, or logging a fixed-shape summary instead of raw exception text.

tests/framework_conformance/test_tool_name_drift.py
```diff
@@ -181,24 +188,37 @@ def test_all_tools_registered(self):
...
-    def test_all_tools_registered(self):
-        """Every non-file tool in TOOL_MODELS has a PydanticAI tool function.
+    _FILE = SRC_ROOT / "integrations" / "pydantic_ai" / "tools.py"
 
-        File tools are excluded: each Band tool is its own
-        ``@platform_tool``-decorated function here, so exposing
-        ``Capability.FILES`` needs a real new function per tool, not just a
-        capability declaration -- not yet done.
-        """
+    def test_derives_tools_from_central_registry(self):
```
AlexanderZ-Band, line 208: nit: `TestPydanticAIToolDrift.test_derives_tools_from_central_registry` duplicates `TestLangGraphToolDrift.test_derives_tools_from_central_registry` verbatim in shape (read file text, assert `'iter_tool_definitions' in source`, near-identical message) — the class docstring even says 'Registry-driven like the LangGraph integration.' Could factor into one shared assertion helper taking the file path and integration label instead of a second hand-copy.

AlexanderZ-Band (commented):
Requesting changes for the confirmed uncaught-exception regression (see inline comment on tools.py:129) — GitHub doesn't allow REQUEST_CHANGES on your own PR, so filing as COMMENT; treat the tools.py:129 issue as blocking. The rest are suggestions/nits.

src/band/integrations/pydantic_ai/tools.py
AlexanderZ-Band: Fixed in 28dc4fb0 — moved `_normalized_result(definition, result)` inside `dispatch`'s existing `try/except Exception` so a decode failure (or any normalization failure) now returns `"Error executing {tool}: {error}"` like every other failure, instead of crashing `agent.run()` uncaught. Added `test_malformed_image_block_becomes_an_llm_readable_error`, which fails without the fix and passes with it.

src/band/integrations/pydantic_ai/tools.py
AlexanderZ-Band: Partially applied in 28dc4fb0, with one change from the suggestion: reusing `resolve_tool_method` from `mcp.engine` directly would give `pydantic_ai/tools.py` a hard dependency on the `mcp` package, which the `pydantic-ai` extra in pyproject.toml deliberately does not include (only `desktop`/`opencode`/`letta`/`claude_sdk`/`crewai` pull in `mcp`) — and relocating the shared helper out of `mcp/engine.py` is out of scope for INT-1375 (explicitly excluded from this ticket). Added a local `_resolve_method()` in `tools.py` with the same actionable-`RuntimeError` shape instead, without the cross-package import.

src/band/integrations/pydantic_ai/tools.py
AlexanderZ-Band: Fixed in 28dc4fb0 — now calls `get_tool_description(definition.name)` instead of reading `schema.__doc__` directly.

src/band/integrations/pydantic_ai/tools.py
AlexanderZ-Band: Fixed in 28dc4fb0 — the dispatch failure log now includes `ctx.tool_call_id` and `exc_info=True`.

AlexanderZ-Band: Declined for now: there's no existing helper that redacts arbitrary exception text (only tool-call arguments are redacted today, via `redact_tool_call_args`), this log line is a net improvement over zero logging for these ~30 tools before this PR, and deciding what's safe to log from an arbitrary `AgentToolsProtocol` exception message is a product/security policy question beyond this PR's scope. Happy to revisit if a specific tool turns out to leak something concrete.

tests/framework_conformance/test_tool_name_drift.py
AlexanderZ-Band: Fixed in 28dc4fb0 — extracted `_assert_derives_from_registry(file_path, integration_label)` and used it from both `TestLangGraphToolDrift` and `TestPydanticAIToolDrift`.

src/band/integrations/pydantic_ai/tools.py
```diff
@@ -0,0 +1,192 @@
...
+        kwargs = _validated_kwargs(definition, schema, arguments)
+        # Resolved outside the try: a missing method is a registry bug, not
+        # something to hand the model as a tool error it could act on.
+        method = _resolve_method(ctx.deps, definition)
+        try:
+            result = await method(**kwargs)
+            # Normalization (e.g. band_read_room_file's image decode) can also
+            # fail on malformed data, so it shares the method call's error
+            # handling rather than crashing the run uncaught.
+            return _normalized_result(definition, result)
```
AlexanderZ-Band, line 222: issue: moving `_normalized_result(definition, result)` inside `dispatch`'s `try` (cycle-1's own fix for the image-decode crash) also means a normalization failure for `band_respond_contact_request` is now caught by the same `except Exception` that gates the room-event report. `band_respond_contact_request` returns a Fern Pydantic model, and `_normalized_result` calls `serialize_tool_result(result)` -> `result.model_dump()` for it. Proven with a real `Agent`+`FunctionModel` repro: `deps.respond_contact_request` succeeds (the approval actually happens, asserted via `assert_called_once_with`), but a `model_dump()` failure on the returned object is caught here, and because `definition.name == BandTool.RESPOND_CONTACT_REQUEST` it unconditionally calls `_send_contact_request_error_event`, posting a room 'error' event and telling the model the response failed — even though the request was, in fact, already answered.

fix: only report to the room when the method call itself failed, not when downstream normalization fails; the two are different failure domains for this one tool.

src/band/integrations/pydantic_ai/tools.py
```diff
@@ -0,0 +1,192 @@
...
+def _dispatcher(
+    definition: ToolDefinition, schema: type[BaseModel]
+) -> Callable[..., Coroutine[Any, Any, Any]]:
+    """The generic tool body: validate, call the bound method, normalize."""
+
+    async def dispatch(ctx: RunContext[AgentToolsProtocol], **arguments: Any) -> Any:
+        kwargs = _validated_kwargs(definition, schema, arguments)
+        # Resolved outside the try: a missing method is a registry bug, not
+        # something to hand the model as a tool error it could act on.
+        method = _resolve_method(ctx.deps, definition)
```
AlexanderZ-Band, line 210: suggestion: `_resolve_method` raises a bare `RuntimeError` before `dispatch`'s `try/except`, so a stale/typo'd registry `method_name` propagates uncaught out of `agent.run()` with no application-level log line — unlike every other dispatch failure, which goes through the `logger.error(..., exc_info=True)` this cycle added. Traced through pydantic-ai's `tool_manager._raw_execute`/`_tool_execution._call_tool`: only `ToolFailed`/`ModelRetry` are caught there, so this `RuntimeError` isn't logged or converted anywhere in this call path before surfacing as an unhandled exception.

fix: log inside `_resolve_method` (name, method_name, deps type) before raising, so this registry-bug path leaves a diagnostic trail too.

src/band/integrations/pydantic_ai/tools.py
```diff
@@ -0,0 +1,192 @@
...
+    deps: AgentToolsProtocol, message: str
+) -> None:
+    """Put a failed ``band_respond_contact_request`` in the room, not just the transcript.
+
+    A contact request the agent failed to answer looks, from the room, exactly
+    like one it never received — so this failure is reported where a human can
+    see it. Best effort: a turn must not fail over its own error reporting.
+    """
+    try:
+        await deps.send_event(message, "error")
```
AlexanderZ-Band: nit: `_send_contact_request_error_event` hardcodes the message-type literal `"error"` instead of the existing `MessageType.ERROR` (StrEnum, `band.core.types`). Other adapters already use the enum for this exact call (e.g. `copilot_sdk.py`, `strands.py`), so this is a scattered magic string where a documented single source of truth already exists.

AlexanderZ-Band (commented):
Contains one blocking issue (the contact-request misreport bug, tools.py:142) — GitHub doesn't allow REQUEST_CHANGES on your own PR, so filing as COMMENT. All 4 findings from this cycle are already fixed in commit to follow.

src/band/integrations/pydantic_ai/tools.py
AlexanderZ-Band: Fixed in 3b626fb2 — split `dispatch`'s single try/except into two: a method-call failure still reports to the room (`notify_room=True`), but a downstream normalization failure now returns a graceful error to the model without claiming the contact request was never answered (`notify_room=False`). Regression test `test_contact_request_success_does_not_report_a_normalization_failure` added, and a post-fix focused Logical Bugs recheck confirmed the original proven flows (image-decode error, genuine contact-request failure) are still intact.

src/band/integrations/pydantic_ai/tools.py
AlexanderZ-Band: Fixed in 3b626fb2 — `_resolve_method` now logs (tool name, method name, deps type) before raising its `RuntimeError`. Added `test_missing_backend_method_raises_an_actionable_error` to cover the branch itself.

src/band/integrations/pydantic_ai/tools.py
AlexanderZ-Band: Fixed in 3b626fb2 — now passes `MessageType.ERROR` instead of the string literal.

AlexanderZ-Band:
Re: the test-coverage finding in the last review body (test_pydantic_ai_adapter.py:1205, _resolve_method's error branch was uncovered) — fixed in 3b626fb2 with test_missing_backend_method_raises_an_actionable_error in TestBuiltinToolExecution.

src/band/integrations/pydantic_ai/tools.py
```diff
@@ -0,0 +1,220 @@
...
+
+
+def _build_tool(definition: ToolDefinition) -> Tool[AgentToolsProtocol]:
+    """Build one native pydantic-ai tool for ``definition``.
+
+    ``Tool.from_schema`` is the public way to hand pydantic-ai a ready-made
+    JSON schema rather than a function for it to introspect — which is what
+    lets one generic dispatcher stand in for a hand-written function per tool.
+    """
+    schema = platform_args_schema(definition.name)
```
AlexanderZ-Band, line 71: issue: The master input models accept unknown fields by default, so both validation passes silently drop an unexpected argument and dispatch the remaining kwargs. A real `band_send_message` tool call with `unexpected` reached `deps.send_message`; the pre-refactor native wrapper returned an `extra_forbidden` retry without dispatch. Reject unknown top-level fields before dispatch and add that real tool-loop regression test.

AlexanderZ-Band (commented):
One confirmed validation regression; fixing it now.

src/band/integrations/pydantic_ai/tools.py
AlexanderZ-Band: Fixed in 66f554cc: the schema-driven boundary now rejects names outside its advertised JSON-schema properties before either validation/dispatch pass. Added a real tool-loop regression test proving the unknown argument produces a retry and `send_message` is not called.

src/band/integrations/pydantic_ai/tools.py
```diff
@@ -0,0 +1,241 @@
...
+    schema: type[BaseModel],
+    arguments: dict[str, Any],
+) -> None:
+    """Reject names outside the JSON schema advertised to the model.
+
+    ``Tool.from_schema`` does not enforce its JSON schema, and the master
+    models intentionally ignore extras for non-tool callers. Native pydantic-ai
+    functions reject extras, so the adapter keeps that tool-call boundary.
+    """
+    accepted = schema.model_json_schema().get("properties", {})
```
AlexanderZ-Band: issue: `_reject_unknown_arguments` (added in 66f554cc) computes its accepted-argument allowlist from `schema.model_json_schema().get("properties", {})`, which only reflects each field's primary `alias` -- never a field's `validation_alias=AliasChoices(...)` secondary names. `AddParticipantInput.identifier`/`RemoveParticipantInput.identifier` (`src/band/runtime/tools/inputs/chat.py:79-104`) accept both `identifier` and `name` via `AliasChoices("identifier", "name")`, and `validate_tool_arguments`'s `model_validate()` call (immediately after, in the same function) honors that -- but this new pre-filter doesn't, so it rejects `name` before validation ever runs.

Proven end-to-end through the real adapter/agent path: a `band_add_participant` call with `{"name": "alice"}` produces `RetryPromptPart('Invalid arguments for band_add_participant: unexpected arguments: name')` and `deps.add_participant` is never called, even though `validate_tool_arguments(BandTool.ADD_PARTICIPANT, schema, {"name": "alice"})` alone resolves it fine to `identifier='alice'`. This is a regression this same commit introduces -- before 66f554cc, `Tool.from_schema`'s pass-through validator let `model_validate()` honor the alias directly.

fix: derive the allowlist from what the model actually accepts (e.g. set `ConfigDict(extra="forbid")` on the schema class built in `platform_args_schema()` and let the single `model_validate()` call do both normalization and rejection) rather than diffing against `model_json_schema()['properties']`, which cannot represent `validation_alias` secondary names.

src/band/integrations/pydantic_ai/tools.py
```diff
@@ -0,0 +1,241 @@
...
+def _validated_kwargs(
+    definition: ToolDefinition,
+    schema: type[BaseModel],
+    arguments: dict[str, Any],
+) -> dict[str, Any]:
+    """``arguments`` as normalized kwargs, or a retry prompt for the model."""
+    try:
+        _reject_unknown_arguments(definition, schema, arguments)
+        return validate_tool_arguments(definition.name, schema, arguments)
+    except ValueError as error:
```
AlexanderZ-Band, line 152: suggestion: `_validated_kwargs`'s `except ValueError` (catching both `_reject_unknown_arguments` and `validate_tool_arguments` failures) re-raises as `ModelRetry` with no logging call anywhere on this path -- unlike `_dispatch_failed`'s two paths, which both log via `logger.error(..., exc_info=True)`. This function runs twice per tool call (the pre-execution hook and dispatch's own re-validation), so every argument-validation failure -- including systemic registry/schema drift -- is invisible server-side; only the model sees the retry prompt.

fix: add a `logger.warning`/`.info` call in the `except ValueError` block before raising `ModelRetry`, naming `definition.name` and the rejected argument text.

tests/adapters/test_pydantic_ai_adapter.py
```diff
@@ -694,158 +769,363 @@ async def test_tool_blurb_is_the_master_docstring(self):
...
+        assert "Invalid arguments for band_send_message" in str(retry.content)
+        assert "mentions" in str(retry.content)
+        deps.send_message.assert_not_called()
 
     @pytest.mark.asyncio
-    async def test_agent_has_file_tools_registered_only_with_capability(self):
-        without_files = PydanticAIAdapter(model="test")
-        await without_files.on_started(agent_name="Probe", agent_description="probe")
-        names = set(without_files._agent._function_toolset.tools)
+    async def test_unknown_arguments_retry_without_dispatching(self):
```
AlexanderZ-Band, line 898: suggestion: `test_unknown_arguments_retry_without_dispatching` only covers a mix of valid + unexpected arguments (`{"content": "hi", "mentions": [], "unexpected": "value"}`). No test drives a call with solely unrecognized keys (e.g. `{"foo": "bar"}`), which is a materially different branch: `_reject_unknown_arguments` short-circuits before pydantic's own required-field check ever runs, so the retry prompt won't mention that `content`/`mentions` are also missing -- worth a parametrized case covering that.

src/band/integrations/pydantic_ai/tools.py
```diff
@@ -0,0 +1,241 @@
...
+    return cast(Callable[..., Coroutine[Any, Any, Any]], method)
+
+
+def _dispatcher(
+    definition: ToolDefinition, schema: type[BaseModel]
+) -> Callable[..., Coroutine[Any, Any, Any]]:
+    """The generic tool body: validate, call the bound method, normalize."""
+
+    async def dispatch(ctx: RunContext[AgentToolsProtocol], **arguments: Any) -> Any:
+        kwargs = _validated_kwargs(definition, schema, arguments)
```
AlexanderZ-Band: suggestion: `dispatch`'s own `_validated_kwargs` call here is documented as defense-in-depth because pydantic-ai discards `args_validator`'s return value, but every current test drives calls through `adapter._agent.run(...)`, where pydantic-ai's `ToolManager` always raises from the pre-execution `args_validator` hook first when arguments are invalid -- so no test can currently tell whether `args_validator` or this fallback guard is what actually rejected bad input. Calling the dispatcher closure directly (bypassing `Tool.from_schema`/`args_validator`) with bad arguments would independently prove the fallback guard works on its own.

src/band/integrations/pydantic_ai/tools.py
```diff
@@ -0,0 +1,241 @@
...
+        logger.error(
+            "%s: method '%s' not found on %s -- registry/protocol drift",
+            definition.name,
+            definition.method_name,
+            type(deps).__name__,
+        )
+        raise RuntimeError(
+            f"{definition.name}: method '{definition.method_name}' not found "
+            f"on {type(deps).__name__}"
+        )
```
AlexanderZ-Band: nit: `_resolve_method` builds the 'method not found' text twice from the same three values -- once as a `%s`-format log string, again as a separately-typed f-string for the raised `RuntimeError` -- so any wording change has to be hand-kept in sync between the two.

fix: build the message once, then `logger.error("%s -- registry/protocol drift", message)` and `raise RuntimeError(message)`.

tests/framework_conformance/test_tool_name_drift.py
```diff
@@ -181,24 +195,32 @@ def test_all_tools_registered(self):
...
+        """Every tool in TOOL_MODELS is built for a fully-capable adapter."""
+        built = {
+            tool.name
+            for tool in build_band_pydantic_ai_tools(
+                AdapterFeatures(capabilities=ALL_CAPABILITIES)
+            )
+        }
+        assert not ALL_TOOL_NAMES - built, (
+            "PydanticAI integration is missing tools for: "
+            f"{sorted(ALL_TOOL_NAMES - built)}."
```
AlexanderZ-Band: nit: `test_all_tools_registered` on `TestPydanticAIToolDrift` computes `ALL_TOOL_NAMES - built` twice -- once in the assert condition, again in the f-string message -- instead of binding it to a `missing` variable once, unlike every sibling drift-test in this file (`missing = X - Y; assert not missing, (...)`).

fix: match the established local pattern -- bind `missing` once and use it in both places.

tests/adapters/test_pydantic_ai_adapter.py
```diff
@@ -694,158 +769,363 @@ async def test_tool_blurb_is_the_master_docstring(self):
...
+        result = await _call_tool(
+            adapter,
+            deps,
+            BandTool.RESPOND_CONTACT_REQUEST,
+            {"action": "approve", "handle": "@ann"},
+        )
+
+        (content,) = _tool_returns(result)
+        assert "nope" in content
+        deps.send_event.assert_called_once_with(content, "error")
```
AlexanderZ-Band: suggestion: `test_contact_request_failure_reaches_the_room` asserts `deps.send_event.assert_called_once_with(content, "error")` with the bare string `"error"`. The production code this test pins was already changed (in an earlier resolved finding) to call `deps.send_event(message, MessageType.ERROR)` specifically to remove this magic string -- this test still hardcodes the string it replaced, passing only because `MessageType` is a `StrEnum`.

fix: import `MessageType` and assert against `MessageType.ERROR` instead of the literal.

AlexanderZ-Band (commented):
Submitting as COMMENT (GitHub blocks REQUEST_CHANGES on ones own PR) — the alias-rejection issue on tools.py:104 is blocking and confirmed end-to-end; will fix under --fix before merge.

src/band/integrations/pydantic_ai/tools.py
AlexanderZ-Band: Fixed in b170e911. Deleted `_reject_unknown_arguments` and its property-name diff entirely; added `_strict_schema()`, a local subclass with `ConfigDict(extra="forbid")`, used for both the advertised JSON schema and the schema passed into validation. The single `model_validate()` call in `validate_tool_arguments` now does the rejecting itself, so it correctly honors `AliasChoices` — `band_add_participant`/`band_remove_participant` with `{"name": "alice"}` now dispatches instead of retrying. Regression test: `test_validation_alias_argument_is_not_rejected_as_unknown`. A focused post-fix Logical Bugs recheck also ran against this fix.

src/band/integrations/pydantic_ai/tools.py
AlexanderZ-Band: Fixed in b170e911. `_validated_kwargs`'s `except ValueError` now logs via `logger.warning` with `definition.name`, `tool_call_id`, and the error before raising `ModelRetry` — both call sites (`_arguments_validator`'s hook and `dispatch`) now thread `ctx.tool_call_id` through.

tests/adapters/test_pydantic_ai_adapter.py
AlexanderZ-Band: Declined as moot. The fix for the alias-rejection issue above deletes `_reject_unknown_arguments`'s separate short-circuit — unknown and missing-required fields are now reported together in one `model_validate()` pass (verified: `{"foo": "bar"}` against `SendMessageInput` now reports missing `content`/`mentions` AND extra `foo` in the same error). The distinct 'incomplete retry prompt' branch this finding was about no longer exists, so a dedicated test for it would just re-exercise the same single code path the existing tests already cover.

src/band/integrations/pydantic_ai/tools.py
AlexanderZ-Band: Declined. This file's own test suite deliberately drives every case through the real adapter/agent path (`adapter._agent.run(...)`) rather than calling internal functions directly — that's this test file's established idiom, and reaching into `_dispatcher`'s closure to test the fallback guard in isolation would break from it for a defense-in-depth path that's already documented in the code (see `_arguments_validator`'s docstring) and was checked via mutation testing during the original implementation (stripping the second validation call was one of the four targeted failures in that pass, per the PR description's test plan).

src/band/integrations/pydantic_ai/tools.py
AlexanderZ-Band: Fixed in b170e911. `_resolve_method` now builds the message once and reuses it for both the log call and the raised `RuntimeError`.

tests/framework_conformance/test_tool_name_drift.py
AlexanderZ-Band: Fixed in b170e911. `test_all_tools_registered` now binds `missing = ALL_TOOL_NAMES - built` once and reuses it in both the assert and the message, matching the sibling drift tests' pattern.

tests/adapters/test_pydantic_ai_adapter.py
AlexanderZ-Band: Fixed in b170e911. The test now imports `MessageType` and asserts `deps.send_event.assert_called_once_with(content, MessageType.ERROR)` instead of the bare string.

src/band/integrations/pydantic_ai/tools.py
```diff
@@ -0,0 +1,245 @@
...
+
+
+def _build_tool(definition: ToolDefinition) -> Tool[AgentToolsProtocol]:
+    """Build one native pydantic-ai tool for ``definition``.
+
+    ``Tool.from_schema`` is the public way to hand pydantic-ai a ready-made
+    JSON schema rather than a function for it to introspect — which is what
+    lets one generic dispatcher stand in for a hand-written function per tool.
+    """
+    schema = _strict_schema(platform_args_schema(definition.name))
```
AlexanderZ-Band: issue: `_strict_schema` (used both for validation and for the model-facing `json_schema=` kwarg passed to `Tool.from_schema`) makes `model_json_schema()` emit `additionalProperties: false` on every built-in tool's advertised schema. Proven against the installed pydantic-ai 2.18.0 source: when `PydanticAIAdapter` is pointed at a Google/Gemini model (a documented, supported configuration per `examples/pydantic_ai/README.md`), pydantic-ai's own `GoogleJsonSchemaTransformer` (`.venv/lib/python3.12/site-packages/pydantic_ai/profiles/google.py:89-124`, invoked on every real request via `Model.customize_request_parameters()`) pops `$schema`/`const`/`discriminator`/`examples`/`title`/`format`/`exclusiveMinimum`/`exclusiveMaximum` -- but never `additionalProperties` -- so this keyword reaches the Gemini API unstripped and breaks tool calls there. This codebase already hit and fixed this exact keyword-on-Gemini problem once, for `GeminiAdapter` (`src/band/adapters/gemini.py:447-452`, `sanitize_tool_schema(..., drop_additional_properties=True)`, asserted by `tests/adapters/test_gemini_adapter.py::TestBuildGeminiTools` -- docstring: "Gemini rejects numeric bounds and additionalProperties on tool params"). `PydanticAIAdapter` has no equivalent guard. OpenAI/Anthropic are unaffected (OpenAI's own strict-mode transformer already forces this itself). Band's own e2e matrix only exercises this adapter against `openai:{model}`, so nothing in-repo currently catches it.

fix: decouple validation strictness from schema advertisement -- use the strict (`extra="forbid"`) schema only for the dispatcher/validator's `model_validate()` call, and advertise the *plain* `schema.model_json_schema()` (no `additionalProperties`) to the model via `Tool.from_schema`'s `json_schema=` kwarg, matching this codebase's own established Gemini precedent.

AlexanderZ-Band (commented):
Submitting as COMMENT (GitHub blocks REQUEST_CHANGES on ones own PR) — this is blocking and will be fixed now.

src/band/integrations/pydantic_ai/tools.py
AlexanderZ-Band: Fixed in 26769e41. Split the strict (extra="forbid") schema's two uses: it now only feeds the dispatcher and the pre-execution args_validator, both purely internal to validation. `Tool.from_schema`'s `json_schema=` kwarg advertises the plain, unmodified master schema again (no `additionalProperties`), matching pre-existing behavior and this codebase's own Gemini precedent. Verified directly against the installed pydantic-ai/Google transformer that the advertised schema no longer carries `additionalProperties` while the alias-honoring rejection still works (`{"name": "alice", "bogus": 1}` still raises `ValidationError`). `test_advertised_schema_is_the_master_schema` reverted to its original assertion, which already fully guards this by equality against the unmodified schema.

src/band/integrations/pydantic_ai/tools.py
```diff
@@ -0,0 +1,254 @@
...
+        # keyword ``adapters/gemini.py`` already has to strip for that
+        # adapter. Strictness only needs to affect validation, not the text
+        # shown to the model.
+        json_schema=schema.model_json_schema(),
+        takes_ctx=True,
+        args_validator=_arguments_validator(definition, strict_schema),
+    )
+
+
+def _strict_schema(schema: type[BaseModel]) -> type[BaseModel]:
```
AlexanderZ-Band, line 99: suggestion: `_strict_schema`'s docstring declares a contract ("Used for validation only ... never for the schema advertised to the model") that nothing in the code enforces -- its return type is plain `type[BaseModel]`, structurally indistinguishable from the advertisable `schema`. This isn't hypothetical: it already happened once in this PR -- an earlier commit fed the strict schema straight into `json_schema=`, breaking Gemini tool calls, caught and reverted one commit later. The underlying risk (two same-typed variables that must be routed to the right one of four call sites) is unchanged, so the same mistake can recur silently -- it would only surface as a Gemini-specific production regression, not a local test failure.

fix: make misrouting fail loudly at the mistake site. Override `model_json_schema` on the subclass `_strict_schema` builds to raise (e.g. `TypeError(f"{schema.__name__}: strict schema is for validation only, never advertisement")`) -- nothing in `src/` currently calls `.model_json_schema()` on the strict variant, so this costs nothing today and converts a repeat of this mistake into an immediate exception instead of a provider-specific regression.

src/band/integrations/pydantic_ai/tools.py
```diff
@@ -0,0 +1,254 @@
...
+            f"on {type(deps).__name__}"
+        )
+        logger.error("%s -- registry/protocol drift", message)
+        raise RuntimeError(message)
+    return cast(Callable[..., Coroutine[Any, Any, Any]], method)
+
+
+def _dispatcher(
+    definition: ToolDefinition, schema: type[BaseModel]
+) -> Callable[..., Coroutine[Any, Any, Any]]:
```
AlexanderZ-Band: nit: `Callable[..., Coroutine[Any, Any, Any]]` is repeated verbatim three times in this file -- `_resolve_method`'s return annotation (line 148), the `cast(...)` in its body (line 162), and `_dispatcher`'s return annotation (line 167). It's the file's only piece of type-level duplication.

fix: add a module-level alias (e.g. `_BoundToolMethod = Callable[..., Coroutine[Any, Any, Any]]`) near the top and use it in all three spots.

tests/adapters/test_pydantic_ai_adapter.py
```diff
@@ -694,158 +770,376 @@ async def test_tool_blurb_is_the_master_docstring(self):
...
+        (retry,) = _parts(result, RetryPromptPart)
+        assert "unexpected" in str(retry.content)
+        deps.send_message.assert_not_called()
 
     @pytest.mark.asyncio
-    async def test_agent_has_file_tools_registered_only_with_capability(self):
-        without_files = PydanticAIAdapter(model="test")
-        await without_files.on_started(agent_name="Probe", agent_description="probe")
-        names = set(without_files._agent._function_toolset.tools)
+    async def test_validation_alias_argument_is_not_rejected_as_unknown(self):
```
AlexanderZ-Band: suggestion: `test_validation_alias_argument_is_not_rejected_as_unknown` only exercises `BandTool.ADD_PARTICIPANT`'s `validation_alias=AliasChoices("identifier", "name")` field. `RemoveParticipantInput.identifier` carries the identical `AliasChoices("identifier", "name")` pattern and is the only other field in the codebase shaped this way, but nothing proves the alias resolves for it too.

fix: parametrize the test over both `(BandTool.ADD_PARTICIPANT, add_participant, role="member")` and `(BandTool.REMOVE_PARTICIPANT, remove_participant)` using the existing `_call_tool`/`_started_adapter` helpers.

AlexanderZ-Band (commented):
No blocking findings this cycle; will apply the 3 suggestions/nits now.

src/band/integrations/pydantic_ai/tools.py
AlexanderZ-Band: Fixed in f3b9002a. `_strict_schema` now overrides `model_json_schema` on the subclass to raise `TypeError` -- verified directly that validation (including the alias case) still works while calling `.model_json_schema()` on the strict variant now fails loudly instead of silently.

src/band/integrations/pydantic_ai/tools.py
AlexanderZ-Band: Fixed in f3b9002a. Added a module-level `_BoundToolMethod = Callable[..., Coroutine[Any, Any, Any]]` alias and used it in all three spots.

tests/adapters/test_pydantic_ai_adapter.py
AlexanderZ-Band: Fixed in f3b9002a. Parametrized the test over both `BandTool.ADD_PARTICIPANT`/`add_participant` and `BandTool.REMOVE_PARTICIPANT`/`remove_participant`, the only two tools sharing the `validation_alias` pattern.

src/band/integrations/pydantic_ai/tools.py
```diff
@@ -0,0 +1,273 @@
...
+    """pydantic-ai's pre-execution argument check for ``definition``.
+
+    Without it nothing would validate at all: ``Tool.from_schema`` installs a
+    pass-through schema validator, so the advertised JSON schema is only ever
+    advice to the model. The dispatcher validates a second time because
+    pydantic-ai discards whatever this returns and calls the tool with the
+    model's raw arguments — this hook can only reject, never normalize.
+    """
+
+    def validate(ctx: RunContext[AgentToolsProtocol], **arguments: Any) -> None:
```
amit-gazal-band, line 174: **High — a non-object tool-arguments payload now aborts the whole turn instead of producing a retry prompt.**

`Tool.from_schema` installs a pass-through `any_schema()` validator, so `ToolManager._validate_tool_args` hands whatever the model emitted straight to `args_validator_func(ctx, **args_dict)`. When a model emits tool args that are valid JSON but not an object (`["hi"]`, `"hi"`, `123` — exactly the malformed-call case the adapter's `retries={"tools": 3}` comment says small models hit), the `**` splat raises `TypeError` inside pydantic-ai. That is neither `ValidationError` nor `ModelRetry`, so it escapes `agent.run_stream_events` and kills `on_message`.

Verified empirically, same adapter, same inputs:

- this branch: `TypeError: ..._arguments_validator.<locals>.validate() argument after ** must be a mapping, not list` — run aborted, all three payloads
- `main`: `RetryPromptPart(content=[{'type': 'dict_type', ... 'msg': 'Input should be an object'}])` — run continues

`_dispatcher`'s `dispatch(ctx, **arguments)` has the same shape, but the validator hook fires first. This can't be fixed inside `validate` (the splat happens at call time) — the tool needs a real object-shaped `SchemaValidator` rather than `from_schema`'s `any_schema()` pass-through, e.g. building the `Tool` with an explicit `FunctionSchema` whose `validator` rejects non-objects.

src/band/integrations/pydantic_ai/tools.py
```diff
@@ -0,0 +1,273 @@
...
+    )
+    # The "Error " prefix is load-bearing: band_tool_errored reads it to
+    # tell a failed Band tool from productive work.
+    message = f"Error executing {definition.name}: {error}"
+    if notify_room and definition.name == BandTool.RESPOND_CONTACT_REQUEST:
+        await _send_contact_request_error_event(ctx.deps, message)
+    return message
+
+
+def _normalized_result(definition: ToolDefinition, result: Any) -> Any:
```
amit-gazal-band, line 265: **Medium — a result-normalization failure after a *successful* `band_send_message` makes the adapter post a spurious "nothing reached the room" error.**

`_normalized_result` now runs `serialize_tool_result` on every built-in tool's return; the old wrappers passed `send_message`'s result through raw. If `model_dump()` raises — the exact case this PR's own `test_contact_request_success_does_not_report_a_normalization_failure` shows is reachable — the tool returns `"Error executing band_send_message: ..."`. `band_tool_errored` matches the `"Error "` prefix, so `is_terminal_success` is `False`, `tool_executed` stays `False`, and `adapters/pydantic_ai.py:600` fires `_report_error(missing_reply_error("Pydantic AI"))` — a room-visible "finished a turn without calling band_send_message" event even though the reply was already posted.

The PR reasons about exactly this split for the contact-request room event (`notify_room=False`), but the same distinction isn't carried into the terminal-work signal.

src/band/integrations/pydantic_ai/tools.py
```diff
@@ -0,0 +1,273 @@
...
+        name=str(definition.name),
+        description=get_tool_description(definition.name).strip(),
+        # The plain schema, not the strict one: ``additionalProperties`` on an
+        # advertised tool schema reaches some providers (e.g. Gemini, via
+        # pydantic-ai's own ``GoogleJsonSchemaTransformer``, which doesn't
+        # strip it) unsanitized and breaks tool calls there -- the same
+        # keyword ``adapters/gemini.py`` already has to strip for that
+        # adapter. Strictness only needs to affect validation, not the text
+        # shown to the model.
+        json_schema=schema.model_json_schema(),
```
amit-gazal-band: **Low — the advertised schemas gain numeric bounds the old ones did not have.**

Diffing the advertised `parameters_json_schema` before/after (all tools, `ALL_CAPABILITIES`): properties, required sets and descriptions are identical, but the new schemas add `"minimum": 1` / `"maximum": 100` on `page`, `page_size`, `limit` (`band_lookup_peers`, `band_list_contacts`, `band_list_tasks`, `band_get_task_history`).

`_NUMERIC_BOUND_KEYWORDS` in `src/band/core/tool_filter.py:20` documents that some providers reject exactly these on integer params ("Gemini, and Anthropic-backed Agno"), and `platform_args_schema`'s `_SanitizedSchema` does **not** pass `drop_numeric_bounds=True`. `PydanticAIAdapter(model="google-gla:gemini-...")` is a supported configuration, and `GoogleJsonSchemaTransformer` only strips `exclusiveMinimum`/`exclusiveMaximum`, not `minimum`/`maximum`.

This matches what LangGraph already ships, so it may be an accepted repo-wide risk — flagging because it is newly true for this adapter. (The `additionalProperties: false` removal and the `$defs`/`$ref` enums are deliberate and look correct.)

tests/framework_conformance/test_tool_name_drift.py
```diff
@@ -39,6 +40,13 @@
 if _HAS_CLAUDE_SDK:
     from band.integrations.claude_sdk.tools import build_band_sdk_tools
 
+try:
+    from band.integrations.pydantic_ai.tools import build_band_pydantic_ai_tools
+
+    _HAS_PYDANTIC_AI = True
+except ImportError:  # pydantic-ai extra, absent from the dev-crewai/dev-parlant lanes
```
amit-gazal-band: **Low — the blanket `except ImportError` silently disables the drift guard in the lane that should run it.**

`_HAS_PYDANTIC_AI` is set from a bare `except ImportError` around the `band.integrations.pydantic_ai.tools` import. Any future import error *inside* that module (a renamed helper in `band.runtime.tools`, say) is indistinguishable from "extra not installed", so `test_all_tools_registered` — the only live proof that every registry tool really gets built — turns into a silent skip in the `dev` lane instead of failing.

Gate on the `pydantic_ai` package itself (`pytest.importorskip` / `importlib.util.find_spec`) and let the integration module's own import errors surface.

amit-gazal-band (changes requested):
Reviewed the full diff against `main`, read the installed pydantic-ai 2.18.0 internals, and ran a before/after schema + behavior probe. Four findings below (one high, one medium, two low).

Positives worth keeping: keyword dispatch removes the real `send_room_file(content, filename, caption, mentions)` positional-reorder hazard; the plain-vs-strict schema split (advertise the sanitized schema, validate with `extra="forbid"`) is correct, and the `model_json_schema` tombstone on the strict subclass is a good guardrail; capability + `include_tools`/`exclude_tools`/`include_categories` filtering now actually applies to this adapter.

src/band/integrations/pydantic_ai/tools.py
AlexanderZ-Band: Fixed in 48cc6b5. Tool.from_schema now receives an object-shaped SchemaValidator, so list, string, and number payloads become model retries before the custom hook is called. test_non_object_arguments_retry_without_dispatching covers all three through the real adapter path.

src/band/integrations/pydantic_ai/tools.py
AlexanderZ-Band: Fixed in 48cc6b5. A normalization failure after a successful terminal tool call now returns a completed-work message rather than the Error-prefixed failure sentinel, so on_message retains the terminal-work signal and does not post a missing-reply error. test_send_message_normalization_failure_counts_as_completed_work covers the real turn flow.

src/band/integrations/pydantic_ai/tools.py
AlexanderZ-Band: Fixed in 48cc6b5. The advertised JSON schema now goes through the shared sanitizer with drop_numeric_bounds=True, preserving the adapter validation schema while removing provider-rejected numeric constraints. The advertised-schema test asserts this exact provider-safe projection for every built-in tool.

tests/framework_conformance/test_tool_name_drift.py
AlexanderZ-Band: Fixed in 48cc6b5. The conformance test now gates only on pydantic_ai package availability with importlib.util.find_spec; once the package is installed, any integration import error fails collection instead of silently skipping the registry-drift guard.

amit-gazal-band (approved):

PR #640, opened 2026-09-17
fix: Align FakeAgentTools participants + mutation tools to real SDK shapes

AlexanderZ-Band:
## Summary
- `FakeAgentTools` is a stateful, contract-faithful test double for participants, message/event sends, room context, and the full contact/contact-request lifecycle, matching the real `AgentTools`/Fern shapes at each boundary.
- Contacts are backed by two directional request stores (`received`/`sent`). `add_contact` auto-approves against an existing contact or a pending reverse request instead of always queuing a new one; `respond_contact_request`/`remove_contact` only resolve currently-actionable records by an id-first, then-handle lookup, so a terminal or unrelated record can never be silently mutated instead. Promoting a request into a contact (handle normalization, required-field validation) is centralized in one helper both paths share.
- `ContactRequestStatus`/`ContactRequestAction` are the single source of truth for the contact-request vocabulary; `ContactRequestSentStatus` and `RespondContactRequestInput.action`'s schema type both derive from them instead of retyping the literals.
- Shared `ParticipantAddResult`/`ParticipantRemoveResult` `TypedDict`s are used by `AgentTools`, `FakeAgentTools`, and `AgentToolsProtocol` so the three can't drift apart; `get_participants` returns real `list[ChatParticipant]` models.
- `platform_tool` now wraps a decorated handler's call and routes its return value through `serialize_tool_result()` itself, so a handler can no longer forget that boundary conversion; all of `pydantic_ai.py`'s tool handlers route through it, with return-type annotations tightened to the real `AgentToolsProtocol` shapes.
- The former private `_normalize_handle` is now public as `strip_handle_prefix`, avoiding a name collision with the pre-existing, semantically-opposite `band.runtime.types.normalize_handle`.
- Shared test seed helpers (`tests/adapters/crewai_flow_support.participant_seed`, `test_pydantic_ai_adapter.agent_tool_functions`) replace per-file duplicates; full state-transition coverage added for the contact lifecycle, mention normalization, and the decorator's auto-serialization.

Implements [INT-1108](https://linear.app/thenvoi/issue/INT-1108/align-fakeagenttools-participants-mutation-tools-to-real-sdk-shapes) from its attached finalized plan. No production behavior changed beyond routing `pydantic_ai`'s handlers through the pre-existing serialization boundary; no unrelated adapters touched.

## Test plan
- [x] `uv run ruff check .`
- [x] `uv run ruff format --check .`
- [x] `uv run pytest tests/ --ignore=tests/integration/ --ignore=tests/e2e/ -q` — 5631 passed, 146 skipped (pre-existing), 0 failed


src/band/core/types.py
```diff
@@ -50,6 +50,28 @@ class ToolEventKey(StrEnum):
 ]
 
 
+class ContactRequestAction(StrEnum):
```
AlexanderZ-Band, line 68: issue: `ContactRequestAction` (approve/reject/cancel) duplicates `RespondContactRequestInput.action`'s existing `Literal["approve", "reject", "cancel"]` (src/band/runtime/tools/inputs/contacts.py:68, untouched by this diff) — two independent definitions of the same closed vocabulary, connected only by this class's docstring ("Mirrors..."), not by code. A future addition/rename of an action on one side won't be caught by the type checker on the other.

Fix: make `RespondContactRequestInput.action` reference `ContactRequestAction` directly (a StrEnum is a valid Pydantic field type) instead of retyping the three literals.

src/band/core/types.py
```diff
@@ -50,6 +50,28 @@ class ToolEventKey(StrEnum):
...
+    Mirrors ``RespondContactRequestInput.action``
+    (``band.runtime.tools.inputs.contacts``).
+    """
+
+    APPROVE = "approve"
+    REJECT = "reject"
+    CANCEL = "cancel"
+
+
+class ContactRequestStatus(StrEnum):
```
AlexanderZ-Band: issue: `ContactRequestStatus` re-declares the same four values (`pending`/`approved`/`rejected`/`cancelled`) already defined 15 lines above as `ContactRequestSentStatus`'s non-`"all"` subset (line 48, pre-existing). The docstring admits the relationship but nothing wires them together in code. This file already has the right pattern for exactly this three lines up: `EventMessageType = Literal[MessageType.THOUGHT, MessageType.ERROR, MessageType.TASK]`, explicitly derived from `MessageType` "so the taxonomy stays single-sourced." The new enum doesn't follow that adjacent precedent.

Fix: derive one from the other the way `EventMessageType` derives from `MessageType`, e.g. `ContactRequestSentStatus = Literal[ContactRequestStatus.PENDING, ContactRequestStatus.APPROVED, ContactRequestStatus.REJECTED, ContactRequestStatus.CANCELLED, "all"]`.

src/band/testing/fake_tools.py
```diff
@@ -303,45 +441,120 @@ async def list_contacts(
...
     async def add_contact(
         self, handle: str, message: str | None = None
-    ) -> dict[str, Any]:
-        return {"id": str(uuid.uuid4()), "status": "pending"}
+    ) -> AddAgentContactResponseData:
+        """Create a pending outgoing request; ``list_contact_requests`` then
+        serves it from the sent-request store, mirroring the real handshake."""
+        request = SentContactRequest(
+            id=str(uuid.uuid4()),
+            inserted_at=datetime(2025, 1, 1, tzinfo=timezone.utc),
```
AlexanderZ-Band: nit: `datetime(2025, 1, 1, tzinfo=timezone.utc)` is written out again here and at line 554 (both new in this diff) — a sentinel value with no fidelity purpose (unlike this file's error-message text, which deliberately echoes real API wording), so there's no reason to retype it at each new call site.

Fix: add a module-level constant next to `_FAKE_ACTOR` (e.g. `_FAKE_TIMESTAMP = datetime(2025, 1, 1, tzinfo=timezone.utc)`) and reference it at both new sites.

tests/adapters/test_crewai_flow_phase3.py
```diff
@@ -61,6 +61,20 @@ class _FakeFlow:
 from band.testing.fake_tools import FakeAgentTools  # noqa: E402
 
 
+def _participant(
```
AlexanderZ-Band: issue: this diff defines the same `_participant(id, handle, name=None, *, type="Agent") -> dict[str, Any]` helper — identical signature, docstring, and body — independently in three sibling files: here, tests/adapters/test_crewai_flow_phase4.py:37, and tests/adapters/test_crewai_flow_phase5.py:58. A future change to what a valid `ChatParticipant` seed needs (as just happened in this very PR) requires three synchronized edits instead of one.

Fix: move `_participant` into one shared location the three files import, and delete the other two copies.

tests/adapters/test_crewai_flow_phase4.py
```diff
@@ -230,14 +244,17 @@ async def test_one_reply_under_join_all_does_not_finalize(self) -> None:
...
-                {"id": "p-b", "handle": "@example/peer-b", "name": "Peer B"},
+                _participant("p-a", "@example/peer-a", "Peer A"),
+                _participant("p-b", "@example/peer-b", "Peer B"),
             ],
             room_context=[
                 {
                     "id": "evt-prior",
                     "message_type": "task",
                     "inserted_at": datetime.now(timezone.utc).isoformat(),
+                    "sender_id": "agent-1",
```
AlexanderZ-Band, line 242: suggestion: the room-context seed dict needing `sender_id`/`sender_type`/`content` (newly required by this diff's `ChatMessage` validation) is hand-typed inline roughly 14 times across test_crewai_flow_phase3/4/5.py (e.g. also lines 335, 405 in this file), while the sibling tests/adapters/test_crewai_flow_state_source.py (touched by this same PR) already factors the equivalent shape into a `_task_event(...)` helper.

Fix: extract a shared `_task_event`-style helper into the same shared location as the `_participant` fix above, and use it at all the inline call sites.

tests/testing/test_fake_tools.py
```diff
@@ -25,6 +26,77 @@ async def store_fact(tools: FakeAgentTools, content: str) -> None:
     )
 
 
+def seeded_participant(
```
AlexanderZ-Band: nit: this diff adds `seeded_participant()`/`seeded_peer()` factories here, but three other files it also touches independently re-invent the same minimal-seed concept as one-off module constants: `ALICE_PARTICIPANT_SEED` in tests/runtime/test_agenttools_contracts.py:13, `ALICE_PARTICIPANT`/`BOB_PEER` in tests/mcp/test_engine.py:70,81, and `ALICE_PARTICIPANT` in tests/adapters/opencode/test_approvals.py:42 — all five encode the same required-field set.

Fix: export `seeded_participant`/`seeded_peer` for reuse (or promote to a shared test-support module) and have the other three files import them instead of redefining equivalent literals.

tests/testing/test_fake_tools.py
```diff
@@ -319,6 +450,183 @@ async def test_pages_serve_distinct_slices(self) -> None:
...
+            room_id=tools.room_id, page=1, page_size=2
+        )
+
+        assert [item["content"] for item in page["data"]] == [
+            "message 0",
+            "message 1",
+        ]
+        assert page["meta"]["total_count"] == 3
+
+    async def test_set_room_context_replaces_and_canonicalizes(self) -> None:
```
AlexanderZ-Band, line 496: issue: `test_set_room_context_replaces_and_canonicalizes` and the adjacent `test_append_room_context_adds_one_canonicalized_item` (line 498) both start from a freshly-constructed `FakeAgentTools()` with an empty room context, then seed a single item and assert only that item is present. Starting from empty makes 'replace' and 'append' produce an identical observable result, so neither test can distinguish the two semantics its name promises — a regression that turned `set_room_context` into an append (or vice versa) would pass both tests unnoticed.

Fix: seed each `FakeAgentTools` with one distinguishable initial context item before calling `set_room_context`/`append_room_context`, then assert the old item is gone for 'replace' and both old+new are present, in order, for 'append'.

src/band/testing/fake_tools.py
```diff
@@ -303,45 +441,120 @@ async def list_contacts(
...
+                    if action == ContactRequestAction.APPROVE
+                    else ContactRequestStatus.REJECTED
+                )
+            case ContactRequestAction.CANCEL:
+                store, handle_field, status = (
+                    self._sent_contact_requests,
+                    "to_handle",
+                    ContactRequestStatus.CANCELLED,
+                )
+            case _:
```
AlexanderZ-Band, line 640: nit: the `case _: raise ValueError(...)` branch is new logic this diff adds (the real `AgentTools.respond_contact_request` never validates `action` locally — it forwards the string straight to the backend). No test in tests/testing/test_fake_tools.py::TestContacts exercises an unrecognized action value.

Fix: add a test calling `respond_contact_request(action="bogus", request_id=...)` asserting it raises `ValueError` matching "Unknown contact request action".

src/band/adapters/pydantic_ai.py
```diff
@@ -475,7 +479,9 @@ async def band_add_contact(
                 message: str | None = None,
             ) -> dict[str, Any] | str:
                 try:
-                    return await ctx.deps.add_contact(handle, message)
+                    return serialize_tool_result(
```
AlexanderZ-Band: issue: this diff routes `band_add_contact` (here), `band_remove_contact` (line 491), and `band_respond_contact_request` (line 524) through `serialize_tool_result()` for the first time. `tests/adapters/test_pydantic_ai_adapter.py` has no test for these three handlers, and its existing tests for the sibling participant handlers stub `ctx.deps` with `AsyncMock(return_value={...})` returning plain dicts — `serialize_tool_result` no-ops on a plain dict, so those tests pass identically with or without the wrapping call. The actual behavior this diff adds — converting a real Fern/Pydantic response into a JSON-safe dict — is unexercised by any test.

Fix: add at least one test per handler group driven by a real `FakeAgentTools` instance (not an `AsyncMock`), so the model-to-dict conversion is actually asserted.

src/band/testing/fake_tools.py
```diff
@@ -303,45 +441,120 @@ async def list_contacts(
...
+                    total_pages=total_pages(len(received), page_size),
                 ),
                 sent=ListAgentContactRequestsResponseMetadataSent(
-                    total=0, total_pages=0
+                    total=len(sent), total_pages=total_pages(len(sent), page_size)
                 ),
             ),
         )
 
     async def respond_contact_request(
```
AlexanderZ-Band, line 615: issue (edge-case flow): `respond_contact_request` resolves its target purely by id/handle via `_find_by_id_or_handle`, with no check on the record's current `status` — a request that already reached a terminal status (`approved`/`rejected`/`cancelled`) remains just as matchable on every later call. Concrete sequence: seed one `pending` received request, call `respond_contact_request(action="reject", request_id=X)` (status -> `rejected`), then `respond_contact_request(action="approve", request_id=X)` on the same id — the second call re-matches the now-`rejected` record, flips its status back to `approved`, and appends a new `AgentContact`. A plain retried `approve` call on the same id likewise appends a second, distinct `AgentContact` (fresh `uuid.uuid4()` each time) for the same handle. `list_contact_requests`'s own docstring says a received request is filtered to `pending` because only pending requests are meant to be actionable, so a terminal request should not remain approvable/rejectable.

Fix: have `respond_contact_request` only match records currently `pending` for approve/reject/cancel, and raise the same 'no response data' `RuntimeError` used for an unresolvable id/handle when the matched record is already terminal.

src/band/testing/fake_tools.py
```diff
@@ -64,6 +79,69 @@ def page_slice(
...
+            )
+        else:
+            handle = normalize_handle(mention)
+            recipients.append(
+                MessageSentResponseRecipientsItem(id=handle, handle=handle)
+            )
+    return recipients
+
+
+def _find_by_id_or_handle(
```
AlexanderZ-Band, line 142: issue (logical bug, proven): `_find_by_id_or_handle` checks `id` then `handle` per-record while iterating a single pass, rather than giving `id` full priority over `handle`. When a caller supplies both an `id` and a `handle` that identify two different seeded records, a record earlier in the list that only matches by `handle` is returned/mutated before the loop ever reaches the record that actually matches the given `id`. Proven: seeding `received_contact_requests=[{id: "req-B", from_handle: "alice"}, {id: "req-A", from_handle: "bob"}]` and calling `respond_contact_request(action="approve", handle="alice", request_id="req-A")` approves and returns `req-B`, silently ignoring the explicit `request_id="req-A"`.

Fix: split into two passes — scan all records for an `id` match first and return immediately if found, then scan for a normalized-handle match. This makes `id` always take precedence over `handle` regardless of iteration order.

src/band/testing/fake_tools.py
```diff
@@ -64,6 +79,69 @@ def page_slice(
     return items[start : start + page_size]
 
 
+def _mention_recipients(
```
AlexanderZ-Band, line 85: issue (logical bug, proven): `_mention_recipients` normalizes a string mention via `normalize_handle` (strips `@`, lowercases) but does not apply the same normalization to the handle/id pulled out of a dict mention. The same logical mention produces different `MessageSentResponseRecipientsItem.id`/`.handle` values depending on caller shape. Proven: `mentions=["@Alice"]` yields recipient `(id="alice", handle="alice")`; `mentions=[{"handle": "@Alice"}]` yields `(id="@Alice", handle="@Alice")` for the same logical mention.

Fix: apply `normalize_handle` to the handle extracted from a dict mention the same way it's applied to a string mention, so both branches converge on one canonical handle.

src/band/testing/fake_tools.py
```diff
@@ -303,45 +441,120 @@ async def list_contacts(
...
+            not_found_message="Failed to respond to contact request - no response data",
+        )
+        request["status"] = status
+        if action == ContactRequestAction.APPROVE:
+            # ReceivedContactRequest doesn't carry the requester's entity
+            # type; extra="allow" lets a seed attach one, defaulting to User.
+            self._contacts.append(
+                AgentContact(
+                    id=str(uuid.uuid4()),
+                    handle=request.get("from_handle") or "",
```
AlexanderZ-Band: issue (logical bug, proven): in the approve branch, `AgentContact(handle=request.get("from_handle") or "", ...)` silently promotes a request seeded/received without `from_handle` (an `Optional[str]` field on the Fern `ReceivedContactRequest`) into a contact with `handle=""`. `AgentContact.handle` is a required `str` on the real Fern model — no genuine backend-approved request could ever produce an empty handle, and this promote-into-a-contact simulation is new logic this diff introduces (the real `AgentTools.respond_contact_request` just forwards to the REST endpoint), so it isn't inherited production-parity fidelity. Proven: seeding a received request with no `from_handle` and approving it produces `self._contacts[0] == {"handle": "", ...}`.

Fix: raise (`ValueError`/`RuntimeError`) when approving a request whose `from_handle` is missing/empty, instead of coercing to `""`.

src/band/runtime/tools/agent.py
```diff
@@ -108,7 +109,7 @@ async def iter_chat_pages(
     )
 
 
-def _normalize_handle(value: str) -> str:
+def normalize_handle(value: str) -> str:
```
AlexanderZ-Band: suggestion: this diff promotes the former private `_normalize_handle` to a public `normalize_handle`, now exported from `band.runtime.tools` and imported by `fake_tools.py`. This collides in name (though not in Python import path) with the pre-existing, unrelated `band.runtime.types.normalize_handle` (src/band/runtime/types.py:41) — which does the opposite (adds a leading `@`) and is imported elsewhere (`src/band/cli/trigger.py`). No current call site imports the wrong one, but the identical name with contradictory semantics across sibling modules is a plausible trap for the next `from band.runtime... import normalize_handle`.

Fix: rename one of the two to be unambiguous — e.g. this one to `strip_handle_prefix`, or the `runtime/types.py` one to `handle_with_at_prefix`.

AlexanderZ-Band (commented):
Automated review pass (my-code-review): 14 findings across 6 angle-groups. 3 proven Logical Bugs defects with repro scripts, 1 demonstrated Edge-Case Flow defect (both would normally warrant REQUEST_CHANGES, but GitHub disallows that on one's own PR), plus Test Quality, Style/Hygiene, and Design/Reuse findings. Fix pass follows.

src/band/core/types.py
AlexanderZ-Band: Fixed in c1bcb50d: `RespondContactRequestInput.action` now uses `ContactRequestActionField`, a `Literal` derived from `ContactRequestAction`'s members (same pattern as `EventMessageType`/`MessageType` above it in the file), instead of retyping the three literals.

src/band/core/types.py
AlexanderZ-Band: Fixed in c1bcb50d: `ContactRequestSentStatus` now derives from `ContactRequestStatus`'s members plus the listing-only `"all"` wildcard, instead of retyping the four values.

src/band/testing/fake_tools.py
AlexanderZ-Band: Fixed in 54966d48: added `_FAKE_TIMESTAMP` next to `_FAKE_ACTOR` and converged every occurrence of the literal in the file, including the 4 pre-existing ones you flagged as optional, not just the two new call sites.

tests/adapters/test_crewai_flow_phase3.py
AlexanderZ-Band: Fixed in d514f675: extracted into `tests/adapters/crewai_flow_support.py`, which phase3/4/5 now import instead of each defining their own copy.

tests/adapters/test_crewai_flow_phase4.py
AlexanderZ-Band: Declined (NOT WORTH IT): the duplication is real, but it's cosmetic inline dict construction across ~14 call sites in tests this PR already touches; the churn of rewriting every seed site to a helper outweighs the benefit for this pass. Not fixing.

tests/testing/test_fake_tools.py
AlexanderZ-Band: Declined (NOT WORTH IT): exporting these for the 3 other files to import would mean editing `tests/runtime/test_agenttools_contracts.py`, `tests/mcp/test_engine.py`, and `tests/adapters/opencode/test_approvals.py` -- none of which this PR otherwise touches -- for a pure style nit. Out of scope for this pass.

tests/testing/test_fake_tools.py
AlexanderZ-Band: Fixed in 54966d48: both tests now seed a distinguishable initial context item, so replace (old item gone) and append (old item still present, new item after it) actually assert different, distinguishing outcomes.

src/band/testing/fake_tools.py
AlexanderZ-Band: Fixed in 54966d48: added `test_respond_contact_request_raises_for_an_unknown_action`, asserting the `bogus` action raises `ValueError` matching "Unknown contact request action".

src/band/adapters/pydantic_ai.py
AlexanderZ-Band: Fixed in a09e089b: added `TestContactTools` in `test_pydantic_ai_adapter.py`, driving all three handlers through a real `FakeAgentTools` instance (not a dict-returning mock) and asserting the result is a plain dict -- so the `serialize_tool_result()` conversion this PR adds is actually exercised.

src/band/testing/fake_tools.py
AlexanderZ-Band: Fixed in 54966d48: `respond_contact_request` now passes `status=ContactRequestStatus.PENDING` into `_find_by_id_or_handle`, so a request already in a terminal status can no longer be re-matched and flipped again. Regression test `test_cannot_respond_to_an_already_terminal_request` fails on the old code and passes now.

src/band/testing/fake_tools.py
AlexanderZ-Band: Fixed in 54966d48: `_find_by_id_or_handle` now runs two full passes -- all records checked for an id match first, then all records checked for a handle match -- so an explicit id always wins regardless of store order. Regression test `test_request_id_takes_precedence_over_an_unrelated_handle_match` reuses your proof scenario, fails on the old code, passes now.

src/band/testing/fake_tools.py
AlexanderZ-Band: Fixed in 54966d48: dict-shaped mentions now normalize their handle the same way string mentions do. Regression test `test_string_and_dict_mentions_normalize_to_the_same_recipient` reuses your proof scenario, fails on the old code, passes now.

src/band/testing/fake_tools.py
AlexanderZ-Band: Fixed in 54966d48: approving a request with no `from_handle` now raises `ValueError` instead of silently creating a contact with `handle=""`. Regression test `test_approving_a_request_with_no_from_handle_raises` fails on the old code, passes now.

src/band/runtime/tools/agent.py
AlexanderZ-Band: Fixed in 54966d48: renamed the newly-public `normalize_handle` to `strip_handle_prefix`, eliminating the name collision with the pre-existing, semantically-opposite `band.runtime.types.normalize_handle`.

src/band/testing/fake_tools.py
```diff
@@ -64,6 +82,94 @@ def page_slice(
...
+            )
+        else:
+            handle = strip_handle_prefix(mention)
+            recipients.append(
+                MessageSentResponseRecipientsItem(id=handle, handle=handle)
+            )
+    return recipients
+
+
+def _find_by_id_or_handle(
```
AlexanderZ-Band, line 142: issue (edge-case flow, confirmed): the previous version of this function filtered `candidates` to the target `status` *before* either the id or handle pass ran, so an explicit `id` naming a record NOT in that status found nothing in the id-pass and silently fell through to the handle-pass, matching a different, unrelated record with the same handle. Concretely: two sent requests to `"frank"` -- `r1` (cancelled) and `r2` (pending) -- then `respond_contact_request(action="cancel", request_id="r1", handle="frank")` cancelled `r2` instead of failing for `r1`. Fixed: the id-pass now scans every record unfiltered and fails fast (raises `not_found_message`) when the id it finds doesn't match the required status, instead of falling back to handle. Regression test: `test_terminal_request_id_does_not_fall_back_to_an_unrelated_handle_match`.

src/band/testing/fake_tools.py
```diff
@@ -301,47 +467,162 @@ async def list_contacts(
...
+            AgentContact(
+                id=str(uuid.uuid4()),
+                handle=strip_handle_prefix(from_handle),
+                name=request.get("from_name"),
+                type=request.get("type", "User"),
+                inserted_at=_FAKE_TIMESTAMP,
+            ).model_dump()
+        )
+
     async def add_contact(
```
AlexanderZ-Band, line 522: issue (edge-case flow / logical bug, confirmed): `add_contact` never implemented the reciprocal auto-accept its own `AddContactInput` docstring documents ("Returns 'approved' when inverse request existed and was auto-accepted") -- adding a contact who already sent a pending request always created a second, redundant outgoing request instead of completing the handshake, and the two stores silently disagreed about whether the contact existed. Fixed: `add_contact` now checks `_received_contact_requests` for a pending reverse request first and completes it if found. Regression test: `test_add_contact_auto_accepts_an_existing_reverse_request`.

src/band/testing/fake_tools.py
```diff
@@ -301,47 +467,162 @@ async def list_contacts(
             ),
         )
 
+    def _promote_received_request_to_contact(self, request: dict[str, Any]) -> None:
```
AlexanderZ-Band, line 484: issue (logical bug, confirmed): `respond_contact_request`'s approve branch built the new `AgentContact.handle` directly from `from_handle` with no `strip_handle_prefix()` call, unlike `add_contact`'s own handle handling in the same file/PR -- approving a request seeded with `from_handle="@alice"` produced a contact with `handle="@alice"` instead of `"alice"`. Fixed: extracted the shared `_promote_received_request_to_contact()` (this line), used by both `respond_contact_request` and `add_contact`'s new reciprocal-approve path, so handle normalization and the from_handle-required check are defined once and applied identically everywhere a received request is promoted.

src/band/testing/fake_tools.py
```diff
@@ -178,66 +298,106 @@ def _require_mentions(
...
+        for cached in self._participants:
+            if matches_identifier(cached, identifier):
+                result = ParticipantAddResult(
+                    id=cached["id"],
+                    name=cached.get("name", identifier),
+                    role=role,
+                    status="already_in_room",
+                )
+                self.participants_added.append(result)
+                return deepcopy(result)
```
AlexanderZ-Band, line 367: nit (runtime safety, confirmed): `add_participant`/`remove_participant` appended the same `ParticipantAddResult`/`ParticipantRemoveResult` dict to `participants_added`/`participants_removed` and returned that identical object, so a caller mutating the returned dict would silently corrupt the tracked entry too. Every other stateful accessor in this file (`store_memory`, `get_task`, `get_board`, etc.) already returns `deepcopy(...)` for this exact shape; these two were the only holdouts. Fixed: now returns `deepcopy(result)` at all three call sites, matching the file's established convention.

tests/adapters/crewai_flow_support.py
```diff
@@ -0,0 +1,19 @@
+"""Shared seed helpers for the CrewAI flow adapter test suite (phase3-5)."""
+
+from __future__ import annotations
+
+from typing import Any
+
+
+def participant_seed(
```
AlexanderZ-Band, line 10: nit (style, confirmed): this helper is imported by name into three sibling test files (`from tests.adapters.crewai_flow_support import _participant`), so a leading underscore signaling "module-private" was backwards from how it's actually used. Renamed to `participant_seed` across all four files.

tests/testing/test_fake_tools.py
```diff
@@ -6,13 +6,17 @@
...
 
+from band.client.rest import ChatParticipant, MessageSentResponseRecipientsItem
 from band.core.exceptions import BandToolError
 from band.core.protocols import AgentToolsProtocol
 from band.runtime.tools import DEFAULT_FILE_CAPTION, serialize_tool_result
 from band.testing import FakeAgentTools
 from tests.content import BLANK_CONTENT_CASES
 
 
+_SEED_INSERTED_AT = "2025-01-01T00:00:00Z"
```
AlexanderZ-Band, line 18: nit (style, confirmed): `seeded_contact`/`seeded_received_request`/`seeded_sent_request` each hardcoded the literal `"2025-01-01T00:00:00Z"` independently -- the same class of duplication `src/band/testing/fake_tools.py`'s own `_FAKE_TIMESTAMP` constant (added this same PR) was introduced to fix, just not applied here. Hoisted `_SEED_INSERTED_AT` and referenced it from all three.

tests/runtime/test_tools_schema.py
```diff
@@ -0,0 +1,105 @@
...
+        """Pre-existing behavior: unrelated to serialization, guarded here too
+        since platform_tool now wraps the function rather than returning it
+        unchanged."""
+
+        @platform_tool
+        async def band_example(identifier: str) -> _Result | str:
+            """Original docstring, discarded."""
+            return _Result(id=identifier)
+
+        assert band_example.__doc__ == get_tool_docstring_with_args("band_example")
```
AlexanderZ-Band, line 105: issue (test quality, confirmed via mutation): this assertion only checked the new docstring `!=` the original literal, which any non-empty replacement satisfies -- injecting a real bug (passing the wrong tool name so the wrong docstring text gets applied) still passed. A stronger, pre-existing sibling test (`test_tool_definitions.py::TestPlatformTool::test_decorator_applies_master_docstring`) already asserts the exact expected value; this now matches that pattern via `get_tool_docstring_with_args("band_example")`.

tests/adapters/test_pydantic_ai_adapter.py
```diff
@@ -60,9 +60,26 @@
 from band.runtime.custom_tools import get_custom_tool_name
 from tests.adapters.usage_events import sent_usage_payloads
 from band.runtime.tools import get_tool_description
+from band.testing import FakeAgentTools
 from tests.framework_configs.adapters import pydantic_ai_probe_tools
 
 
+async def agent_tool_functions(capabilities: Capability) -> dict[str, Any]:
```
AlexanderZ-Band, line 67: issue (design/reuse, confirmed): `TestContactTools._tool_functions()` (added this PR) was byte-for-byte identical to the pre-existing `TestFileTools._tool_functions()`, differing only in which `Capability` it passed to the adapter. Extracted into this shared `agent_tool_functions(capabilities)`, used by both classes; removed both duplicate methods.

tests/adapters/test_crewai_flow_phase3.py
```diff
@@ -386,11 +391,14 @@ async def test_finalized_state_does_not_double_send(self) -> None:
...
         ns = adapter.metadata_namespace
         tools = FakeAgentTools(
-            participants=[{"id": "p1", "handle": "@example/peer"}],
+            participants=[participant_seed("p1", "@example/peer")],
             room_context=[
                 {
                     "id": "evt-prior",
                     "message_type": "task",
+                    "sender_id": "agent-1",
+                    "sender_type": "Agent",
```
AlexanderZ-Band, line 400: suggestion, declined: this `sender_id`/`sender_type`/`content` triad (needed to satisfy `ChatMessage`'s now-stricter validation) is hand-copied ~14 times across phase3/4/5, mirroring `test_crewai_flow_state_source.py`'s own `_task_event()` helper for the identical shape -- the same duplication class this PR's own `crewai_flow_support.py` was introduced to fix for a different literal (`participant_seed`), just not extended to this one. This was also flagged in cycle 1 (declined then too). Declining again: the cleanup is real but touching all 14 sites is more churn than this review-fix pass should carry; worth a follow-up if the team wants it done properly.

tests/testing/test_fake_tools.py
```diff
@@ -25,6 +29,77 @@ async def store_fact(tools: FakeAgentTools, content: str) -> None:
     )
 
 
+def seeded_participant(
```
AlexanderZ-Band: nit, declined: the same "minimal valid Fern-model seed" shape (`seeded_participant`/`seeded_peer` here) is independently re-typed in at least 4 other files this PR touches (`tests/runtime/test_agenttools_contracts.py`, `tests/mcp/test_engine.py`, `tests/mcp/test_shared.py`, `tests/adapters/opencode/test_approvals.py`) plus inline in a couple more. Flagged in cycle 1 (declined) and independently by two cycle-2 groups. Declining again for the same reason -- consolidating means editing several files this PR doesn't otherwise need to touch for a pure style win -- but three independent flags is a real signal; consider a follow-up ticket if this recurs in a future PR.

AlexanderZ-Band (commented):
Cycle 2 of 3 (--cycles 3): 5 parallel angle reviewers + a staged Logical Bugs explorer re-reviewed the code after cycle 1's fixes. 8 confirmed findings, all fixed and pushed (commits 43c731a5, 252881b5, 09382731). Two duplication findings (room-context seed dict across ~14 sites, seeded-participant/peer shape across 5+ files) were declined again -- now flagged independently across cycle 1 and cycle 2's groups, but the cost of touching many otherwise-untouched test files still outweighs the benefit for this pass; worth a follow-up ticket if the team wants it addressed. Also considered and declined as not-worth-it: `add_contact`'s handle lowercasing (pre-existing design already locked in by a passing test, not something this pass should redesign), `ContactRequestActionField`'s enum-instance-vs-str change (verified live: `json.dumps` and `isinstance(x, str)` both treat it identically to the old plain string), an unreachable whitespace-only `from_handle` edge in a fake-only promotion path, and a suggestion to guard `platform_tool` against a non-async handler (no live bug, every real handler is already async).

src/band/testing/fake_tools.py
```diff
@@ -301,47 +488,164 @@ async def list_contacts(
...
+            AgentContact(
+                id=str(uuid.uuid4()),
+                handle=strip_handle_prefix(from_handle),
+                name=request.get("from_name"),
+                type=request.get("type", "User"),
+                inserted_at=_FAKE_TIMESTAMP,
+            ).model_dump()
+        )
+
     async def add_contact(
```
AlexanderZ-Band, line 522: issue (edge-case flow + logical bug, confirmed independently by two reviewers): `add_contact`'s fallback branch created a new pending `SentContactRequest` unconditionally, never checking `self._contacts` first. Calling `add_contact(handle=X)` twice in a row (reciprocal-approve then repeat), or once directly after `X` was separately approved via `respond_contact_request`, left `X` simultaneously an existing contact *and* the target of a live pending outgoing request -- a state no real handshake can produce. Fixed: `add_contact` now checks for an existing contact first and returns `approved` against it instead of minting a redundant request. Regression test: `test_add_contact_is_a_no_op_for_an_existing_contact`.

src/band/testing/fake_tools.py
```diff
@@ -64,6 +82,115 @@ def page_slice(
...
+            )
+        else:
+            handle = strip_handle_prefix(mention)
+            recipients.append(
+                MessageSentResponseRecipientsItem(id=handle, handle=handle)
+            )
+    return recipients
+
+
+def _find_by_handle(
```
AlexanderZ-Band, line 110: issue (style/reuse, confirmed): the `add_contact` fix above initially inlined its two handle-matching scans (existing-contact, reverse-pending-request) as separate `next(...)` comprehensions, each reimplementing the same normalize-both-sides comparison `_find_by_id_or_handle`'s own handle pass already encapsulates -- three copies of the same idiom in ~30 lines. Extracted the find-or-None half into this shared `_find_by_handle()`, now the single definition `_find_by_id_or_handle` and both of `add_contact`'s lookups call.

tests/testing/test_fake_tools.py
```diff
@@ -319,6 +472,339 @@ async def test_pages_serve_distinct_slices(self) -> None:
...
+    async def test_add_contact_auto_accepts_an_existing_reverse_request(self) -> None:
+        """AddContactInput's own docstring: 'Returns approved when inverse
+        request existed and was auto-accepted' -- alice already asked to add
+        me, so my own add_contact(alice) must complete the handshake rather
+        than queue a second, redundant outgoing request. The seeded handle
+        also exercises _promote_received_request_to_contact's own
+        normalization, not just the lookup's."""
+        tools = FakeAgentTools(
+            received_contact_requests=[
+                seeded_received_request("req-1", from_handle="@Alice")
```
AlexanderZ-Band, line 577: issue (test quality, confirmed): neither caller of the shared `_promote_received_request_to_contact` (this test, and `respond_contact_request`'s approve-branch test) seeded an `@`-prefixed or mixed-case `from_handle` -- both used plain `"alice"`, so the helper's own `strip_handle_prefix(from_handle)` call was never actually exercised by either regression test, only the separate lookup's normalization was. Confirmed via mutation: removing that `strip_handle_prefix()` call still passed the full suite before this fix. Both tests now seed `from_handle="@Alice"`; re-verified the mutation now fails both.

src/band/testing/fake_tools.py
AlexanderZ-Band, line 110: suggestion, declined: `_find_by_id_or_handle`'s id-pass does a plain `==` comparison against `id`, while `matches_identifier` (used by `add_participant`/`remove_participant` in the same file family) matches ids case-insensitively -- a genuine inconsistency. Declining: every id in this file is a `str(uuid.uuid4())`, always lowercase and always echoed back verbatim from a prior tool's own JSON output, never user-typed -- there's no reachable flow that would produce a case-mismatched id lookup.

AlexanderZ-Band (commented):
Cycle 3 of 3 (final --cycles 3 pass): 5 parallel angle reviewers + a staged Logical Bugs explorer re-reviewed after cycle 2's fixes. 1 more real bug found and fixed -- notably, two independent agents (Edge-Case Flow and the Logical Bugs explorer) converged on the same defect separately, which is a strong confirmation signal. 1 test-quality gap and 1 style duplication (introduced by this cycle's own fix, caught by a sibling reviewer within the same pass) were also fixed. 1 finding declined as unreachable. This is the last cycle -- stopping here since findings have dropped each round (14 -> 8 -> 4) and this cycle's fixes are now clean.

tests/adapters/crewai_flow_support.py
AlexanderZ-Band, line 10: nit: `participant_seed` duplicates `tests/testing/test_fake_tools.py`'s `seeded_participant` (line 32) — both new in this PR, both building the same six-field `ChatParticipant` seed dict with the identical docstring ("A minimal valid `ChatParticipant` seed for `FakeAgentTools(participants=...)`"), differing only in calling convention (positional vs. keyword-only) and the `type` default ("Agent" vs "User"). Two independent reinventions of the same seed builder within this PR. Consolidate into one shared builder that both `tests/adapters/*` and `tests/testing/*` import, keeping `type` required rather than hiding a domain-specific default inside "shared" code.

src/band/testing/fake_tools.py
```diff
@@ -64,6 +82,115 @@ def page_slice(
     return items[start : start + page_size]
 
 
+def _mention_recipients(
```
AlexanderZ-Band, line 85: nit: `_mention_recipients`'s dict-mention and string-mention branches differ only in how `handle`/`id`/`name` are derived, but each still ends in its own full `MessageSentResponseRecipientsItem(...)` construction. Normalize the mention into a `fields` dict first (`fields = mention if isinstance(mention, dict) else {"handle": mention}`), then do one `strip_handle_prefix(...)` call and one `MessageSentResponseRecipientsItem(id=fields.get("id") or handle, handle=handle, name=fields.get("name"))` construction for both shapes.

src/band/testing/fake_tools.py
```diff
@@ -96,22 +223,36 @@ def __init__(
...
     ):
         self.room_id = room_id
         self._hub_room_id = hub_room_id
         self.messages_sent: list[dict[str, Any]] = []
         self.events_sent: list[dict[str, Any]] = []
-        self._participants: list[dict[str, Any]] = participants or []
-        self._room_context: list[dict[str, Any]] = list(room_context or [])
         # Seeds are validated and canonicalized at seed time (not list time),
         # so every stored record carries the real serialized Fern model shape.
+        self._participants: list[dict[str, Any]] = [
```
AlexanderZ-Band, line 228: nit: this new seed-canonicalization step stores `ChatParticipant.model_validate(p).model_dump()` (6 fields: `handle, id, name, role, status, type`) as `self._participants`, but the real `AgentTools._participants` cache (`src/band/runtime/tools/agent.py:729`) is always projected through the existing `participant_snapshot()` helper (`src/band/runtime/participants.py`) into a fixed 5-field shape (`id, name, type, handle, description`) — dropping `role`/`status` and adding `description`. `add_participant`'s peer-promotion (line ~384) has the same mismatch. No current caller of `.participants` observes the difference (`matches_identifier`/mention-hint code only reads `id/name/handle/type`), but it undercuts this PR's own stated goal of matching real `AgentTools`/Fern shapes for the one place a test can inspect the cache directly. Project through `participant_snapshot()` instead of raw `ChatParticipant.model_dump()` for the single source of truth on this shape.

src/band/testing/fake_tools.py
```diff
@@ -64,6 +82,115 @@ def page_slice(
...
+    """
+    recipients = []
+    for mention in mentions or []:
+        if isinstance(mention, dict):
+            handle = strip_handle_prefix(
+                mention.get("handle") or mention.get("id") or ""
+            )
+            recipients.append(
+                MessageSentResponseRecipientsItem(
+                    id=mention.get("id") or handle,
```
AlexanderZ-Band: issue: `id=mention.get("id") or handle` means an explicit `id` in a dict mention must win over the derived handle, but the only dict-mention test (`test_string_and_dict_mentions_normalize_to_the_same_recipient`) uses `{"handle": "@Alice"}` with no `id` key, so this precedence branch is never exercised. Confirmed by mutation: replacing this line with `id=handle` (always ignoring an explicit `id`) still passes the full 5631-test suite. Add a case with `mentions=[{"id": "user-42", "handle": "@Alice"}]` asserting the recipient's `id == "user-42"` while `handle == "alice"`.

src/band/testing/fake_tools.py
```diff
@@ -301,47 +488,164 @@ async def list_contacts(
...
+        serves the new request from the sent-request store."""
+        existing_contact = _find_by_handle(
+            self._contacts, handle_field="handle", handle=handle
+        )
+        if existing_contact is not None:
+            return AddAgentContactResponseData(
+                id=existing_contact["id"], status=ContactRequestStatus.APPROVED
+            )
+
+        reverse_request = _find_by_handle(
```
AlexanderZ-Band, line 538: issue: `add_contact`'s reverse-request auto-accept only scans `status=ContactRequestStatus.PENDING`, but no test seeds a non-pending (rejected/cancelled) received request from the target handle and calls `add_contact` on it. Confirmed by mutation: removing the `status` filter (matching *any* prior received request regardless of status) still passes the full 5631-test suite — concretely, it would resurrect a `rejected` request into an approved contact instead of creating a fresh pending outgoing request. Add a test seeding a `received_contact_requests` entry with `status="rejected"` for a handle, call `add_contact(handle=...)` on that handle, and assert it creates a *new* pending sent request rather than resurrecting the terminal one.

src/band/testing/fake_tools.py
```diff
@@ -64,6 +82,115 @@ def page_slice(
...
+    ``status``, that is a resolved-but-not-actionable record, not a missing
+    one: it fails the lookup immediately rather than falling through to a
+    handle match on some unrelated record with the same status.
+
+    Shared by every contact/request mutation that resolves its target this
+    way (``remove_contact``, ``respond_contact_request``), so the lookup and
+    its "no response data" failure -- the real tool's own wording when the
+    backend can't find what it was asked to mutate -- stay defined once.
+    """
+    if id is not None:
```
AlexanderZ-Band, line 166: issue: when `id` is given but matches no record at all, `_find_by_id_or_handle` falls through and still tries `handle` — a real, reachable branch shared by `remove_contact` and `respond_contact_request` — but no test calls either with a non-existent `id` *and* a valid `handle` together. Confirmed by mutation: making an unmatched `id` raise immediately (never falling back to `handle`) still passes the full 5631-test suite. Add a test passing a `request_id` matching nothing plus a `handle` that does match a real pending record, asserting it still resolves via the handle fallback — locking in the "id takes precedence, but a miss falls back to handle" contract this function's own docstring describes.

src/band/testing/fake_tools.py
```diff
@@ -301,47 +488,164 @@ async def list_contacts(
...
+        """
+        from_handle = request.get("from_handle")
+        if not from_handle:
+            raise ValueError(
+                "Failed to respond to contact request - malformed request has no from_handle"
+            )
+        self._contacts.append(
+            AgentContact(
+                id=str(uuid.uuid4()),
+                handle=strip_handle_prefix(from_handle),
```
AlexanderZ-Band, line 507: nit: `_promote_received_request_to_contact` sets `name=request.get("from_name")` and `type=request.get("type", "User")` (the deliberate default this method's own docstring calls out), but no test seeds `from_name`/`type` on a received request or checks these fields on the resulting contact — every existing assertion checks only `handle`. Confirmed by mutation: hardcoding `name="WRONG NAME ALWAYS"` and `type="Agent"` here still passes the full 5631-test suite. Extend one existing promotion test to seed `from_name` and assert it lands on the resulting contact, plus a one-line assertion that an unset `type` defaults to `"User"`.

src/band/testing/fake_tools.py
```diff
@@ -301,47 +488,164 @@ async def list_contacts(
             ),
         )
 
+    def _promote_received_request_to_contact(self, request: dict[str, Any]) -> None:
```
AlexanderZ-Band, line 484: issue: `_promote_received_request_to_contact` (called from both `add_contact`'s reverse-auto-accept and `respond_contact_request`'s approve branch) only appends to `self._contacts` and mutates the received request's own status — it never reconciles `self._sent_contact_requests`. Demonstrated flow: seed `sent_contact_requests=[{"id": "s1", "to_handle": "alice", "status": "pending", ...}]` and `received_contact_requests=[{"id": "r1", "from_handle": "alice", "status": "pending", ...}]` on one `FakeAgentTools`, then call either `add_contact(handle="alice")` or `respond_contact_request(action="approve", request_id="r1")`. Verified live: `list_contacts()` returns `alice` as a contact, while `list_contact_requests(sent_status="all")` still reports `s1` as `status="pending"` to the now-already-a-contact `alice` — an incoherent combined state that persists indefinitely. Have this method (or its two call sites) also resolve any matching pending record in `self._sent_contact_requests` for the same handle when promoting, mirroring the same-handle reconciliation `add_contact` already does for a *new* outgoing request.

AlexanderZ-Band (commented):
Second review pass (1 cycle). 8 new findings across Style & Hygiene, Design & Reuse, Test Quality, and Edge-Case Flow Analysis (3 of them blocking issue: severity in Test Quality/Edge-Case Flow — REQUEST_CHANGES isn't available on one's own PR, so submitting as COMMENT; proceeding to fix all 8).

tests/adapters/crewai_flow_support.py
AlexanderZ-Band: Fixed — moved the seed-builder into a new shared `tests/testing/support.py::seeded_participant`; `participant_seed` here now delegates to it, keeping its own explicit `type="Agent"` default local to this CrewAI-scoped wrapper rather than hiding it in shared code. No call sites needed to change.

src/band/testing/fake_tools.py
AlexanderZ-Band: Fixed — normalized both mention shapes into one `fields` dict up front, so there's a single `MessageSentResponseRecipientsItem(...)` construction for both.

src/band/testing/fake_tools.py
AlexanderZ-Band: Declined — verified this fix would regress `get_participants()`. The real `ChatParticipant` model requires `role`/`status` (no defaults); `participant_snapshot()`'s 5-field projection drops both. Real `AgentTools` avoids this collision by keeping two separate representations (the full REST response `get_participants()` returns, and a separately-fetched snapshot used only internally for mention resolution) — but `FakeAgentTools` uses the same `self._participants` list for both purposes, so projecting it through `participant_snapshot()` would break the existing `test_get_participants_returns_chat_participant_models` (a validation error on the now-missing required fields). Replicating the real implementation's two-representation split here, for a shape divergence with no actual consumer, isn't worth the added complexity.

src/band/testing/fake_tools.py
AlexanderZ-Band: Added `test_a_dict_mentions_explicit_id_wins_over_its_handle` covering this precedence branch.

src/band/testing/fake_tools.py
AlexanderZ-Band: Added `test_add_contact_does_not_auto_accept_a_terminal_reverse_request` covering the status guard.

src/band/testing/fake_tools.py
AlexanderZ-Band: Added `test_an_unresolvable_request_id_falls_back_to_a_matching_handle` covering this fallback.

src/band/testing/fake_tools.py
AlexanderZ-Band: Added `test_approving_a_received_request_propagates_name_and_type` and `test_approving_a_received_request_defaults_type_to_user` covering both fields.

src/band/testing/fake_tools.py
AlexanderZ-Band: Fixed — `_promote_received_request_to_contact` now also resolves any matching pending record in `self._sent_contact_requests` for the same handle, marking it approved. Added two regression tests (via `add_contact`'s reverse-accept path and via `respond_contact_request`'s approve branch) — both verified to fail on the pre-fix code and pass after.

amit-gazal-band (approved):

PR #642, opened 2026-09-18
fix: Migrate band-python-kit spec.yaml to the current Docker sbx v2 schema

AlexanderZ-Band:
## Summary
- `spec.yaml` declared `schemaVersion: "2"` but used v1-era field names (`sandbox.aiFilename`, `commands.startup`, `caps.network`, `agentContext`), which `sbx` v0.42.1+ rejects. Migrates to `agentInstructions`, `permissions.network`, and `sandbox.entrypoint`.
- **Not just a rename**: live testing against real `sbx` v0.43.0 found the literal rename target (`setup.startup`) reproduces old behavior exactly, but `sbx` v0.43.0 auto-stops an idle, unattached sandbox within ~60-90s regardless of a live background process inside. `sandbox.entrypoint` (an attached foreground process, launched via `sbx run`) is the correct mapping — README/RELEASING.md document the new "needs a host-side process holding the attach open" requirement.
- `tests/docker/toolkit/sbx_cli.py` gained public `spawn_attached_run`/`stop_attached_run` helpers (pty-backed); `Sandbox.create()` uses them so the live never-in-VM/round-trip proofs still get a running agent to message.
- **Self-review pass** caught `examples/sandbox/self-registration/demo.py` making the same now-invalid "sbx create alone boots the agent" assumption — it would have hung waiting for a reply. Fixed with the same attach helper, live-verified end to end.
- Dropped "verified live"/"confirmed (live, during the migration)" narration from comments and docs — they state facts, not how the facts were checked.
- Pins the newly-verified-latest-stable `sbx` v0.43.0 in `RELEASING.md` and the `kit-publish.yml` header comment (was stale at v0.35.0).

Full design and the live-verification log are in the [attached plan on INT-1477](https://linear.app/thenvoi/issue/INT-1477/migrate-band-python-kit-to-the-current-docker-sbx-schema).

## Test plan
- [x] `sbx kit validate docker/band_python_kit` — passes clean on `sbx` v0.43.0 (previously failed with 4 unknown-field errors)
- [x] `uv run pytest tests/docker/test_kit_spec.py tests/docker/test_nevervm_contracts.py -v` — 11 passed
- [x] `uv run pytest tests/ --ignore=tests/integration/ --ignore=tests/e2e/` — 5602 passed, 0 failed
- [x] `ruff check`, `ruff format --check`, `pyrefly check` — clean
- [x] Live `sbx create` + pty-attached `sbx run` against the real repo spec — clean launcher-failure message, exit 1
- [x] `tests/docker/test_kit_proxy_managed_live.py` — 3/3 passed live (injection path, key-absence, round-trip)
- [x] `tests/docker/test_kit_self_registration_live.py` — 1/1 passed live (registration + idempotency + round-trip)
- [x] `examples/sandbox/self-registration/demo.py` — run live end to end, echo reply confirmed, full cleanup verified
- [x] Manual quickstart flow (provision → `sbx create` → attach via `sbx run` → `band-trigger` → observed echo reply via REST) against the real dev deployment
- No residual sandboxes/secrets/agents/rooms left after any run

https://claude.ai/code/session_01UPGczNQcBV2MoTUcAmYYS5


nir-singher-band:
Heads-up for whoever merges this: `thenvoi/fern#348` ships a docs workaround for the exact schema mismatch this PR fixes (published `bandhq/band-python-kit:3.1.1` declares `schemaVersion: "2"` but still uses v1-era fields, which `sbx` v0.42.1+ rejects). It's flagged with a `<Note>` tied to this PR, and we're shipping it now — ahead of this merging — because of a marketing campaign running with the Docker team that needs the guide live.

**Once this merges and ships in a released kit version**, `fern/docs/pages/integrations/sandboxes/docker-sbx-kit.mdx` needs:

1. Delete the "Prepare the compatible kit specification" step entirely (the `curl` for the old spec, the `sed` patching `schemaVersion`/image placeholder, `sbx kit validate/inspect` on the local copy) and its `<Note>`.
2. Replace every `sbx run "$BAND_KIT_DIR" --name my-band-agent "$HOME/my-band-agent"` with `sbx run "docker.io/bandhq/band-python-kit:${BAND_KIT_VERSION}" --name my-band-agent "$HOME/my-band-agent"` (referencing the published kit directly, no local spec) — this appears in "Start and attach to the Docker Sandbox" and "Existing Band Identity: Create Only the Sandbox."
3. Drop every `export BAND_KIT_DIR=...` line alongside those, in both of the above sections plus the three Keychain-recovery command blocks under "Recover from a Keychain Failure" that also reference `$BAND_KIT_DIR`.
4. Update the `invalid spec.yaml` troubleshooting row (currently points at "Prepare the compatible kit specification") to point at the new direct-validate command instead: `sbx kit validate "docker.io/bandhq/band-python-kit:${BAND_KIT_VERSION}"`.
5. Bump the pinned `BAND_KIT_VERSION` used throughout the guide to the first release containing this fix.

Will file a Linear ticket blocked on this PR merging so the revert doesn't get lost. Ping @AlexanderZ-Band (reviewed the fern PR) when this ships.

examples/sandbox/self-registration/demo.py
```diff
@@ -219,13 +223,16 @@ async def run(kit: str) -> None:
 
     room_id: str | None = None
     agent_id: str | None = None
+    attach: tuple[subprocess.Popen[bytes], int] | None = None
```
AlexanderZ-Band: nit: This attach-handle shape `tuple[subprocess.Popen[bytes], int]` is spelled out here, and again in `spawn_attached_run`'s return annotation and `stop_attached_run`'s parameter list in `tests/docker/toolkit/sbx_cli.py` (lines 277, 311) - three independent copies of the same shape with no shared name. A change to either element (e.g. widening the fd to a small dataclass) has to be caught by hand at all three sites instead of one. Consider a `type AttachHandle = tuple[subprocess.Popen[bytes], int]` alias in `sbx_cli.py` that all three sites reference.

tests/docker/toolkit/sbx_cli.py
```diff
@@ -270,6 +274,52 @@ def allow_network_for_hosts(hosts: Iterable[str], *, kit: Path | str) -> Iterato
         yield
 
 
+def spawn_attached_run(name: str) -> tuple[subprocess.Popen[bytes], int]:
```
AlexanderZ-Band: suggestion: `spawn_attached_run`/`stop_attached_run` are a manually-paired acquire/release function pair, but this same file already establishes the idiom for exactly this shape (acquire a resource, guarantee release even on exception) as an `@contextmanager` - see `custom_secret` (line 158), `allow_network` (line 215), `allow_network_for_hosts` (line 260). The free-function pair pushes the pairing discipline onto every caller instead of the type system: `Sandbox.create` (line 376) gets it right via try/finally, but nothing stops a future caller from calling `spawn_attached_run` and forgetting `stop_attached_run` - `demo.py` already tracks a nullable `attach` by hand to cope with that. Consider `@contextmanager def attached_run(name): ... finally: stop_attached_run(...)`, with `Sandbox.create` nesting into it as `with attached_run(name) as (process, fd):`.

AlexanderZ-Band: issue: `spawn_attached_run` (here) and `stop_attached_run` (line 311) are new public functions with non-trivial logic - pty allocation, subprocess spawn wired to the pty fds, and a terminate()/wait(timeout)/kill()/wait(timeout) escalation - with zero test coverage anywhere in `tests/`. The escalation path (what happens when `terminate()` doesn't stop the process within `ATTACH_STOP_TIMEOUT_S`) is exactly the kind of behavior that silently regresses without a test. This file already has a precedent for unit-testing its pure/host-side pieces against a local stand-in rather than the real `sbx` CLI (`tests/docker/test_sbx_cli.py`). Add coverage that spawns `spawn_attached_run` against a harmless real command and asserts: the returned controller fd is a live/readable pty fd, `stop_attached_run` terminates a cooperative process cleanly, the kill() escalation fires when the child ignores SIGTERM, and the fd is closed afterward.

tests/docker/toolkit/sbx_cli.py
```diff
@@ -270,6 +274,52 @@ def allow_network_for_hosts(hosts: Iterable[str], *, kit: Path | str) -> Iterato
...
+    `pty` is POSIX-only (no `termios` on Windows) and imported here rather
+    than at module level, so this module still imports cleanly for
+    collection on Windows CI — every sandbox-marked test that reaches this
+    function is already gated behind `sbx_available()`/`SANDBOX_TESTS_ENABLED`
+    and never runs there.
+    """
+    import pty  # noqa: PLC0415 -- POSIX-only, deferred so this module still collects on Windows
+
+    controller_fd, sandbox_fd = pty.openpty()
+    process = subprocess.Popen(
```
AlexanderZ-Band: issue: `controller_fd, sandbox_fd = pty.openpty()` (above) allocates two fds before this `subprocess.Popen(...)` call. If `Popen` raises (e.g. `sbx` not on PATH, or fd-table exhaustion), `os.close(sandbox_fd)` below never executes and `controller_fd` is never returned to any caller to close - both fds leak silently, with no exception handling or logging around the call. Wrap the `Popen` call so both fds are closed on failure (and log the failure) before re-raising.

tests/docker/toolkit/sbx_cli.py
```diff
@@ -270,6 +274,52 @@ def allow_network_for_hosts(hosts: Iterable[str], *, kit: Path | str) -> Iterato
...
+
+def stop_attached_run(process: subprocess.Popen[bytes], controller_fd: int) -> None:
+    """Tear down a `spawn_attached_run` session before `sbx rm`."""
+    if process.poll() is None:
+        process.terminate()
+        try:
+            process.wait(timeout=ATTACH_STOP_TIMEOUT_S)
+        except subprocess.TimeoutExpired:
+            process.kill()
+            process.wait(timeout=ATTACH_STOP_TIMEOUT_S)
```
AlexanderZ-Band: issue: this escalation (`terminate()` -> `wait()` -> on `TimeoutExpired` -> `kill()` -> `wait()`) never catches a second `TimeoutExpired` from this line. If the process is still unresponsive after SIGKILL (e.g. stuck in D-state), the exception propagates out of `stop_attached_run` and the `os.close(controller_fd)` below is skipped, leaking the pty fd on top of an incomplete caller cleanup. Separately, none of terminate/kill/either wait outcome is logged, so an operator can't tell from logs whether a session exited gracefully or had to be force-killed. Wrap the final wait so `os.close` always runs, and add a log line on the kill-escalation path.

tests/docker/toolkit/sbx_cli.py
```diff
@@ -321,9 +373,11 @@ def create(
             check=True,
             timeout=CREATE_TIMEOUT_S,
         )
+        attach_process, attach_fd = spawn_attached_run(name)
```
AlexanderZ-Band: issue: `sbx create` (above) is not inside the `try/finally` that starts right after this line - only `spawn_attached_run(name)` and everything after are covered. If `spawn_attached_run` raises here, the sandbox `name` was already created by `sbx create` but the `finally` block's `stop_attached_run` + `sbx rm -f` never run, and nothing logs that a real (billable) sandbox was left running with no attach and no cleanup path. Move the `sbx create` call inside (or ahead of) a try that guarantees `sbx rm -f name` on any exception from `spawn_attached_run`, and log a warning naming the orphaned sandbox if cleanup is skipped.

AlexanderZ-Band: issue (edge-case flow): starting state - no sandbox exists. Trigger sequence - `subprocess.run([SBX, "create", ...])` succeeds and creates a real sandbox named `name`, then this `spawn_attached_run(name)` call raises (e.g. `pty.openpty()` hits a fd/pty limit, or `Popen` raises `FileNotFoundError`/`OSError`) before the `try/finally` below is entered. Incorrect outcome: the exception propagates straight out of the `Sandbox.create` context manager, so neither `stop_attached_run` nor `sbx rm -f` ever runs - the just-created sandbox is left running, unattached, and untracked by the caller. (Same root cause as the adjacent Runtime Safety finding on this line - one fix should close both.)

AlexanderZ-Band (commented):
Submitting my-code-review pass 1 findings (see review body + 7 inline comments). REQUEST_CHANGES isn't permitted on one's own PR, so posting as COMMENT; treat the 5 issue: findings as blocking for the --fix pass that follows.

examples/sandbox/self-registration/demo.py
AlexanderZ-Band: Fixed. The duplicated `tuple[Popen, int]` shape is gone entirely, not just aliased — demo.py no longer touches the tuple at all; it now just does `with attached_run(name):`. The tuple lives only internally between the new private `_spawn_pty_process`/`_stop_pty_process` helpers in sbx_cli.py.

tests/docker/toolkit/sbx_cli.py
AlexanderZ-Band: Fixed. Replaced spawn_attached_run/stop_attached_run with a single `@contextmanager attached_run(name)`, matching this file's existing custom_secret/allow_network idiom. Both Sandbox.create and demo.py now use it, so cleanup is guaranteed by the `with` protocol instead of each caller pairing calls by hand.

AlexanderZ-Band: Fixed. Added 5 tests in tests/docker/test_sbx_cli.py covering the pty spawn/stop lifecycle: a live pty read, the fd-leak-on-Popen-failure regression, cooperative terminate, the SIGKILL escalation, and attached_run itself — none need the real sbx CLI, gated by a new requires_posix_pty marker (same style as the file's other requires_posix_* markers).

tests/docker/toolkit/sbx_cli.py
AlexanderZ-Band: Fixed. `_spawn_pty_process` now wraps the Popen call in try/except and closes both fds before re-raising on failure. Added a regression test (test_spawn_pty_process_closes_both_fds_when_the_child_fails_to_start, fd-count based) — verified by hand that it fails against the old code (2 fds leaked per failed spawn) and passes against the fix (0 leaked).

tests/docker/toolkit/sbx_cli.py
AlexanderZ-Band: Fixed. `_stop_pty_process` now wraps the whole terminate/kill escalation in try/finally so `os.close(controller_fd)` always runs regardless of how the escalation ends, and logs a warning both when the SIGKILL escalation fires and if the process is still alive after SIGKILL. Covered by test_stop_pty_process_escalates_to_sigkill_when_sigterm_is_ignored.

tests/docker/toolkit/sbx_cli.py
AlexanderZ-Band: Fixed. Sandbox.create now enters attached_run(name) inside the same try whose finally runs `sbx rm -f`, so any exception raised while attaching (including one from _spawn_pty_process itself) still triggers cleanup of the already-created sandbox. Same root cause as the adjacent edge-case-flow finding below — one fix covers both.

AlexanderZ-Band: Fixed together with the Runtime Safety finding above on this same line — Sandbox.create's sbx rm now runs on any exception from the attach step, so the leaked-sandbox flow described here no longer reproduces.

PR #612, opened 2026-09-06
feat: Adopt AgentFailure/send_failure across Python adapters

AlexanderZ-Band:
## Summary

Adopts band-sdk-core's shared `AgentFailure` shape (`provider`/`code`/`message`/`detail`)
across band-sdk-python adapters, replacing the free-text `"error"` event and
its ad hoc per-adapter metadata keys.

## Changes

- Bump `band-sdk-core` to 2.3.0
- `send_failure`/`to_failure_event` on `AgentToolsProtocol`, `AgentTools`, `FakeAgentTools`
- `DeliveryFailedError`/`deliver_reply` for Band-delivery vs. provider-failure separation
- Migrated adapters: Anthropic, Gemini, Google ADK, Claude SDK, Copilot SDK, LangGraph, Letta, A2A, A2A Gateway, CrewAI, CrewAI Flow, Pydantic AI, Parlant, Strands, Agno, OpenCode, Codex, ACP client
- ACP client: turn-level timeout (`turn_timeout_s`, default 300s) around `_runtime.prompt`
- A2A Gateway: structured failure channel on `PendingA2ATask.fail`, credential redaction on both the adapter's own exception path and the peer-forwarded-failure relay path
- Codex: removed the old remediation/suggested-action policy; turn-timeout, error-notification, and turn/completed failure paths each report exactly once
- Shared `FAILURE_CODE_TIMEOUT` constant; `MessageType.ERROR` used instead of a raw `"error"` literal
- `to_failure_event` blank-content check uses `has_visible_content()`
- Centralized `reported_failures()`/`events_of_type()` test helpers in `band.testing`
- A2A adapter: terminal task-event emission isolated from the try block's exception in `finally`
- CrewAI Flow: `record_failed` caps the room-visible message at 500 chars, preserves the full message in `AgentFailure.detail`

## Breaking Changes

- **Codex**: `CodexAdapterConfig.structured_errors` (previously `bool = True`, documented in `docs/adapters/codex.md`) is removed. The model uses `extra="forbid"`, so any existing deployment config, constructor kwarg, or `CODEX_STRUCTURED_ERRORS` env var still setting this field will now raise `pydantic.ValidationError` at construction instead of being silently ignored. No deprecation alias/passthrough is added — this repo's own conventions reject backwards-compat shims in favor of a clean break. **The eventual merge commit needs a `!` on its type (e.g. `refactor!:`) and a `BREAKING CHANGE:` footer** so release-please's changelog actually reflects this.

## Out of scope

[...]

PR #644, opened 2026-09-19
feat: Support dynamic ACP session configuration

AlexanderZ-Band:
Closes INT-1501

## Summary
- expose one generic ACP select-catalog resolver for provider-advertised model, reasoning-effort, and future configuration options
- revalidate each selection against the refreshed catalog returned by `session/set_config_option`
- cover generic and Copilot ACP adapters, restored sessions, grouped lists, dynamic catalog changes, and visible failures with real-wire harness tests
- add a runnable generic ACP bridge example that logs each provider catalog and safely applies an advertised model or reasoning preference

## Validation
- `uv run ruff check .`
- `uv run pyrefly check`
- `uv run pytest tests/integrations/acp tests/adapters/test_copilot_acp_adapter.py -q --no-cov`
- `uv run pytest tests/example_agents/test_acp_dynamic_config.py tests/example_agents/test_surface_guards.py tests/integrations/acp/test_session_config.py -q --no-cov`


src/band/integrations/acp/session_config.py
```diff
@@ -0,0 +1,161 @@
...
+            ) from error
+        except Exception as error:
+            raise ACPConfigError(
+                session_id=session_id,
+                option_id=option_id,
+                selected_value=selected_value,
+                message=str(error),
+            ) from error
+
+        if response is None or not isinstance(response.config_options, list):
```
AlexanderZ-Band: `config_options` being a list is not sufficient at this custom-transport boundary. A list containing malformed entries is accepted after the final selection (so the turn prompts), and causes an uncaught `AttributeError` on a later selection. Validate every refreshed entry as a `SessionConfigOption` and raise `ACPConfigError` before replacing the catalog. Please add no-prompt coverage for malformed final and intermediate responses.

src/band/integrations/acp/session_config.py
```diff
@@ -0,0 +1,161 @@
...
+            raise ACPConfigError(
+                session_id=session_id,
+                option_id=option_id,
+                selected_value=selected_value,
+                message=(
+                    f'ACP config option "{option_id}" did not return a refreshed '
+                    "catalog."
+                ),
+            )
+        catalog = tuple(response.config_options)
```
AlexanderZ-Band: A complete typed response that leaves this select option's `current_value` unchanged is treated as success, so the adapter prompts with the old effort/model. The refreshed catalog is the setter acknowledgement; require the selected option to be present as a select and reflect `selected_value` before continuing.

src/band/integrations/acp/client_adapter.py
```diff
@@ -566,10 +593,14 @@ async def _get_or_create_session(self, room_id: str) -> tuple[str, bool]:
 
             mcp_servers = await self._session_mcp_servers()
 
-            session_id = await self._runtime.create_session(
+            session = await self._runtime.create_session_response(
                 cwd=self._cwd,
                 mcp_servers=mcp_servers,
             )
+            session_id = session.session_id
+            await self._configure_session(
```
AlexanderZ-Band, line 750: This awaits the caller resolver and remote setter while holding the adapter-wide `_session_lock`. A slow configuration decision for one room blocks unrelated rooms from creating or restoring sessions and prevents cleanup from acquiring the lock. Coalesce establishment per room, retaining the shared lock only for shared state, so configuration I/O does not serialize rooms.

AlexanderZ-Band, line 750: If configuration fails here, the newly-created remote session has not been mapped and is never cancelled/closed; retries create another remote session. Clean up a newly-created session best-effort on this failure path before surfacing the configuration error, without touching restored sessions.

src/band/integrations/acp/client_adapter.py
```diff
@@ -586,6 +617,78 @@ async def _session_mcp_servers(self) -> list[object]:
...
+        except (asyncio.CancelledError, KeyboardInterrupt):
+            raise
+        except Exception as error:
+            raise ACPConfigError(
+                session_id=session_id,
+                option_id="resolver",
+                selected_value="",
+                message=f"ACP session configuration resolver failed: {error}",
+            ) from error
+        if not selections:
```
AlexanderZ-Band: A falsy invalid resolver result such as `[]`, `""`, or `0` bypasses `apply_session_config_selections()` and silently prompts using provider defaults. Only `None` should opt out; pass every other result to the helper so its mapping validation produces the documented visible configuration error.

AlexanderZ-Band (commented):
Cycle 1 review: configuration failures must remain visible and leak-free, and configuration must not serialize independent rooms.

tests/integrations/acp/test_session_config.py
```diff
@@ -0,0 +1,195 @@
...
+    @pytest.mark.asyncio
+    async def test_generic_harness_applies_the_remote_effort_catalog(self) -> None:
+        effort = select_option("reasoning_effort", "medium", ["medium", "high"])
+        agent = FakeACPAgent(config_options=[effort]).will_say("Configured")
+
+        async def resolve_config(request: ACPConfigRequest) -> dict[str, str]:
+            assert request.config_options == (effort,)
+            return {"reasoning_effort": "high"}
+
+        async with acp_adapter(agent, resolve_session_config=resolve_config) as session:
```
AlexanderZ-Band, line 213: The real-wire success test checks that configuration happened eventually, but not that it happened before the first prompt. Assert the selected current value or a unified RPC sequence from the prompt handler so moving the setter after `prompt` cannot keep this test green.

tests/integrations/acp/acp_toolkit/agent.py
```diff
@@ -25,13 +25,20 @@
...
+    SessionConfigOptionSelect,
+    SetSessionConfigOptionResponse,
     ToolCallUpdate,
 )
 from mcp import ClientSession
 from mcp.client.streamable_http import streamable_http_client
 
 
 PromptHandler = Callable[["FakeACPAgent", str], Awaitable[None]]
+SessionConfigOption = SessionConfigOptionSelect | SessionConfigOptionBoolean
```
AlexanderZ-Band: nit: The fake agent repeats the production catalog union. Import the shared `SessionConfigOption` alias so a future catalog addition cannot make production validation and harness typing silently diverge.

tests/integrations/acp/acp_toolkit/agent.py
```diff
@@ -77,6 +88,11 @@ def on_prompt(self, handler: PromptHandler) -> PromptHandler:
         self._custom = handler
         return handler
 
+    def on_config_option(self, handler: ConfigOptionHandler) -> ConfigOptionHandler:
+        """Set dynamic behavior for the next ``session/set_config_option`` call."""
```
AlexanderZ-Band: nit: This handler is retained and used for every later configuration call, rather than only the next one. Please describe its persistent behavior (or clear it after one call if one-shot semantics were intentional).

src/band/integrations/acp/session_config.py
AlexanderZ-Band: Fixed in 6db6104. `refreshed_catalog()` now requires every entry to be a supported ACP option before later selection lookup, with final and intermediate malformed-catalog regressions.

src/band/integrations/acp/session_config.py
AlexanderZ-Band: Fixed in 6db6104. `refreshed_catalog()` now requires the selected option to remain a select and acknowledge the requested current value before prompting; covered by a stale-current-value regression.

src/band/integrations/acp/client_adapter.py
AlexanderZ-Band: Fixed in 6db6104. Room setup is now coalesced by a per-room in-flight task; the shared lock only guards state. A real-wire two-room test proves a blocked resolver in one room does not delay the other.

AlexanderZ-Band: Fixed in 6db6104. Fresh setup now uses an async context manager that sends stable `session/cancel` if configuration fails or is cancelled before publication. The fake agent and regressions verify this lifecycle.

src/band/integrations/acp/client_adapter.py
AlexanderZ-Band: Fixed in 6db6104. Only `None` now opts out; all other resolver results reach the mapping validator. The real-wire falsy-list regression asserts an error, no prompt, and session cancellation.

tests/integrations/acp/test_session_config.py
AlexanderZ-Band: Fixed in 6db6104. The prompt handler now asserts the config request at prompt entry, so configuration moved after prompt would fail the real-wire test.

tests/integrations/acp/acp_toolkit/agent.py
AlexanderZ-Band: Fixed in 6db6104. The fake agent imports the production `SessionConfigOption` alias as the single catalog contract.

tests/integrations/acp/acp_toolkit/agent.py
AlexanderZ-Band: Fixed in 6db6104. The handler documentation now accurately states that it governs every later config-option call.

src/band/integrations/acp/client_adapter.py
```diff
@@ -551,33 +581,136 @@ async def _get_or_start_band_mcp_server(self) -> LocalMcpServerConfig:
...
+    ) -> AsyncIterator[NewSessionResponse]:
+        """Yield a new session, cancelling it unless initialization completes."""
+        session = await self._runtime.create_session_response(
+            cwd=self._cwd,
+            mcp_servers=mcp_servers,
+        )
+        try:
+            yield session
+        except BaseException:
+            await self._cancel_fresh_session(session.session_id)
```
AlexanderZ-Band: `session/cancel` only aborts an active prompt; this fresh session has never prompted, so it does not dispose of the remote session. Track the advertised `sessionCapabilities.close` capability in the runtime and use capability-gated `session/close` here; retain cancellation only as an active-turn operation.

src/band/integrations/acp/client_adapter.py
```diff
@@ -551,33 +581,136 @@ async def _get_or_start_band_mcp_server(self) -> LocalMcpServerConfig:
...
+            if initializer is None:
+                initializer = asyncio.create_task(
+                    self._initialize_session(room_id, history),
+                    name=f"acp-session:{room_id}",
+                )
+                self._session_initializers[room_id] = initializer
 
-            mcp_servers = await self._session_mcp_servers()
+        try:
+            return await asyncio.shield(initializer)
```
AlexanderZ-Band: Shielding lets this initializer outlive a cancelled sole caller. It stays in `_session_initializers`, so the next room turn waits behind abandoned resolver work (and a completed failure can be replayed indefinitely). Give initialization explicit waiter ownership: cancel/remove it when its last waiter leaves, while preserving sharing for concurrent same-room turns; also remove completed entries independently of waiters.

tests/integrations/acp/test_session_config.py
```diff
@@ -0,0 +1,355 @@
...
+            session_id="session-1",
+            config_options=[grouped_model],
+            selections={"model": "large"},
+            set_option=set_option,
+        )
+
+        set_option.assert_awaited_once_with("session-1", "model", "large")
+
+    @pytest.mark.asyncio
+    async def test_fails_when_a_prior_selection_removes_a_later_effort_option(
```
AlexanderZ-Band, line 128: The dynamic catalog transition is only covered through `AsyncMock`. Add a real-wire `FakeACPAgent.on_config_option` case where selecting model removes effort, asserting a visible error, no prompt, and fresh-session lifecycle cleanup.

tests/integrations/acp/test_session_config.py
```diff
@@ -0,0 +1,355 @@
...
+            await fake.say(session_id, "Configured")
+
+        async def resolve_config(request: ACPConfigRequest) -> dict[str, str]:
+            assert request.config_options == (effort,)
+            return {"reasoning_effort": "high"}
+
+        async with acp_adapter(agent, resolve_session_config=resolve_config) as session:
+            reply = await session.send("Configure the session")
+
+        assert reply.messages[0]["content"] == "Configured"
```
AlexanderZ-Band: nit: Use `Reply.texts` rather than indexing reply message dictionaries, so this test asserts the harness's observable content projection rather than payload layout. Apply the same cleanup to the other new success assertions.

AlexanderZ-Band (commented):
Cycle 2 review: session cleanup must use the advertised lifecycle capability, and initializer ownership must survive interrupted turns safely.

src/band/integrations/acp/client_adapter.py
AlexanderZ-Band: Fixed in 8017b16. Runtime records `sessionCapabilities.close`; the fresh-session context manager now invokes capability-gated `session/close`, with real-wire capability and lifecycle coverage.

src/band/integrations/acp/client_adapter.py
AlexanderZ-Band: Fixed in 8017b16. A `SessionInitializer` now tracks same-room waiters. The last departing waiter removes and cancels pending setup; concurrent waiters still share it. The interrupted-turn regression proves retry is not blocked.

tests/integrations/acp/test_session_config.py
AlexanderZ-Band: Fixed in 8017b16. A real-wire `on_config_option` test removes effort after model selection and verifies the visible no-prompt failure plus fresh-session close.

tests/integrations/acp/test_session_config.py
AlexanderZ-Band: Fixed in 8017b16. The configuration tests consistently use the harness `Reply.texts` projection for response assertions.

src/band/integrations/acp/client_adapter.py
```diff
@@ -110,6 +133,16 @@ def new_message_marker() -> str:
 SpawnProcess = Callable[..., object]
 
 
+def session_config_options(response: object) -> list[SessionConfigOption] | None:
```
AlexanderZ-Band: nit: Initial/load catalog parsing duplicates the refreshed-catalog validation. Centralize the parser in `session_config` so the accepted ACP catalog shape cannot drift between session creation, restoration, and setter acknowledgements.

src/band/integrations/acp/client_adapter.py
```diff
@@ -586,6 +747,78 @@ async def _session_mcp_servers(self) -> list[object]:
...
+        self,
+        room_id: str,
+        session_id: str,
+        config_options: list[SessionConfigOption] | None,
+    ) -> None:
+        """Apply caller-selected values from the session's live ACP catalog."""
+        if self._resolve_session_config is None:
+            return
+
+        if not config_options:
```
AlexanderZ-Band, line 869: issue: ACP normalizes malformed initial `configOptions` entries to an empty list before this adapter sees them. This guard treats that as absent and prompts without calling the resolver, despite the documented visible malformed-response failure. Preserve malformed-catalog state at the parsing boundary or narrow the documented guarantee; validating here alone cannot distinguish it from a genuinely absent catalog.

src/band/integrations/acp/client_adapter.py
```diff
@@ -586,6 +747,78 @@ async def _session_mcp_servers(self) -> list[object]:
...
+                    session_id=session_id,
+                    config_options=catalog,
+                )
+            )
+        except (asyncio.CancelledError, KeyboardInterrupt):
+            raise
+        except Exception as error:
+            raise ACPConfigError(
+                session_id=session_id,
+                option_id="resolver",
```
AlexanderZ-Band: nit: `resolver` is a diagnostic pseudo-option id also defined in `session_config`. Give it one internal constant so resolver errors and their tests use a single vocabulary.

tests/integrations/acp/test_e2e_codex_acp.py
```diff
@@ -218,6 +255,70 @@ async def test_codex_acp_prompt_and_collect(acp_runtime: ACPRuntime) -> None:
...
+        return await acp_runtime.set_config_option(
+            session_id=session_id,
+            config_id=option_id,
+            value=value,
+        )
+
+    await apply_session_config_selections(
+        session_id=session.session_id,
+        config_options=config_options,
+        selections=non_current_config_selections(config_options),
```
AlexanderZ-Band: issue: This chooses both values from the initial catalog, even though a model selection may change available reasoning levels. The production helper correctly rejects that transition, so a correct provider can make this E2E fail before its two-turn assertions. Apply one selection per session or derive the later value from the refreshed catalog.

tests/integrations/acp/test_client_runtime.py
```diff
@@ -696,6 +696,19 @@ async def test_create_session_and_prompt_use_active_connection(self) -> None:
         mock_conn.new_session.assert_awaited_once_with(cwd="/tmp", mcp_servers=[])
         mock_conn.prompt.assert_awaited_once()
 
+    @pytest.mark.asyncio
+    async def test_close_session_uses_the_active_connection_when_supported(
```
AlexanderZ-Band, line 700: issue: The close lifecycle test covers only the advertised-capability path. Add the default unsupported case and assert no `session/close` RPC is sent, so a regression cannot call an unsupported remote method during cleanup.

AlexanderZ-Band (commented):
Five verified cycle-1 findings for remediation.

src/band/integrations/acp/client_adapter.py
AlexanderZ-Band: Fixed in fa0164a. `session_config_options()` now owns typed catalog parsing for initial/load and refreshed setter responses.

src/band/integrations/acp/client_adapter.py
AlexanderZ-Band: Verified: the ACP dependency discards malformed initial entries before this SDK can distinguish them from an absent catalog. The adapter keeps the correct empty-catalog no-op; docs now limit the visible malformed-response guarantee to observable `session/set_config_option` responses.

src/band/integrations/acp/client_adapter.py
AlexanderZ-Band: Fixed in fa0164a. `RESOLVER_CONFIG_OPTION_ID` is now the single production and test vocabulary for resolver failures.

tests/integrations/acp/test_e2e_codex_acp.py
AlexanderZ-Band: Fixed in fa0164a. The live flow is parametrized by one advertised option per fresh session, so model and effort are each exercised without assuming cross-catalog compatibility.

tests/integrations/acp/test_client_runtime.py
AlexanderZ-Band: Fixed in fa0164a. Added the default capability-negative regression asserting no `session/close` RPC is sent.

src/band/integrations/acp/client_adapter.py
```diff
@@ -551,33 +583,154 @@ async def _get_or_start_band_mcp_server(self) -> LocalMcpServerConfig:
...
+    ) -> AsyncIterator[NewSessionResponse]:
+        """Yield a new session, closing it unless initialization completes."""
+        session = await self._runtime.create_session_response(
+            cwd=self._cwd,
+            mcp_servers=mcp_servers,
+        )
+        try:
+            yield session
+        except BaseException:
+            await self._close_fresh_session(session.session_id)
```
AlexanderZ-Band: issue: If an agent advertises `session/close` but never answers it, fresh-session cleanup waits forever and the original configuration failure never reaches the room. Bound this best-effort close and log a timeout so `_report_config_error` remains reachable.

tests/integrations/acp/test_session_config.py
```diff
@@ -0,0 +1,433 @@
...
+            interrupted_turn = asyncio.create_task(session.send("Interrupted"))
+            await first_resolver_started.wait()
+            interrupted_turn.cancel()
+            with pytest.raises(asyncio.CancelledError):
+                await interrupted_turn
+
+            reply = await session.send("Retry")
+
+        assert reply.texts == ["Configured"]
+        assert [item["session_id"] for item in agent.sessions] == [
```
AlexanderZ-Band: nit: This lifecycle assertion reaches into the fake agent's raw session dictionaries. Add an intent-named session-id projection beside `prompt_texts()` and assert through it, so the test is decoupled from harness storage layout.

AlexanderZ-Band (commented):
Two verified cycle-2 findings for remediation.

src/band/integrations/acp/client_adapter.py
AlexanderZ-Band: Fixed in fca2102. Fresh-session close is bounded by `SESSION_CLOSE_TIMEOUT_SECONDS`, logs timeout, and returns so the original configuration error remains reportable; the blocked-close regression covers it.

tests/integrations/acp/test_session_config.py
AlexanderZ-Band: Fixed in fca2102. `FakeACPAgent.session_ids()` now owns the lifecycle projection and the retry assertion no longer reads raw harness dictionaries.

src/band/integrations/acp/client_adapter.py
```diff
@@ -551,33 +584,163 @@ async def _get_or_start_band_mcp_server(self) -> LocalMcpServerConfig:
...
```
AlexanderZ-Band: issue: Cancellation of an initializing turn is delayed by the bounded close. If the agent never answers `session/close`, the cancelled turn waits for five seconds and blocks the next same-room turn. Preserve immediate cancellation while scheduling the bounded best-effort close independently.

AlexanderZ-Band (commented):
One verified cycle-3 finding for remediation.

src/band/integrations/acp/client_adapter.py
AlexanderZ-Band: Fixed in a7468cc. Cancellation schedules the already bounded fresh-session close and re-raises immediately; the blocked-close regression proves cancellation completes before cleanup.

examples/acp/clients/generic.py
```diff
@@ -65,6 +76,53 @@ class Settings(BaseSettings):
...
+    preferences: Mapping[str, str],
+) -> dict[str, str]:
+    """Choose the first configured value that this session advertises."""
+    catalog = advertised_config_values(request)
+    logger.info("ACP session '%s' config options: %s", request.session_id, catalog)
+
+    for option_id, selected_value in preferences.items():
+        if selected_value in catalog.get(option_id, ()):
+            return {option_id: selected_value}
+        if option_id in catalog:
```
AlexanderZ-Band: issue: A configured preference is silent when the harness does not advertise that option id at all. For example, `ACP_MODEL` against an effort-only harness returns `{}` without revealing that the requested setting was unavailable. Emit a warning for the absent-option case and cover it, so the generic bridge does not appear to have applied a default.

AlexanderZ-Band (commented):
One in-scope observability issue found in the dynamic ACP example.

examples/acp/clients/generic.py
AlexanderZ-Band: Fixed in 43d5bb2. The generic resolver now warns when a configured option id is absent, then continues to the next preference; the regression asserts both the warning and the fallback to advertised reasoning effort.

tests/example_agents/test_acp_dynamic_config.py
```diff
@@ -0,0 +1,114 @@
...
+            select_option("reasoning_effort", "medium", ["medium", "high"]),
+        ),
+    )
+
+    selections = await generic_acp.choose_session_config(
+        request,
+        preferences={"model": "large", "reasoning_effort": "high"},
+    )
+
+    assert selections == {"reasoning_effort": "high"}
```
AlexanderZ-Band, line 110: nit: This warning assertion reconstructs a log projection from raw records. Scope capture to warnings for the example logger with `caplog.at_level(...)`, then assert `caplog.messages` directly; that expresses the observable outcome without coupling the test to record layout.

AlexanderZ-Band (commented):
One test-quality cleanup found in cycle two.

tests/example_agents/test_acp_dynamic_config.py
AlexanderZ-Band: Fixed in c20b4b1. The test now scopes capture to warning-level records from the example logger and asserts the captured messages directly.

src/band/integrations/acp/client_adapter.py
```diff
@@ -551,33 +584,166 @@ async def _get_or_start_band_mcp_server(self) -> LocalMcpServerConfig:
...
+    ) -> AsyncIterator[NewSessionResponse]:
+        """Yield a new session, closing it unless initialization completes."""
+        session = await self._runtime.create_session_response(
+            cwd=self._cwd,
+            mcp_servers=mcp_servers,
+        )
+        try:
+            yield session
+        except asyncio.CancelledError:
+            asyncio.create_task(self._close_fresh_session(session.session_id))
```
AlexanderZ-Band: issue: `asyncio.create_task(self._close_fresh_session(session.session_id))` discards its Task with no stored reference and no done-callback. Per asyncio's own docs, a Task with no other reference can be garbage-collected before it runs, silently dropping the best-effort close this branch exists to perform (the exact leak #4/#9/#18/#20 already fixed for other paths). It also has no exception visibility if `_close_fresh_session` ever raises unexpectedly, and nothing drains it on `cleanup_all`/shutdown. `src/band/integrations/slack/adapter.py` already has the fix for this exact shape (`_background_tasks` set + `add_done_callback(self._on_background_task_done)` + draining the set on shutdown) — reuse that pattern here instead of a second, unretained `create_task` call.

AlexanderZ-Band (commented):
my-code-review (pass 1/3): 1 finding (Runtime Safety) — fixing now under --fix.

src/band/integrations/acp/client_adapter.py
AlexanderZ-Band: Fixed in c1b350d9: added `_background_tasks`/`_track_background_task`/`_on_background_task_done` to `ACPClientAdapter` (mirrors the existing pattern in `integrations/slack/adapter.py`) and switched the CancelledError branch of `_fresh_session` to use it instead of a bare `asyncio.create_task`. `cleanup_all` now drains pending background tasks via `_drain_background_tasks` before the runtime stops.

While implementing this, found and fixed a second, more serious bug in my own first draft of `_drain_background_tasks`: on Python 3.13+/3.14, `asyncio.gather()` resolves eagerly/synchronously when every awaited future is already done (confirmed by reading CPython's `asyncio/tasks.py` `gather()` source), so a naive `while tasks: await asyncio.gather(*tasks)` never actually suspends and can spin forever without ever letting the done-callback shrink the set. Reproduced the hang directly against a real event loop, then fixed `_drain_background_tasks` to explicitly discard its awaited snapshot instead of depending on callback timing. Added two regression tests (`test_cancelled_fresh_session_close_is_tracked_not_lost`, `test_cleanup_all_waits_for_background_close_tasks`) that reproduce this exact call pattern and would hang/fail without the fix.

src/band/integrations/acp/client_adapter.py
```diff
@@ -607,33 +640,213 @@ async def _get_or_start_band_mcp_server(self, room_id: str) -> LocalMcpServerCon
...
-                "Created ACP session %s for room %s (mcp_servers=%d)",
+                "Persisted ACP session %s is unavailable for room %s; using a new session",
                 session_id,
                 room_id,
-                len(mcp_servers),
             )
-            return session_id, True
+            return None
+
+        await self._configure_session(
```
amit-gazal-band: If `_configure_session` raises here (e.g. a persisted model is no longer advertised), the successfully loaded session is neither recorded nor closed. Fresh sessions get closed via `_fresh_session` in the same situation, but this one stays open in the agent process for the life of the runtime, and the next turn creates a new session on top of it. Consider closing the loaded session on failure, mirroring `_close_fresh_session`.

src/band/integrations/acp/client_adapter.py
```diff
@@ -642,6 +855,71 @@ async def _session_mcp_servers(self, room_id: str) -> list[object]:
...
+            ),
+        )
+
+    async def _report_config_error(
+        self,
+        tools: AgentToolsProtocol,
+        error: ACPConfigError,
+    ) -> None:
+        logger.warning("ACP session configuration failed: %s", error)
+        await tools.send_event(
```
amit-gazal-band: This posts via the raising `tools.send_event(..., message_type="error")` rather than the best-effort `send_failure`/`AgentFailure` path used elsewhere since #612, and retypes `"error"` instead of `MessageType.ERROR`. If the REST post itself fails (transient 5xx), that exception escapes `on_message` outside the turn's error handler, so the real configuration error never reaches the room and is replaced by an unrelated one.

src/band/integrations/acp/client_adapter.py
```diff
@@ -642,6 +855,71 @@ async def _session_mcp_servers(self, room_id: str) -> list[object]:
...
+        runtime: ACPRuntime,
+        room_id: str,
+        session_id: str,
+        config_options: tuple[SessionConfigOption, ...] | None,
+    ) -> None:
+        """Apply caller-selected values from the session's live ACP catalog."""
+        if self._resolve_session_config is None:
+            return
+
+        if not config_options:
```
amit-gazal-band: When the session advertises no `configOptions`, this returns without calling `resolve_session_config`. `docs/acp.md` says the callback runs once after each new/restored session and that a selection fails the turn visibly rather than silently falling back. With e.g. `ACP_MODEL` set against an agent that exposes no catalog, nothing is logged and every turn quietly runs on the default model. Either call the resolver with an empty catalog (so it can raise) or adjust the docs to match.

src/band/integrations/acp/client_adapter.py
```diff
@@ -607,33 +640,213 @@ async def _get_or_start_band_mcp_server(self, room_id: str) -> LocalMcpServerCon
...
         """
-        if room_id in self._room_to_session:
-            return self._room_to_session[room_id], False
-
-        runtime = await self._runtime_for(room_id)
-        mcp_servers = await self._session_mcp_servers(room_id)
         async with self._session_lock:
             if room_id in self._room_to_session:
                 return self._room_to_session[room_id], False
+            initializer = self._session_initializers.get(room_id)
```
amit-gazal-band: A failed `SessionInitializer` stays in `_session_initializers` until its last waiter's `finally` reacquires `_session_lock`. A second turn for the same room that wins the lock first sees the already-failed task, bumps `waiters`, and gets the old `ACPConfigError` without retrying, so both turns post the same configuration error. Consider treating a `done()` task with an exception as absent here and starting a fresh initializer.

amit-gazal-band (dismissed):
LGTM. The SessionInitializer locking, the cancellation paths (including cleanup racing or cancelling setup) and draining the background close tasks all hold up, and `acp>=0.11.0` ships `set_config_option`, `close_session` and every schema class imported here. Four low-severity notes inline, none blocking. The restored-session leak and the error-reporting path are the most worth a follow-up.

src/band/integrations/acp/client_adapter.py
AlexanderZ-Band: Fixed in 0beff10: failed configuration of a restored session now closes that session before the error is reported.

src/band/integrations/acp/client_adapter.py
AlexanderZ-Band: Fixed in 0beff10: configuration failures now use send_failure with an ACP AgentFailure detail, preserving the structured context while making reporting best-effort.

src/band/integrations/acp/client_adapter.py
AlexanderZ-Band: Fixed in 0beff10: the resolver now receives an empty catalog as well, so unsupported selections fail visibly instead of silently using the default.

src/band/integrations/acp/client_adapter.py
AlexanderZ-Band: Fixed in 0beff10: a completed failed or cancelled initializer is discarded under the session lock, allowing the next turn to create a fresh session.

PR #645, opened 2026-09-20
fix: harden opencode turn and approval lifecycle

AlexanderZ-Band:
## Summary
- isolate each OpenCode turn's mutable delivery state
- make approval abandonment explicit and retry MCP registration until connected
- protect fake identity and group-start cleanup boundaries

## Validation
- `uv run pytest tests/adapters/opencode tests/testing/test_fake_tools.py tests/skills/bughunting/test_runner.py -q`
- `uv run ruff check .`
- `uv run pyrefly check`


src/band/adapters/opencode/adapter.py
```diff
@@ -899,102 +954,115 @@ async def _ensure_session(
...
                 "OpenCode turn timed out for room %s (session=%s)",
                 room_id,
-                room_state.session_id,
+                turn.session_id,
             )
-            await self._abort_session(room_state, "timed-out")
-            if room_state.tools:
-                await room_state.tools.send_event(
+            await self._abort_turn(turn, "timed-out")
+            if turn.tools:
```
AlexanderZ-Band: issue: `_watch_turn_completion`'s `except asyncio.TimeoutError:` branch (994-1008) calls `await self._abort_turn(...)`, `await turn.tools.send_event("OpenCode timed out before completing the turn.", "error")` (this line), and `await self._emit_turn_usage(turn)` (line 1008) with no surrounding try/except — unlike the `else:` branch immediately below (1009-1017), which wraps its equivalent `_deliver_fallback_text`/`_emit_turn_usage` calls in `try/except Exception: logger.exception(...); await self._report_delivery_failure(...)`.

Proven with a repro test (`FakeAgentTools.send_event` raising for the timeout message, `turn_timeout_s` forced low, no `session.idle` ever sent): the exception escapes the detached `asyncio.create_task`-launched watcher uncaught (`watcher_task.exception()` returns the injected error; `on_message`'s own task sees nothing), `_emit_turn_usage` never runs (`tools.events_sent == []` despite `Emit.USAGE`), and there is no compensating room-visible failure report on this path, unlike what the success path produces via `_report_delivery_failure`. `finally:` still clears turn state, so `on_message` itself doesn't hang.

This is exactly the ticket item this function was supposed to close ("the turn-timeout branch's send_event/_emit_turn_usage are unguarded") — it wasn't.

Fix: wrap the `send_event`/`_emit_turn_usage` calls in this branch in the same `try/except Exception: logger.exception(...); await self._report_delivery_failure(room_state.room_id, turn)` pattern the success branch already uses.

src/band/adapters/opencode/adapter.py
```diff
@@ -488,8 +516,10 @@ async def on_message(
             # guard until turn_timeout_s. Drop the turn state first (that cancels
             # the watcher), then ask OpenCode to stop working.
             if turn_future is not None:
-                self._clear_turn_state(room_state, expected_future=turn_future)
-                await self._abort_session(room_state, "interrupted")
+                assert turn is not None
+                await turn.approvals.abandon()
+                self._clear_turn_state(room_state, expected_turn=turn)
+                await self._abort_turn(turn, "interrupted")
```
AlexanderZ-Band: issue: in `on_message`'s `except asyncio.CancelledError:` branch, `await turn.approvals.abandon()` (line 520) already calls `self._ports.abort_session()` → `_abort_owner_turn` → `_abort_turn(turn, "approval abandoned")` whenever something was actually pending. The very next lines then unconditionally call `await self._abort_turn(turn, "interrupted")` (this line) again for the same turn/session — a second `client.abort_session(session_id)` call every time a turn is interrupted, not only when an approval was pending.

Each call is independently guarded/logged, so nothing crashes, but this is a wasted round trip on every interrupt, and if OpenCode's abort endpoint treats an already-aborted session as an error, it produces a misleading `logger.exception("Failed to abort %s OpenCode session %s", "interrupted", ...)` traceback for expected, already-handled behavior.

Fix: don't call abort twice — either skip the explicit `_abort_turn(turn, "interrupted")` when `abandon()` already aborted (have it report whether it aborted), or make `_abort_turn`/`abort_session` idempotent per turn (e.g. a `TurnState.aborted` flag).

src/band/adapters/opencode/approvals.py
```diff
@@ -287,7 +306,9 @@ async def try_handle_reply(self, content: str, sender_id: str | None) -> bool:
         mentions = [{"id": sender_id}] if sender_id else []
 
         approval = parse_permission_reply(command)
-        if approval and self._permissions:
+        if approval and (
```
AlexanderZ-Band: issue: the "a reply naming an unknown/stale request id gets feedback instead of being silently forwarded" fix only recognizes an id the *current* `RoomApprovals` instance's `_handled_permission_ids`/`_handled_question_ids` remembers (`self._permissions or approval.request_id in self._handled_permission_ids`, and the analogous question-rejection check at 365-374). But `_begin_turn` (adapter.py:957-974) rebuilds `room_state.approvals` from scratch every turn via `_new_approvals`, discarding the previous instance's handled-id memory entirely.

Proven with two passing/failing tests against PR head: (1) a single `RoomApprovals` with zero pending permissions — `try_handle_reply("approve never-asked-id", ...)` returns `False` (not consumed, no feedback) even for a same-turn, never-seen id; (2) a real `OpencodeAdapter` driven through 3 turns — turn 1 asks+resolves `perm-1` (confirmed it lands in turn 1's `_handled_permission_ids`); turn 2 (any message) triggers `_begin_turn`, replacing `room_state.approvals` with a fresh instance (confirmed `perm-1` is absent from its handled set); turn 3 resends `"approve perm-1"` — it gets forwarded to OpenCode as a brand-new model prompt (`fake_client.prompt_calls[2]` literally contains `'approve perm-1'`), with no "no longer pending" notice ever posted.

Fix: track handled/recently-seen ids at a scope that survives `_begin_turn`'s reconstruction (e.g. on `RoomState`, or explicitly seed a new `RoomApprovals`'s handled-id sets from the prior instance at turn boundaries).

src/band/adapters/opencode/approvals.py
```diff
@@ -247,9 +262,13 @@ async def on_permission_asked(self, request: OpencodePermissionRequest) -> None:
 
     async def on_question_asked(self, request: OpencodeQuestionRequest) -> None:
         request_id = request.id
-        if not request_id:
+        if not request_id or not request.questions:
```
AlexanderZ-Band, line 289: issue: the `if not request_id or not request.questions: return` guard (this line) — added so a malformed `question.asked` with no questions can't park the turn — has zero test coverage. Every `on_question_asked`/`event_question` call anywhere in `tests/adapters/opencode/test_approvals.py` supplies at least one question; none passes `questions=[]` to prove the ask is dropped instead of creating a `PendingQuestion`, starting an expiry timer, and parking the turn on a human. This is one of this PR's own explicit ticket items.

Add: `await approvals.on_question_asked(OpencodeQuestionRequest(id="q-x", questions=[]))` then assert `approvals.awaiting_human()` stays `False` and no room message was posted.

src/band/adapters/opencode/approvals.py
```diff
@@ -217,6 +227,11 @@ async def on_permission_asked(self, request: OpencodePermissionRequest) -> None:
             await self._approve_own_band_tool(request_id)
             return
 
+        existing = self._permissions.get(request_id)
```
AlexanderZ-Band: issue: the cancel-before-overwrite fix on a redelivered ask (`existing = self._permissions.get(request_id); _cancel_timeout(existing); if existing and existing.replying: return`, this line, and the analogous block at 268-271 for questions) — plus the new `PendingPermission.replying`/`PendingQuestion.replying` in-flight guard used by `_reply_permission`/`_reply_question`/`_reject_question` — has zero test coverage. No test in the file calls `on_permission_asked`/`on_question_asked` twice with the SAME request id; every test uses distinct ids. This is another of this PR's own explicit ticket items ("a redelivered ask with the same request id orphans the first expiry timer, which then auto-rejects the replacement").

Add: using the existing `BlockingReplyClient` pattern, redeliver the same id mid-reply and assert the redelivery is ignored (no new `PendingPermission` overwrites the in-flight one); separately, redeliver the same id with no reply in flight and assert the first entry's `timeout_task` gets cancelled instead of leaking.

src/band/adapters/opencode/approvals.py
```diff
@@ -385,77 +423,81 @@ async def _notify_room(self, text: str, mentions: list[dict[str, str]]) -> None:
...
         self._questions.clear()
         # No ask is parked anymore -- release any watcher waiting on us.
         self._release_from_human()
 
+    async def abandon(self) -> None:
+        """Stop a parked session after local approval state is discarded."""
+        was_pending = bool(self._permissions or self._questions)
+        self.cancel()
+        if was_pending:
+            await self._ports.abort_session()
```
AlexanderZ-Band, line 552: issue: `abandon()`'s `await self._ports.abort_session()` call (this line, gated on `was_pending`) is this PR's explicit ticket item ("abandonment never tells OpenCode... make it actually abort the session"), but the two tests that exercise the exact condition it's for — `test_cleanup_with_pending_permission` and `test_cleanup_with_pending_question` (both call `on_cleanup` while a permission/question is genuinely pending) — never assert on `fake_client.aborted_sessions`. Both tests pass identically even with `await room_state.approvals.abandon()` deleted from `on_cleanup` entirely.

Add to both tests, after `on_cleanup("room-1")`: `assert fake_client.aborted_sessions == ["sess-1"]` (or the actual session id) to prove cleanup now aborts the OpenCode session instead of only clearing local state.

tests/adapters/opencode/helpers.py
```diff
@@ -360,7 +360,7 @@ async def abort_session(self, session_id: str) -> None:
     async def register_mcp_server(self, *, name: str, url: str) -> dict[str, Any]:
         self.registered_mcp_servers.append({"name": name, "url": url})
         self.serve_registrations[name] = url
-        return {"name": name, "url": url}
+        return {name: {"status": "connected"}}
```
AlexanderZ-Band: issue: `FakeOpencodeClient.register_mcp_server` now always returns `{name: {"status": "connected"}}`, so no test can ever reach `_register_mcp_backend`'s new branch (`adapter.py:706`, `if result.get(name, {}).get("status") != "connected": logger.warning(...); return`) that leaves `_registered_client` unset on a non-connected result. This is this PR's own explicit ticket item ("retry MCP registration until it actually reports connected") — the retry itself (every `_ensure_client_started` call re-attempts registration while `_registered_client is not client`) is likewise never exercised.

Add: a `FakeOpencodeClient` variant whose `register_mcp_server` returns `{name: {"status": "pending"}}` once then `{name: {"status": "connected"}}`; assert the first `_ensure_client_started()` leaves the client unregistered and a subsequent call marks it registered.

src/band/testing/fake_tools.py
```diff
@@ -172,7 +179,7 @@ def _require_mentions(
             raise BandToolError(
                 append_mention_handles_hint(
                     "At least one mention is required",
-                    available_mention_handles(self._participants),
+                    available_mention_handles(self._participants, self.agent_id),
```
AlexanderZ-Band, line 333: issue: `available_mention_handles(self._participants, self.agent_id)` (this line) is supposed to exclude the agent's own handle from the "available handles" hint, matching production (`runtime/tools/agent.py`). No test in `tests/testing/test_fake_tools.py` (the file this PR's own validation command names) constructs `FakeAgentTools` with a non-`None` `agent_id` — both existing no-mentions tests leave it at the default, so the exclusion wiring never actually runs with a real id. A regression in this specific line would go undetected by the whole repo (the one place it's incidentally exercised, `tests/mcp/test_shared.py::test_invoke_send_message_error_enriched_with_available_handles`, goes through a subclass that overrides `send_message` and never calls `_require_mentions`).

Add: `FakeAgentTools(participants=[{"id": "self", "handle": "@self"}, {"id": "user-1", "handle": "@alice"}], agent_id="self")`, call `send_message` with no mentions, and assert the raised `BandToolError`'s hint contains `@alice` but not `@self`.

.claude/skills/bug-hunting-via-example/scripts/runner.py
```diff
@@ -1018,8 +1017,9 @@ async def run_group(
     repo: Path,
     results: list[Result],
 ) -> None:
-    running = await start_group_examples(plan, resources, settings, repo, results)
+    running: dict[str, RunningExample] = {}
```
AlexanderZ-Band, line 1015: suggestion: moving `running: dict[str, RunningExample] = {}` inside `run_group`'s `try` (this line) fixes a real leak (a cancellation during `start_group_examples` used to skip the `finally: await stop_examples(...)` cleanup for already-spawned examples), but no test in `tests/skills/bughunting/test_runner.py` proves it — the closest test, `test_group_startup_failure_is_reported_per_example`, only simulates ordinary per-item `RuntimeError`s, which the per-item `except Exception` already caught correctly even before this fix.

Consider adding: a fake `start_example` that succeeds for the first spec then raises something the per-item `except Exception` won't catch (e.g. `asyncio.CancelledError`) for the second, and assert a fake `stop_example` still ran for the first spec's `RunningExample`.

src/band/adapters/opencode/adapter.py
```diff
@@ -625,37 +684,31 @@ async def _ensure_mcp_backend(self) -> BandMCPBackend:
...
+                result = await client.register_mcp_server(
+                    name=self._mcp_server_name, url=local_server.sse_url
+                )
+            except Exception:
+                logger.exception(
+                    "Failed to register MCP server %s with OpenCode",
+                    self._mcp_server_name,
+                )
+                return
+            if result.get(self._mcp_server_name, {}).get("status") != "connected":
```
AlexanderZ-Band: suggestion: the registration-success check hardcodes the literal `"connected"` as the sole definition of OpenCode's MCP registration-status vocabulary. The test double independently re-types the same literal (`tests/adapters/opencode/helpers.py:363`, `{name: {"status": "connected"}}`) with nothing tying the two together — a typo in either place would fail silently. This codebase already has precedent for wrapping OpenCode's own wire vocabulary in a `StrEnum` (`OpencodeToolStatus`).

Consider a small constant or `StrEnum` for this status vocabulary that both the adapter's check and the fake reference.

src/band/adapters/opencode/adapter.py
```diff
@@ -89,81 +91,100 @@ class RoomState:
...
-    def begin_turn(self, sender_id: str | None) -> None:
+
+@dataclass
+class RoomState:
+    room_id: str
+    session_id: str | None = None
+    tools: AgentToolsProtocol | None = None
+    turn: TurnState | None = None
+    turn_task: asyncio.Task[None] | None = None
+    # Bound in _get_or_create_room_state, immediately after construction.
```
AlexanderZ-Band: nit: this comment ("Bound in `_get_or_create_room_state`, immediately after construction.") no longer tells the whole story — `_begin_turn` (line 967) also reassigns `room_state.approvals` on every turn, not just once at construction. A reader trusting this comment would assume the field is stable for the room's lifetime.

src/band/adapters/opencode/approvals.py
```diff
@@ -506,6 +552,36 @@ def _fail_request(
         _cancel_timeout(self._questions.pop(request_id, None))
         self._release_if_idle()
         self._ports.fail_turn(message)
+        await self._ports.abort_session()
+
+    @asynccontextmanager
+    async def _permission_reply(
+        self, action: str, request_id: str
```
AlexanderZ-Band, line 694: nit: `_permission_reply` (557-570) and `_question_reply` (572-584) are near-identical `@asynccontextmanager`s — the `except Exception: await self._fail_request(...); raise ApprovalReplyError from error` tail is byte-identical in both; only the precondition (whether `session_id` is also required) and the yielded shape differ. Consider factoring the shared tail into one small helper both wrap, leaving only the differing precondition checks separate (a naive "nest one CM inside the other" would double-fire `_fail_request`, so the shared piece needs to be the tail only).

tests/adapters/opencode/test_lifecycle.py
```diff
@@ -99,16 +106,26 @@ async def test_new_turn_does_not_wipe_prior_turns_pending_usage(
...
-    room_state.usage_by_message["msg-1"] = TurnUsage(input_tokens=100, output_tokens=20)
+    first_turn = adapter._begin_turn(
+        room_state,
+        session_id="sess-1",
+        client=FakeOpencodeClient(),
+        tools=tools_protocol(tools),
+        sender_id="user-1",
+    )
+    first_turn.usage_by_message["msg-1"] = TurnUsage(input_tokens=100, output_tokens=20)
     # What on_message hands this turn's watch task.
```
AlexanderZ-Band: nit: this comment ("What on_message hands this turn's watch task.") is now orphaned — it used to sit directly above `first_turn_usage = room_state.usage_by_message`, explaining that snapshot line. This PR replaced that line (usage is now read straight off `first_turn`) but left the comment behind; it now explains nothing about the line below it (a `# The next turn begins...` comment already covers that point).

src/band/adapters/opencode/adapter.py
```diff
@@ -899,102 +954,115 @@ async def _ensure_session(
...
+    def _begin_turn(
+        self,
+        room_state: RoomState,
+        *,
+        session_id: str,
+        client: OpencodeClientProtocol,
+        tools: AgentToolsProtocol,
+        sender_id: str | None,
+    ) -> TurnState:
+        owner: list[TurnState] = []
```
AlexanderZ-Band: suggestion: the `owner: list[TurnState] = []` cell (populated via `owner.append(turn)` only after `_new_approvals`/`begin_turn` run) is unnecessary indirection. `_get_or_create_room_state` already builds the room's first `RoomApprovals` with the simpler `lambda: state.turn` (line 607) — no cell — and `RoomState.begin_turn` performs no `await` and invokes no approval port before assigning `self.turn`, so there's no window in which `lambda: room_state.turn` would observe anything different from what the owner-cell observes. `lambda: room_state.turn` is a drop-in replacement matching the established pattern, and is arguably more correct: once `_clear_turn_state` sets `room_state.turn = None`, it reports "no turn" immediately, whereas the cell keeps pointing at the finished turn until the next `_begin_turn`.

tests/adapters/opencode/test_approvals.py
```diff
@@ -107,11 +108,16 @@ def make_room_approvals(
             turn_mentions=lambda: [],
             release_turn_wait=release_turn_wait,
             fail_turn=fail_turn,
+            abort_session=abort_session or _ignore_abort,
```
AlexanderZ-Band: nit: `abort_session` is typed `Callable[[], Awaitable[None]] | None = None` and resolved via `abort_session or _ignore_abort` (this line) only because `_ignore_abort` is defined after `make_room_approvals`, so it can't be a literal default. Its two sibling ports in the same signature (`release_turn_wait`, `fail_turn`) use a real callable directly as the default. Moving `_ignore_abort`'s definition above `make_room_approvals` would let `abort_session: Callable[[], Awaitable[None]] = _ignore_abort` be a plain default, matching its neighbors and dropping the `Optional`/`or` indirection.

AlexanderZ-Band (commented):
Personal review pass (my-code-review, --fix). 8 issues (2 proven with passing/failing repro tests), 3 suggestions, 4 nits — see inline comments. (Submitted as COMMENT, not REQUEST_CHANGES: GitHub disallows requesting changes on one's own PR.)

src/band/adapters/opencode/adapter.py
AlexanderZ-Band: Fixed in 2e607842: wrapped the timeout branch's send_event/_emit_turn_usage calls in the same try/except Exception: logger.exception(...); await self._report_delivery_failure(...) pattern the success branch already used.

src/band/adapters/opencode/adapter.py
AlexanderZ-Band: Fixed in 2e607842: abandon() now returns whether it actually issued an abort, and the interrupt handler only makes its own explicit abort call when abandon() didn't already do it.

src/band/adapters/opencode/approvals.py
AlexanderZ-Band: Fixed in 2e607842, and more directly than the fix_note suggested: rather than carrying handled-id memory across the per-turn RoomApprovals rebuild, removed that bookkeeping entirely and broadened both guards to give "no longer pending" feedback for any named id that matches nothing currently pending (whether never asked, or already resolved) -- same outcome, simpler mechanism, no cross-turn state needed. Covered by two new tests: test_reply_naming_a_never_asked_id_is_consumed and test_reply_naming_an_already_resolved_id_is_consumed.

src/band/adapters/opencode/approvals.py
AlexanderZ-Band: Fixed in 2e607842: added test_malformed_question_with_no_questions_is_ignored.

src/band/adapters/opencode/approvals.py
AlexanderZ-Band: Fixed in 2e607842: added test_redelivered_permission_while_reply_in_flight_is_ignored and test_redelivered_permission_cancels_previous_timeout.

src/band/adapters/opencode/approvals.py
AlexanderZ-Band: Fixed in 2e607842: both test_cleanup_with_pending_permission and test_cleanup_with_pending_question now assert fake_client.aborted_sessions after on_cleanup.

tests/adapters/opencode/helpers.py
AlexanderZ-Band: Fixed in 2e607842: FakeOpencodeClient gained a register_mcp_statuses param (popped per call, defaulting to connected once exhausted) and a new test_mcp_registration_retries_until_connected exercises the non-connected-then-connected retry path.

src/band/testing/fake_tools.py
AlexanderZ-Band: Fixed in 2e607842: added test_rejection_excludes_the_agents_own_handle.

.claude/skills/bug-hunting-via-example/scripts/runner.py
AlexanderZ-Band: Fixed in 2e607842: added test_group_startup_cancellation_still_stops_already_started_examples (a CancelledError mid-startup, which the per-item except Exception doesn't catch, still reaches run_group's finally: stop_examples for the already-started example).

src/band/adapters/opencode/adapter.py
AlexanderZ-Band: Fixed in 2e607842: added MCP_REGISTRATION_CONNECTED_STATUS in adapter.py and imported it into the test double instead of re-typing the literal.

src/band/adapters/opencode/adapter.py
AlexanderZ-Band: Fixed in 2e607842: updated the comment to note _begin_turn's per-turn rebind.

src/band/adapters/opencode/approvals.py
AlexanderZ-Band: Fixed in 2e607842: factored the shared except/raise tail into a small _reply_guard context manager both _permission_reply/_question_reply wrap (only the yield differs now).

tests/adapters/opencode/test_lifecycle.py
AlexanderZ-Band: Fixed in 2e607842: removed the orphaned comment.

src/band/adapters/opencode/adapter.py
AlexanderZ-Band: Fixed in 2e607842: dropped the owner list-cell, _begin_turn now passes lambda: room_state.turn directly, matching _get_or_create_room_state's existing pattern.

tests/adapters/opencode/test_approvals.py
AlexanderZ-Band: Fixed in 2e607842: moved _ignore_abort above make_room_approvals and defaulted abort_session to it directly.

src/band/adapters/opencode/adapter.py
```diff
@@ -745,7 +828,10 @@ async def _handle_event(self, event: OpencodeEvent) -> None:
             case QuestionAskedEvent():
                 await room_state.approvals.on_question_asked(event.properties)
             case SessionErrorEvent():
-                room_state.last_error_message = describe_error(event.properties.error)
+                if room_state.turn is not None:
+                    room_state.turn.last_error_message = describe_error(
+                        event.properties.error
+                    )
                 self._finish_turn(room_state)
```
AlexanderZ-Band, line 861: issue: every other turn-terminating path in this file (interrupt/CancelledError -> abandon(), timeout -> the loop invariant in _await_turn guarantees no ask is parked, cleanup -> abandon()) abandons pending permissions/questions before releasing the turn. `SessionErrorEvent` is the one path that doesn't -- it only sets `last_error_message` and calls `_finish_turn`, leaving any permission/question still parked on a human untouched.

Proven with a repro against the real production path: park a permission on turn A (`room_state.approvals.on_permission_asked`), let turn A finish via a session error without cancelling approvals (matching this PR's own comment in `_begin_turn`: "e.g. a SessionErrorEvent finishes turn_future without cancelling approvals"), begin turn B (rebinding `room_state.approvals`), then send the human's real reply through `on_message` -> `room_state.approvals.try_handle_reply("approve perm-1", ...)`. It gets consumed and answered with `OpenCode approval \`perm-1\` is no longer pending.` -- a false claim, since `turn_a.approvals._permissions` still has it. Neither client ever receives a reply; the ask only ever resolves via its own expiry timer, not the human's actual answer.

AlexanderZ-Band (commented):
Personal review pass (my-code-review, --fix), against the full PR diff including the double-cancellation-wedge fix from this pass. One confirmed issue found and fixed (see inline). Everything else scanned clean under this pass's six angle-groups: turn-ownership pinning (owner-cell), room-scoped known-id tracking, the new _mcp_lifecycle_lock/_state_lock ordering in _shutdown_client/_register_mcp_backend, the _claim/_registry_for/_resolve_turn refactors, and the new double-cancellation regression test all verified correct against the real code.

src/band/adapters/opencode/adapter.py
AlexanderZ-Band: Fixed in ef541312: SessionErrorEvent's handler now calls `await room_state.turn.approvals.abandon()` before `_finish_turn`, matching every other terminal path. Regression test `test_session_error_abandons_a_pending_permission` (tests/adapters/opencode/test_turns.py) reproduces the bug against the real on_message/_handle_event path (fails without the fix: the reply is swallowed with a false "no longer pending" message and neither client gets a reply) and passes with it.

src/band/adapters/opencode/approvals.py
```diff
@@ -287,7 +324,9 @@ async def try_handle_reply(self, content: str, sender_id: str | None) -> bool:
         mentions = [{"id": sender_id}] if sender_id else []
 
         approval = parse_permission_reply(command)
-        if approval and self._permissions:
+        if approval and (
```
AlexanderZ-Band: issue: `try_handle_reply` enters the permission branch whenever any permission is pending (`self._permissions or named id in known_permission_ids`). `reject q-1` parses as `PermissionCommand(reject, 'q-1')`; `_resolve_permission('q-1')` is None, so the named-id fallback posts "OpenCode approval `q-1` is no longer pending" and returns True. `_resolve_question` / `_reject_question` never run, so a live question stays parked.

Take the permission branch only when the named id is a pending/known permission. Let `reject <question-id>` fall through to question handling when that id is a live question.

src/band/adapters/opencode/approvals.py
```diff
@@ -217,12 +243,18 @@ async def on_permission_asked(self, request: OpencodePermissionRequest) -> None:
             await self._approve_own_band_tool(request_id)
             return
 
+        existing = self._permissions.get(request_id)
```
AlexanderZ-Band: issue: `on_permission_asked` calls `_cancel_timeout(existing)` before the `existing.replying` guard (same shape at `approvals.py:285` for questions). When the in-flight reply is `_expire_permission` itself (timeout task already `_claim`ed, awaiting `client.reply_permission`), that cancel kills the timeout-reply. `CancelledError` is `BaseException`, so `_reply_guard`'s `except Exception` does not run `_fail_request`; `_forget` never runs; the pending stays `replying=True`. Later `approve req-1` `_claim`-fails and is consumed (`return True`) with no OpenCode POST and no room feedback; the turn never unparks.

Do not cancel the timeout while `existing.replying` is True. Also treat `CancelledError` on the reply path as a failed claim and `_forget`/unpark.

src/band/adapters/opencode/adapter.py
```diff
@@ -899,102 +993,124 @@ async def _ensure_session(
...
+        sender_id: str | None,
+    ) -> TurnState:
+        # Pinned to THIS turn specifically, not "whichever turn room_state.turn
+        # currently is": an ask left pending when a later turn begins (e.g. a
+        # SessionErrorEvent finishes turn_future without cancelling approvals)
+        # must keep acting on its own turn's client/session, never a
+        # successor's -- an ambient lookup would silently misroute a reply or
+        # abort onto the wrong turn.
+        owner: list[TurnState] = []
+        room_state.approvals = self._new_approvals(
```
AlexanderZ-Band: issue: `_begin_turn` installs a fresh `RoomApprovals` without `abandon()`/`cancel()` of the previous instance. `on_message` only `try_handle_reply`s `room_state.approvals` (the newest). `SessionIdleEvent` calls `_finish_turn` only (no abandon), so `turn_future` can be done while a permission is still parked; the busy-guard checks only `turn_future.done()`, so a second prompt starts turn B. `approve perm-1` then hits B's empty `RoomApprovals`, matches room-scoped `known_permission_ids`, and posts no-longer-pending without replying to OpenCode. A's orphaned ask stays parked. `test_orphaned_turns_pending_permission_replies_against_its_own_client` misses this because it calls `turn_a.approvals.try_handle_reply` directly.

Abandon/cancel the previous `RoomApprovals` in `_begin_turn` before replacing it, and/or treat `awaiting_human()` as busy in `on_message`. Named replies should reach any still-pending turn in the room, not only the ambient instance.

src/band/adapters/opencode/adapter.py
```diff
@@ -515,6 +566,7 @@ async def on_cleanup(self, room_id: str) -> None:
             should_shutdown = not self._rooms
 
         if room_state:
+            await room_state.approvals.abandon()
```
AlexanderZ-Band: issue: `on_cleanup` awaits `room_state.approvals.abandon()` (which awaits `abort_session` when an ask is pending) and only then calls `_clear_turn_state`. A `CancelledError` at that abort — agent shutdown cancelling cleanup — skips watcher cancellation and the `_shutdown_client` that follows. The room is already popped from `_rooms`, so the detached turn watcher keeps running and the shared OpenCode client/MCP backend are never stopped. The interrupt handler on this same change wraps the matching await in `try`/`finally` so `_clear_turn_state` still runs; cleanup does not.

Match the interrupt handler: `try: await abandon()` / `finally: _clear_turn_state(room_state)`. If this was the last room, keep `_shutdown_client` in an outer `finally` so cancellation during abandon still tears down the shared client.

src/band/adapters/opencode/adapter.py
```diff
@@ -625,64 +703,69 @@ async def _ensure_mcp_backend(self) -> BandMCPBackend:
...
+                return
+            status = result.get(self._mcp_server_name, {}).get("status")
+            if status != MCP_REGISTRATION_CONNECTED_STATUS:
+                logger.warning(
+                    "MCP server %s was not connected (status=%s)",
+                    self._mcp_server_name,
+                    status,
+                )
+                return
+            async with self._state_lock:
```
AlexanderZ-Band, line 746: issue: `_register_mcp_backend` logs on exception and on a non-`connected` status, then on success only assigns `_registered_client` under `_state_lock`. The previous `logger.info` for a successful register is gone. The retry-until-connected path therefore has no log line when it finally succeeds, so CI/production logs cannot tell a recovered registration from one that never connected.

After `status == MCP_REGISTRATION_CONNECTED_STATUS` (and/or after recording `_registered_client`), log at info with the MCP server name and status.

src/band/adapters/opencode/approvals.py
```diff
@@ -335,6 +384,17 @@ async def try_handle_reply(self, content: str, sender_id: str | None) -> bool:
                 )
             return True
 
+        if _is_question_rejection(command) and len(command.split()) > 1:
```
AlexanderZ-Band: issue: `try_handle_reply` now consumes a named `reject <id>` that matches no pending question (when another question is pending, or the id is in `_known_question_ids`) and posts `QUESTION_NO_LONGER_PENDING_TEMPLATE` instead of forwarding it as a model prompt. Permission stale/unknown ids have three tests (`test_reply_to_nonmatching_request_id_is_consumed`, `test_reply_naming_a_never_asked_id_is_not_consumed`, `test_reply_naming_an_already_resolved_id_is_consumed`); this question branch has none. Deleting the block, or routing `reject q-stale` into a fresh prompt, would not fail the suite — including `test_reject_by_id_targets_a_pending_question_not_a_stale_permission`, which only covers a still-pending question.

Add the question counterparts: named reject of a never-asked id is not consumed; named reject of an already-resolved (or other-pending) question id is consumed and the room gets the no-longer-pending line; assert `question_rejections` stays empty.

src/band/adapters/opencode/adapter.py
```diff
@@ -573,37 +625,67 @@ async def _get_or_create_room_state(self, room_id: str) -> RoomState:
...
+                tools=lambda: self._owner_tools(owner),
+                turn_mentions=lambda: self._owner_mentions(owner),
+                release_turn_wait=lambda: self._release_turn_wait_for_owner(owner),
+                fail_turn=lambda message: self._fail_turn_for_owner(owner, message),
+                abort_session=lambda: self._abort_owner_turn(
+                    owner, "approval abandoned"
+                ),
+                is_own_band_tool=self._is_own_band_tool,
+            ),
+            known_permission_ids=state.known_permission_ids,
```
AlexanderZ-Band, line 669: issue: `_new_approvals` threads `state.known_permission_ids` / `state.known_question_ids` into each replacement `RoomApprovals` so a later turn still treats an earlier request id as known. Every stale-id test builds an isolated `make_room_approvals()` and never calls `_begin_turn` / `on_message` after a resolve. Dropping those kwargs (each new instance would get a fresh `set()`) would still pass: `test_reply_naming_an_already_resolved_id_is_consumed` never leaves the original instance, and `test_orphaned_turns_pending_permission_replies_against_its_own_client` replies on `turn_a.approvals` rather than `room_state.approvals` (the object `on_message` actually consults after a rebind).

Drive it through the adapter: complete a manual permission (or question) via `on_message`, let the next turn `_begin_turn` (new `RoomApprovals`, same room sets), then send `approve <old-id>` / `reject <old-id>` on that room and assert the no-longer-pending room message and no extra `prompt_async`.

src/band/adapters/opencode/approvals.py
```diff
@@ -506,6 +587,39 @@ def _fail_request(
         _cancel_timeout(self._questions.pop(request_id, None))
         self._release_if_idle()
         self._ports.fail_turn(message)
+        await self._ports.abort_session()
```
AlexanderZ-Band, line 681: issue: `_fail_request` now awaits `self._ports.abort_session()` after `fail_turn`, so a failed OpenCode permission/question reply stops the session instead of leaving it parked. `test_failed_auto_replies_fail_only_the_affected_turn` still uses `make_room_approvals`'s default `_ignore_abort` and only asserts the `fail_turn` strings; `test_abandoning_a_request_stops_its_expiry_timer` (the client-gone path that reaches `_fail_request`) passes `abort_session=_ignore_abort` explicitly. Removing line 590 would not fail any test, so a failed auto-reply can still leave OpenCode waiting.

Pass a recording `abort_session` into those two `RoomApprovals` setups and assert it ran once per failed reply (and that `test_failed_auto_replies_fail_only_the_affected_turn` still isolates `fail_turn` to the affected request).

src/band/adapters/opencode/adapter.py
AlexanderZ-Band: issue: `_begin_turn` replaces `room_state.approvals` with a new `RoomApprovals`, which is what resets `_human_wait_total` so turn N's deliberation is not added to turn N+1's `_await_turn` deadline (`started + turn_timeout_s + approvals.human_wait_seconds`). `test_manual_approval_pauses_the_turn_timeout` and `test_approval_wait_does_not_shorten_the_resumed_turn` only cover that clock inside a single turn. Reusing the same `RoomApprovals` across turns (so banked wait leaks) would still pass, and a later turn with no human ask would inherit the previous turn's wait and not time out when it should.

Two sequential `on_message` turns: turn 1 parks on a permission long enough to bank wait, then completes; turn 2 has `turn_timeout_s` well below that banked wait and never goes idle. Assert turn 2 aborts on the compute budget (session aborted / timed-out event), not after banked-wait + timeout.

AlexanderZ-Band: issue: `_begin_turn` replaces `room_state.approvals` with a fresh `RoomApprovals` and does not `abandon()`/`cancel()` the previous instance. `on_message` only calls `try_handle_reply` on the current `room_state.approvals`, and a superseded watcher's `_clear_turn_state(expected_turn=...)` returns without touching the orphan. `_fail_request` still `fail_turn`s (resolves `turn_future`) after popping only the failed id, so any sibling permission/question stays parked with a live expiry timer. A later non-reply room message then passes the busy guard (`turn_future` is done), rebinds, and the leftover ask is no longer reachable from the room; `known_permission_ids` is room-scoped, so a later `approve <id>` is answered with "no longer pending" while the orphaned timer can still `reply_permission` / `_fail_request` / `abort_session` against the successor turn's session (same `session_id`). This is the same unreachability `test_session_error_abandons_a_pending_permission` calls a bug for `SessionErrorEvent`; that terminal path now abandons, but `_begin_turn` plus `_fail_request` leftover still creates it. Before this PR a room had one `RoomApprovals` for its lifetime, so a leftover ask stayed on the object `on_message` consults even if the watcher skipped clear.

Starting state: one OpenCode turn with two parked manual permissions. Triggering sequence: (1) user `approve perm-1`; the reply POST fails; `_fail_request` pops only perm-1, leaves perm-2 parked with its expiry timer, `fail_turn`s, then awaits `abort_session`. (2) That `on_message` returns without waiting for the watcher. (3) A following non-reply room message passes the busy guard and `_begin_turn` installs a new `RoomApprovals`. (4) The watcher's `_clear_turn_state(expected_turn=turn_a)` no-ops because `room_state.turn` is now turn B. Incorrect outcome: `approve perm-2` hits turn B's empty pending map, matches the shared `known_permission_ids`, and the room is told the ask is no longer pending; perm-2's timer remains on the orphaned instance and can still reply to or abort turn B's session.

src/band/adapters/opencode/approvals.py
```diff
@@ -392,81 +452,102 @@ def cancel(self) -> None:
         # No ask is parked anymore -- release any watcher waiting on us.
         self._release_from_human()
 
+    async def abandon(self) -> bool:
```
AlexanderZ-Band, line 533: nit: `abandon()` discards pending asks and may abort the OpenCode session with no log. Interrupt, `SessionErrorEvent`, and `on_cleanup` all go through it, while `session.idle` and the timeout path emit turn-phase breadcrumbs. A stuck-turn diagnosis from CI logs cannot see that a turn ended by abandon/abort rather than idle or timeout.

Log at info when `was_pending` is true (room id from ports, and that the session is being aborted). Callers can stay silent.

src/band/adapters/opencode/approvals.py
```diff
@@ -247,14 +279,19 @@ async def on_permission_asked(self, request: OpencodePermissionRequest) -> None:
 
     async def on_question_asked(self, request: OpencodeQuestionRequest) -> None:
         request_id = request.id
-        if not request_id:
+        if not request_id or not request.questions:
```
AlexanderZ-Band, line 289: nit: `on_question_asked` returns immediately when `request.questions` is empty, with no log. A live OpenCode sending a malformed `question.asked` is indistinguishable from the event never arriving, which is the failure this branch was added to absorb.

Log at warning or debug with the request id before returning.

src/band/adapters/opencode/adapter.py
```diff
@@ -899,102 +993,124 @@ async def _ensure_session(
...
+        room_state: RoomState,
+        *,
+        session_id: str,
+        client: OpencodeClientProtocol,
+        tools: AgentToolsProtocol,
+        sender_id: str | None,
+    ) -> TurnState:
+        # Pinned to THIS turn specifically, not "whichever turn room_state.turn
+        # currently is": an ask left pending when a later turn begins (e.g. a
+        # SessionErrorEvent finishes turn_future without cancelling approvals)
```
AlexanderZ-Band: nit: `_begin_turn`'s pin comment cites `SessionErrorEvent` as the case that finishes `turn_future` without cancelling approvals. The same PR now calls `abandon()` on that path before `_finish_turn`, so the example is false. A reader will treat session-error as a still-open hole and miss whatever remaining reason the pin actually has.

Drop the `SessionErrorEvent` example. Keep the why (an ask that outlives `_begin_turn` must still address its own client/session), without naming a path this PR already closed.

src/band/adapters/opencode/approvals.py
AlexanderZ-Band: nit: `on_permission_asked` and `on_question_asked` both newly implement the redelivered-ask rule as the same four-line shape: lookup by id, `_cancel_timeout(existing)`, return if `existing.replying`, then insert a replacement (also `src/band/adapters/opencode/approvals.py:285`). The in-flight guard and timer cancel are one comparison; a later change has to be edited in both methods or the two ask types silently diverge.

Extract one helper, e.g. `_supersede_ask(registry, request_id) -> bool`, that cancels the previous timer and returns False when a reply is in flight. Both ask handlers call it before constructing the replacement.

src/band/adapters/opencode/adapter.py
```diff
@@ -573,37 +625,67 @@ async def _get_or_create_room_state(self, room_id: str) -> RoomState:
...
+                    owner, "approval abandoned"
+                ),
+                is_own_band_tool=self._is_own_band_tool,
+            ),
+            known_permission_ids=state.known_permission_ids,
+            known_question_ids=state.known_question_ids,
+        )
+
+    @staticmethod
+    def _owner_session_id(owner: TurnOwner) -> str | None:
```
AlexanderZ-Band: nit: `_new_approvals` reaches the TurnOwner through four static methods (`_owner_session_id` here, `_owner_tools` at 658, `_owner_client` at 663, `_owner_mentions` at 668) that are the same `turn = owner(); return turn.<attr> if turn else <default>` shape. There is no shared helper, so a fifth port or a change to the no-turn default has to copy the pattern again.

Replace the four methods with one helper, e.g. `_if_owner(owner, pick, default)`, and pass `lambda t: t.session_id` (and the other attributes) at the ApprovalPorts construction site.

src/band/adapters/opencode/adapter.py
```diff
@@ -899,102 +993,124 @@ async def _ensure_session(
...
+        tools: AgentToolsProtocol,
+        sender_id: str | None,
+    ) -> TurnState:
+        # Pinned to THIS turn specifically, not "whichever turn room_state.turn
+        # currently is": an ask left pending when a later turn begins (e.g. a
+        # SessionErrorEvent finishes turn_future without cancelling approvals)
+        # must keep acting on its own turn's client/session, never a
+        # successor's -- an ambient lookup would silently misroute a reply or
+        # abort onto the wrong turn.
+        owner: list[TurnState] = []
```
AlexanderZ-Band: suggestion: `_begin_turn` allocates a one-element list as a late-binding cell (`owner: list[TurnState] = []`, then `lambda: owner[0] if owner else None`, then `owner.append(turn)`) so ApprovalPorts can pin this turn before the TurnState exists. That cell (and the comment forbidding `lambda: room_state.turn`) exists only because `TurnState.approvals` is a required constructor argument, while `RoomState.approvals` in this same file already uses `field(init=False)` for the identical construction-order problem. Making `TurnState.approvals` `init=False` too would let `_begin_turn` construct the turn first and pass `lambda: turn`, which stays pinned to this instance after a later `_begin_turn` replaces `room_state.turn` — without the list cell.

Give `TurnState.approvals` `field(init=False)` like `RoomState.approvals`; construct the TurnState first, bind `_new_approvals(..., lambda: turn)`, then assign `room_state.turn` / `room_state.approvals`. Do not switch the owner to ambient `lambda: room_state.turn`.

tests/adapters/opencode/test_approvals.py
```diff
@@ -293,6 +328,124 @@ async def test_mention_only_question_reply_requests_a_real_answer() -> None:
...
+    """A permission redelivered with the SAME id (no reply in flight) must
+    cancel the FIRST ask's expiry timer, not leave it running to later
+    auto-reject the replacement."""
+    client = FakeOpencodeClient()
+    approvals = make_room_approvals(cast(OpencodeClientProtocol, client))
+
+    await approvals.on_permission_asked(
+        OpencodePermissionRequest(id="req-1", permission="bash")
+    )
+    first_timer = approvals._permissions["req-1"].timeout_task
```
AlexanderZ-Band: nit: `test_redelivered_permission_cancels_previous_timeout` reads `approvals._permissions["req-1"].timeout_task`, waits until that Task is `.done()`, and asserts the replacement's Task is a different object that is not done. That restates the timer swap; it does not observe the behavior in the docstring — that the first expiry must not later auto-reject the replacement. With the default 300s `approval_wait_timeout_s`, a first timer that was not cancelled would not fire during this test either; `.done()` is also true if the expiry actually ran. There is no question-id counterpart.

Use a short `approval_wait_timeout_s`, redeliver the same id, wait past the first timeout, then approve the replacement and assert no reject was sent and the approve landed. Mirror it for questions.

AlexanderZ-Band (commented):
Cycle 1 of a my-code-review pass against INT-1132. Three proven logic bugs in the new named-id / redelivery / per-turn approvals wiring, plus the matching test-coverage holes and a cleanup-path cancel that can skip teardown.

src/band/adapters/opencode/adapter.py:1267 — nit: `_emit_turn_usage` still documents a `room_state.usage_by_message` dict that a later turn may replace. This PR moved usage onto `TurnState`, deleted that field, and the method now takes `turn` and reads `turn.usage_by_message`. The sentence describes a capture-the-dict workaround that no longer exists. Delete the `room_state.usage_by_message` paragraph; keep only that summing `turn.usage_by_message` is a no-op when usage emit is off or nothing was captured.

src/band/adapters/opencode/approvals.py
AlexanderZ-Band: Fixed in 7a825ead: `try_handle_reply` takes the permission branch only when the named id is a pending/known permission (or the command is unnamed while one is pending). `reject q-1` while `req-1` is parked now falls through to `_reject_question`. Covered by `test_reject_question_id_is_not_stolen_by_a_pending_permission`.

src/band/adapters/opencode/approvals.py
AlexanderZ-Band: Fixed in 7a825ead: `_supersede_ask` checks `existing.replying` before `_cancel_timeout`, so a same-id redelivery cannot kill an in-flight `_expire_permission` claim. Covered by `test_redelivered_permission_does_not_cancel_in_flight_expiry`.

src/band/adapters/opencode/adapter.py
AlexanderZ-Band: Fixed in 7a825ead: `SessionIdleEvent` now `abandon()`s any still-parked ask before `_finish_turn` (same as the other terminal paths), and `on_message`'s busy-guard also treats `awaiting_human()` as busy so a leftover cannot be rebound out from under the room. Covered by `test_session_idle_abandons_a_pending_permission`.

src/band/adapters/opencode/adapter.py
AlexanderZ-Band: Fixed in 7a825ead: `on_cleanup` wraps `abandon()` in `try`/`finally` so `_clear_turn_state` still runs, and keeps `_shutdown_client` in an outer `finally` so a cancel during abort still tears down the shared client. Covered by `test_cleanup_cancel_during_abandon_still_tears_down`.

src/band/adapters/opencode/adapter.py
AlexanderZ-Band: Fixed in 7a825ead: successful MCP registration now logs at info with the server name and status.

src/band/adapters/opencode/approvals.py
AlexanderZ-Band: Fixed in 7a825ead: added `test_named_reject_of_a_never_asked_question_is_not_consumed` and `test_named_reject_of_an_already_resolved_question_is_consumed`.

src/band/adapters/opencode/adapter.py
AlexanderZ-Band: Fixed in 7a825ead: `test_stale_approval_id_survives_a_later_turn` drives approve → next `_begin_turn` → `approve <old-id>` through `on_message` and asserts the no-longer-pending room message with no extra `prompt_async`.

src/band/adapters/opencode/approvals.py
AlexanderZ-Band: Fixed in 7a825ead: `test_failed_auto_replies_fail_only_the_affected_turn` and `test_abandoning_a_request_stops_its_expiry_timer` now record `abort_session`; `_fail_request` also `cancel()`s leftover siblings after aborting the session. Covered by `test_failed_reply_cancels_sibling_asks`.

src/band/adapters/opencode/adapter.py
AlexanderZ-Band: Fixed in 7a825ead: `test_new_turn_does_not_inherit_previous_human_wait` asserts the successor `RoomApprovals.human_wait_seconds == 0` after a turn that banked deliberation.

AlexanderZ-Band: Fixed in 7a825ead: `_fail_request` now `cancel()`s leftover siblings after aborting the session (they can never be answered). Covered by `test_failed_reply_cancels_sibling_asks`. Combined with session.idle abandon and the `awaiting_human()` busy-guard so a later `_begin_turn` cannot orphan a live ask.

src/band/adapters/opencode/approvals.py
AlexanderZ-Band: Fixed in 7a825ead: `abandon()` logs at info with the room id when it actually aborts.

src/band/adapters/opencode/approvals.py
AlexanderZ-Band: Fixed in 7a825ead: empty `question.asked` now logs a warning with the request id and room before returning.

src/band/adapters/opencode/adapter.py
AlexanderZ-Band: Fixed in 7a825ead: dropped the SessionErrorEvent example from the pin comment; it now states the remaining why (an ask that outlives this `_begin_turn` must keep its own client/session).

src/band/adapters/opencode/approvals.py
AlexanderZ-Band: Fixed in 7a825ead: the shared redelivery rule is now `_supersede_ask`, used by both `on_permission_asked` and `on_question_asked`.

src/band/adapters/opencode/adapter.py
AlexanderZ-Band: Fixed in 7a825ead: the four `_owner_*` getters are one `_if_owner` helper.

src/band/adapters/opencode/adapter.py
AlexanderZ-Band: Fixed in 7a825ead: `TurnState.approvals` is `init=False`; `_begin_turn` constructs the turn first and binds `_new_approvals(..., lambda: turn)` so the owner stays pinned without a list cell.

tests/adapters/opencode/test_approvals.py
AlexanderZ-Band: Fixed in 7a825ead: the test now waits past the first timer's deadline (and not the replacement's), asserts no auto-reject, then approves the replacement. Mirrored for questions.

src/band/adapters/opencode/approvals.py
```diff
@@ -335,8 +389,39 @@ async def try_handle_reply(self, content: str, sender_id: str | None) -> bool:
...
+                # feedback, not a fresh prompt for the model.
+                await self._notify_room(
+                    QUESTION_NO_LONGER_PENDING_TEMPLATE.format(request_id=request_id),
+                    mentions,
+                )
+                return True
+
         return False
 
+    def _permission_command_applies(self, approval: PermissionCommand) -> bool:
```
AlexanderZ-Band, line 434: issue: `_permission_command_applies` falls through for every permission verb whose named id is a live/known question. `reject q-1` is the shared grammar that must reach `_reject_question`; `approve q-1` / `always q-1` are not. With perm `req-1` and question `q-1` both pending, those commands now submit free-text answers (`[["approve q-1"]]`) and leave the permission parked.

Only return False for the `reject` verb. Keep `approve`/`always <question-id>` on the permission branch (unknown-id / no-longer-pending).

src/band/adapters/opencode/adapter.py
```diff
@@ -745,14 +832,30 @@ async def _handle_event(self, event: OpencodeEvent) -> None:
...
                     "OpenCode turn: session.idle room=%s session=%s",
                     room_state.room_id,
                     event.session_id,
                 )
+                # Idle is a terminal path: a permission/question still parked
+                # on a human can never be answered on this turn once the
+                # session has gone idle, and a later _begin_turn would make
+                # it unreachable from on_message.
+                if room_state.turn is not None:
+                    await room_state.turn.approvals.abandon()
```
AlexanderZ-Band, line 873: issue: `SessionIdleEvent` now always `abandon()`s if a turn exists. `abandon()` treats any dict entry as parked (`was_pending = bool(self._permissions or self._questions)`), but an in-flight human reply stays in that dict with `replying=True` until `_forget`. Idle during `await client.reply_permission` therefore aborts a session the human already claimed; `cancel()` does not stop the in-flight POST, which still goes out afterward and the room is told the approval was handled.

Abandon on idle only when an ask is actually waiting on a human (`not pending.replying`). In-flight replies should just `_finish_turn`.

AlexanderZ-Band (commented):
Focused Logical Bugs pass after the cycle-1 fixes. Two new defects the routing/idle-abandon changes introduced; H1–H3 still hold.

src/band/adapters/opencode/approvals.py
AlexanderZ-Band: Fixed in 616db51c: `_permission_command_applies` now falls through only for `reject`. `approve`/`always <question-id>` stay on the permission branch (no-longer-pending). Covered by `test_approve_question_id_does_not_answer_the_question`.

src/band/adapters/opencode/adapter.py
AlexanderZ-Band: Fixed in 616db51c: `abandon()` aborts only when an unanswered ask remains (`not pending.replying`). Idle during an in-flight human reply finishes the turn without aborting. Covered by `test_abandon_does_not_abort_an_in_flight_reply` and `test_session_idle_does_not_abort_an_in_flight_approval`.

src/band/adapters/opencode/approvals.py
```diff
@@ -335,8 +389,41 @@ async def try_handle_reply(self, content: str, sender_id: str | None) -> bool:
...
```
AlexanderZ-Band, line 434: issue: `_permission_command_applies` only routes `approve`/`always` when `_permissions` or `_known_permission_ids` is populated. With **only** a question pending, `approve q-1` / bare `approve` fall through to `_resolve_question` and are POSTed as literal answers. Repro: `/private/tmp/pr-645-skeptic-repros/h1_h2_h3_h4_h10_h11.py` (H1/H2 CONFIRMED).

src/band/adapters/opencode/approvals.py
```diff
@@ -287,7 +331,7 @@ async def try_handle_reply(self, content: str, sender_id: str | None) -> bool:
         mentions = [{"id": sender_id}] if sender_id else []
 
         approval = parse_permission_reply(command)
```
AlexanderZ-Band, line 337: issue: with one permission and one question pending, bare `reject` always rejects the permission only; the question stays parked. Repro: h1 script (H3 CONFIRMED).

src/band/adapters/opencode/approvals.py
```diff
@@ -335,8 +389,41 @@ async def try_handle_reply(self, content: str, sender_id: str | None) -> bool:
...
+        """Whether a parsed approve/always/reject targets a permission.
+
+        ``reject <id>`` is also the question-rejection grammar. A named id
+        that is a live or known question (and not a permission) must fall
+        through so ``_resolve_question`` / ``_reject_question`` can run.
+        ``approve`` / ``always`` are not shared; those stay on this branch
+        so they are not submitted as free-text question answers.
+        """
+        named = approval.request_id
+        if (
```
AlexanderZ-Band, line 444: issue: when the same id exists in both `_permissions` and `_questions`, `reject <id>` never reaches question rejection. Repro: h1 script (H4 CONFIRMED).

src/band/adapters/opencode/approvals.py
```diff
@@ -392,81 +479,112 @@ def cancel(self) -> None:
...
+        """Stop a parked session after local approval state is discarded.
+
+        Returns whether an ask was actually pending (and so the session was
+        aborted), so a caller with its own unconditional abort afterward
+        (on_message's interrupt handler) can skip a redundant one.
+        """
+        # An in-flight reply stays in the dict with ``replying=True`` until
+        # ``_forget``; aborting that session would cancel work the human
+        # already claimed. Only unanswered asks need the session stopped.
+        was_pending = any(
```
AlexanderZ-Band: issue: `abandon()` aborts when any ask has `replying=False` even if another ask has an in-flight HTTP reply (`replying=True`). Repro: h3_h4_h5_h7.py (H5 CONFIRMED). Same root on `SessionErrorEvent` at adapter.py:845.

src/band/adapters/opencode/adapter.py
```diff
@@ -385,7 +426,10 @@ async def on_message(
         if await room_state.approvals.try_handle_reply(msg.content, msg.sender_id):
             return
 
-        if room_state.turn_future and not room_state.turn_future.done():
+        if room_state.turn and (
```
AlexanderZ-Band, line 436: issue: busy guard allows a new turn while the prior watcher is still delivering fallback text (H7 CONFIRMED in h3_h4_h5_h7.py).

src/band/adapters/opencode/adapter.py
```diff
@@ -745,14 +832,30 @@ async def _handle_event(self, event: OpencodeEvent) -> None:
...
                 logger.info(
                     "OpenCode turn: session.idle room=%s session=%s",
                     room_state.room_id,
                     event.session_id,
                 )
+                # Idle is a terminal path: a permission/question still parked
+                # on a human can never be answered on this turn once the
+                # session has gone idle, and a later _begin_turn would make
+                # it unreachable from on_message.
+                if room_state.turn is not None:
```
AlexanderZ-Band, line 872: issue: `SessionIdleEvent` abandons/finishes the current turn with no stale-idle guard; duplicate idle after `_begin_turn` hits the successor turn (H8 CONFIRMED).

src/band/adapters/opencode/adapter.py
```diff
@@ -1017,7 +1139,7 @@ def deadline() -> float:
         while True:
             try:
                 await asyncio.wait_for(
-                    asyncio.shield(turn_future), max(deadline() - loop.time(), 0.0)
+                    asyncio.shield(turn.turn_future), max(deadline() - loop.time(), 0.0)
                 )
                 return
             except asyncio.TimeoutError:
```
AlexanderZ-Band: issue: hung permission/question HTTP reply leaves `_await_turn` blocked in `wait_until_idle()` forever (H9 CONFIRMED).

tests/adapters/opencode/test_approvals.py
```diff
@@ -228,20 +255,49 @@ async def test_polite_permission_reply_uses_pending_request() -> None:
...
     client = FakeOpencodeClient()
     approvals = make_room_approvals(cast(OpencodeClientProtocol, client))
 
     await approvals.on_permission_asked(
         OpencodePermissionRequest(id="req-current", permission="bash")
     )
 
-    assert not await approvals.try_handle_reply("approve req-stale", "user-1")
+    assert await approvals.try_handle_reply("approve req-stale", "user-1")
+    assert client.permission_replies == []
```
AlexanderZ-Band, line 278: issue: `test_reply_to_nonmatching_request_id_is_consumed` does not assert room feedback (`no longer pending`) unlike sibling tests.

tests/adapters/opencode/test_turns.py
```diff
@@ -374,6 +375,310 @@ async def test_session_error_emits_error_event(make_adapter, tools) -> None:
     assert "boom" in error_events[0]["content"].lower()
 
 
+async def test_session_error_abandons_a_pending_permission(make_adapter, tools) -> None:
```
AlexanderZ-Band, line 402: issue: terminal-path abandon tests cover permissions only, not parked questions on session.error/idle.

AlexanderZ-Band (commented):
Cycle 1: confirmed approval-routing and turn-lifecycle defects (see inline).

src/band/adapters/opencode/approvals.py
AlexanderZ-Band: Fixed in 231c99b8: permission grammar for once/always with questions pending; question-only approve uses hint not free-text answers. Tests: test_approve_question_id_with_only_a_question_pending, test_bare_approve_with_only_a_question_pending.

src/band/adapters/opencode/approvals.py
AlexanderZ-Band: Fixed in 231c99b8: bare reject with both pending types posts dual disambiguation hint. Test: test_bare_reject_with_both_permission_and_question_pending.

src/band/adapters/opencode/approvals.py
AlexanderZ-Band: Fixed in 231c99b8: reject id prefers question rejection when id is live in _questions. Test: test_reject_shared_id_rejects_the_question.

src/band/adapters/opencode/approvals.py
AlexanderZ-Band: Fixed in 231c99b8: abandon skips abort_session when any ask is replying. Test: test_abandon_does_not_abort_while_another_reply_is_in_flight.

src/band/adapters/opencode/adapter.py
AlexanderZ-Band: Declined in 231c99b8: turn_task busy guard blocked legitimate back-to-back turns; overlap during slow fallback remains without watcher reorder.

src/band/adapters/opencode/adapter.py
AlexanderZ-Band: Declined in 231c99b8: session.idle has no prompt correlation in the wire event; stale idle indistinguishable from idle-while-parked without protocol metadata.

src/band/adapters/opencode/adapter.py
AlexanderZ-Band: Fixed in 231c99b8: cancelled expiry timers cleared; claim releases human-wait gate so _await_turn is not stuck during in-flight HTTP replies.

tests/adapters/opencode/test_approvals.py
AlexanderZ-Band: Fixed in 231c99b8: test_reply_to_nonmatching_request_id_is_consumed now asserts room feedback.

tests/adapters/opencode/test_turns.py
AlexanderZ-Band: Fixed in 231c99b8: added session.error/session.idle question abandon tests.

src/band/adapters/opencode/approvals.py
```diff
@@ -391,81 +530,113 @@ def cancel(self) -> None:
...
+        was_pending = any(not pending.replying for pending in pending_entries)
+        self.cancel()
+        if was_pending and not in_flight:
+            logger.info(
+                "OpenCode turn: abandon pending approvals room=%s",
+                self._ports.room_id,
             )
-            return
+            await self._ports.abort_session()
+        return was_pending
```
amit-gazal-band: **(1, high) The interrupt path can leave the session running.**

`abandon()` returns `was_pending` (there is an unanswered ask), but it only calls `abort_session` when `was_pending and not in_flight`. The interrupt handler in `adapter.py:551-555` treats the return value as "already aborted" (which is what this docstring promises) and skips `_abort_turn`.

Scenario: reply to `p1` is in flight (`replying=True`) and `p2` or a question is still waiting on the human. The user interrupts. `abandon()` skips the abort because of the in-flight reply and returns `True`, so the handler skips `_abort_turn` too. Nothing aborts the session: the turn state is cleared locally, and OpenCode keeps running or keeps waiting on an ask the room can no longer answer.

The interrupt path is also inconsistent: with *only* an in-flight reply, `abandon()` returns `False` and the handler aborts anyway. So on interrupt, the rule against aborting a claimed reply is already ignored, except in this mixed case, where nothing aborts at all.

Suggested fix: return whether the abort actually happened, and update `test_abandon_does_not_abort_while_another_reply_is_in_flight` (which currently asserts `abandon() is True` with no abort).

src/band/adapters/opencode/approvals.py
```diff
@@ -334,8 +414,51 @@ async def try_handle_reply(self, content: str, sender_id: str | None) -> bool:
...
+        so they are not submitted as free-text question answers.
+        """
+        named = approval.request_id
+        if (
+            named is not None
+            and approval.reply == "reject"
+            and named in self._questions
+        ):
+            return False
+        if approval.reply in ("once", "always"):
```
amit-gazal-band: **(2, low, UX) Question answers that start with `approve`/`always` get swallowed, with a misleading reply.**

I understand keeping permission words out of question answers is deliberate (see the docstring and the `test_approve_question_id_*` / `test_bare_approve_*` tests). The cost is that a real free-text answer like "Always use spaces" is claimed here, `parse_permission_reply` reads `use` as the request id, and the room gets "OpenCode approval `use` is no longer pending." The question is never answered, and the message doesn't tell the user why.

Narrow option: treat approve/always as a command only when the trailing token is empty, `please`, or a known permission/question id. Otherwise let it fall through to `_resolve_question` as free text. At minimum, reply with the question-command hint here instead of "no longer pending".

src/band/adapters/opencode/approvals.py
```diff
@@ -246,14 +286,23 @@ async def on_permission_asked(self, request: OpencodePermissionRequest) -> None:
 
     async def on_question_asked(self, request: OpencodeQuestionRequest) -> None:
         request_id = request.id
-        if not request_id:
+        if not request_id or not request.questions:
```
amit-gazal-band: **(3, low) A `question.asked` with an id but an empty `questions` list is now dropped.**

On `main`, only a missing id returned early. An empty list was still registered, so `auto_reject` rejected it immediately and otherwise the room could reject it. Here it is dropped with just a warning. OpenCode stays blocked on that request id with no reply/reject and no expiry timer, so the turn hangs until `turn_timeout_s` (300s by default) and is then aborted as a timeout.

Probably rare in practice, but this is easy to fix: return early only for a missing id, and reject the question when the id is present but `questions` is empty. The warning text ("no questions") is also misleading when the real problem is a missing `request_id`.

amit-gazal-band (dismissed):
Approving. Three comments inline: one real bug in the interrupt path (1), and two lower-priority items (2, 3).

src/band/adapters/opencode/approvals.py
AlexanderZ-Band: Fixed in a330620: abandon now reports whether it actually aborted and aborts whenever an unanswered ask remains, including alongside an in-flight reply.

src/band/adapters/opencode/approvals.py
AlexanderZ-Band: Fixed in a330620: an identified question with an empty questions list is registered only long enough to send OpenCode an explicit rejection, so it cannot leave the turn blocked.

src/band/adapters/opencode/approvals.py
AlexanderZ-Band: Fixed in a330620: approve/always only remain control commands for known request IDs; otherwise they fall through as natural-language question answers.

amit-gazal-band (approved):

PR #646, opened 2026-09-20
feat(integrations): expose Claude response effort

AlexanderZ-Band:
## Summary
- expose Claude Agent SDK response effort on ClaudeSDKAdapter
- forward the typed value to ClaudeAgentOptions
- cover the constructor contract and SDK option forwarding

## Validation
- 124 Claude adapter tests
- 683 framework conformance tests
- Ruff

PR #647, opened 2026-09-20
feat(integrations): expose Claude response effort

AlexanderZ-Band:
## Summary
- expose Claude Agent SDK response effort on ClaudeSDKAdapter
- forward the typed value to ClaudeAgentOptions
- cover the constructor contract and SDK option forwarding
- source the effort literal from `claude_agent_sdk`'s own `EffortLevel` type instead of a hand-typed duplicate, and bump the `claude_sdk` extra's floor pin to the version that exports it (0.2.82)

## Validation
- 124 Claude adapter tests
- 683 framework conformance tests
- Ruff, pyrefly

## Note
Replaces #646, which GitHub auto-closed after a branch rename broke its head-ref link (no commits lost — same two commits, plus the `EffortLevel` follow-up fix).

Linear: INT-1506

https://claude.ai/code/session_01E3bZSx75vgPyczVTTudNzm


pyproject.toml
```diff
@@ -85,7 +85,7 @@ langgraph = [
     "beautifulsoup4>=4.12.0",
 ]
 claude_sdk = [
-    "claude-agent-sdk>=0.1.81",  # Floor: ResultMessage.errors/api_error_status (used by the failure path). Requires Node.js 20+ and @anthropic-ai/claude-code CLI
+    "claude-agent-sdk>=0.2.82",
```
AlexanderZ-Band: issue: This line's `# Floor: ResultMessage.errors/api_error_status (used by the failure path). Requires Node.js 20+ and @anthropic-ai/claude-code CLI` comment was dropped entirely when the floor bumped from `>=0.1.81` to `>=0.2.82`. That loses two things: the original floor's rationale, and the standalone Node.js 20+ / `@anthropic-ai/claude-code` CLI requirement note (unrelated to the floor bump, and not documented anywhere else). No new `# Floor: ...` comment was added to explain the new pin either, even though this repo's own convention (e.g. the `mcp~=1.28.1` floor comments elsewhere in this file and in packages/band-mcp/pyproject.toml) documents every floor pin's rationale, and INT-1506's own acceptance criteria explicitly ask for 'pyproject.toml floor pin bumped with a `# Floor: ...` comment matching the existing convention.'

Verified: `claude-agent-sdk` 0.1.81 has no `EffortLevel` name at all (the `effort` field on `ClaudeAgentOptions` was hand-typed inline as `Literal["low","medium","high","xhigh","max"] | None`); 0.2.82 (PyPI's very next release after 0.1.81) is exactly where `EffortLevel` first appears as an importable `TypeAlias`, exported from both `claude_agent_sdk.types` and the package's top-level `__init__.py`. So the floor bump itself is correct — only the comment documenting it (and the unrelated Node.js/CLI note it silently carried) is missing.

Fix: restore a `# Floor: ...` comment on this line documenting why 0.2.82 is now required (EffortLevel exported as a named type), and re-attach the Node.js 20+ / @anthropic-ai/claude-code CLI requirement note that was dropped.

tests/adapters/test_claude_sdk_adapter.py
```diff
@@ -326,6 +326,22 @@ async def test_fallback_model_is_forwarded(self):
             assert sdk_options.model == "opus"
             assert sdk_options.fallback_model == "sonnet"
 
+    @pytest.mark.asyncio
+    async def test_effort_is_forwarded(self):
```
AlexanderZ-Band, line 353: nit: `test_effort_is_forwarded` has no docstring, unlike its immediate siblings in this class — `test_explicit_model_is_forwarded` ("Explicit model= should land in ClaudeAgentOptions.model.") and `test_fallback_model_is_forwarded` (a one-line regression-guard explanation). Every other on_started-forwarding test in this file states what it verifies.

Fix: add a one-line docstring, e.g. "effort= should land in ClaudeAgentOptions.effort.", for consistency with the class's documented-intent convention.

AlexanderZ-Band (commented):
Automated personal-checklist review (my-code-review, pass 1). Six angle-groups checked; this is a small, low-risk passthrough diff (a single optional constructor field forwarded once at adapter startup). Two non-blocking findings.

pyproject.toml
AlexanderZ-Band: Fixed: restored a `# Floor:` comment on this line documenting the new rationale (EffortLevel exported as a named type starting in 0.2.82) and re-attaching the Node.js 20+ / @anthropic-ai/claude-code CLI requirement note that was dropped.

tests/adapters/test_claude_sdk_adapter.py
AlexanderZ-Band: Fixed: added a one-line docstring ("effort= should land in ClaudeAgentOptions.effort.") matching the sibling tests' convention.

amit-gazal-band (approved):

PR #649, opened 2026-09-21
feat: Add Cursor ACP support

AlexanderZ-Band:
## Summary
- add the public Cursor ACP adapter backed by `agent acp`
- support dynamic provider-defined session configuration plus manual or automatic Cursor decisions
- wire Cursor into documentation, the baseline matrix, and backend-lane CI

## Validation
- `uv run ruff check .`
- `uv run pyrefly check`
- `uv run pytest tests/integrations/acp/ tests/adapters/test_cursor_acp_adapter.py tests/baseline/test_registry.py -q`
- `uv run pytest tests/framework_conformance/test_adapter_conformance.py tests/framework_conformance/test_config_drift.py tests/framework_conformance/test_files_image_passthrough_matrix.py -q`
- `uv run pytest tests/e2e/baseline/ -k cursor_acp --collect-only -q`

Stacked on #644.


src/band/integrations/acp/client_profiles.py
```diff
@@ -49,62 +52,111 @@ async def ext_notification(
...
-                )
-                return {"outcome": {"type": "selected", "optionId": option_id}}
-            return {"outcome": {"type": "cancelled"}}
-
-        if method == "cursor/create_plan":
-            return {"outcome": {"type": "approved"}}
-
-        return {}
+        logger.debug("Cursor ACP extension method: %s", method)
+        if method not in {"cursor/ask_question", "cursor/create_plan"}:
```
AlexanderZ-Band: nit: `cursor/ask_question` and `cursor/create_plan` are repeated in `CursorACPAdapter._resolve_extension_method` (`src/band/adapters/cursor_acp.py:221-223`). The profile admission check and adapter dispatch can drift. Define the extension identifiers once and share them.

src/band/adapters/cursor_acp.py
```diff
@@ -0,0 +1,545 @@
...
+            case "auto_accept":
+                return select_allow_option_id(request.options)
+            case "auto_decline":
+                return None
+            case "manual":
+                turn = self._active_turn_for(request.room_id, request.session_id)
+                if turn is None:
+                    return None
+                choices = {"permission": self._option_ids(request.options)}
+                result = await self._wait_for_decision(
```
AlexanderZ-Band, line 223: issue: Manual permission resolution waits up to five minutes while `ACPCollectingClient.request_permission()` holds the session ingestion lock. Concurrent Cursor notifications and session updates queue behind that wait, so live progress narration disappears and tasks accumulate. Release the lock during the user-mediated wait; retain atomicity only for the denied-permission narration.

src/band/adapters/cursor_acp.py
```diff
@@ -0,0 +1,545 @@
...
+        kind: DecisionKind,
+        turn: _CursorTurn,
+        prompt: str,
+        choices: dict[str, tuple[str, ...]] | None = None,
+        multi_select: frozenset[str] = frozenset(),
+    ) -> object | None:
+        token = uuid4().hex[:8]
+        future: asyncio.Future[object] = asyncio.get_running_loop().create_future()
+        self._evict_oldest_decision()
+        self._pending_decisions[token] = _PendingDecision(
```
AlexanderZ-Band: issue: If sending a manual decision prompt fails, `_wait_for_decision()` returns with this token still registered. It remains listed and consumes pending-decision capacity. Clean up and settle the pending decision on every exit after registration.

src/band/integrations/acp/client_profiles.py
```diff
@@ -49,62 +52,111 @@ async def ext_notification(
...
+        self._todos: dict[str, tuple[str, str]] = {}
+
+    @property
+    def extension_session_id(self) -> str | None:
+        """The session receiving Cursor notifications without a session id."""
+        return self._session_id
+
+    def bind_session(self, session_id: str | None) -> None:
+        """Bind extension notifications to the serialized Cursor turn's session."""
+        self._session_id = session_id
```
AlexanderZ-Band, line 76: issue: Binding a new ACP session leaves the prior session's todo state active. A first `merge: true` update in the new session emits old tasks into the new room. Scope todo state per session, or reset it whenever the bound session changes.

src/band/adapters/cursor_acp.py
```diff
@@ -0,0 +1,545 @@
...
+                        + self._question_summary(params)
+                    ),
+                )
+                return (
+                    result
+                    if isinstance(result, dict)
+                    else {"outcome": {"outcome": "cancelled"}}
+                )
+
+    async def _resolve_plan(
```
AlexanderZ-Band, line 284: issue: The advertised plan and permission decision policies have no behavioral coverage: the tests do not settle manual plan accept/reject, manually select a permission option, or assert automatic approval outcomes. A reversed plan outcome or broken positive permission path would pass. Add decision tests through the room-command path and direct auto-policy assertions.

src/band/adapters/cursor_acp.py
```diff
@@ -0,0 +1,545 @@
...
+            )
+        except Exception:
+            logger.warning("Could not deliver Cursor %s decision prompt", kind)
+            return None
+        try:
+            return await asyncio.wait_for(
+                future, timeout=self._config.decision_timeout_s
+            )
+        except TimeoutError:
+            await turn.tools.send_message(
```
AlexanderZ-Band: issue: A failure while sending the timeout notice escapes instead of returning the cancelled decision outcome. That turns a transient post-timeout room-delivery failure into an ACP RPC failure. Make the timeout notification best-effort and still cancel the Cursor request.

AlexanderZ-Band (commented):
Six verified in-scope issues require fixes before approval.

src/band/integrations/acp/client_profiles.py
AlexanderZ-Band: Fixed: shared Cursor extension method constants now drive both profile admission and adapter dispatch.

src/band/adapters/cursor_acp.py
AlexanderZ-Band: Fixed: manual decision waits now run outside the session ingestion lock; only denied-permission narration is serialized under that lock, with a regression test for live updates during the wait.

src/band/adapters/cursor_acp.py
AlexanderZ-Band: Fixed: pending decisions are now removed and settled from one finally block on every exit, including room-prompt delivery failure.

src/band/integrations/acp/client_profiles.py
AlexanderZ-Band: Fixed: Cursor todo state is keyed by ACP session and discarded during room and adapter teardown; a cross-session merge regression test covers the flow.

src/band/adapters/cursor_acp.py
AlexanderZ-Band: Fixed: added behavioral coverage for manual permission selection, manual plan acceptance, automatic plan outcomes, automatic permission policies, and failed/timeout decision delivery.

src/band/adapters/cursor_acp.py
AlexanderZ-Band: Fixed: timeout-notice delivery is best-effort and logged; the decision still returns cancellation, covered by a regression test.

src/band/adapters/cursor_acp.py
```diff
@@ -0,0 +1,559 @@
...
+        return frozenset(
+            question_id
+            for question in questions
+            if isinstance(question, dict)
+            and question.get("allowMultiple") is True
+            and isinstance((question_id := question.get("id")), str)
+        )
+
+    @staticmethod
+    def _question_summary(params: dict[str, object]) -> str:
```
AlexanderZ-Band, line 509: nit: `_question_summary` independently reparses the Cursor question/options payload already parsed by `_question_choices`, using a different accepted shape. Keep that projection in one helper so displayed choices and command validation cannot drift.

AlexanderZ-Band (commented):
One hygiene cleanup found in cycle 2.

src/band/adapters/cursor_acp.py
AlexanderZ-Band: Fixed: one shared question projection now supplies both validation choices and manual prompt rendering.

src/band/adapters/cursor_acp.py
```diff
@@ -0,0 +1,562 @@
...
+    approval_mode: ApprovalMode = "manual"
+    question_mode: QuestionMode = "manual"
+    plan_mode: PlanMode = "manual"
+    decision_timeout_s: float = 300.0
+    max_pending_decisions: int = 10
+    decision_authorized_senders: frozenset[str] | None = None
+
+
+@dataclass
+class _CursorTurn:
```
AlexanderZ-Band: nit: `_CursorTurn` and `_PendingDecision` (line 73) violate the repository's plain-PascalCase class-name convention. Rename them to `CursorTurn` and `PendingDecision`.

src/band/integrations/acp/client_runtime.py
```diff
@@ -641,6 +649,11 @@ async def ext_method(
 
     async def ext_notification(self, method: str, params: dict[str, object]) -> None:
         session_id = str(params.get("sessionId") or params.get("session_id") or "")
+        if not session_id:
+            profile_session_id = getattr(self._profile, "extension_session_id", None)
```
AlexanderZ-Band: nit: This reflection-based Cursor notification routing silently drops notifications if the profile property changes. Model the optional notification-session hook on `ACPClientProfile`, give the no-op profile a `None` implementation, and call it directly.

src/band/adapters/cursor_acp.py
```diff
@@ -0,0 +1,562 @@
...
+            for question in questions
+            if isinstance(question, dict)
+            and question.get("allowMultiple") is True
+            and isinstance((question_id := question.get("id")), str)
+        )
+
+    @staticmethod
+    def _question_summary(params: dict[str, object]) -> str:
+        return "; ".join(
+            f"{question_id}: {prompt} ({', '.join(option_ids)})"
```
AlexanderZ-Band: issue: Manual question prompts discard Cursor's semantic option labels and show only opaque IDs. A room participant cannot make an informed selection before the blocking request times out. Retain labels in the shared projection, render `id=label`, and add a regression assertion.

AlexanderZ-Band (commented):
Three final-cycle findings require a small cleanup batch.

src/band/adapters/cursor_acp.py
AlexanderZ-Band: Fixed: renamed the helper classes to `CursorTurn` and `PendingDecision` throughout the adapter and tests.

src/band/integrations/acp/client_runtime.py
AlexanderZ-Band: Fixed: `ACPClientProfile` now explicitly declares `extension_session_id`; the no-op profile returns `None` and runtime routing uses the typed hook directly.

src/band/adapters/cursor_acp.py
AlexanderZ-Band: Fixed: shared question details retain labels; manual prompts render `option-id=label`, with assertions covering opaque option IDs.

src/band/adapters/cursor_acp.py
```diff
@@ -0,0 +1,567 @@
...
+                not question_id
+                or question_id not in choices
+                or not option_ids
+                or (len(option_ids) > 1 and question_id not in multi_select)
+                or any(
+                    option_id not in choices[question_id] for option_id in option_ids
+                )
+            ):
+                return _INVALID_DECISION
+            selected[question_id] = option_ids
```
AlexanderZ-Band, line 432: issue: Repeating a question assignment silently overwrites the earlier answer. `/cursor answer <token> mode=agent mode=plan` selects `plan` rather than rejecting the ambiguous command. Reject duplicate question IDs before mutating the selected mapping.

AlexanderZ-Band (commented):
Focused logical recheck found one ambiguous manual-answer flow.

src/band/adapters/cursor_acp.py
AlexanderZ-Band: Fixed: duplicate question IDs now reject the command before an earlier answer can be overwritten, with a regression test.

src/band/adapters/cursor_acp.py
```diff
@@ -0,0 +1,582 @@
...
+            return True
+        result = self._command_result(action, words[3:], pending)
+        if result is _INVALID_DECISION:
+            await tools.send_message(
+                f"That command is not valid for Cursor {pending.kind} decision `{token}`."
+            )
+            return True
+        if not pending.future.done():
+            pending.future.set_result(result)
+        await tools.send_message(f"Cursor {pending.kind} decision `{token}` resolved.")
```
amit-gazal-band: **[A] Control-command replies carry no mentions, and the real `AgentTools` rejects them.**

Every `tools.send_message(...)` in `_handle_control_message` (the decisions list, "not pending", "not authorized", "not valid", and this "resolved." ack) is sent without `mentions`. `AgentTools._resolve_required_mentions` raises `BandToolError("At least one mention is required")`, and production passes plain `AgentTools.from_context(ctx)` with nothing adding a default mention.

For a valid `/cursor select`, the future is resolved on line 386 *before* this ack, so the decision does settle, but `on_message` then raises and the message cycle fails. For every error path the user gets no feedback at all.

**Confirmed:** with a real `AgentTools("room-1", MagicMock(), participants=[...])`, both `/cursor decisions` and `/cursor select nope allow-1` raise `BandToolError: At least one mention is required`. With the fake agent requesting permission and `TranscriptTools` enforcing the real mention contract, a valid `/cursor select <token> allow-1` approves the tool (`agent.approved is True`) and then raises `BandToolError` out of `on_message`.

Suggested fix: mention `msg.sender_id` on these replies.

src/band/adapters/cursor_acp.py
```diff
@@ -0,0 +1,582 @@
...
+        msg: PlatformMessage,
+        tools: AgentToolsProtocol,
+        history: ACPClientSessionState,
+        participants_msg: str | None,
+        contacts_msg: str | None,
+        *,
+        is_session_bootstrap: bool,
+        room_id: str,
+    ) -> None:
+        if await self._handle_control_message(msg, tools, room_id):
```
amit-gazal-band: **[1] A `/cursor` reply can never reach the decision it's meant to resolve.**

`ExecutionContext._process_loop` runs one message cycle per room at a time (`_run_cycle` awaits the handler task). While turn 1 is parked in `_wait_for_decision`, the user's `/cursor select …` is a *new message in the same room*, so it waits in `ctx.queue` until turn 1 finishes, which only happens when `decision_timeout_s` expires. With the defaults, every manual permission, question and plan times out and is cancelled or denied, and the late reply then gets "decision is not pending".

**Confirmed:** I started a real `ExecutionContext` (mocked link, `/next` → `None`) whose `on_execute` calls `CursorACPAdapter.on_message`, with `decision_timeout_s=1.0` and the fake agent requesting permission. I enqueued `run it`, waited for the token, then enqueued `/cursor select <token> allow-1`. The result was `agent.approved is False`, the reply was handled ≥0.8s later (after the timeout), and the room shows "timed out" then "`<token>` is not pending". **Control:** the same reply sent concurrently, bypassing the room loop, gives `agent.approved is True`, so the room queue is the cause, not the command parsing.

This needs a path that handles `/cursor` commands outside the room's serialized cycle (for example intercepting them before they're enqueued, or resolving from the receive side, the same way `interrupt()` works).

src/band/adapters/cursor_acp.py
```diff
@@ -0,0 +1,582 @@
...
+            return
+
+        async with self._turn_lock:
```
amit-gazal-band: **[2] `_turn_lock` is shared across all rooms.**

The lock is created once per adapter (line 103), so while room A waits up to `decision_timeout_s` for a decision (and because of [1] it always waits the full time), every Cursor turn in rooms B, C, … is blocked. The base `ACPClientAdapter` already runs a runtime per room, so this undoes that isolation.

**Confirmed:** two rooms on one adapter with `decision_timeout_s=1.0`. The fake agent's first prompt requests permission (room A), and later prompts just say "hello from B". Room B's plain turn completed only after ≥0.8s, i.e. after room A's decision timed out.

I think this needs per-room turn state (a lock and active turn keyed by `room_id`) instead of a single `_active_turn`. The session binding in `CursorACPClientProfile` would need the same treatment.

src/band/adapters/cursor_acp.py
```diff
@@ -0,0 +1,582 @@
...
+    api_key: str | None = None
+    auth_token: str | None = None
+    custom_section: str = ""
+    inject_band_tools: bool = True
+    mcp_servers: list[dict[str, object]] | None = None
+    resolve_session_config: SessionConfigResolver | None = None
+    approval_mode: ApprovalMode = "manual"
+    question_mode: QuestionMode = "manual"
+    plan_mode: PlanMode = "manual"
+    decision_timeout_s: float = 300.0
```
amit-gazal-band: **[3] Waiting on a decision counts against the turn timeout.**

`decision_timeout_s` defaults to 300, the same as the parent's `turn_timeout_s`, and the decision wait happens inside the prompt that `ACPClientAdapter.on_message` bounds with `asyncio.wait(..., timeout=self._turn_timeout_s)`. A user who approves at ~290s leaves Cursor about 10s before `ACPTurnTimeoutError` and `on_cleanup`, and any `decision_timeout_s ≥ turn_timeout_s` has no effect.

**Confirmed:** with `decision_timeout_s=10.0` and `_turn_timeout_s` scaled down to 0.5, the turn raised "timed out" in under 5s, before the decision window closed.

**Correction to my original note:** `turn_timeout_s` *can* be set. `CursorACPAdapter(turn_timeout_s=900)._turn_timeout_s == 900` works at runtime through `**features`, but `FeatureKwargs` doesn't declare it, so type checkers reject that call. Consider either exposing it on `CursorACPAdapterConfig` with validation that `decision_timeout_s < turn_timeout_s`, or pausing the turn deadline while a decision is pending.

src/band/adapters/cursor_acp.py
```diff
@@ -0,0 +1,582 @@
...
+        )
+
+    @staticmethod
+    def _question_summary(params: dict[str, object]) -> str:
+        return "; ".join(
+            f"{question_id}: {prompt} ({', '.join(f'{option_id}={label}' for option_id, label in options)})"
+            for question_id, (prompt, options) in CursorACPAdapter._question_details(
+                params
+            ).items()
+            if prompt is not None
```
amit-gazal-band: **[7] A question with no string `prompt` is hidden but still required.**

This filter drops those questions from the summary shown to the user, but `_question_choices` still includes them and `_answer_result` requires `set(selected) == set(choices)`. The user never sees the hidden question's id or options, so every `/cursor answer` they can build is rejected as invalid.

**Confirmed:** with `{"questions": [{"id": "visible", "prompt": "Pick", ...}, {"id": "hidden", "options": [...]}]}`, the summary doesn't contain `hidden`, `choices` does, and `_answer_result(["visible=a"], choices, frozenset()) is _INVALID_DECISION`.

Suggest either rendering these questions with a fallback label (e.g. the id) or leaving them out of `choices` as well.

src/band/integrations/acp/client_profiles.py
```diff
@@ -49,66 +61,126 @@ async def ext_notification(
...
-            return {"outcome": {"type": "cancelled"}}
-
-        if method == "cursor/create_plan":
-            return {"outcome": {"type": "approved"}}
-
-        return {}
+        logger.debug("Cursor ACP extension method: %s", method)
+        if method not in {CURSOR_ASK_QUESTION_METHOD, CURSOR_CREATE_PLAN_METHOD}:
+            return {}
+        if self._resolve_method is None:
```
amit-gazal-band: **[4] Regression for the bridge path: `resolve_acp_client_profile("cursor")` (line 194) returns `CursorACPClientProfile()` with no resolver.**

`examples/acp/clients/bridge_architecture.py` uses that factory. Every `cursor/ask_question` and `cursor/create_plan` now returns `cancelled` here, where `main` previously auto-answered or approved.

**Confirmed:** `await resolve_acp_client_profile("cursor").ext_method(...)` returns `{"outcome": {"outcome": "cancelled"}}` for both methods.

**Correction to my original note:** session-less `cursor/task` and `cursor/generate_image` notifications being dropped is *not* a regression; `main`'s `ext_notification` also returned early without a `sessionId`. See the next comment for todos.

src/band/integrations/acp/client_profiles.py
```diff
@@ -49,66 +61,126 @@ async def ext_notification(
...
+
+        if method == "cursor/generate_image":
+            return self._image_chunks(params)
 
         return []
 
+    def _todo_chunks(self, params: dict[str, object]) -> list[CollectedChunk]:
+        """Apply Cursor's replace-or-merge todo update and render its state."""
+        todos = params.get("todos")
+        session_id = self._session_id
```
amit-gazal-band: **[4, continued] Todos only render when a session is bound, even if the notification has its own `sessionId`.**

`_todo_chunks` keys on `self._session_id` rather than the notification's `sessionId`. Nothing on the bridge path calls `bind_session`, so `cursor/update_todos` never renders there, a regression from `main`.

**Confirmed:** on the factory profile, `ext_notification("cursor/update_todos", {"sessionId": "s-1", "todos": [{"id": "t", "content": "do it", "status": "pending"}]})` returns `[]`.

Suggest preferring `params["sessionId"]` and falling back to the bound session.

src/band/integrations/acp/client_profiles.py
```diff
@@ -3,16 +3,24 @@
...
+CursorMethodResolver = Callable[[str, dict[str, object]], Awaitable[dict[str, object]]]
+CURSOR_ASK_QUESTION_METHOD = "cursor/ask_question"
+CURSOR_CREATE_PLAN_METHOD = "cursor/create_plan"
+
 
 class ACPClientProfile(Protocol):
     """Extension hook surface for runtime-specific ACP behavior."""
 
+    @property
+    def extension_session_id(self) -> str | None: ...
```
amit-gazal-band: **[8] Compat: `extension_session_id` is now a required member of the `ACPClientProfile` protocol.**

`ACPCollectingClient.ext_notification` (`client_runtime.py:655`) reads it whenever a notification has no session id, so a custom profile written against the previous protocol raises.

**Confirmed:** a profile class with only `ext_method`/`ext_notification`, passed to `ACPCollectingClient`, raises `AttributeError: ... extension_session_id` on `ext_notification("cursor/task", {"description": "d"})`.

A `getattr(self._profile, "extension_session_id", None)` in the runtime, or a changelog note, would cover it.

.github/scripts/setup-cursor.sh
```diff
@@ -0,0 +1,19 @@
+#!/usr/bin/env bash
+# Install Cursor CLI for the Cursor ACP baseline backend.
+set -euo pipefail
+
+: "${CURSOR_API_KEY:?CURSOR_API_KEY is required for Cursor ACP E2E auth}"
```
amit-gazal-band: **[5] A missing secret fails the whole `backends` lane instead of skipping the Cursor cells.**

Under `set -euo pipefail`, `: "${CURSOR_API_KEY:?...}"` exits non-zero, so the setup step fails, and with it the codex, opencode and copilot_acp cells. That defeats `Dep.CURSOR_CLI` in `deps.py`, which is built to skip Cursor cleanly when the key is absent. On the nightly run it would also block the baseline-green mark that `release-gate.yml` reads.

**Confirmed:** running the script with `CURSOR_API_KEY` removed from the environment exits non-zero with "CURSOR_API_KEY is required for Cursor ACP E2E auth" on stderr.

**Correction to my original note:** `e2e.yml` only runs on the nightly schedule or manual dispatch, so fork PRs aren't affected; the risk is a nightly or dispatch run where `E2E_CURSOR_API_KEY` isn't configured.

Suggest warning and `exit 0` (or skipping the install) when the key is unset, and letting `Dep.CURSOR_CLI` handle the skip.

tests/e2e/baseline/toolkit/builders.py
```diff
@@ -551,6 +551,42 @@ def _build_copilot_acp(
...
+    from band.adapters.cursor_acp import (  # noqa: PLC0415 -- isolates the ACP extra from other framework builders
+        CursorACPAdapter,
+        CursorACPAdapterConfig,
+    )
+
+    sandbox = tempfile.mkdtemp(prefix="band-e2e-cursor-acp-")
+    config_kwargs: dict[str, Any] = {
+        "api_key": s.backends.cursor_api_key,
+        "custom_section": prompt or "",
+        "cwd": sandbox,
```
amit-gazal-band: **[6] The E2E builder leaves `approval_mode`, `question_mode` and `plan_mode` on `"manual"`.**

Nothing in the baseline matrix answers `/cursor …` prompts, and because of [1] nothing could. Any cell where Cursor asks permission (e.g. before a Band MCP tool call), asks a question, or proposes a plan will stall for 300s and then fail or deny.

**Confirmed (config only):** `_build_cursor_acp(BaselineSettings(), prompt=None, features=None)._config` has all three modes set to `"manual"`. I did **not** run a live Cursor cell, so whether a given cell actually triggers a decision is unverified.

Suggest `approval_mode="auto_accept"`, `question_mode="auto_first"`, `plan_mode="auto_accept"` for the matrix, plus a dedicated smoke for the manual flow once [A] and [1] are fixed.

amit-gazal-band (changes requested):
Requesting changes. Two bugs mean the default `manual` decision flow can't work in a live room: control replies raise on the real `AgentTools`, and a reply is queued behind the turn it's meant to unblock. Details are inline.

**How this was checked:** I wrote a local reproduction suite against this head (`045222b5`). The fake Cursor agent comes from `tests/integrations/acp/acp_toolkit` (real ACP JSON-RPC over an in-process socketpair, plus an `authenticate` stub). Where it mattered I used a real `ExecutionContext` and the real `AgentTools`. Each test asserts the buggy behavior, so a passing test confirms the finding. All 12 pass, and the PR's own 19 Cursor tests still pass. I can share the file if that's useful.

| # | Finding | Status |
|---|---|---|
| A | Control-command replies have no mentions, so real `AgentTools` raises | Confirmed (real `AgentTools`) |
| 1 | A `/cursor` reply is queued behind the turn waiting for it | Confirmed (real `ExecutionContext`) |
| 2 | `_turn_lock` is adapter-wide, so one pending decision blocks every room | Confirmed |
| 3 | Time spent waiting on a decision counts against the turn timeout | Confirmed; `turn_timeout_s` works at runtime but isn't typed |
| 4 | Bridge `cursor` profile cancels questions and plans and never renders todos | Confirmed |
| 5 | `setup-cursor.sh` hard-fails the backends lane when the secret is missing | Confirmed |
| 6 | E2E builder leaves all three decision modes on `manual` | Config confirmed; live stall not exercised |
| 7 | A question with no prompt is hidden from the user but still required | Confirmed |
| 8 | Profile protocol now requires `extension_session_id` | Confirmed (compat note) |

Findings A and 1 are the blockers. The unit tests miss both because they call `_handle_control_message` directly from a side task with a fake `send_message` that doesn't enforce mentions.

src/band/adapters/cursor_acp.py
AlexanderZ-Band: Fixed in 6a7af3e8: every reply in `_handle_control_message` (decisions list, not-pending, not-authorized, invalid-command, resolved ack) now mentions the sender, matching the mention discipline `_wait_for_decision` already used for its own prompts.

src/band/adapters/cursor_acp.py
AlexanderZ-Band: Fixed in 6a7af3e8, using OpencodeAdapter's turn_release_future pattern as precedent: `on_message` now runs the ACP turn as a detached task and returns the moment a decision opens (an early-release future resolved from `_wait_for_decision` right after the prompt posts), instead of blocking for the whole turn. That frees the room's message queue for the reply well before `decision_timeout_s`. Added a regression test that drives two sequential `on_message` calls the way `ExecutionContext` would and hand-verified it fails with a `TimeoutError` against the pre-fix code.

src/band/adapters/cursor_acp.py
AlexanderZ-Band: Confirmed as a real, separate defect, but it turned out to need more than a per-room lock: `ext_method`/`ext_notification` on `ACPClientProfile` carry no session/room identifier, so the single shared `CursorACPClientProfile` instance can only route correctly because `_turn_lock` keeps one room active at a time — that's a public Protocol change (`client_profiles.py`), not an adapter-local fix like the others here. Filed as https://linear.app/thenvoi/issue/INT-1540 with the root-cause writeup and fix sketch rather than folding it into this PR.

src/band/adapters/cursor_acp.py
AlexanderZ-Band: Fixed in 6a7af3e8: `turn_timeout_s` is now a typed `CursorACPAdapterConfig` field (default 900s) passed through explicitly to the base class, with `_validate_config` enforcing `decision_timeout_s < turn_timeout_s`.

src/band/adapters/cursor_acp.py
AlexanderZ-Band: Fixed in 6a7af3e8: `_question_details` (the single source both `_question_choices` and `_question_summary` read from) now falls back to the question's id as its displayed label when no string `prompt` is given, so every question required by `_answer_result` is also visible.

src/band/integrations/acp/client_profiles.py
AlexanderZ-Band: Fixed in 6a7af3e8: with no resolver, `ext_method` now answers unattended (first advertised option per question, plan auto-accepted) instead of cancelling every request, restoring `main`'s prior standalone behavior under the current outcome shape.

src/band/integrations/acp/client_profiles.py
AlexanderZ-Band: Fixed in 6a7af3e8: `_todo_chunks` now prefers the notification's own `sessionId`/`session_id`, falling back to the bound session — the same precedence `ACPCollectingClient.ext_notification` already uses for chunk routing.

src/band/integrations/acp/client_profiles.py
AlexanderZ-Band: Fixed in 6a7af3e8: the session-less fallback now reads `getattr(self._profile, "extension_session_id", None)` instead of a direct attribute access, so a profile written against the pre-PR protocol no longer raises `AttributeError`.

.github/scripts/setup-cursor.sh
AlexanderZ-Band: Fixed in 6a7af3e8: the script now checks for `CURSOR_API_KEY` up front and exits 0 with a log line when it's unset, instead of hard-failing under `set -euo pipefail` — `Dep.CURSOR_CLI`'s skip logic can now do its job. Verified locally that the script exits 0 without attempting the install when the key is absent.

tests/e2e/baseline/toolkit/builders.py
AlexanderZ-Band: Fixed in 6a7af3e8: `_build_cursor_acp` now sets `approval_mode="auto_accept"`, `question_mode="auto_first"`, `plan_mode="auto_accept"` in the baseline config.

src/band/adapters/cursor_acp.py
```diff
@@ -0,0 +1,664 @@
...
+        finally:
+            pending = self._pending_decisions.pop(token, None)
+            if pending is not None and not pending.future.done():
+                pending.future.set_result(None)
+
+    async def _handle_control_message(
+        self, msg: PlatformMessage, tools: AgentToolsProtocol, room_id: str
+    ) -> bool:
+        words = strip_leading_mentions(msg.content).strip().split()
+        if not words or words[0].lower() != "/cursor":
```
AlexanderZ-Band: issue: The `/cursor` command vocabulary (`select`, `deny`, `answer`, `accept`, `reject`, `decisions`) is spelled out twice with no shared source -- once as literal match targets in `_command_result` (lines 469-484) and again, independently typed, inside every prompt string shown to the room (lines 290-293, 334-337, 360-362, 434). A rename or typo in one place silently desyncs from the other: the user follows the prompt text and the parser rejects it as invalid.

Root-cause fix: define the command verbs once (a StrEnum or a small verb -> description mapping) and build both the match arms and each prompt's `/cursor ...` fragment from it, so the two can't drift.

src/band/integrations/acp/client_adapter.py
```diff
@@ -517,15 +539,101 @@ async def handler(
...
+            if isinstance(option, Mapping):
+                option_id = option.get("optionId", option.get("option_id"))
+            else:
+                option_id = option.option_id
+            if isinstance(option_id, str):
+                option_ids.add(option_id)
+        return option_ids
+
+    @staticmethod
+    def _resolve_transport(
```
AlexanderZ-Band: issue: `_resolve_transport` (new in this PR) is dead code -- nothing in the repository calls it. It also disagrees with the adapter's real behavior: `__init__` (lines 249-256) unconditionally raises `ValueError` for `host`/`port`/`spawn_process`, so there is no path through which `_resolve_transport`'s TCP-vs-stdio selection logic can ever run.

Root-cause fix: delete `_resolve_transport`; the constructor's existing inline checks are the actual (and only) transport validation.

src/band/integrations/acp/client_adapter.py
```diff
@@ -517,15 +539,101 @@ async def handler(
...
+        if not isinstance(options, (list, tuple)):
+            return ()
+        return tuple(
+            option
+            for option in options
+            if isinstance(option, (PermissionOption, Mapping))
+        )
+
+    @staticmethod
+    def _permission_option_ids(options: tuple[PermissionOptionValue, ...]) -> set[str]:
```
AlexanderZ-Band: issue: option-id extraction from a permission's offered options is implemented three times with subtly different behavior: `ACPClientAdapter._permission_option_ids` (this line, returns an unordered `set`), `CursorACPAdapter._option_ids` (src/band/adapters/cursor_acp.py:599, returns an ordered `tuple`, bare attribute access instead of `getattr`), and `select_allow_option_id` (client_runtime.py:51, a related but distinct one-option extraction). All three are new or newly exercised by this PR and walk the same `Mapping`-or-object option shape.

Root-cause fix: factor one ordered extraction helper (e.g. in client_runtime.py, already imported by both adapters) and have `_permission_option_ids` wrap it in `set(...)` and `CursorACPAdapter._option_ids` call it directly, instead of maintaining three parallel walks.

src/band/adapters/cursor_acp.py
```diff
@@ -0,0 +1,664 @@
...
+        params: dict[str, object],
+    ) -> dict[str, tuple[str, tuple[tuple[str, str], ...]]]:
+        """Project the valid question IDs, prompts, and labeled options once.
+
+        Every entry here is required for a complete ``/cursor answer`` (see
+        ``_answer_result``), so each gets a displayable prompt -- falling
+        back to its id -- rather than being silently absent from the summary
+        while still being required.
+        """
+        questions = params.get("questions")
```
AlexanderZ-Band: issue: Cursor's `questions` payload (`{id, prompt, options:[{id,label}], allowMultiple}`) is independently re-parsed four times by this PR's new code: `_question_details` (this line), `_multiple_choice_questions` (line 578), `_question_summary` (recomputes `_question_details` on every call), and `CursorACPClientProfile._auto_answer_question` (client_profiles.py:106) -- each hand-rolls its own `isinstance(question, Mapping)`/`question.get("id")`/options-walk. Two real defects already live inside this duplicated parsing (see the two Logical Bugs findings on `_question_details` in this review), which is exactly the kind of drift four independent copies of the same walk invites.

Root-cause fix: parse `params["questions"]` once into a small typed structure (id, prompt, allow_multiple, options) in client_profiles.py, which already owns Cursor's wire format, and have all four call sites consume that one parse instead of each re-walking the raw dict.

src/band/adapters/cursor_acp.py
```diff
@@ -0,0 +1,664 @@
...
+    ) -> None:
+        config = config or CursorACPAdapterConfig()
+        self._validate_config(config)
+        self._config = config
+        self._cursor_profile = CursorACPClientProfile(self._resolve_extension_method)
+        self._turn_lock = asyncio.Lock()
+        self._active_turn: CursorTurn | None = None
+        self._pending_decisions: dict[str, PendingDecision] = {}
+        env = self._cursor_env(config)
+        workspace_for_room = config.workspace_for_room
```
AlexanderZ-Band: nit: The `cwd`/`workspace_for_room` mutual-exclusion block (lines 113-117) is a verbatim copy of `CopilotACPAdapter.__init__`'s equivalent block, including the exact error string, instead of a shared helper in `band.workspaces` (which already owns `create_room_workspace_resolver`).

Fix: add a small `workspace_resolver_for(cwd, workspace_for_room)` helper to `band/workspaces.py` and have both `CopilotACPAdapter.__init__` and `CursorACPAdapter.__init__` call it.

tests/framework_conformance/test_adapter_conformance.py
```diff
@@ -274,6 +274,11 @@ def test_accepts_files_capability(self, adapter_config):
         "wraps the ACP client adapter, which shares the same MCP engine fix; "
         "not a separate ADAPTER_CONFIGS entry, so it has no probe of its own",
     ),
+    ExcludedAdapter(
```
AlexanderZ-Band: nit: The `ExcludedAdapter` reason string for `Adapter.CURSOR_ACP` (lines 278-280) is an exact copy-paste of the `Adapter.COPILOT_ACP` entry immediately above it (lines 273-275): "wraps the ACP client adapter, which shares the same MCP engine fix; not a separate ADAPTER_CONFIGS entry, so it has no probe of its own".

Fix: pull the shared reason into one named constant referenced by both `ExcludedAdapter(...)` calls.

src/band/adapters/cursor_acp.py
```diff
@@ -0,0 +1,664 @@
...
+            turn
+            if turn and (turn.room_id, turn.session_id) == (room_id, session_id)
+            else None
+        )
+
+    def _is_authorized(self, sender_id: str | None) -> bool:
+        allowed = self._config.decision_authorized_senders
+        return allowed is None or sender_id in allowed
+
+    def _evict_oldest_decision(self) -> None:
```
AlexanderZ-Band: issue: `_evict_oldest_decision` (624-630), `_cancel_room_decisions` (632-637), and `_cancel_all_decisions` (639-643) all silently resolve a pending Cursor permission/question/plan future to `None` -- with no logging anywhere in any of the three, unlike the timeout path in `_wait_for_decision`, which both logs and notifies the room. Trigger: a room exceeds `max_pending_decisions` (default 10) and the oldest still-outstanding decision is evicted, or `on_cleanup`/`cleanup_all` fires while a decision is outstanding. An operator debugging why a Cursor permission/plan was silently denied has no trace to follow.

Fix: add a `logger.info`/`warning` at each cancellation site naming room_id, decision kind, and token -- mirror the existing timeout-notice logging pattern already used in `_wait_for_decision`.

src/band/adapters/cursor_acp.py
```diff
@@ -0,0 +1,664 @@
...
+                tools,
+                history,
+                participants_msg,
+                contacts_msg,
+                is_session_bootstrap=is_session_bootstrap,
+                room_id=room_id,
+            )
+        )
+        self._background_tasks.add(turn_task)
+        turn_task.add_done_callback(self._on_background_task_done)
```
AlexanderZ-Band: issue: The detached `turn_task` (lines 188-201) is tracked through the shared, generic `_background_tasks`/`_on_background_task_done` mechanism (client_adapter.py:950-956), whose only failure signal is `logger.warning("ACP background task failed", exc_info=error)` with zero room/session/decision context -- indistinguishable from any other background task's failure. `on_cleanup` (lines 256-262) explicitly documents that this detached task can keep running and fail after the room has moved on, so this is the only diagnostic signal an operator gets for that documented failure mode.

Fix: tag the failure with room_id/session_id before or instead of relying solely on the generic sink -- e.g. a dedicated done-callback for `turn_task`, or catch/log inside `_run_turn`'s `finally` before re-raising.

src/band/adapters/cursor_acp.py
```diff
@@ -0,0 +1,664 @@
...
+            mcp_servers=config.mcp_servers,
+            resolve_session_config=config.resolve_session_config,
+            resolve_permission=self._resolve_cursor_permission,
+            turn_timeout_s=config.turn_timeout_s,
+            **features,
+        )
+
+    @staticmethod
+    def _validate_config(config: CursorACPAdapterConfig) -> None:
+        if not config.command:
```
AlexanderZ-Band: issue: `_validate_config` has three `ValueError` branches with zero test coverage: empty `command` (136-137), `decision_timeout_s <= 0` (140-141), and `max_pending_decisions <= 0` (144-145). Only the ambiguous-auth, ambiguous-workspace, and decision/turn-timeout-ordering branches are tested, though the ticket explicitly calls out "configuration validation" as expected coverage.

Fix: add parametrized tests asserting `pytest.raises(ValueError, match=...)` for each of the three branches.

src/band/adapters/cursor_acp.py
```diff
@@ -0,0 +1,664 @@
...
+                    )
+                except Exception:  # noqa: BLE001 -- best-effort room notify; the decision has already timed out, so a delivery failure here changes nothing
+                    logger.warning("Could not deliver Cursor %s timeout notice", kind)
+                return None
+        finally:
+            pending = self._pending_decisions.pop(token, None)
+            if pending is not None and not pending.future.done():
+                pending.future.set_result(None)
+
+    async def _handle_control_message(
```
AlexanderZ-Band: issue: `_handle_control_message`, the room-facing `/cursor` command router this PR introduces, has several branches with no end-to-end test: the bare `/cursor`/`/cursor decisions` listing path (429-431 -> `_list_decisions`), the `len(words) < 3` help-text fallback (432-437), and the `_INVALID_DECISION` -> "not valid for Cursor {kind} decision" reply (452-457). The one duplicate-answer test exercises `_answer_result` directly, bypassing dispatch, so the actual room reply a user would see for a malformed command is never asserted.

Fix: send `/cursor`, `/cursor decisions`, a 2-word command, and a structurally invalid decision command through `_handle_control_message` and assert on the tool's sent message.

src/band/adapters/cursor_acp.py
```diff
@@ -0,0 +1,664 @@
...
+    async def _resolve_cursor_permission(
+        self, request: ACPPermissionRequest
+    ) -> str | None:
+        match self._config.approval_mode:
+            case "auto_accept":
+                return select_allow_option_id(request.options)
+            case "auto_decline":
+                return None
+            case "manual":
+                turn = self._active_turn_for(request.room_id, request.session_id)
```
AlexanderZ-Band: suggestion: the manual-mode silent-deny path in `_resolve_cursor_permission` when `_active_turn_for` finds no matching active turn for the request's room/session (280-282, returns `None` with no room message) has no test.

Fix: test `_resolve_cursor_permission` with `approval_mode='manual'` and no matching `_active_turn` set, asserting the result is `None` and nothing is sent to the room.

src/band/adapters/cursor_acp.py
```diff
@@ -0,0 +1,664 @@
...
+            if turn and (turn.room_id, turn.session_id) == (room_id, session_id)
+            else None
+        )
+
+    def _is_authorized(self, sender_id: str | None) -> bool:
+        allowed = self._config.decision_authorized_senders
+        return allowed is None or sender_id in allowed
+
+    def _evict_oldest_decision(self) -> None:
+        if len(self._pending_decisions) < self._config.max_pending_decisions:
```
AlexanderZ-Band: suggestion: `_evict_oldest_decision` -- a real behavior where the oldest pending decision is silently cancelled to make room for a new one -- has no test at all.

Fix: configure `max_pending_decisions=1`, open two decisions, and assert the first's result resolves to `None` (cancelled) once the second is created.

src/band/integrations/acp/client_profiles.py
```diff
@@ -49,66 +61,168 @@ async def ext_notification(
...
+                chunk_type=ChunkType.PLAN,
+                content="\n".join(lines),
+                metadata={"cursor_todos": True},
+            )
+        ]
+
+    @staticmethod
+    def _task_chunks(params: dict[str, object]) -> list[CollectedChunk]:
+        """Render Cursor's documented subagent-task notification."""
+        description = params.get("description")
```
AlexanderZ-Band: nit: `_task_chunks`/`_image_chunks`'s empty/missing-description no-op guards (201-203, 214-216) and the `in_progress`/`cancelled` todo status marks (180-183; only `completed`/`pending` are exercised) are untested.

Fix: add cases for an empty-description task/image notification returning `[]`, and a todo update with `status='in_progress'`/`'cancelled'` asserting the `~`/`-` marks.

src/band/adapters/cursor_acp.py
```diff
@@ -0,0 +1,664 @@
...
+        result = self._command_result(action, words[3:], pending)
+        if result is _INVALID_DECISION:
+            await tools.send_message(
+                f"That command is not valid for Cursor {pending.kind} decision `{token}`.",
+                mentions=mentions,
+            )
+            return True
+        if not pending.future.done():
+            pending.future.set_result(result)
+        await tools.send_message(
```
AlexanderZ-Band: issue: A decision that times out can produce a misleading confirmation if a reply for the same token lands while the timeout notice is still being sent. In `_wait_for_decision` (404-420), when `asyncio.wait_for(future, ...)` times out, `future` is already cancelled/done at that instant, but the token is only popped from `_pending_decisions` in the outer `finally` (417-420) -- which runs *after* the `await turn.tools.send_message(...)` timeout notice (409-413). During that await, `_handle_control_message` (422-463) still finds the token pending; its `if not pending.future.done()` guard (458) only prevents a crash, so it still falls through to send "Cursor {kind} decision resolved" (460-462) -- telling the user their `/cursor select|deny|accept|reject|answer` was applied when it was actually already discarded by the timeout.

Fix: pop the token from `_pending_decisions` synchronously as soon as the future is observed done (right after catching `TimeoutError`, before awaiting the notice send) -- mirroring how `_evict_oldest_decision`/`_cancel_room_decisions` already pop with no intervening await. Additionally/alternatively, branch `_handle_control_message`'s reply on whether the future was already done at lookup, so it never claims a discarded answer succeeded.

src/band/integrations/acp/client_profiles.py
```diff
@@ -49,66 +61,168 @@ async def ext_notification(
...
+            and isinstance((content := todo.get("content")), str)
+            and isinstance((status := todo.get("status")), str)
+        }
+        if params.get("merge") is True:
+            current_todos = self._todos_by_session.setdefault(session_id, {})
+            current_todos.update(updates)
+        else:
+            self._todos_by_session[session_id] = updates
+        current_todos = self._todos_by_session[session_id]
+        if not current_todos:
```
AlexanderZ-Band: issue: `_todo_chunks` returns `[]` whenever the updated `current_todos` is empty (this line), including right after a legitimate Cursor `cursor/update_todos` notification that clears the list (`todos: []`). The session's internal state is correctly cleared, but no chunk is emitted, so the room keeps displaying the last-rendered, now-stale todo checklist with no signal it was cleared. Reproduced: two chunks are collected for a populated list, zero for the clearing update, while `_todos_by_session` correctly shows `{}`.

Fix: when the update legitimately transitions to an empty todo list, still emit a chunk (e.g. an explicit cleared-checklist rendering) instead of returning `[]`; only skip emission when there was never any prior or new todo state at all.

src/band/adapters/cursor_acp.py
```diff
@@ -0,0 +1,664 @@
...
+                    option_id,
+                    label
+                    if isinstance((label := option.get("label")), str)
+                    else option_id,
+                )
+                for option in options
+                if isinstance(option, Mapping)
+                and isinstance((option_id := option.get("id")), str)
+            )
+            if option_details:
```
AlexanderZ-Band: issue: `_question_details` skips adding a question to `details` when its options list is empty (`if option_details:` on this line), contradicting its own docstring ("Every entry here is required ... rather than being silently absent from the summary while still being required"). Since `_question_choices`/`_answer_result`'s completeness check derive from this same filtered `details`, a malformed question with an unparseable/empty options list is dropped from the manual prompt entirely, and the `answered` outcome sent back to Cursor looks complete but is missing an answer for a question Cursor actually asked. Reproduced with a two-question payload where one has valid options and one has `options: []`: `_answer_result` returns a successful `answered` outcome covering only the first question.

Fix: don't let an empty options list silently exclude the question from `details`; either always include it (so it counts toward the completeness check and gets an explicit unanswerable path) or route the whole exchange to `cancelled` when any question can't be answered.

src/band/adapters/cursor_acp.py
```diff
@@ -0,0 +1,664 @@
...
+
+        Every entry here is required for a complete ``/cursor answer`` (see
+        ``_answer_result``), so each gets a displayable prompt -- falling
+        back to its id -- rather than being silently absent from the summary
+        while still being required.
+        """
+        questions = params.get("questions")
+        if not isinstance(questions, list):
+            return {}
+        details: dict[str, tuple[str, tuple[tuple[str, str], ...]]] = {}
```
AlexanderZ-Band: issue: `_question_details` keys its result by `question_id` (this line, and the assignment at 569-572), so two entries in the same `cursor/ask_question` payload sharing an `id` silently collapse to whichever is processed last -- the earlier question's prompt and options vanish with no error or signal, and only the later question is ever presented or required for a complete answer. Reproduced: a two-entry payload with duplicate id `q1` yields a single `details['q1']` holding only the second entry's prompt/options.

Fix: detect duplicate question ids while building `details` and either reject/cancel the exchange as malformed, or explicitly keep-first-and-log, rather than a silent last-write-wins assignment.

AlexanderZ-Band (commented):
Automated multi-angle review (Style & Hygiene, Design & Reuse, Runtime Safety & Observability, Test Quality, Edge-Case Flow Analysis, Logical Bugs).

17 findings: 12 issue, 3 nit, 2 suggestion. Three are proven Logical Bugs (with repro scripts): `_todo_chunks` silently drops a legitimate "clear all todos" update; `_question_details` silently drops a question with an empty/unparseable options list from a multi-question prompt, so the `answered` outcome sent back to Cursor is missing an answer Cursor actually required; and a duplicate question id in the same payload silently collapses to the later entry. One Edge-Case Flow finding: a timeout and a late user reply for the same decision token can race, so the user is told their reply was "resolved" when it was actually discarded by the timeout. Two Runtime Safety findings are both missing-logging gaps (decision eviction/cancellation, and the detached turn task's generic failure log lacking room/session context). The rest are duplication/dead-code (Style & Hygiene, Design & Reuse -- most notably `_resolve_transport` being unreachable dead code, and permission option-id extraction being implemented three separate times) and missing test coverage (Test Quality) for validation branches, control-command dispatch paths, and decision eviction.

All findings were verified against the real file content at this PR's head commit and are in scope of this diff (Linear INT-1513). All 21 threads from the prior review round remain resolved and were re-checked live -- none reopened.

src/band/adapters/cursor_acp.py
AlexanderZ-Band: Fixed in f15c1e01: the `/cursor` command vocabulary is now a single `CursorCommandWord` StrEnum; both `_command_result`'s match arms and every prompt string build from it.

src/band/integrations/acp/client_adapter.py
AlexanderZ-Band: Fixed in f15c1e01: deleted `_resolve_transport`.

src/band/integrations/acp/client_adapter.py
AlexanderZ-Band: Fixed in f15c1e01: factored a single `permission_option_ids` helper in `client_runtime.py` (`select_allow_option_id`'s id resolution now shares it too); `_permission_option_ids` and the Cursor adapter's permission-option lookup both call it directly.

src/band/adapters/cursor_acp.py
AlexanderZ-Band: Fixed in f15c1e01: added `parse_cursor_questions`/`CursorQuestion` in `client_profiles.py` as the single parse of Cursor's `questions` payload; `_resolve_question`, `_question_summary`, and `_auto_answer_question` all consume it now (the old `_question_details`/`_question_choices`/`_multiple_choice_questions` are gone). This also fixes the two Logical Bugs findings below that lived inside the old duplicated parsing.

src/band/adapters/cursor_acp.py
AlexanderZ-Band: Fixed in f15c1e01: added `workspace_resolver_for(cwd, workspace_for_room)` to `band/workspaces.py`; both `CursorACPAdapter` and `CopilotACPAdapter` call it instead of each re-typing the mutual-exclusion check.

tests/framework_conformance/test_adapter_conformance.py
AlexanderZ-Band: Fixed in f15c1e01: pulled the shared reason into `_WRAPS_ACP_CLIENT_ADAPTER_REASON`, referenced by both `ExcludedAdapter(...)` calls.

src/band/adapters/cursor_acp.py
AlexanderZ-Band: Fixed in f15c1e01: `_evict_oldest_decision`, `_cancel_room_decisions`, and `_cancel_all_decisions` each log the room, decision kind, and token before cancelling.

src/band/adapters/cursor_acp.py
AlexanderZ-Band: Fixed in f15c1e01: `on_message`'s detached `turn_task` now has its own `_on_turn_task_done` callback that logs a room-tagged warning on failure, instead of relying solely on the generic, contextless `_on_background_task_done` sink.

src/band/adapters/cursor_acp.py
AlexanderZ-Band: Fixed in f15c1e01: added tests for the empty-`command`, `decision_timeout_s<=0`, and `max_pending_decisions<=0` branches.

src/band/adapters/cursor_acp.py
AlexanderZ-Band: Fixed in f15c1e01: added tests driving `/cursor`, `/cursor decisions`, a 2-word command, and a structurally invalid decision command through `_handle_control_message`, asserting on the room reply.

src/band/adapters/cursor_acp.py
AlexanderZ-Band: Fixed in f15c1e01: added `test_manual_permission_denies_silently_with_no_matching_active_turn`.

src/band/adapters/cursor_acp.py
AlexanderZ-Band: Fixed in f15c1e01: added `test_exceeding_max_pending_decisions_evicts_the_oldest`.

src/band/integrations/acp/client_profiles.py
AlexanderZ-Band: Fixed in f15c1e01: added tests for `_task_chunks`/`_image_chunks`'s empty-description no-op and the `in_progress`/`cancelled` todo marks.

src/band/adapters/cursor_acp.py
AlexanderZ-Band: Fixed in f15c1e01: `_wait_for_decision` now pops the token from `_pending_decisions` synchronously as soon as `TimeoutError` is caught, before awaiting the (best-effort) timeout-notice send -- closing the window where a late room reply for the same token could find it still pending and be told 'resolved' after the timeout had already discarded it. Regression test `test_a_late_reply_during_the_timeout_notice_is_not_reported_as_resolved`, verified to fail on the pre-fix code with the old 'resolved' message.

src/band/integrations/acp/client_profiles.py
AlexanderZ-Band: Fixed in f15c1e01: `_todo_chunks` now emits an explicit '(no todos)' chunk when a real update clears the list to empty, instead of silently emitting nothing and leaving the stale checklist rendered. Regression test `test_ext_notification_cursor_update_todos_clearing_the_list_still_renders`.

src/band/adapters/cursor_acp.py
AlexanderZ-Band: Fixed in f15c1e01, as part of the `parse_cursor_questions` consolidation (see the Design & Reuse finding above): a question with no valid options is no longer silently dropped -- the whole exchange now cancels rather than reporting a false 'answered' outcome missing that question's answer. Regression test `test_a_question_with_no_answerable_options_cancels_the_whole_exchange`, verified against your repro shape.

src/band/adapters/cursor_acp.py
AlexanderZ-Band: Fixed in f15c1e01, same consolidation: `parse_cursor_questions` now keeps a question's first occurrence and drops (with a logged warning) a later duplicate id, instead of a silent last-write-wins collapse. Regression test `test_a_duplicate_question_id_keeps_the_first_occurrence`, verified against your repro shape.

amit-gazal-band (approved):

AlexanderZ-Band:
Follow-up filed: [INT-1542](https://linear.app/thenvoi/issue/INT-1542) — extract a shared chat-mediated decision registry that both `OpencodeAdapter`'s `RoomApprovals` and `CursorACPAdapter`'s `PendingDecision` machinery build on, instead of two independent hand-rolled implementations of the same pattern. Root-caused after two review rounds on this PR each found fresh defects in Cursor's copy of that mechanism.

PR #654, opened 2026-09-22
feat: Add OMP ACP support to Python SDK

AlexanderZ-Band:
## Summary
- Add a thin, stdio-only `OmpACPAdapter` over the shared ACP client lifecycle (no duplicated session/MCP/cleanup logic), pinned for CI to `@oh-my-pi/pi-coding-agent@18.2.8`.
- Enforce a final `--approval-mode always-ask`, fail-closed Approve/Deny form elicitation via the new shared `PermissionResolver`, and normalize OMP `xd://mcp__…` device writes before narration.
- Register OMP across baseline/E2E (`Dep.OMP` on the backends lane), docs, example, and CI setup.

Stacked on #644 (`feat/support-dynamic-acp-session-configuration-INT-1501`) — rebase onto `main` after that merges.

Fixes INT-1502.

## Test plan
- [x] `uv run pytest tests/adapters/test_omp_acp_adapter.py tests/integrations/test_omp.py tests/integrations/acp/ tests/baseline/ tests/framework_conformance/test_adapter_conformance.py tests/test_lazy_exports.py --no-cov`
- [x] `uv run --all-packages pyrefly check`
- [x] `uv run pytest tests/ --ignore=tests/integration/ --ignore=tests/e2e/ -q --no-cov`
- [ ] Live backends lane with Bun ≥ 1.3.14 + OMP **18.2.8** + provider key for `OMP_MODEL`:
  `E2E_TESTS_ENABLED=true BAND_E2E_LANE=backends uv run pytest tests/e2e/baseline/ -v -s --no-cov`
  (local run hit a turn timeout on host `omp/17.2.9`; construction cell passed)

Made with [Cursor](https://cursor.com)


src/band/integrations/acp/client_runtime.py
```diff
@@ -576,19 +594,56 @@ async def request_permission(  # type: ignore[override]  # ACP Client uses speci
...
+                session_id=session_id,
+                tool_call=tool_call,
+                narrate_permission=narrate_permission,
+                **kwargs,
+            )
 
         logger.debug("Auto-cancelling permission request for session %s", session_id)
         return cancel_permission()
 
+    async def create_elicitation(  # type: ignore[override]
```
AlexanderZ-Band, line 630: issue: `create_elicitation` invokes the handler with no session-lock / `narrate_*` wrapper, unlike `request_permission` which serializes denied narration via `narrate_permission` under `_session_lock`. OMP denial then awaits `emitter.open_permission` directly, so a concurrent `session_update` sink post can overlap the synthetic denied tool_call/tool_result pair. Proven: concurrent session_update + deny elicitation → elicitation_peak=2 vs permission_peak=1.

Mirror `request_permission`: pass a lock-holding narrate callback into elicitation handlers (or await denial narration under `_session_lock` inside `create_elicitation`).

src/band/adapters/omp_acp.py
```diff
@@ -0,0 +1,227 @@
...
+            custom_section=config.custom_section,
+            resolve_session_config=config.resolve_session_config,
+            resolve_permission=config.resolve_permission,
+            client_capabilities=_OMP_FORM_CAPABILITIES,
+            use_unstable_protocol=True,
+            spawn_process=spawn_process,
+            **features,
+        )
+
+    def _build_runtime(self, spawn_process: SpawnProcess | None) -> ACPRuntime:
```
AlexanderZ-Band: issue: `OmpACPAdapter._build_runtime` re-copies the entire parent `ACPClientAdapter._build_runtime` body solely to swap `client_factory`, and reaches into module-private `_resolve_launcher` via a deferred import. Any future runtime kwarg has to be hand-synced again; Copilot never needed this pattern.

Add an overridable client-factory hook on `ACPClientAdapter` that base `_build_runtime` already calls; override only that hook and delete the duplicated method.

src/band/adapters/omp_acp.py
```diff
@@ -0,0 +1,227 @@
...
+            field = approve_deny_form_field(requested_schema)
+            assert field is not None
+            synthetic_call = ACPToolCall(
+                tool_call_id=omp_elicitation_call_id(session_id),
+                name="omp_approval_form",
+                arguments={"message": message},
+            )
+            options = (
+                PermissionOption(
+                    optionId="omp-approve",
```
AlexanderZ-Band: issue: Permission/elicitation wire ids `"omp-approve"`, `"omp-deny"`, and synthetic tool name `"omp_approval_form"` are magic strings (also compared at line 190; echoed in tests). Form labels already live as `OMP_FORM_APPROVE` / `OMP_FORM_DENY`, but the optionIds that must match them do not.

Add `OMP_APPROVE_OPTION_ID` / `OMP_DENY_OPTION_ID` / `OMP_APPROVAL_FORM_TOOL_NAME` in `band.integrations.omp` and reference them from adapter + tests.

src/band/integrations/omp.py
```diff
@@ -0,0 +1,225 @@
...
+import re
+from collections.abc import Collection, Mapping, Sequence
+from uuid import uuid4
+
+from band.runtime.tools import BAND_MCP_SERVER_NAME, canonicalize_mcp_tool_name
+
+DEFAULT_OMP_ACP_COMMAND: tuple[str, ...] = (
+    "omp",
+    "acp",
+    "--approval-mode",
```
AlexanderZ-Band: issue: `DEFAULT_OMP_ACP_COMMAND` and `OMP_UNSAFE_APPROVAL_FLAGS` retype `--approval-mode` / `always-ask` / `--yolo` / `--auto-approve` as bare literals even though `OMP_APPROVAL_MODE_FLAG` / `OMP_APPROVAL_MODE_ALWAYS_ASK` exist nearby — and `validate_omp_command` hardcodes another bare-flag tuple. Default spawn, finalize append, and unsafe rejection can drift apart.

Build default/unsafe joined forms from the shared constants.

src/band/integrations/omp.py
```diff
@@ -0,0 +1,225 @@
...
+        "--approval-mode=yolo",
+    }
+)
+
+XD_URL_PREFIX = "xd://"
+
+OMP_FORM_APPROVE = "Approve"
+OMP_FORM_DENY = "Deny"
+
+OMP_PINNED_PACKAGE = "@oh-my-pi/pi-coding-agent@18.2.8"
```
AlexanderZ-Band, line 50: issue: `OMP_PINNED_PACKAGE` is defined but never referenced; `.github/scripts/setup-omp.sh` and `docs/acp.md` retype `@oh-my-pi/pi-coding-agent@18.2.8` independently. Bumping the Python constant does nothing to CI/docs.

Assert the script default against this constant (or drive the script from it), or drop the unused constant and keep one pin site.

src/band/adapters/omp_acp.py
```diff
@@ -0,0 +1,227 @@
...
+            return DeclineElicitationResponse(action="decline")
+
+        return handler
+
+    @staticmethod
+    def _elicitation_scope(
+        mode: object, kwargs: dict[str, object]
+    ) -> tuple[str, object | None]:
+        """Resolve session id + schema from ACP ``mode`` or test kwargs."""
+        session_id = str(
```
AlexanderZ-Band: issue: `_elicitation_scope` reimplements the same `mode.session_id` / `session_id` / `sessionId` fallback chain already in `ACPCollectingClient.create_elicitation`, plus a parallel schema alias dance. A new ACP wire alias learned in one place will not update the other.

Extract one shared helper, or pass `session_id` into `_make_elicitation_handler` from `_install_turn_handlers` (which already has it).

src/band/integrations/acp/client_adapter.py
```diff
@@ -400,6 +430,29 @@ async def on_message(
...
+        *,
+        emitter: RoomTurnEmitter,
+        room_id: str,
+        session_id: str,
+    ) -> None:
+        self._runtime.set_permission_handler(
+            session_id,
+            self._make_permission_handler(emitter, room_id),
+        )
+        elicitation_handler = self._make_elicitation_handler(emitter, room_id)
```
AlexanderZ-Band: nit: `_install_turn_handlers` already has the turn `session_id` but calls `_make_elicitation_handler(emitter, room_id)` without it, so `OmpACPAdapter` re-derives the same id by copying `create_elicitation`'s getattr chain. Pass `session_id` into the install hook and close over it.

tests/e2e/baseline/toolkit/builders.py
```diff
@@ -542,6 +542,73 @@ def _build_copilot_acp(
...
+        DEFAULT_OMP_MODEL,
+        omp_model_provider,
+        omp_provider_api_key_env,
+        omp_provider_env,
+    )
+
+    model = s.backends.omp_model.strip() or DEFAULT_OMP_MODEL
+    env_key = omp_provider_api_key_env(omp_model_provider(model))
+    creds = s.llm_credentials
+    api_key = {
```
AlexanderZ-Band: issue: The ANTHROPIC/OPENAI/GEMINI → credentials mapping is duplicated verbatim in `deps.py` `_omp_provider_api_key`. Adding a provider updates spawn env or the Dep.OMP gate but not both.

Extract one shared helper used by both `omp_acp_env` and `_omp_provider_api_key`.

src/band/integrations/acp/client_adapter.py
```diff
@@ -400,6 +430,29 @@ async def on_message(
...
+        if elicitation_handler is not None:
+            self._runtime.set_elicitation_handler(session_id, elicitation_handler)
```
AlexanderZ-Band: issue: `_install_turn_handlers` now registers an elicitation handler, but no test asserts that path — `test_permission_handler_wired_on_message` only checks `_permission_handlers`. Dropping `set_elicitation_handler` would still pass the suite.

Add an on_message wiring test for `OmpACPAdapter` that asserts the runtime client has an elicitation handler for the session after the turn starts.

src/band/integrations/acp/client_runtime.py
```diff
@@ -576,19 +594,56 @@ async def request_permission(  # type: ignore[override]  # ACP Client uses speci
...
+        # same room session that registered them.
+        session_id = str(
+            getattr(mode, "session_id", None)
+            or kwargs.get("session_id")
+            or kwargs.get("sessionId")
+            or ""
+        )
+        handler = self._elicitation_handlers.get(session_id)
+        if handler is not None:
+            return await handler(message=message, mode=mode, **kwargs)
```
AlexanderZ-Band: issue: `create_elicitation` / `set_elicitation_handler` / `reset_session` clearing of `_elicitation_handlers` have almost no unit coverage beyond the OMP adapter happy-path with a pre-registered handler. Auto-decline when no handler is set, clearing on `reset_session`, and `ACPRuntime.set_elicitation_handler` delegation are untested.

Add client_runtime tests mirroring the existing permission-handler coverage.

src/band/integrations/acp/client_runtime.py
```diff
@@ -732,13 +799,16 @@ async def start(self, *, respawn: bool = False) -> None:
...
+                **spawn_kwargs,
             ),
         )
         self._ctx = ctx
         try:
             self._conn, _ = await ctx.__aenter__()
-            init_response = await self._conn.initialize(protocol_version=1)
+            init_kwargs: dict[str, Any] = {"protocol_version": 1}
+            if self._client_capabilities is not None:
+                init_kwargs["client_capabilities"] = self._client_capabilities
```
AlexanderZ-Band, line 830: issue: Form-elicitation advertising depends on `client_capabilities` reaching `initialize`, and OMP needs `use_unstable_protocol` on the builtin spawn. Construction tests only assert stored fields; nothing asserts `initialize(..., client_capabilities=...)` or `use_unstable_protocol=True` in spawn kwargs.

Extend `ACPRuntime.start` tests for those two wire paths.

tests/integrations/acp/test_client_adapter.py
```diff
@@ -884,6 +897,48 @@ async def test_permission_handler_wired_on_message(
         # Permission handler should have been set for this session
         assert len(adapter_with_mocks._runtime._client._permission_handlers) > 0
 
+    @pytest.mark.asyncio
+    async def test_permission_resolver_receives_only_advertised_choices(self) -> None:
```
AlexanderZ-Band, line 1025: issue: New PermissionResolver coverage only calls private `_resolve_permission_option` directly. A handler that stops consulting `_resolve_permission` would still pass these two tests.

Add a test that installs `resolve_permission`, fires `request_permission` through the wired handler, and asserts the advertised option id / cancel+narration outcome.

tests/e2e/baseline/smoke/adapters/test_omp_acp.py
```diff
@@ -0,0 +1,52 @@
...
+from tests.e2e.baseline.toolkit.user_ops import UserOps
+
+BAND_EVENT_TOOL_NAME = "band_send_event"
+
+
+@with_adapters(Adapter.OMP_ACP, **TOOL_AGENT)
+@flaky_model("OMP may occasionally miss an explicit tool-only request")
+@pytest.mark.timeout(extra=180)
+@pytest.mark.asyncio(loop_scope="session")
+async def test_omp_acp_band_tool_call_is_narrated(
```
AlexanderZ-Band, line 41: issue: Plan acceptance centers on always-ask + PermissionResolver/form elicitation, but the only backends-lane live smoke checks Band tool narration. No e2e covers an OMP approval form reaching the room (approve or deny) with a real resolver.

Add a backends smoke that supplies `resolve_permission`, prompts a gated tool that triggers OMP's Approve/Deny form, and asserts permission narration / declined outcome.

tests/adapters/test_omp_acp_adapter.py
```diff
@@ -0,0 +1,279 @@
...
+)
+from band.integrations.acp.client_adapter import ACPClientAdapter, ACPPermissionRequest
+from band.integrations.acp.client_runtime import ACPCollectingClient
+from band.integrations.acp.session_config import ACPConfigRequest
+from band.integrations.omp import DEFAULT_OMP_ACP_COMMAND as INTEGRATION_DEFAULT
+
+
+class TestOmpACPAdapterConstruction:
+    def test_is_acp_client_adapter(self) -> None:
+        assert issubclass(OmpACPAdapter, ACPClientAdapter)
```
AlexanderZ-Band: nit: `assert issubclass(OmpACPAdapter, ACPClientAdapter)` restates the class declaration and cannot fail for a behavioral reason. Delete it; keep construction/elicitation tests.

tests/adapters/test_omp_acp_adapter.py
```diff
@@ -0,0 +1,279 @@
...
+        assert chunks[0].tool is not None
+        assert chunks[0].tool.name == "band_send_message"
+
+        text_update = MagicMock()
+        text_update.session_update = "agent_message_chunk"
+        text_update.content = MagicMock(text="hello from omp")
+        await client.session_update("sess", text_update)
+        await client.flush("sess")
+        assert client.get_collected_text("sess") == "hello from omp"
+        assert isinstance(client, ACPCollectingClient)
```
AlexanderZ-Band: nit: `assert isinstance(client, ACPCollectingClient)` is tautological — `client` was constructed as `OmpACPCollectingClient`. Remove it; keep the tool-name/collected-text checks.

tests/e2e/baseline/toolkit/deps.py
```diff
@@ -274,6 +356,11 @@ def _letta_available(settings: BaselineSettings) -> bool:
     Dep.COPILOT_CLI: DepSpec(
         _copilot_cli_available, "Copilot CLI not found on PATH", lane=Lane.BACKENDS
     ),
+    Dep.OMP: DepSpec(
+        _omp_available,
+        "Bun >= 1.3.14, working `omp`/`omp acp`, and a provider key for OMP_MODEL not set",
```
AlexanderZ-Band: nit: `Dep.OMP`'s skip reason hardcodes `Bun >= 1.3.14` while this module already imports `OMP_MIN_BUN` for the real gate. Interpolate `OMP_MIN_BUN` into the reason string.

src/band/integrations/omp.py
```diff
@@ -0,0 +1,225 @@
...
+    tail = path[len(XD_URL_PREFIX) :]
+    if not tail.startswith("mcp__"):
+        return None
+    remainder = tail.removeprefix("mcp__")
+    if "__" in remainder:
+        server, tool = remainder.split("__", 1)
+        if server and tool:
+            return f"{server}-{tool}"
+        return None
+    # ``mcp__band_band_send_message`` → ``band-band_send_message``
```
AlexanderZ-Band: nit: The inline comment narrates what the next three lines already do; the docstring already covers the spelling variants. Delete the line comment.

src/band/integrations/acp/client_runtime.py
```diff
@@ -91,6 +95,19 @@ def cancel_permission() -> dict[str, object]:
     return {"outcome": {"outcome": "cancelled"}}
 
 
+async def narrate_denied_elicitation_form(
```
AlexanderZ-Band: suggestion: `narrate_denied_elicitation_form` is a one-caller duck-typed wrapper around `open_permission(..., outcome="cancelled")` and lives in the generic ACP runtime beside transport plumbing. Inline `await emitter.open_permission(...)` in `OmpACPAdapter` (mirroring the permission handler), or type it as `RoomTurnEmitter` if it is meant to stay shared.

AlexanderZ-Band (commented):
Cycle 1 review — see inline findings (elicitation lock + test gaps + DRY/design). Cannot REQUEST_CHANGES on own PR; treating blocking issues as must-fix.

src/band/integrations/acp/client_runtime.py
AlexanderZ-Band: Fixed in 96b5ba85: `create_elicitation` now passes a lock-holding `narrate_elicitation` callback (mirrors `request_permission`); OMP denial runs `open_permission` through it. Concurrency test covers the path.

src/band/adapters/omp_acp.py
AlexanderZ-Band: Fixed in 96b5ba85: deleted the override; base `_build_runtime` uses overridable `_runtime_client_factory`, which OMP specializes.

src/band/adapters/omp_acp.py
AlexanderZ-Band: Fixed in 96b5ba85: `OMP_APPROVE_OPTION_ID` / `OMP_DENY_OPTION_ID` / `OMP_APPROVAL_FORM_TOOL_NAME` in `band.integrations.omp`; adapter + tests reference them.

src/band/integrations/omp.py
AlexanderZ-Band: Fixed in 96b5ba85: default command and unsafe flag set are built from shared `OMP_APPROVAL_MODE_*` / bare-flag constants; `validate_omp_command` uses that set.

src/band/integrations/omp.py
AlexanderZ-Band: Fixed in 96b5ba85: unit test asserts `.github/scripts/setup-omp.sh` contains `OMP_PINNED_PACKAGE`.

src/band/adapters/omp_acp.py
AlexanderZ-Band: Fixed in 96b5ba85: factory receives `session_id`; shared `elicitation_requested_schema` / `elicitation_session_id` helpers replace the duplicated chain.

src/band/integrations/acp/client_adapter.py
AlexanderZ-Band: Fixed in 96b5ba85: `_make_elicitation_handler` now takes `session_id` from `_install_turn_handlers`.

tests/e2e/baseline/toolkit/builders.py
AlexanderZ-Band: Fixed in 96b5ba85: shared `tests/e2e/baseline/toolkit/omp_credentials.py` used by both `omp_acp_env` and Dep.OMP.

src/band/integrations/acp/client_adapter.py
AlexanderZ-Band: Fixed in 96b5ba85: added `test_elicitation_handler_wired_on_message` for OmpACPAdapter.

src/band/integrations/acp/client_runtime.py
AlexanderZ-Band: Fixed in 96b5ba85: added decline-when-unset, set+reset clear, and ACPRuntime forward tests.

src/band/integrations/acp/client_runtime.py
AlexanderZ-Band: Fixed in 96b5ba85: ACPRuntime.start tests assert `initialize(..., client_capabilities=...)` and `use_unstable_protocol` in spawn kwargs.

tests/integrations/acp/test_client_adapter.py
AlexanderZ-Band: Fixed in 96b5ba85: `test_permission_resolver_deny_cancels_via_request_permission` exercises the wired handler path.

tests/e2e/baseline/smoke/adapters/test_omp_acp.py
AlexanderZ-Band: Fixed in 96b5ba85: added `test_omp_acp_form_elicitation_denied_is_narrated` (deny resolver + permission narration).

tests/adapters/test_omp_acp_adapter.py
AlexanderZ-Band: Fixed in 96b5ba85: removed the tautological issubclass assertion.

tests/adapters/test_omp_acp_adapter.py
AlexanderZ-Band: Fixed in 96b5ba85: removed the tautological isinstance assertion.

tests/e2e/baseline/toolkit/deps.py
AlexanderZ-Band: Fixed in 96b5ba85: Dep.OMP skip reason interpolates `OMP_MIN_BUN`.

src/band/integrations/omp.py
AlexanderZ-Band: Fixed in 96b5ba85: removed the narrating comment.

src/band/integrations/acp/client_runtime.py
AlexanderZ-Band: Fixed in 96b5ba85: helper deleted; OMP inlines `emitter.open_permission(...)` under `narrate_elicitation`.

src/band/integrations/omp.py
```diff
@@ -0,0 +1,240 @@
...
+        return None
+    remainder = tail.removeprefix("mcp__")
+    if "__" in remainder:
+        server, tool = remainder.split("__", 1)
+        if server and tool:
+            return f"{server}-{tool}"
+        return None
+    prefix = f"{BAND_MCP_SERVER_NAME}_"
+    if remainder.startswith(prefix) and remainder != prefix:
+        return f"{BAND_MCP_SERVER_NAME}-{remainder.removeprefix(prefix)}"
```
AlexanderZ-Band: issue: OMP 18.2.8 `createMCPToolName('band','band_send_message')` strips the redundant `band_` tool prefix and mints `mcp__band_send_message` (not `mcp__band_band_send_message`). For `xd://mcp__band_send_message`, `_xd_mcp_wire_tool_name` returns `band-send_message`; `canonicalize_mcp_tool_name` does not map that into `own_names`, so `normalize_omp_mcp_device_call` leaves name=`write`. Unit tests only cover the non-minted spellings.

Invert the pinned mint's redundant-prefix strip (or accept a remainder already in `own_names`) before the current `band_`→suffix split.

src/band/adapters/omp_acp.py
```diff
@@ -0,0 +1,207 @@
...
+        canonicalize_tool_name: Callable[[str], str] | None = None,
+    ) -> None:
+        super().__init__(canonicalize_tool_name=canonicalize_tool_name)
+        self._own_tool_names = own_tool_names
+
+    def _tool_call_chunk(self, update: object):
+        chunk = super()._tool_call_chunk(update)
+        if chunk.tool is None or not isinstance(chunk.tool, ACPToolCall):
+            return chunk
+        name, args = normalize_omp_mcp_device_call(
```
AlexanderZ-Band, line 73: issue: `OmpACPCollectingClient._tool_call_chunk` only rewrites via `normalize_omp_mcp_device_call`, which no-ops unless args carry an `xd://` path. A direct ACP tool_call title `mcp__band_send_message` stays as `mcp__*`; `is_room_posting_tool` returns False → room text-fallback suppression misses (duplicate room delivery risk).

Peel/canonicalize `mcp__` titles the same way as device paths so registered Band tools map into `own_names`.

tests/e2e/baseline/smoke/adapters/test_omp_acp.py
```diff
@@ -0,0 +1,131 @@
...
+            cwd=sandbox,
+            env=omp_acp_env(settings, omp_agent_home_dir(sandbox)),
+            resolve_permission=deny,
+        )
+    )
+
+
+@lane(Lane.BACKENDS)
+@requires(Dep.OMP)
+@flaky_infra("OMP form elicitation depends on the model attempting a gated action")
```
AlexanderZ-Band: issue: `@flaky_infra(...)` misclassifies model nondeterminism as infra — `flaky_infra` sets `rerun_except=["AssertionError"]`, so a model that never attempts the gated tool fails once with no retry. Use `@flaky_model(...)` so AssertionError is retried.

tests/e2e/baseline/smoke/adapters/test_omp_acp.py
```diff
@@ -0,0 +1,131 @@
...
+                "You must attempt the delete; do not only describe it.",
+                mention_id=agent.id,
+                mention_name=agent.name,
+            )
+            await capture.wait_for_processed(
+                mid, agent.id, deadline_s=BUDGET.deadline_s
+            )
+            tool_calls = await capture.events(MessageType.TOOL_CALL, sender_id=agent.id)
+
+        tool_calls.assert_contains_any([OMP_APPROVAL_FORM_TOOL_NAME])
```
AlexanderZ-Band, line 132: issue: Form-denial e2e only asserts a TOOL_CALL mentioning `omp_approval_form`. Denial narration is the cancelled pair (`tool_call` + `tool_result` with `outcome="cancelled"` / Permission cancelled). Assert a matching cancelled TOOL_RESULT as well.

tests/adapters/test_omp_acp_adapter.py
```diff
@@ -0,0 +1,307 @@
...
+        handler = adapter._make_elicitation_handler(MagicMock(), "room-1", "sess-1")
+        response = await handler(
+            message="?",
+            mode="form",
+            requested_schema={"properties": {}},
+        )
+        assert isinstance(response, DeclineElicitationResponse)
+
+    @pytest.mark.asyncio
+    async def test_denial_uses_unique_elicitation_ids(self) -> None:
```
AlexanderZ-Band, line 263: issue: OMP form denial is only exercised by calling `_make_elicitation_handler` directly (hits the `narrate_elicitation is None` branch). Live ACP goes through `create_elicitation` which injects the locked narrator — that branch is never unit-tested for deny. Add a deny analogue of the approve `create_elicitation` test.

tests/adapters/test_omp_acp_adapter.py
```diff
@@ -0,0 +1,307 @@
...
+            mode="form",
+            requested_schema=schema,
+        )
+        assert isinstance(first, DeclineElicitationResponse)
+        assert isinstance(second, DeclineElicitationResponse)
+        calls = emitter.open_permission.await_args_list
+        assert (
+            calls[0].kwargs["call"].tool_call_id != calls[1].kwargs["call"].tool_call_id
+        )
+        assert calls[0].kwargs["call"].name == OMP_APPROVAL_FORM_TOOL_NAME
```
AlexanderZ-Band, line 290: issue: `test_denial_uses_unique_elicitation_ids` never asserts `outcome="cancelled"` on `open_permission`. Add that assert on both calls.

tests/adapters/test_omp_acp_adapter.py
```diff
@@ -0,0 +1,307 @@
...
+class TestOmpACPAdapterConstruction:
+    def test_stdio_only_no_tcp_fields(self) -> None:
+        adapter = OmpACPAdapter()
+        assert adapter._host is None
+        assert adapter._port is None
+
+    def test_default_command_gets_final_always_ask(self) -> None:
+        adapter = OmpACPAdapter()
+        assert adapter._command[-2:] == ["--approval-mode", "always-ask"]
+        assert list(INTEGRATION_DEFAULT) == list(DEFAULT_OMP_ACP_COMMAND)
```
AlexanderZ-Band: issue: `assert list(INTEGRATION_DEFAULT) == list(DEFAULT_OMP_ACP_COMMAND)` is a re-export identity check and cannot fail for a behavioral reason. Delete it; keep the trailing always-ask command assert.

src/band/integrations/omp.py
```diff
@@ -0,0 +1,240 @@
...
+    """Validate ``command`` and append a final ``always-ask`` override."""
+    validate_omp_command(command)
+    finalized = list(command)
+    finalized.extend((OMP_APPROVAL_MODE_FLAG, OMP_APPROVAL_MODE_ALWAYS_ASK))
+    return finalized
+
+
+def omp_elicitation_call_id(session_id: str) -> str:
+    """Stable synthetic tool-call id namespace for declined OMP forms."""
+    return f"omp-elicitation:{session_id}:{uuid4()}"
```
AlexanderZ-Band: nit: `omp-elicitation:` is a bare f-string prefix also asserted in tests. Give it one constant (e.g. `OMP_ELICITATION_CALL_ID_PREFIX`) shared by producer and assertions.

src/band/integrations/omp.py
```diff
@@ -0,0 +1,240 @@
...
+    if not isinstance(field_name, str):
+        return False
+    field_map = _schema_as_mapping(field_schema)
+    if field_map is None:
+        return False
+    enum = field_map.get("enum")
+    return enum == [OMP_FORM_APPROVE, OMP_FORM_DENY]
+
+
+def approve_deny_form_field(requested_schema: object) -> str | None:
```
AlexanderZ-Band: nit: `approve_deny_form_field` re-runs the full schema walk that `is_omp_approve_deny_form` just did. Collapse to one parser returning the field name or None.

src/band/integrations/omp.py
```diff
@@ -0,0 +1,240 @@
...
+
+    OMP registers MCP tools as ``mcp__<server>_<tool>`` (single underscore after
+    the server) and may also surface the double-underscore ``mcp__server__tool``
+    spelling. Both map to the hyphen-joined form ``canonicalize_mcp_tool_name``
+    already understands for the Band loopback server.
+    """
+    if not path.startswith(XD_URL_PREFIX):
+        return None
+    tail = path[len(XD_URL_PREFIX) :]
+    if not tail.startswith("mcp__"):
```
AlexanderZ-Band: nit: After centralizing `XD_URL_PREFIX`, the MCP marker `"mcp__"` is still a bare literal at startswith/removeprefix. Add one named constant beside `XD_URL_PREFIX`.

src/band/integrations/acp/client_runtime.py
```diff
@@ -576,19 +603,61 @@ async def request_permission(  # type: ignore[override]  # ACP Client uses speci
...
-            # keeps the pair atomic between narration posts.
-            async with self._session_lock(session_id):
-                return await handler(
-                    options=options,
-                    session_id=session_id,
-                    tool_call=tool_call,
-                    **kwargs,
-                )
+
+            async def narrate_permission(action: Awaitable[None]) -> None:
```
AlexanderZ-Band: nit: `narrate_permission` and `narrate_elicitation` are identical session-lock wrappers (and the Narrator type aliases match). One shared `_session_narrator(session_id)` helper keeps the lock gate in a single place.

tests/adapters/test_omp_acp_adapter.py
```diff
@@ -0,0 +1,307 @@
...
+
+class TestOmpACPAdapterConstruction:
+    def test_stdio_only_no_tcp_fields(self) -> None:
+        adapter = OmpACPAdapter()
+        assert adapter._host is None
+        assert adapter._port is None
+
+    def test_default_command_gets_final_always_ask(self) -> None:
+        adapter = OmpACPAdapter()
+        assert adapter._command[-2:] == ["--approval-mode", "always-ask"]
```
AlexanderZ-Band: nit: Assertions retype `--approval-mode` / `always-ask` / Approve/Deny even though `OMP_APPROVAL_MODE_*` / `OMP_FORM_*` exist. Import and assert against those constants.

AlexanderZ-Band (commented):
Cycle 2 review — see inline findings.

src/band/integrations/omp.py
AlexanderZ-Band: Fixed in follow-up commit: shared `normalize_omp_mcp_tool_name` accepts OMP 18.2.8 mint remainder `band_send_message` so `xd://mcp__band_send_message` rewrites to `band_send_message`.

src/band/adapters/omp_acp.py
AlexanderZ-Band: Fixed: `OmpACPCollectingClient` canonicalizes `mcp__` titles through the shared normalizer.

tests/e2e/baseline/smoke/adapters/test_omp_acp.py
AlexanderZ-Band: Fixed: switched to `@flaky_model` so AssertionError is retried.

tests/e2e/baseline/smoke/adapters/test_omp_acp.py
AlexanderZ-Band: Fixed: e2e also asserts cancelled form TOOL_RESULT / permission_outcome.

tests/adapters/test_omp_acp_adapter.py
AlexanderZ-Band: Fixed: added `test_denial_via_create_elicitation_narrates_cancelled`.

tests/adapters/test_omp_acp_adapter.py
AlexanderZ-Band: Fixed: both open_permission calls assert outcome cancelled.

tests/adapters/test_omp_acp_adapter.py
AlexanderZ-Band: Fixed: removed the re-export identity assert.

src/band/integrations/omp.py
AlexanderZ-Band: Fixed: `OMP_ELICITATION_CALL_ID_PREFIX` shared by producer and tests.

src/band/integrations/omp.py
AlexanderZ-Band: Fixed: single parser; `is_omp_approve_deny_form` delegates.

src/band/integrations/omp.py
AlexanderZ-Band: Fixed: `XD_MCP_PREFIX` constant.

src/band/integrations/acp/client_runtime.py
AlexanderZ-Band: Fixed: shared `_session_narrator(session_id)` helper.

tests/adapters/test_omp_acp_adapter.py
AlexanderZ-Band: Fixed: adapter tests import/assert `OMP_*` constants.

src/band/adapters/omp_acp.py
```diff
@@ -0,0 +1,207 @@
...
+        *,
+        additional_tools: list[CustomToolDef] | None = None,
+        spawn_process: SpawnProcess | None = None,
+        **features: Unpack[FeatureKwargs],
+    ) -> None:
+        config = config or OmpACPAdapterConfig()
+        super().__init__(
+            command=finalize_omp_command(config.command),
+            env=config.env,
+            cwd=config.cwd,
```
amit-gazal-band: **Bug (high): `ACPClientAdapter.__init__` now rejects any non-`None` `cwd`.** Since #629 it raises `ValueError("cwd is not supported; use workspace_for_room or the default")`, and `OmpACPAdapterConfig` has no `workspace_for_room` field, so there is no working way to set the workspace.

Reproduced:
```
OmpACPAdapter(config=OmpACPAdapterConfig())          # builds fine
OmpACPAdapter(config=OmpACPAdapterConfig(cwd=tmp))   # ValueError: cwd is not supported; use workspace_for_room or the default
```
This breaks `examples/acp/clients/omp.py` (`cwd=settings.acp_agent_cwd`), `_build_omp_acp` in the baseline builders (`"cwd": sandbox`, so every OMP matrix cell fails), and `_denying_omp_adapter` in `test_omp_acp.py`.

Suggested fix: do what `CopilotACPAdapter` does. Map `cwd` to `create_room_workspace_resolver(cwd)`, add a `workspace_for_room` field to the config, and add a unit test that builds the adapter with `cwd`.

tests/e2e/baseline/toolkit/deps.py
```diff
@@ -189,6 +194,64 @@ def _codex_cli_available(settings: BaselineSettings) -> bool:
...
+    binary = (
+        settings.backends.omp_command.split()[0]
+        if settings.backends.omp_command.strip()
+        else "omp"
+    )
+    if shutil.which(binary) is None:
+        return False
+    command = settings.backends.omp_command.strip() or "omp"
+    base = command.split()
+    for args in (["--version"], ["acp", "--help"]):
```
amit-gazal-band: **Bug (medium): the dependency check and the builder read `OMP_COMMAND` differently.** Here it's treated as the bare binary and `acp --help` is appended. `_build_omp_acp` uses it as the full spawn command (`tuple(omp_command.split())`), which matches the settings comment "override the binary + args".

Reproduced with `subprocess.run` mocked:
- `OMP_COMMAND="omp acp"`: the check runs `omp acp acp --help`, which fails, so every OMP test is skipped without a clear reason.
- `OMP_COMMAND="omp"`: the check passes, but the adapter spawns `['omp', '--approval-mode', 'always-ask']` without the `acp` subcommand.

Pick one meaning and use it in both places. Treating it as the full command is probably the right one, with the check running the command with `--help` instead of adding the `acp` subcommand again.

tests/e2e/baseline/toolkit/omp_credentials.py
```diff
@@ -0,0 +1,26 @@
...
+from band.integrations.omp import (
+    DEFAULT_OMP_MODEL,
+    omp_model_provider,
+    omp_provider_api_key_env,
+)
+from tests.e2e.baseline.settings import BaselineSettings
+
+
+def omp_provider_api_key(settings: BaselineSettings) -> str:
+    """The provider API key ``OMP_MODEL`` needs, or ``""`` when unset/unsupported."""
```
amit-gazal-band: **Docstring doesn't match behavior (low).** `omp_model_provider` and `omp_provider_api_key_env` raise for a malformed or unsupported model, so the `case _: return ""` branch is never reached for those inputs:
```
OMP_MODEL="bedrock/x"         -> ValueError: Unsupported OMP model provider 'bedrock'; supported: ...
OMP_MODEL="gemini-2.5-flash"  -> ValueError: OMP model must look like "provider/model", got "gemini-2.5-flash"
```
`_omp_available` calls this during `Dep.OMP` evaluation, after the Bun and CLI checks. So on a machine that has OMP installed, a bad `OMP_MODEL` crashes the dependency check instead of skipping the test with the Dep reason. Catch `ValueError` here and return `""`, or fix the docstring.

src/band/integrations/omp.py
```diff
@@ -0,0 +1,255 @@
...
+    tokens = list(command)
+    index = 0
+    while index < len(tokens):
+        token = tokens[index]
+        if token in OMP_BARE_UNSAFE_APPROVAL_FLAGS:
+            raise ValueError(f"Unsafe OMP flag {token!r} is not allowed")
+        if token == OMP_APPROVAL_MODE_FLAG:
+            mode = tokens[index + 1] if index + 1 < len(tokens) else ""
+            joined = f"{OMP_APPROVAL_MODE_FLAG} {mode}".strip()
+            if joined in OMP_UNSAFE_APPROVAL_FLAGS:
```
amit-gazal-band: **Safety check can be bypassed by changing the case (low).** The lookup in `OMP_UNSAFE_APPROVAL_FLAGS` is case-sensitive, both here and in the `=` branch at line 123, even though `_APPROVAL_MODE_EQ` itself matches case-insensitively.

Reproduced:
```
--approval-mode yolo    -> rejected
--approval-mode=write   -> rejected
--approval-mode YOLO    -> ACCEPTED -> [..., '--approval-mode', 'YOLO', '--approval-mode', 'always-ask']
--approval-mode=Write   -> ACCEPTED
```
Safety then depends entirely on OMP honouring the trailing `always-ask` (last flag wins). If OMP ever reads flag values case-insensitively or takes the first occurrence, the "reject unsafe modes" guarantee no longer holds. Lowercasing `mode` before both lookups fixes it.

amit-gazal-band (changes requested):
Requesting changes: the adapter throws as soon as it's constructed with `cwd`, and that breaks the example and every OMP e2e cell. There are also three smaller issues, noted inline.

I reproduced all four against `bd667dd` with a script that runs the real code. The PR's 370 OMP unit tests pass, but none of them builds the adapter with `cwd`, so the first bug isn't caught by any test that CI runs.

src/band/adapters/omp_acp.py
AlexanderZ-Band: Fixed in 56956a0: mirrored `CopilotACPAdapter`'s `cwd` -> `workspace_for_room` pattern. `OmpACPAdapterConfig` gained a `workspace_for_room` field; the adapter maps `cwd` through `create_room_workspace_resolver(cwd)` (raising if both are set) instead of passing `cwd=` straight through to `ACPClientAdapter.__init__`. Added `test_cwd_becomes_a_room_workspace_root` and `test_cwd_and_workspace_for_room_together_is_rejected`.

tests/e2e/baseline/toolkit/deps.py
AlexanderZ-Band: Fixed in 56956a0: `_omp_cli_responds` now treats `OMP_COMMAND` as the full base command (matching `_build_omp_acp` and the settings comment), defaulting to `["omp", "acp"]` when unset, and probes it once with `--help` instead of unconditionally appending `acp --help` on top. Added `test_omp_cli_responds_defaults_base_to_omp_acp` and `test_omp_cli_responds_treats_override_as_full_base_command` (mocking `shutil.which`/`subprocess.run`) asserting the exact probe command in both the default and override case.

tests/e2e/baseline/toolkit/omp_credentials.py
AlexanderZ-Band: Fixed in 56956a0: `omp_provider_api_key` now catches the `ValueError` from `omp_model_provider`/`omp_provider_api_key_env` and returns `""`, matching what the docstring already promised, so a malformed/unsupported `OMP_MODEL` reports `Dep.OMP` as unavailable instead of crashing the dependency check. Added `test_omp_provider_api_key_returns_empty_for_unsupported_provider` and `..._for_malformed_model` covering both repro cases.

src/band/integrations/omp.py
AlexanderZ-Band: Fixed in 56956a0: lowercased `mode` before both the space-form and `=`-form membership checks in `validate_omp_command`, so it's consistently case-insensitive like `_APPROVAL_MODE_EQ`'s own regex already is. Extended `test_validate_rejects_unsafe_flags` with `--approval-mode YOLO`/`Write` and `--approval-mode=YOLO`/`=Write` cases; confirmed they fail against the pre-fix code and pass against the fix.

tests/e2e/baseline/settings.py
```diff
@@ -164,6 +164,11 @@ class Backends(BaseSettings):
     # BYOK (see toolkit/builders.py copilot_acp_env), like the copilot_sdk builder.
     copilot_command: str = ""  # COPILOT_COMMAND (override the `copilot` binary + args)
 
+    # OMP (oh-my-pi) over ACP (`omp_acp` adapter). Defaults to `omp acp` with
+    # provider-qualified OMP_MODEL; override the binary + args via OMP_COMMAND.
+    omp_command: str = ""  # OMP_COMMAND
+    omp_model: str = "google/gemini-2.5-flash"  # OMP_MODEL
```
AlexanderZ-Band: nit: `Backends.omp_model` defaults to the literal `"google/gemini-2.5-flash"`, hand-copying `DEFAULT_OMP_MODEL` from `src/band/integrations/omp.py`. Every caller already implements "empty falls back to `DEFAULT_OMP_MODEL`" (`omp_acp_env` and `omp_provider_api_key` both do `s.backends.omp_model.strip() or DEFAULT_OMP_MODEL`), the same pattern `codex_model: str = ""` uses two fields above. Because this field's default is non-empty, that fallback branch never fires, leaving two independently-typed copies of the same model string that can silently drift.

Fix: default `omp_model` to `""` (matching `codex_model`/`copilot_command`) and let the existing `.strip() or DEFAULT_OMP_MODEL` fallback supply the real default from `band.integrations.omp`.

.github/scripts/setup-omp.sh
```diff
@@ -0,0 +1,36 @@
+#!/usr/bin/env bash
+# Install Bun and pin OMP (@oh-my-pi/pi-coding-agent) for the `backends` e2e lane.
+set -euo pipefail
+
+OMP_PACKAGE="${OMP_PACKAGE:-@oh-my-pi/pi-coding-agent@18.2.8}"
+OMP_MIN_BUN="${OMP_MIN_BUN:-1.3.14}"
```
AlexanderZ-Band: nit: `OMP_MIN_BUN="${OMP_MIN_BUN:-1.3.14}"` duplicates the Python constant `OMP_MIN_BUN = "1.3.14"` (`src/band/integrations/omp.py`). The sibling literal (`OMP_PACKAGE`/`OMP_PINNED_PACKAGE`) is guarded against drift by `tests/integrations/test_omp.py::test_setup_omp_script_pins_same_package`, but no equivalent assertion checks the Bun version default still matches `OMP_MIN_BUN`, so the two can silently diverge.

Fix: extend that test (or add a sibling) to also assert `OMP_MIN_BUN in script`.

tests/e2e/baseline/toolkit/builders.py
```diff
@@ -551,6 +551,67 @@ def _build_copilot_acp(
...
+def omp_agent_home_dir(work_dir: str) -> str:
+    """Create and return the ``pi-coding-agent-home`` subdirectory of ``work_dir``."""
+    home = os.path.join(work_dir, "pi-coding-agent-home")
+    os.makedirs(home, exist_ok=True)
+    return home
+
+
+def omp_acp_env(s: BaselineSettings, agent_home: str) -> dict[str, str]:
+    """Hermetic OMP child env: model, provider key, and isolated agent state dir."""
+    from band.integrations.omp import (  # noqa: PLC0415 -- keep builder imports lazy like sibling adapters
```
AlexanderZ-Band: nit: every other deferred import in this file (~20 of them, e.g. lines 59, 80, 149, 503) uses the single established phrasing `# noqa: PLC0415 -- isolates the <name> extra from the other frameworks this file builds`. The three new OMP imports break that convention: this line uses different wording ("keep builder imports lazy like sibling adapters"), and lines 567 and 590 carry no rationale comment at all.

Fix: match the established phrasing, e.g. `# noqa: PLC0415 -- isolates the omp_acp extra from the other frameworks this file builds`, on all three (563, 567, 590).

tests/framework_conformance/test_toolkit_helpers.py
```diff
@@ -96,6 +98,60 @@ def test_is_letta_cloud_rejects_non_cloud(url: str) -> None:
...
+    ):
+        assert _omp_cli_responds(_settings_with_omp_command("")) is True
+
+    run.assert_called_once()
+    assert run.call_args.args[0] == ["omp", "acp", "--help"]
+
+
+def test_omp_cli_responds_treats_override_as_full_base_command() -> None:
+    # A custom OMP_COMMAND already carries its own subcommand; the probe must not
+    # append "acp" a second time (that produced "omp acp acp --help" before the fix).
```
AlexanderZ-Band: nit: this comment narrates a since-fixed bug ("that produced ... before the fix") rather than stating the current invariant, which this project's comment-style rule disallows (state facts about the code as it is now, never narrate how it got there).

Fix: drop "before the fix" and state the invariant directly, e.g. "...a second time, or the process receives `omp acp acp --help`."

tests/e2e/baseline/toolkit/deps.py
```diff
@@ -189,6 +194,57 @@ def _codex_cli_available(settings: BaselineSettings) -> bool:
...
+            timeout=15,
+        )
+    except (OSError, subprocess.TimeoutExpired):
+        return False
+    if completed.returncode != 0:
+        return False
+    match = re.search(r"(\d+\.\d+\.\d+)", completed.stdout)
+    if not match:
+        return False
+    found = tuple(int(part) for part in match.group(1).split("."))
```
AlexanderZ-Band: suggestion: `_bun_meets_min_version` regexes Bun's `--version` output and hand-compares int tuples to gate `Dep.OMP`. `packaging>=23.0` is already a direct `dev` extra dependency of this project and is already used here for exactly this kind of version-floor comparison (the CI coverage-gate version guard uses `packaging.version.Version`). Since the `backends` lane that gates `Dep.OMP` installs the `dev` extra, `packaging.version.Version` is available and would replace the hand-rolled regex/tuple parsing with the version-comparison tool this repo already relies on elsewhere.

Fix: `from packaging.version import Version` and `Version(match.group(1)) >= Version(min_version)`.

src/band/integrations/acp/client_runtime.py
```diff
@@ -35,6 +37,9 @@
 ACP_STDIO_LIMIT_BYTES = 16 * 1024 * 1024
 ACP_SESSION_LOAD_TIMEOUT_SECONDS = 5.0
 PermissionHandler = Callable[..., Awaitable[dict[str, object]]]
+PermissionNarrator = Callable[[Awaitable[None]], Awaitable[None]]
+ElicitationHandler = Callable[..., Awaitable[object]]
+ElicitationNarrator = Callable[[Awaitable[None]], Awaitable[None]]
```
AlexanderZ-Band: nit: `PermissionNarrator` and `ElicitationNarrator` are both `Callable[[Awaitable[None]], Awaitable[None]]` — identical types, both produced by the same `_session_narrator()` factory; the only difference is which kwarg name (`narrate_permission` vs `narrate_elicitation`) each is bound to. Two separately-declared aliases for one type add a name to keep in sync for no behavioral difference.

Fix: declare one shared `Narrator = Callable[[Awaitable[None]], Awaitable[None]]` alias and reuse it for both handler types' narrate-callback parameter.

src/band/integrations/acp/client_adapter.py
```diff
@@ -512,20 +577,74 @@ async def handler(
...
+        options: object,
+        room_id: str,
+        session_id: str,
+    ) -> str | None:
+        """Return a validated permission choice, or cancel the request."""
+        if self._resolve_permission is None:
+            return select_allow_option_id(options)
+
+        offered = self._permission_options(options)
+        option_id = await self._resolve_permission(
```
AlexanderZ-Band: issue: `_resolve_permission_option` awaits the caller-supplied `resolve_permission` callback here and raises bare `ValueError`s below (lines 617, 619-621) with no try/except and no `logger` call anywhere on this path. `request_permission` is dispatched by the `acp` transport as its own independent task, separate from `on_message`'s `prompt_task`, so neither the resolver's own exceptions nor these validation errors ever reach `on_message`'s `except Exception as e: logger.exception("ACP agent error")`. They're instead caught generically by `acp`'s own dispatch loop, which logs only `"Unhandled error while handling request method=%s"` via the root logger (no room_id/session_id/tool/adapter context) and returns a bare JSON-RPC internal error to the agent. Compare this to the sibling `resolve_session_config` hook: its failures are caught in `_configure_session`, converted to `ACPConfigError`, logged via `logger.warning("ACP session configuration failed: %s", error)`, and reported to the room via `tools.send_failure(...)` (`_report_config_error`). A broken or misconfigured `resolve_permission` callback fails completely silently from Band's own logs and the room's perspective.

Fix: wrap the resolver invocation/validation in try/except here, log with this module's own logger (with room_id/session_id/tool context), and surface it to the room the same way `_report_config_error` does for `resolve_session_config`.

src/band/adapters/omp_acp.py
```diff
@@ -0,0 +1,218 @@
...
+                        kind="reject_once",
+                    ),
+                )
+                option_id = await self._resolve_permission_option(
+                    call=synthetic_call,
+                    options=options,
+                    room_id=room_id,
+                    session_id=session_id,
+                )
+            if option_id == OMP_APPROVE_OPTION_ID:
```
AlexanderZ-Band: issue: OMP's elicitation-form handler never logs its approve/deny decision. When `self._resolve_permission` is `None` (no `PermissionResolver` configured — a supported, documented configuration), `option_id` stays `None` and every OMP approval form is auto-declined below with zero log output anywhere, not even at DEBUG. An operator running `OmpACPAdapter` without wiring `resolve_permission` would see every OMP action requiring approval silently fail, with nothing in Band's logs explaining why. The parallel, near-identical `_make_permission_handler` in `client_adapter.py` logs `"Permission request: tool=%s, session=%s, room=%s, option=%s"` for every resolved permission request regardless of outcome — this OMP elicitation path has no equivalent.

Fix: add a logger call recording the resolved `option_id` (or that no resolver is configured, so this always fail-closes) on both the accept and decline branches, mirroring `_make_permission_handler`'s logging.

src/band/integrations/acp/client_runtime.py
```diff
@@ -578,19 +616,51 @@ async def request_permission(  # type: ignore[override]  # ACP Client uses speci
...
-            # A denied request posts a tool_call/tool_result pair; the lock
-            # keeps the pair atomic between narration posts.
-            async with self._session_lock(session_id):
-                return await handler(
-                    options=options,
-                    session_id=session_id,
-                    tool_call=tool_call,
-                    **kwargs,
-                )
+            return await handler(
```
AlexanderZ-Band: issue: this PR narrows the per-session lock in `request_permission` and the new `create_elicitation` so only the final narration write (via `narrate_permission`/`narrate_elicitation`, i.e. `_session_narrator`) is serialized — the resolver await itself now runs unlocked, unlike the pre-PR code which held the session lock around the whole handler. This narrowing exists specifically to accommodate the new `resolve_permission`/`PermissionResolver` hook, which can await an arbitrarily slow external decision (e.g. a human clicking Approve/Deny in the room). `acp` dispatches `session_update` (notification) and `request_permission`/`create_elicitation` (requests) as independent, concurrently-running tasks, so this is a reachable interleaving: Task A is the permission handler for tool call T1, parked awaiting a slow `resolve_permission`; Task B is `session_update` for a second, concurrently-issued tool call T2 in the same session, which only needs the lock briefly to post its own tool_call/tool_result. Task B can acquire, post, and release the lock while Task A is still awaiting its slow resolver, so T2's narration reaches the room before T1's eventual denial narration, even though T1 was requested first on the wire. This contradicts this class's own docstring ("lock waiters wake FIFO and the tasks start in wire-arrival order, so room posts keep the stream's causal order"), which was accurate under the old full-handler lock but is no longer accurate once a slow resolver is configured.

Fix: either narrow the docstring's claim to match the new, weaker guarantee (mutual exclusion of narration writes, not causal ordering across a slow resolve), or have the handler reserve its place in room-post order before awaiting the slow resolver.

tests/e2e/baseline/toolkit/deps.py
```diff
@@ -189,6 +194,57 @@ def _codex_cli_available(settings: BaselineSettings) -> bool:
     return _cli_on_path(settings.backends.codex_command, "codex")
 
 
+def _bun_meets_min_version(min_version: str) -> bool:
```
AlexanderZ-Band: issue: `_bun_meets_min_version` (new in this PR) is the sole gate behind `Dep.OMP` via `_omp_available`, which decides whether the entire OMP CI lane runs or silently skips-with-reason. It has no test at any level: no case covers `shutil.which("bun")` returning `None`, `subprocess.run` raising `OSError`/`TimeoutExpired`, a non-zero return code, an unparsable `--version` string, or the `found >= required` tuple comparison itself. A bug in any of these (e.g. an inverted comparison, or a regex that misses a real Bun version string) would make `Dep.OMP` report unavailable forever with no failing test to surface it — the OMP lane would just quietly never run.

Fix: add unit tests next to the existing `_omp_cli_responds` tests in `tests/framework_conformance/test_toolkit_helpers.py`: Bun missing from PATH, subprocess raising OSError/TimeoutExpired, non-zero return code, unparsable version output, and both sides of the version comparison.

src/band/adapters/omp_acp.py
```diff
@@ -0,0 +1,218 @@
...
+                return AcceptElicitationResponse(
```
AlexanderZ-Band: issue: the elicitation handler's approve branch (`AcceptElicitationResponse(...)`) has no e2e coverage against a real `omp acp` process. `test_omp_acp_form_elicitation_denied_is_narrated` proves the deny half of this fail-closed mechanism live, but the approve half is only exercised by unit tests against a manually constructed `ElicitationFormSessionMode`. A mismatch between the payload OMP actually expects on approval and what this handler returns would pass CI while breaking every approve-configured deployment in production.

Fix: add an e2e smoke mirroring `test_omp_acp_form_elicitation_denied_is_narrated` but with a `resolve_permission` that returns `OMP_APPROVE_OPTION_ID`, asserting the requested action actually completes rather than being cancelled.

src/band/integrations/acp/client_adapter.py
```diff
@@ -512,20 +577,74 @@ async def handler(
...
+            ACPPermissionRequest(
+                room_id=room_id,
+                session_id=session_id,
+                tool_call=call,
+                options=offered,
+            )
+        )
+        if option_id is None:
+            return None
+        if not isinstance(option_id, str):
```
AlexanderZ-Band: suggestion: this `not isinstance(option_id, str)` guard (new in this PR, guarding a resolver that mistakenly returns e.g. `True` instead of an option id) is untested. The sibling guard just below it, for an option id not in the offered set, is tested (`test_permission_resolver_invalid_option_raises`), but this one is not.

Fix: add a test alongside `test_permission_resolver_invalid_option_raises` where the resolver returns a non-str value and assert the `ValueError` with this message.

src/band/integrations/acp/client_adapter.py
```diff
@@ -489,17 +551,20 @@ async def handler(
...
             del kwargs
             call = ACPToolCall.from_acp(
                 tool_call, canonicalize=self._canonical_tool_name
             )
 
-            # Auto-approve by selecting one of the agent's offered allow options;
-            # an ACP grant must reference an offered optionId (not a bare
-            # "allowed"), or the agent can't parse the response and aborts.
-            option_id = select_allow_option_id(options)
+            option_id = await self._resolve_permission_option(
```
AlexanderZ-Band: issue: starting state — a turn is awaiting a permission/elicitation decision; the handler is suspended at `await self._resolve_permission_option(...)`, itself awaiting the caller-supplied `resolve_permission`, which by design (per this PR) can be a human-driven Approve/Deny decision that outlives a turn.

Triggering sequence: the turn exceeds `_turn_timeout_s` (or `runtime.prompt(...)` raises), and `on_message` calls `self.on_cleanup(room_id)` -> `runtime.stop()` -> the ACP connection's `__aexit__()` -> the transport's task supervisor shutdown. Each incoming `request_permission`/`create_elicitation` JSON-RPC request runs as its own independent task (not part of `prompt_task`), so shutdown reaches directly into the still-parked `_resolve_permission_option` await and cancels it — `prompt_task.cancel()` never touches it.

Incorrect outcome: the handler is cancelled before ever reaching `emitter.open_permission(...)`, so no denial narration (tool_call/tool_result pair, or for OMP's form, the decline) is ever posted. Reproduced directly: constructing `ACPClientAdapter(resolve_permission=<hangs>)`, invoking `_make_permission_handler` as a standalone task, then cancelling it the way the transport's shutdown does, yields zero permission events on `FakeAgentTools` — only the unrelated session bookkeeping `task` event. The room is left with the agent's already-narrated tool_call forever unresolved, with no record the decision was ever pending or cancelled.

Fix: track in-flight permission/elicitation handler tasks (similar to `_background_tasks`) and in `on_cleanup`/turn-timeout, cancel + await them with a bounded grace window that still runs `emitter.open_permission(...)`/the elicitation decline post, rather than letting the connection teardown silently kill them first. Applies to both `_make_permission_handler` (here) and `OmpACPAdapter._make_elicitation_handler` (`src/band/adapters/omp_acp.py`), which share `_resolve_permission_option`.

src/band/integrations/omp.py
```diff
@@ -0,0 +1,255 @@
...
+    "google": "GEMINI_API_KEY",
+    "openai": "OPENAI_API_KEY",
+    "openrouter": "OPENROUTER_API_KEY",
+    "xai": "XAI_API_KEY",
+    "deepseek": "DEEPSEEK_API_KEY",
+    "mistral": "MISTRAL_API_KEY",
+    "groq": "GROQ_API_KEY",
+    "together": "TOGETHER_API_KEY",
+    "fireworks": "FIREWORKS_API_KEY",
+    "cohere": "COHERE_API_KEY",
```
AlexanderZ-Band: issue: `_OMP_PROVIDER_API_KEY_ENV` lists `"cohere": "COHERE_API_KEY"`, but OMP (the pinned `@oh-my-pi/pi-coding-agent` CLI this adapter spawns) has no top-level `cohere` provider at all. Cross-checked against the installed CLI's real provider registry (72 provider ids: anthropic, azure, openai, google, openrouter, xai, deepseek, mistral, groq, together, fireworks, etc.) — there is no `cohere` entry, and the literal string `COHERE_API_KEY` does not appear anywhere in the bundle. `cohere` only exists in OMP as a model family reachable through other providers (e.g. `openrouter/cohere/command-a`), never as a provider id of its own. `omp_provider_api_key_env`/`omp_provider_env` are consumed by `examples/acp/clients/omp.py` to validate `OMP_MODEL` before spawning the real OMP subprocess with a built env — so a user who sets `OMP_MODEL=cohere/command-r-plus` gets no SDK-level rejection; the SDK builds `{OMP_MODEL: "cohere/command-r-plus", COHERE_API_KEY: <key>}` and hands it to OMP, which has no such provider and no use for that env var, deferring the failure to an opaque OMP-side error instead of the SDK's own actionable `ValueError`. Reproduced: `omp_provider_api_key_env('cohere')` returns `'COHERE_API_KEY'` with no exception, and `omp_provider_env(model='cohere/command-r-plus', api_key='sk-fake')` returns `{'OMP_MODEL': 'cohere/command-r-plus', 'COHERE_API_KEY': 'sk-fake'}`.

Fix: drop the fabricated `"cohere"` entry (cohere is reachable in OMP only as a model family through other providers, not as a provider id), so `omp_provider_api_key_env("cohere")` raises `ValueError` like any other unsupported provider.

src/band/integrations/omp.py
```diff
@@ -0,0 +1,255 @@
...
+    return {"OMP_MODEL": model, env_key: api_key}
+
+
+def validate_omp_command(command: Sequence[str]) -> None:
+    """Reject spawn commands that disable OMP's always-ask safety."""
+    tokens = list(command)
+    index = 0
+    while index < len(tokens):
+        token = tokens[index]
+        if token in OMP_BARE_UNSAFE_APPROVAL_FLAGS:
```
AlexanderZ-Band: issue: `validate_omp_command`'s case-insensitivity is inconsistent across the three unsafe-flag spellings it checks. The bare-flag check here and the flag-name check for the space-separated form (`if token == OMP_APPROVAL_MODE_FLAG`) both compare the raw token with no `.lower()`, while only the value side (`mode.lower()`) and the `=`-form regex (compiled with `re.IGNORECASE`) are case-normalized. So `--YOLO`, `--Auto-Approve`, and the two-token `--APPROVAL-MODE yolo` all slip through uncaught, while `--approval-mode=YOLO` (equals form) is correctly caught. Reproduced: `validate_omp_command(("omp", "acp", "--YOLO"))`, `(..., "--Auto-Approve")`, and `(..., "--APPROVAL-MODE", "yolo")` all raise nothing, while the lowercase controls and the `=`-form equivalents correctly raise. (Checked the real OMP CLI's own arg parser too — it's equally case-sensitive on both flag names and mode values, so this is a dormant inconsistency in this PR's own validation logic rather than a live bypass of OMP's actual behavior today.)

Fix: normalize case once at the top of the loop (`token.lower()` for the bare-flag and flag-name comparisons), matching what's already done for the mode value and the `=`-form regex, so all three forms are handled the same way.

amit-gazal-band (dismissed):

